PR #9, opened 2017-04-24
Jenkins build role

tatitati:
- Added funcionality to allow jenkins upload assets to S3
- fixed problem with gulp failing when using @import. The reason is that is case sensitive, so was not finding the folders to import.

## Don't merge. This is the approach with cosmos-build. But we need to migrate from cosmos-build to cosmos-release to avoid the versioning folder name


programmes-developers:

PR #11, opened 2017-04-24
Fix to solve problem with case sensitive folders name

tatitati:


programmes-developers:

PR #12, opened 2017-04-24
Redirect for static assets to s3 when production environment

tatitati:


programmes-developers:

PR #13, opened 2017-04-26
Updated frontend with orbit-client changes

tatitati:
Updated frontend with orbit-client changes to use the new PSR-6 interfaces


programmes-developers:

BPScott:
Nice. I believe this means we're no longer using `doctrine/doctrine-cache-bundle` so that can be removed from our dependancies.

PR #15, opened 2017-05-03
Drop IE7 support in the grid framework

BPScott:
Moving to use box-sizing:border-box means we can remove that inner div
that we previously needed to create fixed width gutters.

### DIFF
--- a/src/Ds2013/Core/Gel/Objects/_grid.scss
+++ b/src/Ds2013/Core/Gel/Objects/_grid.scss
@@ -1,36 +1,38 @@
 /**
  * Grid object.
  *
- * An IE 7 compatible implementation of a fluid grid with fixed gutter sizes.
+ * An IE 8 compatible implementation of a fluid grid with fixed gutter sizes.
  * Use in conjunction with the width objects to set the sizing of your grid
  * items.
  *
- * The .grid__inner div is required for IE7 compatibility. If we drop support
- * for IE 7 then we can use box-sizing:border-box; and apply the padding
- * directly to the .grid element, thus removing the need for the nested div.
+ * The original grid from programmes had IE7 compatibility by using an inner
+ * `grid__inner` div rather than using box-sizing: border-box.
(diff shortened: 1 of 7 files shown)


programmes-developers:

src/Ds2013/Core/Gel/Objects/_grid.scss
```diff
@@ -1,36 +1,38 @@
...
  *
- * An IE 7 compatible implementation of a fluid grid with fixed gutter sizes.
+ * An IE 8 compatible implementation of a fluid grid with fixed gutter sizes.
  * Use in conjunction with the width objects to set the sizing of your grid
  * items.
  *
- * The .grid__inner div is required for IE7 compatibility. If we drop support
- * for IE 7 then we can use box-sizing:border-box; and apply the padding
- * directly to the .grid element, thus removing the need for the nested div.
+ * The original grid from programmes had IE7 compatibility by using an inner
```
votemike: Is this comment useful?

src/Ds2013/Core/Gel/Objects/_widths.scss
```diff
@@ -1,68 +1,77 @@
 /**
  * Widths.
  *
- * Human readable width abstractions.
+ * In the old system we used word based names for the classes e.g. "one-third"
```
votemike: Are comments about how a system used to work useful?

src/Ds2013/Core/Gel/Objects/_grid.scss
BPScott: I was thinking that in the move most of the the CSS classes shall be a copy+paste, however as this bit is deviating from what already exists it'd be worth pointing that divergence out to save people thinking "I copy+pasted something that mentions grid__inner but it doesn't do anything".

votemike: Fair enough

PR #16, opened 2017-05-05
Schedules skeleton

tatitati:
# in progress


programmes-developers:

src/Ds2013/Organism/Broadcast/BroadcastPresenter.php
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    public function getEndAt($format): string
+    {
+        $endAt = $this->broadcast->getEndAt();
+
+        return $endAt->format($format);
+    }
+
+    public function getBroadcast()
```
BPScott: We don't want to expose the domain objects to the templates as then the templates are reliant on the domain objects. The aim of having these presenters is that the templates rely only on the interface of the presenter rather than anything deeper. Remove this.

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,51 @@
...
     {
-        // TOD if date is absent use today's date
-        // TODO see if we can create a valid date from $date else throw 404
-        // TODO make sure date is in a sensible range - year is between 1900
-        // and 2100 else throw 404 (no point querying the DB for dates we know
-        // won't have any results)
-        // TODO lookup Service based on pid
-        // TODO get Broadcasts for this date (tv day or radio day)
+        // validation date
+        if (!is_null($date)) {
```
votemike: Change this to `is_null` and swap the code in the braces? Otherwise the else statement is `not not null`?

src/Ds2013/Organism/Broadcast/BroadcastPresenter.php
```diff
@@ -0,0 +1,144 @@
...
+    {
+        $options = $this->getOptions();
+        if (isset($options[$keyOption])) {
+            return $options[$keyOption];
+        }
+
+        return null;
+    }
+
+    public function getCss($key)
```
BPScott: These should be separate functions.

Currently you could put a bad key in and it'll complain at you

src/Ds2013/Organism/Broadcast/BroadcastPresenter.php
```diff
@@ -0,0 +1,144 @@
...
+            return $cssClasses[$key];
+        }
+
+        return null;
+    }
+
+    public function hasProgrammeLiveBroadcasts()
+    {
+        $has_live_broadcast = false;
+        if (isset($this->broadcast)
```
votemike: I can't read this `if` statement. Is it possible to refactor it in to a private method?

src/Ds2013/Organism/Broadcast/broadcast.html.twig
```diff
@@ -0,0 +1,67 @@
...
+        <div class="broadcast__info grid {{ broadcastPresenter.getCss('infoClasses') }}">
+            <div class="grid__inner">
+                <!-- start time and zone -->
+                {% if broadcastPresenter.getStartAt('c') is not null %}
+                    <h3 class="broadcast__time gamma" data-timezone="true" property="startDate" datatype="xsd:dateTime" content={{ broadcastPresenter.getStartAt('c') }}>
+                        <span class="timezone--time">
+                            {{ broadcastPresenter.getStartAt('H:i') }}
+                        </span>
+                        {%  if broadcastPresenter.getOption('show_date') == true %}
+                            <br />
```
votemike: Do we need to use `<br/>`s for spacing? Can we not use CSS?

src/Ds2013/Organism/Broadcast/broadcast.html.twig
```diff
@@ -0,0 +1,67 @@
...
+                        {% else %}
+                            {{ tr('on_air') }}
+                        {% endif %}
+                    </div>
+                {% endif %}
+            </div>
+        </div>
+
+        <div class="broadcast__programme grid {{ broadcastPresenter.getCss('programmeClasses') }}" rev="publication">
+            <div class="grid__inner">
```
BPScott: These grid__inner divs are no longer needed in the grid setup as of f864f40, so they can be removed

src/Ds2013/Organism/Broadcast/broadcast.html.twig
```diff
@@ -0,0 +1,67 @@
...
+
+        <div class="broadcast__info grid {{ broadcastPresenter.getCss('infoClasses') }}">
+            <div class="grid__inner">
+                <!-- start time and zone -->
+                {% if broadcastPresenter.getStartAt('c') is not null %}
+                    <h3 class="broadcast__time gamma" data-timezone="true" property="startDate" datatype="xsd:dateTime" content={{ broadcastPresenter.getStartAt('c') }}>
+                        <span class="timezone--time">
+                            {{ broadcastPresenter.getStartAt('H:i') }}
+                        </span>
+                        {%  if broadcastPresenter.getOption('show_date') == true %}
```
BPScott: No need for the == true. Use `{%  if broadcastPresenter.getOption('show_date') %}`

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -1,5 +1,38 @@
 {% extends 'base-ds2013.html.twig' %}
 
 {% block body %}
-Schedules By Day
+    {% include 'schedules/nav.html.twig' %}
+    {% include 'schedules/regionpick.html.twig' %}
+
+    {% block broadcasts %}
```
BPScott: Why do you need this block?

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -1,5 +1,38 @@
...
+                                <li {{ id }} {{ class }}>
+                                    {{ds2013('broadcast', broadcast, {
+                                        'highlight_box_classes' : 'highlight-box--list'
+                                    })}}
+                                </li>
+                        {% endfor %}
+
+                    </ol>
+                {% else %}
+                    NO SHEDULE TEMPLATE
```
votemike: SCHEDULE

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,51 @@
...
+            $date = new DateTimeImmutable();
+        }
+
+
+        /** @var EntityService $service */
+        $service = $servicesService->findByPidFull($pid);
+        $servicesInNetwork = $servicesService->findAllInNetwork($service->getNetwork()->getNid());
+
+        // set day interval time
+        $startDate = $date->setTime(0,0,0);
```
votemike: Carbon is good for dealing with things like startOfDay(). Do we want to think about using that? http://carbon.nesbot.com/docs/#api-modifiers

src/Ds2013/Organism/Broadcast/BroadcastPresenter.php
```diff
@@ -0,0 +1,144 @@
...
+        parent::__construct($ds2013, $options);
+        $this->broadcast = $broadcast;
+    }
+
+    public function getName(): string
+    {
+        return $this->broadcast->getService()->getName();
+    }
+
+    public function getSynopsis(): bool
```
votemike: This method sounds like a sentence or paragraph will be returned. Is there a better name for the method?

src/Ds2013/Organism/Broadcast/BroadcastPresenter.php
```diff
@@ -0,0 +1,144 @@
...
+    {
+        return $this->options['showSynopsis'];
+    }
+
+    public function getLocale(): string
+    {
+        return $this->presenterFactory->getLocale();
+    }
+
+    public function getStartAt($format): string
```
votemike: Do we want to pass in the format? Or do we want to enforce it so that it is constant across the codebase?

src/Ds2013/Organism/Broadcast/broadcast.html.twig
```diff
@@ -0,0 +1,67 @@
...
+                            <br />
+                            <span class="micro timezone--date">
```
votemike: Should this be a `<time>` tag?

src/Ds2013/Organism/Broadcast/BroadcastPresenter.php
```diff
@@ -0,0 +1,144 @@
...
+        if (isset($cssClasses[$key])) {
+            return $cssClasses[$key];
+        }
+
+        return null;
+    }
+
+    public function hasProgrammeLiveBroadcasts()
+    {
+        $has_live_broadcast = false;
```
BPScott: PSR2 so camelCase for variable names.

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,103 @@
...
+
+    private function _getBroadcastPeriod($broadcast)
+    {
+        $broadcastHourStart = $broadcast->getStartAt()->format('H');
+
+        switch ($broadcastHourStart)  {
+            case ($broadcastHourStart < 6 || $broadcastHourStart >23):
+                $period = 'late';
+                break;
+            case ($broadcastHourStart < 6):
```
votemike: Won't this always be caught by the case above?

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,51 @@
...
-        // and 2100 else throw 404 (no point querying the DB for dates we know
-        // won't have any results)
-        // TODO lookup Service based on pid
-        // TODO get Broadcasts for this date (tv day or radio day)
+        // validation date
+        if (!is_null($date)) {
+            $date = DateTimeImmutable::createFromFormat('Y-m-d', $date);
+            $dateErrors = DateTimeImmutable::getLastErrors();
+            if ($dateErrors['warning_count'] != 0 || $dateErrors['error_count'] != 0 || $date->format('Y') < 1900) {
+                throw $this->createNotFoundException('Invalid date');
```
BPScott: That exception message would never be exposed to the user, it'd only appear within our logs and we don't really care that somebody got a 404 because they requested `2017-01-99`.

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,103 @@
...
+            'network'=>$service->getNetwork(),
+            'service' => $service,
+            'servicesInNetwork'=>$servicesInNetwork,
+            'groupedBroadcasts' => $broadcastsGroupedByPeriodOfDay,
         ]);
     }
+
+    private function groupBroadcastsByPeriodOfDay($broadcasts, DateTimeImmutable $selectedDate = null)
+    {
+        $periods_of_day = [
```
BPScott: Follow PSR2 - use camelCase for variable names.

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,103 @@
...
+            // }
+
+            $period = $this->_getBroadcastPeriod($broadcast);
+            $periods_of_day[$period][] = $broadcast;
+        }
+
+        return array_filter($periods_of_day);
+    }
+
+    private function _getBroadcastPeriod($broadcast)
```
BPScott: No need to start private functions with _.

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,103 @@
...
+    {
+        $broadcastHourStart = $broadcast->getStartAt()->format('H');
+
+        switch ($broadcastHourStart)  {
+            case ($broadcastHourStart < 6 || $broadcastHourStart >23):
+                $period = 'late';
```
BPScott: If we're breaking in every case, could setting the variable be replaced with an early return, and going on from that would a set of if statements be easier to read over a switch/case?

src/Ds2013/Organism/Broadcast/BroadcastPresenter.php
BPScott: Agreed, I'd rather this returns the DateTime object, then in the template we can use the `format` filter. `broadcast.startAt|format('Y-m-d')`

programmes-developers:

PR #17, opened 2017-05-08
PROGRAMMES-5528, PROGRAMMES-5518 Redis on production and sandbox

Tasiobg:
- Add symfony configuration to use redis on production server. Similar configuration can be applied to the sandbox so we can develop and test with redis cluster if needed
- Add composer package "ocramius/proxy-manager" which implements lazy loading and also enabled lazy loading to redis to avoid unnecessary connections
- Whitelist virtualbox interface into app_dev.php so we are allowed to access sandbox from our local computer (php-fpm requirement)


programmes-developers:

web/app_dev.php
```diff
@@ -12,7 +12,7 @@
 // Feel free to remove this, extend it, or make something more sophisticated.
 if (isset($_SERVER['HTTP_CLIENT_IP'])
     || isset($_SERVER['HTTP_X_FORWARDED_FOR'])
-    || !(in_array(@$_SERVER['REMOTE_ADDR'], ['127.0.0.1', '::1']) || php_sapi_name() === 'cli-server')
+    || !(in_array(@$_SERVER['REMOTE_ADDR'], ['192.168.193.1', '127.0.0.1', '::1']) || php_sapi_name() === 'cli-server')
```
votemike: What is this I.P. address? Should this be in here if we are "coding in the open"?

BPScott: It's the sandbox IP address. Rather than replying on this being the external hostname it might be better if we set an environment variable that then gets passed to php-fpm that then lets us use debug mode. That would be impossible to fake

Tasiobg: anything within 192.168.0.0-192.168.255.255 is private network, IP packets addressed from them cannot be transmitted through the public Internet, more info: https://en.wikipedia.org/wiki/Private_network

@BPScott, no one can access app_dev.php but could be good idea to remove it as a step on the deployment process. As far as I know this file shouldn't be deployed. Anyway I'll pass this parameter into the fpm env variables

Bjorn told me that jenkins job may attempt to use redis while it is not available on the docker container, I'm doing some changes to fix this. should I close this PR and create new one with the changes or how should I procced?

votemike: Why are you suppressing errors?

Tasiobg: This is the default symfony file, I have only added the IP.
https://github.com/symfony/symfony-standard/blob/master/web/app_dev.php

After you rise the question a was a bit curious and I found the reason, REMOTE_ADDR is only available when running the script through a webserver, command line wont find this variable.

votemike: Fair enough. It's grim. I wonder why they did that and didn't just check with `isset`.

PR #18, opened 2017-05-09
PROGRAMMES-5528, PROGRAMMES-5518 Redis on production and sandbox

Tasiobg:
- Added symfony configuration to use cache on different environment. By default I have set parameters to use null adapter on sandbox and redis on production
- Add composer package "ocramius/proxy-manager" which implements lazy loading and also enabled lazy loading to redis to avoid unnecessary connections
- Whitelist virtualbox interface into app_dev.php so we are allowed to access sandbox from our local computer


programmes-developers:

BjornTwachtmann:
Weren't we going to do this without the lazy-loading until we could look into any effects that might have in another ticket?

Other than that this looks good, though we will need to change the pages-service stuff for the new cache abstraction.

Tasiobg:
code comments updated

app/config/services.yml
```diff
@@ -15,6 +15,18 @@ services:
     cache.null_provider:
         class: Symfony\Component\Cache\Adapter\NullAdapter
 
+    # we need an instance of RedisCluster, default Redis class doesn't support cluster mode.
+    # RedisCluster class support most Redis operations
+    cache.default_redis_provider:
+        lazy: true #lazy load otherwise a connection to the redis server will be stabilized with any request
```
BPScott: Remove the lazy loading and `ocramius/proxy-manager` for now so we have a baseline to develop against

web/app_dev.php
```diff
@@ -12,7 +12,7 @@
 // Feel free to remove this, extend it, or make something more sophisticated.
 if (isset($_SERVER['HTTP_CLIENT_IP'])
     || isset($_SERVER['HTTP_X_FORWARDED_FOR'])
-    || !(in_array(@$_SERVER['REMOTE_ADDR'], ['127.0.0.1', '::1']) || php_sapi_name() === 'cli-server')
+    || !(in_array(@$_SERVER['REMOTE_ADDR'], ['192.168.193.1', '127.0.0.1', '::1']) || php_sapi_name() === 'cli-server')
```
BPScott: This is unrelated to the redis config stuff, it should probably be a separate commit where we work out how to use debug mode from nginx.

PR #20, opened 2017-05-09
Schedules controllers

tatitati:


programmes-developers:

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -1,5 +1,41 @@
...
+                            </li>
+                        {% endfor %}
+
+                    </ol>
+                </li>
+            {% endfor %}
+
+        </ol>
+    {% else %}
+        NO SHEDULE TEMPLATE
```
votemike: SCHEDULE

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,100 @@
...
 namespace App\Controller;
 
+use BBC\ProgrammesPagesService\Domain\Entity\Broadcast;
+use BBC\ProgrammesPagesService\Domain\Entity\Service as EntityService;
 use BBC\ProgrammesPagesService\Domain\ValueObject\Pid;
+use BBC\ProgrammesPagesService\Service\BroadcastsService;
+use BBC\ProgrammesPagesService\Service\ServicesService;
+use DateInterval;
+use DateTimeImmutable;
+use BBC\ProgrammesPagesService\Domain\Entity\Network;
```
votemike: Should all the BBC be near each other?

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,100 @@
...
+            // }
+
+            $period = $this->_getBroadcastPeriod($broadcast, $selectedDate);
+            $periods_of_day[$period][] = $broadcast;
+        }
+
+        return array_filter($periods_of_day);
+    }
+
+    private function _getBroadcastPeriod($broadcast, DateTimeImmutable $selectedDate)
```
votemike: I feel this is the job of the Broadcast, not the controller. I could be wrong.

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,100 @@
...
+            'date' => $dateSelected,
+            'service' => $service,
+            'servicesInNetwork'=>$servicesInNetwork,
+            'groupedBroadcasts' => $broadcastsGroupedByPeriodOfDay,
         ]);
     }
+
+    private function groupBroadcastsByPeriodOfDay($broadcasts, DateTimeImmutable $selectedDate = null)
+    {
+        $periods_of_day = [
```
votemike: camelCase please

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,100 @@
...
     {
-        // TOD if date is absent use today's date
-        // TODO see if we can create a valid date from $date else throw 404
-        // TODO make sure date is in a sensible range - year is between 1900
-        // and 2100 else throw 404 (no point querying the DB for dates we know
-        // won't have any results)
-        // TODO lookup Service based on pid
-        // TODO get Broadcasts for this date (tv day or radio day)
+        // validation date
+        if (!is_null($dateSelected)) {
```
votemike: Reverse this check?

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -1,5 +1,41 @@
...
+
+            <!-- periods          -->
+            {% for periodName, broadcastInPeriod in groupedBroadcasts %}
+                <li id="{{ periodName }}" class="schedule-skip-link">
+                    <h2 class="delta text--shout island--horizontal component">{{ periodName }}</h2>
+                    <ol class="highlight-box-wrapper">
+                        {% set allow_on_now_class = true %}
+
+                        <!-- broadcasts in period-->
+                        {% for broadcast in broadcastInPeriod %}
```
votemike: This is a for nested in a for. Should some of this be split out into a separate template for clarity?

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,100 @@
...
+            $dateSelected = new DateTimeImmutable($dateSelected);
+            $dateErrors   = DateTimeImmutable::getLastErrors();
```
votemike: Move this in to a private method? `validateDate()` or something?

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -1,5 +1,41 @@
 {% extends 'base-ds2014.html.twig' %}
 
 {% block body %}
-Schedules By Day
+    {% include 'schedules/nav.html.twig' %}
+    {% include 'schedules/regionpick.html.twig' %}
```
BPScott: To make it clearer what includes need, i reckon we should make heavy use of the [`with ... only` syntax](https://twig.sensiolabs.org/doc/2.x/tags/include.html)

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -1,5 +1,41 @@
...
+
+    {% if groupedBroadcasts | length > 0 %}
+        <ol class="list-unstyled g-c-l">
+
+            <!-- periods -->
```
BPScott: Use twig comments if you need to comment things

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -1,5 +1,41 @@
 {% extends 'base-ds2014.html.twig' %}
 
 {% block body %}
-Schedules By Day
+    {% include 'schedules/nav.html.twig' %}
+    {% include 'schedules/regionpick.html.twig' %}
+
+    {% if groupedBroadcasts | length > 0 %}
```
BPScott: just {% groupedBroadcasts %} would work

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -1,5 +1,41 @@
...
+        <ol class="list-unstyled g-c-l">
+
+            <!-- periods -->
+            {% for periodName, broadcastInPeriod in groupedBroadcasts %}
+                <li id="{{ periodName }}" class="schedule-skip-link">
```
BPScott: We should come up with a standard on how we do variables. Personally I think I prefer `{{thing}}` over `{{ thing }}`.

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -1,5 +1,41 @@
...
+                    <h2 class="delta text--shout island--horizontal component">{{ periodName }}</h2>
+                    <ol class="highlight-box-wrapper">
+                        {% set allow_on_now_class = true %}
+
+                        <!-- broadcasts in period-->
+                        {% for broadcast in broadcastInPeriod %}
+                            {% set id = '' %}
+                            {% set class = '' %}
+                            {% set is_now = broadcast.getStartAt()|date('U') < "now"|date('U') and broadcast.getEndAt()|date('U') > "now"|date('U') %}
```
BPScott: We want to try and avoid as much logic in the view as possible. Try having an isNow / isOnAir helper in the broadcast presenter to deal with this 

Or perhaps doing this logic in the controller, and passing through a reference to the currently "on air" broadcast.

app/Resources/views/schedules/regionpick.html.twig
```diff
@@ -0,0 +1,22 @@
+{% if servicesInNetwork|length == 0 %}
```
BPScott: Does this actually work? I don't think twig has a way of exiting out of templates, that was an SPV thing.

app/config/services.yml
```diff
@@ -105,3 +105,11 @@ services:
     BBC\ProgrammesPagesService\Service\ProgrammesService:
         autowire: false
         factory: [ '@BBC\ProgrammesPagesService\Service\ServiceFactory', 'getProgrammesService']
+
+    BBC\ProgrammesPagesService\Service\ServicesService:
```
BPScott: Can we keep this service listing alphabetical?

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,100 @@
...
+            if ($dateErrors['warning_count'] != 0 || $dateErrors['error_count'] != 0
```
BPScott: Check is_null($date) rather than checking $dateErrors

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,100 @@
...
+                || $dateSelected->format('Y') < 1900 || $dateSelected->format('Y') > 2100) {
+                throw $this->createNotFoundException('Invalid date');
+            }
+        } else {
+            $dateSelected = new DateTimeImmutable();
+        }
+
+
+        /** @var EntityService $service */
+        $service = $servicesService->findByPidFull($pid);
```
BPScott: Need to throw a 404 is the service doesn't exist

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,100 @@
...
+
+    private function _getBroadcastPeriod($broadcast, DateTimeImmutable $selectedDate)
+    {
+        $dayStart = $selectedDate->setTime(0,0,0);
+        $dayEnd = $selectedDate->setTime(23,59,59);
+
+        $startBroadcast = $broadcast->getStartAt();
+        $startBroadcastHour = $startBroadcast->format('H');
+
+        switch ($startBroadcastHour)  {
```
BPScott: Would a set of if statements be easier to read than a switch/case?

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,67 @@
...
-        // TOD if date is absent use today's date
-        // TODO see if we can create a valid date from $date else throw 404
-        // TODO make sure date is in a sensible range - year is between 1900
-        // and 2100 else throw 404 (no point querying the DB for dates we know
-        // won't have any results)
-        // TODO lookup Service based on pid
-        // TODO get Broadcasts for this date (tv day or radio day)
+        if (is_null($dateSelected)) {
+            $dateSelected = new DateTimeImmutable();
+        } elseif (!is_null($dateSelected) && !$this->isValidDate($dateSelected)) {
```
votemike: If the code gets to this point then `$dateSelected` won't be null. So you don't need to check again.

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,67 @@
...
         return $this->renderWithChrome('schedules/by_day.html.twig', [
-            'service' => null,
+            'date' => $dateSelected,
+            'service' => $service,
+            'servicesInNetwork'=>$servicesInNetwork,
+            'groupedBroadcasts' => $broadcastsGroupedByPeriodOfDay,
         ]);
     }
+
+    private function isValidDate($dateSelected)
```
votemike: Typehints for `$dateSelected` and the return type?

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,67 @@
...
+            $dateSelected = new DateTimeImmutable();
+        } elseif (!is_null($dateSelected) && !$this->isValidDate($dateSelected)) {
+            throw $this->createNotFoundException('Invalid date');
+        } else {
+            $dateSelected = new DateTimeImmutable($dateSelected);
+        }
+
+        /** @var EntityService $service */
+        $service = $servicesService->findByPidFull($pid);
+        if (is_null($service)) {
```
votemike: @BPScott Should findByPidFull not throw the exception if the service is not found? Otherwise the controllers will be full of these `is_null()`s.

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,67 @@
...
+
+    private function isValidDate($dateSelected)
+    {
+        try {
+            $date = new DateTimeImmutable($dateSelected);
+        } catch (Exception $e) {
+            return false;
+        }
+
+        if ($date->format('Y') < 1900 || $date->format('Y') > 2100) {
```
votemike: Why are these invalid?

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,67 @@
...
+        /** @var EntityService $service */
+        $service = $servicesService->findByPidFull($pid);
+        if (is_null($service)) {
+            throw $this->createNotFoundException('Service not found');
+        }
+
+        $servicesInNetwork = $servicesService->findAllInNetwork($service->getNetwork()->getNid());
+
+        // set day interval time
+        $startDate = $dateSelected->setTime(0, 0, 0);
```
votemike: How does this deal with being in GMT or BST.  
And how does this play nicely twice a year when they switch over?

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,66 @@
...
         return $this->renderWithChrome('schedules/by_day.html.twig', [
-            'service' => null,
+            'date' => $date,
+            'service' => $service,
+            'servicesInNetwork'=>$servicesInNetwork,
+            'groupedBroadcasts' => $broadcastsGroupedByPeriodOfDay,
         ]);
     }
+
+    private function isValidDate($dateSelected) : bool
```
votemike: `string $dateSelected`?

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,66 @@
...
+        $service = $servicesService->findByPidFull($pid);
+        if (is_null($service)) {
+            throw $this->createNotFoundException('Service not found');
+        }
+
+        $servicesInNetwork = $servicesService->findAllInNetwork($service->getNetwork()->getNid());
+
+        // set day interval time
+        $startDate = $date->setTime(0, 0, 0);
+        if ($service->getNetwork()->getType() == 'tv') {
```
votemike: What happened to the constant for 'tv'?

src/Ds2014/Organism/Broadcast/BroadcastPresenter.php
```diff
@@ -0,0 +1,113 @@
...
+        array $options = []
+    ) {
+        $defaultOptions = [
+            'highlight_box_classes' => '',
+            'container_classes' => '',
+            'is_stacked' => false,
+            'show_synopsis' => true,
+            'show_date' => false,
+            'force_service' => false,
+            'show_programme_image' => true, // null = live_only, true = always, false = never
```
votemike: I feel these 3 options should be constants. Yes? No?

src/Ds2014/Organism/Broadcast/BroadcastPresenter.php
```diff
@@ -0,0 +1,113 @@
...
+            return $options[$keyOption];
+        }
+
+        return null;
+    }
+
+    public function getServiceSchemaUrl()
+    {
+        $pid = $this->getBroadcast()->getService()->getPid();
+        return 'http://www.bbc.co.uk/programmes/' . $pid;
```
votemike: This looks like it's for the live site. What about locally and test?

src/Ds2014/Organism/Broadcast/BroadcastPresenter.php
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    public function isNow()
+    {
+        $now = new DateTimeImmutable();
+        return $this->getBroadcast()->getStartAt() <= $now  and $now < $this->getBroadcast()->getEndAt();
```
votemike: `&&`?

src/Ds2014/Organism/Broadcast/BroadcastPresenter.php
```diff
@@ -0,0 +1,113 @@
...
+    {
+        return $this->getBroadcast()->getEndAt() < new DateTimeImmutable();
+    }
+
+    public function isTvBroadcast()
+    {
+        return $this->getBroadcast()->getProgrammeItem()->getNetwork()->getMedium() == 'tv';
+    }
+
+    public function isRadioBroadcast()
```
votemike: return typehint for this and other methods?

src/Controller/SchedulesByDayController.php
tatitati: yeah, I know is necessary to add it (this and a few more). But I didn't want to create noise in my pages-service PR initially with stuff that was not used yet in frontend. I will add them in the future. I will keep this idea in mind.

src/Ds2014/PresenterFactory.php
```diff
@@ -55,4 +55,15 @@ public function programmePresenter(
             $options
         );
     }
+
+    public function broadcastPresenter(
+        \BBC\ProgrammesPagesService\Domain\Entity\Broadcast $broadcast,
+        array $options = []
+    ): Organism\Broadcast\BroadcastPresenter {
+        return new Organism\Broadcast\BroadcastPresenter(
```
votemike: `use` these instead of writing the long name?

src/Ds2014/Organism/Broadcast/BroadcastPresenter.php
tatitati: Good point. So far i feel is quite random the amount of options that we can pass......I will manage it in a better way. Thanks!

src/Ds2014/Organism/Broadcast/BroadcastPresenter.php
tatitati: well, I copied the one that is being used (also as a constant in programme). I feel is something to ask to @BPScott 
https://github.com/bbc/programmes/blob/develop/webapp/php/sharedlib/BBC/Programmes/Programme.php#L235

src/Controller/SchedulesByDayController.php
BPScott: We know we don't have any schedules for those date formats, so we can throw a 404 early, rather than having to go make some DB calls

src/Ds2014/Organism/Broadcast/BroadcastPresenter.php
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    public function getEndAt($format): string
+    {
+        $endAt = $this->broadcast->getEndAt();
+
+        return $endAt->format($format);
+    }
+
+    public function getBroadcast()
```
BPScott: We shouldn't expose this

src/Ds2014/Organism/Broadcast/BroadcastPresenter.php
```diff
@@ -0,0 +1,113 @@
...
+        parent::__construct($ds2013, $options);
+        $this->broadcast = $broadcast;
+    }
+
+    public function getServiceName(): string
+    {
+        return $this->broadcast->getService()->getName();
+    }
+
+    public function getLocale(): string
```
BPScott: This isn't used

src/Ds2014/Organism/Broadcast/BroadcastPresenter.php
```diff
@@ -0,0 +1,113 @@
...
+    {
+        return $this->presenterFactory->getLocale();
+    }
+
+    public function getStartAt($format): string
```
BPScott: StartAt and EndAt should return DateTimeImmutable instances, and then you can use the [|date](https://twig.sensiolabs.org/doc/2.x/filters/date.html) filter in twig to format them

src/Ds2014/Organism/Broadcast/BroadcastPresenter.php
```diff
@@ -0,0 +1,113 @@
...
+
+        return $endAt->format($format);
+    }
+
+    public function getBroadcast()
+    {
+        return $this->broadcast;
+    }
+
+    public function getOption($keyOption)
```
BPScott: See commints in old PR

src/Ds2014/Organism/Broadcast/BroadcastPresenter.php
```diff
@@ -0,0 +1,113 @@
...
+    {
+        return $this->getBroadcast()->getProgrammeItem()->getNetwork()->getMedium() == 'tv';
+    }
+
+    public function isRadioBroadcast()
+    {
+        return $this->getBroadcast()->getProgrammeItem()->getNetwork()->getMedium() == 'radio';
+    }
+
+    public function isGapBroadcast()
```
BPScott: Remove this for now, we'll be implementing gaps in a different way

src/Ds2014/Organism/Broadcast/broadcast.html.twig
```diff
@@ -0,0 +1,75 @@
...
+    } %}
+
+    {{- cssByKey[key]|join(' ') -}}
+{% endmacro %}
+
+{% import _self as macro %}
+{% set presenter = broadcast %}
+{% set id= presenter.getOption('allow_on_now_class') and presenter.isNow() ? 'id=on-now' : null %}
+
+<li {{id}} {{macro.getCssByKey(presenter, 'broadcast_on_now')}}>
```
BPScott: The base item these templates should be divs. This currently means broadcasts have to be inside a list, rather than being something that could be rendered standalone.

src/Ds2014/Organism/Broadcast/broadcast.html.twig
```diff
@@ -0,0 +1,75 @@
...
+                        <meta property="name" content="{{presenter.getServiceName()}}" />
+                    </span>
+
+                    {% if presenter.isGapBroadcast() %}
+                        <div class="broadcast__live centi br-box-subtle">
+                            {{tr('off_air')}}
+                        </div>
+                    {% elseif presenter.isNow() %}
+                        <div class="broadcast__live centi br-box-highlight">
+                            {%  if presenter.isTvBroadcast() %}
```
BPScott: This might be a bit nicer to read as a terniary statement `{{tr(presenter.isTvBroadcast() ? 'on_now' :  'on_air')}}`

src/Ds2014/Organism/Broadcast/BroadcastPresenter.php
```diff
@@ -0,0 +1,113 @@
...
+{
+    /** @var Broadcast */
+    private  $broadcast;
+
+    public function __construct(
+        PresenterFactory $ds2013,
+        Broadcast $broadcast,
+        array $options = []
+    ) {
+        $defaultOptions = [
```
BPScott: Puts these in the options property rather than here, then the parent constructor can already deal with the option merging stuff

src/Controller/SchedulesByDayController.php
tatitati: @votemike these constants are created/used in pages-service. We want to keep this library as much simple as possible, and keep it small. So we are not adding these constants into that library.

src/Ds2013/Organism/Broadcast/BroadcastPresenter.php
```diff
@@ -0,0 +1,105 @@
...
+    {
+        return $this->broadcast->getProgrammeItem()->getNetwork()->getMedium() == 'radio';
+    }
+
+    /**
+     * @return null|string
+     */
+    public function getInfoClassesCss()
+    {
+        return $this->getOption('is_stacked') ? 'one-quarter bpb2-one-sixth bpw-one-sixth' : null;
```
BPScott: For these css utils, they're going to be used in template output so they should alway return strings. If you return null it's going to get casted to empty string when you output the template anyway.

Return an empty string rather than null.

src/Ds2013/Organism/Broadcast/BroadcastPresenter.php
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /**
+     * @return string|null
+     */
+    public function getObjectClassesCss()
```
BPScott: This shall always return a string to typehint

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,147 @@
...
     {
-        // TOD if date is absent use today's date
-        // TODO see if we can create a valid date from $date else throw 404
-        // TODO make sure date is in a sensible range - year is between 1900
-        // and 2100 else throw 404 (no point querying the DB for dates we know
-        // won't have any results)
-        // TODO lookup Service based on pid
-        // TODO get Broadcasts for this date (tv day or radio day)
+        if (is_null($date)) {
+            $date = new DateTimeImmutable();
```
BPScott: use `ApplicationTime::getTime()` rather than creating a DateTimeImmutable. This shall allow us to fix the concept of "now" to a fixed point when we write fixtures and tests so that we can write tests against things that are "on now" or not.

src/Ds2013/Organism/Broadcast/broadcast.html.twig
```diff
@@ -0,0 +1,33 @@
+{% set presenter = broadcast %}
+
+<div class="broadcast__info grid {{presenter.getObjectClassesCss()}}" vocab="http://schema.org/" typeof="BroadcastEvent">
+    <div class="grid-wrapper">
+        <div class="broadcast__info grid {{presenter.getInfoClassesCss()}}">
+                {#start time and zone#}
+                <h3 class="broadcast__time gamma" data-timezone="true" property="startDate" datatype="xsd:dateTime" content={{ presenter.getStartAt()|date('c') }}>
+                    <span class="timezone--time">{{presenter.getStartAt()|date('H:i')}}</span>
```
BPScott: Twig has a bit of functionality that lets you treat getters as properties, e.g. you can do {{presenter.startAt}} instead of {{presenter.getStartAt()}}. It's a little bit shorter an easier to read which I quite like.

We should agree on if we want to do that, or always have longform function names.

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,147 @@
...
+        } elseif (!$this->isValidDate($date)) {
+            throw $this->createNotFoundException('Invalid date');
+        } else {
+            $date = new DateTimeImmutable($date);
+        }
+
+        $service = $servicesService->findByPidFull($pid);
+        if (is_null($service || !$this->isServiceAvailableOnDate($service, $date))) {
```
votemike: AHH! Why is there logic inside an `is_null()`? If it does actually need to be there, I think it should be extracted in to a variable.

src/Controller/SchedulesByDayController.php
votemike: I see. I'm not sure I agree with this check and exception. 

- I don't think the code should "know" about what data is or isn't in the DB except if it has queried it.
- < 1900 and > 2100 are still valid dates. So why should the page 404 instead of showing a page saying no broadcasts are scheduled.

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -1,5 +1,35 @@
...
 
 {% block body %}
-Schedules By Day
+    {% include 'schedules/components/nav.html.twig' %}
+    {% include 'schedules/components/regionpick.html.twig' with { 'servicesInNetwork': servicesInNetwork, 'service': service } only %}
+
+    {% if groupedBroadcasts %}
+        <ol class="list-unstyled g-c-l">
+
+            {% for periodName, broadcastInPeriod in groupedBroadcasts %}
```
votemike: I think a single loop of this for is displaying a broadcastPeriod. Should we have a twig template for a broadcast period?

src/Controller/SchedulesByDayController.php
votemike: Also, `$this->isServiceAvailableOnDate` never returns `null`...

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,147 @@
...
+            // }
+
+            $period = $this->getBroadcastPeriodWord($broadcast, $selectedDate);
+            $intervalsDay[$period][] = $broadcast;
+        }
+
+        return array_filter($intervalsDay);
+    }
+
+    private function getBroadcastPeriodWord(Broadcast $broadcast, DateTimeImmutable $selectedDate): string
```
votemike: Should the `Broadcast` not be in charge of deciding it's word?

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,147 @@
...
+    /**
+     * @param array $broadcasts
+     * @return Broadcast|null
+     */
+    private function getBroadcastedNowFromSchedule(array $broadcasts)
+    {
+        $now = new DateTimeImmutable();
+
+        foreach ($broadcasts as $broadcast) {
+            if ($broadcast->getStartAt() <= $now && $broadcast->getEndAt() < $now) {
```
votemike: Should this be:  
`if ($broadcast->getStartAt() <= $now && $broadcast->getEndAt() > $now) {`?

app/Resources/views/schedules/by_day.html.twig
BPScott: The broadcast period thing is only used in this one place, nowhere else in the codebase. Nested loops aren't great but I don't think splitting this out shall help in the understanding of it. I don't think its worth splitting it out.

src/Controller/SchedulesByDayController.php
BPScott: The bucketing of broadcasts into periods only happens on the day schedule. I don't think it's worth putting something into the shared object if it is only used in one place, otherwise we run the risk of the Broadcast object becoming enormous.

src/Controller/SchedulesByDayController.php
BPScott: Agreed that both cases should show the 'no broadcasts page'. This would need to go after you get the service to show that contextual page, at which point we've got the info to check the service start/end dates so this is superflous.

That 'no broadcasts' page should have a status of 404 - we don't want to add 1900-01-01 and other bad dates to google's search index. But that's a problem we can handle when we sort out the no broadcasts page.

Lesson learned: I should stop with the premature optimising :)

src/Controller/SchedulesByDayController.php
BPScott: Uncaught exceptions would mean the page would return 500 errors. A non-existant service pid is a reasonable input to ask for, and we should return a 404 if a pid doesn't exist.

To do that we'd need something like this, which looks even worse:

```php
try {
    $service = $servicesService->findByPidFull($pid);
} catch (\Exception) {
    throw $this->createNotFoundException('Invalid service');
}
```

votemike:
It 11:33, but the programme on between 10:00 and 10:45 is marked as "On now".  
I assume this is something to do with GMT/BST

votemike:
My programme list (for TV) seems to go from 5 until midnight with no information between midnight and 5.

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -1,5 +1,29 @@
...
+
+    <ol class="list-unstyled g-c-l">
+        {% for periodName, broadcastInPeriod in groupedBroadcasts %}
+            <li id="{{periodName}}" class="schedule-skip-link">
+                <h2 class="delta text--shout island--horizontal component">{{periodName}}</h2>
+                <ol class="highlight-box-wrapper">
+
+                    {#list of broadcasts in day period#}
+                    {% for broadcast in broadcastInPeriod %}
+                        <li {{ broadcastedNow is not null and broadcast == broadcastedNow ? 'id=on-now class=schedule-skip-link-on-now' : null }} >
```
votemike: Does `broadcastedNow` need to be an object? Seems like an extra query. Can't you just check the start and end time?

app/Resources/views/schedules/components/none_broadcasts.twig
```diff
@@ -0,0 +1 @@
+NOT FOUND BROADCASTS FOR THIS DAY
```
votemike: I can't seem to trigger this.

src/Controller/SchedulesByDayController.php
```diff
@@ -2,22 +2,144 @@
...
+            return 'morning';
+        }
+
+        if ($startBroadcastHour < 18) {
+            return 'afternoon';
+        }
+
+        if ($startBroadcastHour <= 23 && $startBroadcast > $selectedDayStart) {
+            return 'evening';
+        }
```
votemike: Throw an exception if it doesn't hit any of the `if`s?

src/Ds2013/Organism/Broadcast/broadcast.html.twig
```diff
@@ -0,0 +1,33 @@
+{% set presenter = broadcast %}
+
+<div class="{{presenter.getObjectClassesCss()}}" vocab="http://schema.org/" typeof="BroadcastEvent">
```
votemike: Should there be some more metadata in the markup if we are using this schema?

PR #23, opened 2017-05-12
Adding other services block

votemike:
Waiting on reply to JIRA comment https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5512?focusedCommentId=3417246&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-3417246


programmes-developers:

votemike:
Can't figure out how to delete comments. But no longer waiting for any responses to ticket questions...

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -24,6 +24,26 @@
         {% endfor %}
     </ol>
 
+    {# If there is only 1 service in the network, it is the service we are currently viewing #}
+    {% if servicesInNetwork|length >= 2 %}
+        {% if service.network.medium == constant('BBC\\ProgrammesPagesService\\Domain\\Enumeration\\NetworkMediumEnum::TV') %}
```
BPScott: Would putting this check in the Controller be nicer? Exposing 'isTvNetwork' passed into the view perhaps? The view layer feels like the wrong layer of abstraction to be checking PHP constants.

votemike: Yeah, I guess that would be better.

votemike:
R.E. the CSS. Since the programmes stuff wasn't styled, I assumed this didn't need to be styled yet.

BPScott:
Added som small niggles. This is :+1: now.

PR #24, opened 2017-05-12
Pass in Translate to presenters only when required

BPScott:
Now that the ds2013() twig function exists we don't need to do weird
things with passing the PresenterFactory into every presenter because
the twig template deals with the creation of any dependant presenters
(e.g. rending an image presenter within a programme presenter).
No presenters shall need the Factory passed in as an argument.

However we still sometimes want to use the translation function within
some Presenters. So create a TranslatablePresenterTrait for these
presenters to use. These presenters should pass an instance of Translate
in as the first argument (not a hard and fast rule but seems like a
sensible convention).

### DIFF
--- a/src/Ds2013/Organism/Broadcast/BroadcastPresenter.php
+++ b/src/Ds2013/Organism/Broadcast/BroadcastPresenter.php
@@ -3,7 +3,6 @@ declare(strict_types = 1);
 namespace App\Ds2013\Organism\Broadcast;
 
 use App\Ds2013\Presenter;
-use App\Ds2013\PresenterFactory;
 use BBC\ProgrammesPagesService\Domain\ApplicationTime;
 use BBC\ProgrammesPagesService\Domain\Entity\Broadcast;
 use DateTimeImmutable;
@@ -24,11 +23,10 @@ class BroadcastPresenter extends Presenter
     private $now;
 
     public function __construct(
-        PresenterFactory $ds2013,
         Broadcast $broadcast,
         array $options = []
     ) {
-        parent::__construct($ds2013, $options);
+        parent::__construct($options);
         $this->broadcast = $broadcast;
         $this->now = ApplicationTime::getTime();
     }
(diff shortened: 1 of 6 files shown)


programmes-developers:

votemike:
Does the twig translation filter not cover every situation we need it to?

BjornTwachtmann:
This looks like a sensible approach. 

Two questions.

1. Can DesignSystemPresenterExtension not simply use the trait here? It looks like the same method. Or am I being obtuse?

2. Was the presenter factory not being used for anything other than translations in these classes? I imagine at some point we might want presenters to be able to create presenters, but can re-add the functionality then if necessary I suppose.

BPScott:
@votemike: The twig |trans filter doesn't quite handle all of the use cases that we need, so we ended up rolling our own. @BjornTwachtmann can give you more details on why we ended up rolling our own as he did most of the work around the translations. We need to pass in `Translate` into presenters because sometimes there is complex logic that requires a translation function within the presenter, e.g. getting the date in the ProgrammePresenter: https://github.com/bbc/programmes-frontend/pull/25/files#diff-bf08fd6f03b03c05ad75d6dbd45110a6R294

@bjorn: `DesignSystemPresenterExtension` could use the trait, it's the same `tr` function. It's `setTranslate` is slightly different (as it needs to update the translate instance within the ds2013PresenterFactory) but that's trivial to override. That not-being-a-perfect-drop-in bit put me off including it initially but you're right, using the trait in the extension would be nicer.

In the old Amen world we needed to pass in the factory because presenters that called other presenters (e.g. programme, calling image)[https://github.com/bbc/rmp-amen/blob/144cb53fb1ed62b06585f4103b1c16d422d04294/lib/src/RMP/Amen/Organism/Programme/ProgrammePresenter.php#L153] the Presenter had to return an instance of a presenter. In the new world the ds2103() function is available to the templates, so the presenter returns a domain object.

With that need gone, the only thing the factory is needed for is getting ahold of the Translate instance. At which point, just pass in Translate when required (i.e. this PR).

BPScott:
Have renamed `TranslatablePresenterTrait` to `TranslatableTrait` and the DesignSystemExtension now uses that trait.

tatitati:
Yeah, I think is what we need. and is still easy to modify if we have to.

BPScott:
@votemike, @felipeparaujo, @tatitati, any WIP presenters you've got shall have to be updated as their constructors have changed (for the simpler! No need to pass in a PresenterFactory if you don't need translations).

PR #25, opened 2017-05-15
Programme item presenter

tatitati:


programmes-developers:

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -24,23 +72,285 @@ public function __construct(
         $this->programme = $programme;
     }
 
-    public function getTitle(): string
+    public function getTitle(): array
```
votemike: If this returns an array, should it be `getTitles()`?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -24,23 +72,285 @@ public function __construct(
...
+
+    public function getAmountEpisodes(): ?int
+    {
+        return $this->programme->getParent()->getExpectedChildCount();
     }
 
     public function getSynopsis(): string
     {
-        return $this->programme->getShortSynopsis();
+        $truncateLength = ($this->getOption('truncate')) ? $this->getOption('truncate') * 1.5 : null;
```
votemike: What is the 1.5 magic? Should this be a constant? Or should it have a comment?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -24,23 +72,285 @@ public function __construct(
...
+    {
+        $titleLinkUrl = [
+            '/programmes',
+            (string) $this->programme->getPid(),
+        ];
+        // TODO: we dont have yet routes for programmes
+        return implode('/', $titleLinkUrl);
+    }
+
+    public function getObjectClasses(): string
```
votemike: `getObjectCssClasses `?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -24,23 +72,285 @@ public function __construct(
...
+        switch ($programme) {
+            case ($programme instanceof Series):
+                return self::SERIES;
+            case ($programme instanceof Brand):
+                return self::BRAND;
+            case ($programme instanceof Episode):
+                return self::EPISODE;
+            case ($programme instanceof Clip):
+                return self::CLIP;
+        }
```
votemike: Throw an Exception by default?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -24,23 +72,285 @@ public function __construct(
...
+                return $this->getNetworkMedium() == self::TV ? 'TVSeason' : 'Season';
+            case self::EPISODE:
+                return 'Episode';
+            case self::CLIP:
+                return 'Clip';
+        }
     }
 
-    public function getLocale(): string
+    private function secondsToReadableTime($seconds, $separator = ':'): string
```
votemike: This looks like something that someone else will have already done better. Is there a composer package to do this? It will probably be more stable and reduce the amount of code we have to maintain.

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -24,23 +72,285 @@ public function __construct(
...
+        $hours = floor($seconds / $secondsInDay);
+        $seconds = $seconds - ($hours * $secondsInDay);
+        $minutes = str_pad((string) floor($seconds / 60), 2, '0', STR_PAD_LEFT);
+        $seconds = str_pad((string) ($seconds%60), 2, '0', STR_PAD_LEFT);
+
+        $time = (($hours) ? $hours . $separator : '') . $minutes . $separator . $seconds;
+        return $time;
+    }
+
+    private function getWordyDate(DateTimeImmutable $programmeReleaseDate, $useRelativeDays = true): string
```
votemike: I think a lot of this could be made simpler by using [Carbon](http://carbon.nesbot.com/docs/)

src/Ds2013/Organism/Broadcast/BroadcastPresenter.php
```diff
@@ -48,10 +48,9 @@ public function getEndAt(): DateTimeImmutable
         return $this->broadcast->getEndAt();
     }
 
-    public function getServiceSchemaUrl(): string
+    public function getSchemaUrl(): string
```
votemike: What is schema URL?

src/Ds2013/Presenter.php
```diff
@@ -102,6 +114,27 @@ protected function tr(
...
+        }
+    }
+
+    protected function truncateText(string $text, ?int $length, $suffix = '...'): string
+    {
+        if (strlen($text) <= $length) {
+            return $text;
+        }
+
+        $length = strrpos($text, ' ', -(strlen($text) - $length));
```
votemike: Use `rtrim` instead?

votemike:
```
if (mb_strwidth($value, 'UTF-8') <= $limit) {
    return $value;
}
return rtrim(mb_strimwidth($value, 0, $limit, '', 'UTF-8')).$end;
```

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -24,23 +71,290 @@ public function __construct(
...
+
+        return $icon;
+    }
+
+    public function isProgrammeTleo(): bool
+    {
+        return is_null($this->programme->getParent());
+    }
+
+    private function getRecursiveTitles($programme, array $stackedTitles = []): array
```
votemike: I'm not entirely sure what this method is doing. Could it have a clearer name? Or a docblock?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -24,23 +71,290 @@ public function __construct(
...
+
+    public function getNumberCurrentEpisode(): ?int
     {
-        return $this->programme->getTitle();
+        return $this->programme->getPosition();
+    }
+
+    public function getAmountEpisodes(): ?int
+    {
+        return $this->programme->getParent()->getExpectedChildCount();
```
votemike: What if `->getParent()` returns null? An error will occur when trying to call `->getExpectedChildCount()` on null.

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -24,23 +71,290 @@ public function __construct(
...
     }
 
     public function getSynopsis(): string
     {
-        return $this->programme->getShortSynopsis();
+        $truncateLength = ($this->getOption('truncate')) ? $this->getOption('truncate') * 1.5 : null;
```
votemike: Instead of setting $truncateLength to null and truncating if there is no option, why don't you just return `$this->programme->getShortSynopsis()` straight away?

src/Ds2013/Presenter.php
```diff
@@ -102,6 +114,27 @@ protected function tr(
...
+    {
+        switch ($entity) {
+            case ($entity instanceof Programme):
+                return 'http://www.bbc.co.uk/programmes/' . $entity->getPid();
+            case ($entity instanceof Broadcast):
+                return 'http://www.bbc.co.uk/programmes/' . $entity->getService()->getPid();
+        }
+    }
+
+    protected function truncateText(string $text, ?int $length, $suffix = '...'): string
```
votemike: Why would `$length` be null?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -24,23 +71,290 @@ public function __construct(
...
+        return $this->truncateText($this->programme->getShortSynopsis(), $truncateLength);
+    }
+
+    public function showDuration()
+    {
+        return $this->getOption('show_duration') || $this->getTypeFromProgramme() == self::CLIP;
```
votemike: You need to pass the programme in to `$this->getTypeFromProgramme()`

src/Ds2013/Organism/Programme/ProgrammePresenter.php
votemike: Typehint `$programme`?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -24,23 +71,290 @@ public function __construct(
...
+            return $this->getRecursiveTitles($programme->getParent(), $stackedTitles);
+        }
+
+        // sort from container to item
+        return array_reverse($stackedTitles);
+    }
+
+    private function isProgrammeContainer(): bool
+    {
+        return $this->getTypeFromProgramme($this->programme) == self::BRAND || $this->getTypeFromProgramme($this->programme) == self::SERIES;
```
votemike: Maybe use `in_array()` instead of two `==`, then you can avoid the effort of running `$this->getTypeFromProgramme()` twice.

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -24,23 +71,290 @@ public function __construct(
...
+    }
+
+    private function getNetworkMedium(): string
+    {
+        return $this->programme->getNetwork()->getMedium();
```
votemike: What if `$this->programme->getNetwork()` is null?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -24,23 +71,290 @@ public function __construct(
...
+            case self::SERIES:
+                return $this->getNetworkMedium() == self::TV ? 'TVSeason' : 'Season';
+            case self::EPISODE:
+                return 'Episode';
+            case self::CLIP:
+                return 'Clip';
+        }
+    }
+
+    private function secondsToReadableTime($seconds, $separator = ':'): string
```
votemike: Typehint `int`?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
tatitati: Each title is an array which contains PID, title text, type programme and something else. So we return only one of those arrays in this case (1 title) , no multiple. I agree, is confuse, maybe we can create a new object for title, but I think it should be discussed. Maybe is used very little and we can just use this stuff.

src/Ds2013/Organism/Programme/ProgrammePresenter.php
tatitati: By design for any reason we were using this constant to set the truncate length. I just set the same. But I don't understand why is used.

https://github.com/bbc/programmes/blob/develop/webapp/php/sharedmodules/programmes/views/partials/generic/programme-object.spv#L480

src/Ds2013/Organism/Programme/ProgrammePresenter.php
votemike: Is 'title' referring to the description of the programme? Or the programme itself?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
tatitati: We build the css inside the presenter to avoid much more logic in the twig template. I used the same variable name than in the original. We can change the name, but I think is gonna be confuse however. Programme-wrapper (but that is a broadcast), and programme has content and body....is a little hard to decide the name


https://github.com/bbc/programmes/blob/develop/webapp/php/sharedmodules/programmes/views/partials/generic/programme-object.spv#L98

src/Ds2013/Organism/Programme/ProgrammePresenter.php
votemike: @BPScott @BjornTwachtmann Any idea?

src/Ds2013/Organism/Broadcast/BroadcastPresenter.php
tatitati: Schema is the url for that service (if we speak of broadcasts), and the url for that programme (if we speak of programmes).

https://github.com/bbc/programmes/blob/develop/webapp/php/sharedlib/BBC/Programmes/Programme.php#L235

https://github.com/bbc/programmes/blob/develop/webapp/php/sharedlib/BBC/Programmes/Service.php#L1167

src/Ds2013/Organism/Programme/ProgrammePresenter.php
votemike: Your link doesn't have the same name....

src/Ds2013/Organism/Broadcast/BroadcastPresenter.php
votemike: I don't quite see why 'schema' is in the name, it doesn't seem to do anything to do with a schema....  
http://www.dictionary.com/browse/schema

src/Ds2013/Organism/Programme/ProgrammePresenter.php
tatitati: Titles hare hierarchical. They are like this tree: "brand/serie/episode"
If you see the schedules page you will see something like this:

# < brand title >
## < serie title > , < episode title >

Becuase the programme we are displaying is an specific epidose, we have to go recursively to the parent in order to take all the titles.

votemike: I see.... Very confusing to a newcomer. @BPScott is there some clearer way we could describe this?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
BPScott: Use the class hierarchy. `return $this->programme instanceof ProgrammeContainer.`

src/Ds2013/Organism/Programme/ProgrammePresenter.php
BPScott: getTitle() and getSubtitle() in the Presenter, what they return shall change depending on if the `bottom_up` option is set. What Amen does may or may not be useful: https://github.com/bbc/rmp-amen/blob/master/lib/src/RMP/Amen/Helper/Titles.php

src/Ds2013/Presenter.php
```diff
@@ -102,6 +114,27 @@ protected function tr(
         return $this->presenterFactory->getTranslate()->translate($key, $substitutions, $numPlurals, $domain);
     }
 
+    protected function getSchemaUrlFromEntity($entity): string
+    {
+        switch ($entity) {
+            case ($entity instanceof Programme):
+                return 'http://www.bbc.co.uk/programmes/' . $entity->getPid();
```
BjornTwachtmann: Hardcoding URLs in the presenters is an extremely bad idea. We want to leave this up to the symfony routing logic. Chage this function to return a pid and pass that into the path() function in the Twig view with the relevant route.

src/Ds2013/Presenter.php
tatitati: when is null the function calculates a proper length

votemike: I don't understand....

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -4,25 +4,34 @@
...
+    <div class="grid-wrapper">
+        <ol class="list-unstyled g-c-l">
+            {% for periodName, broadcastInPeriod in groupedBroadcasts %}
+                <li id="{{periodName}}" class="schedule-skip-link">
+                    <h2 class="delta text--shout island--horizontal component">{{periodName}}</h2>
+                    <ol class="highlight-box-wrapper">
+
+                        {#list of broadcasts in day period#}
+                        {% for broadcast in broadcastInPeriod %}
+                            <li {{ broadcast == broadcastedNow ? 'id=on-now class=schedule-skip-link-on-now' : null }} >
```
votemike: Do you need to compare a whole object? Can't you just check to see if this broadcast is on now. Rather than loading the broadcast that is on now and comparing to that.

src/Ds2013/Core/Gel/Objects/_media.scss
```diff
@@ -0,0 +1,74 @@
...
+}
+
+
+/**
+ * `.img`s in `.islet`s need an appropriately sized margin.
+ */
+.islet .media__img {
+    margin-right: $basic-half-spacing;
+}
+.islet .media__img--rev {
```
votemike: This can be combined with the statement above can't it?

src/Ds2013/Organism/Broadcast/broadcast.html.twig
```diff
@@ -11,7 +11,7 @@
 
                 {#end time and zone#}
                 <meta property="endDate" datatype="xsd:dateTime" content="{{ broadcast.getEndAt()|date('c')}}" />
-                <span property="publishedOn" typeof="BroadcastService" resource="{{broadcast.getServiceSchemaUrl()}}>">
+                <span property="publishedOn" typeof="BroadcastService" resource="{{ path('find_by_pid', {'pid' : broadcast.getPid()}) }}>">
```
votemike: Does `{'pid' : broadcast.pid}` not work?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -4,15 +4,60 @@
 
 use App\Ds2013\Presenter;
 use App\Ds2013\TranslatableTrait;
-use BBC\ProgrammesPagesService\Domain\Entity\Programme;
 use RMP\Translate\Translate;
+use BBC\ProgrammesPagesService\Domain\Entity\Programme;
+use BBC\ProgrammesPagesService\Domain\Entity\Episode;
```
votemike: I believe these are supposed to be in alphabetical order.

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -28,23 +73,208 @@ public function __construct(
         $this->programme = $programme;
     }
 
-    public function getTitle(): string
+    public function getTitle(): array
     {
-        return $this->programme->getTitle();
+        $stackedTitles = $this->getRecursiveTitles($this->programme);
+        if ($this->getOption('bottom_up') || $this->getTypeFromProgramme($this->programme) == self::CLIP) {
```
votemike: Does `$this->programme` not know what type it is?

src/Ds2013/Organism/Programme/programme.html.twig
```diff
@@ -1,8 +1,132 @@
...
+
+                {#BROADCAST EVENT#}
+                {#TO-DO: broadcasts events will be implemented in the future. Dependency of programme with broadcasts.#}
+
+                {#SIMULCAST#}
+                {#TO-DO: Dependency of programme with broadcasts.#}
+
+                {#AVAILABILITY#}
+                {% if programme.getOption('show_child_availability') %}
+                    {% if programme.isProgrammeTleo() == false or programme.isAvailable() == false %}
```
votemike: Can these two ifs be combined?

PR #27, opened 2017-05-16
Use isActiveAt / isOnAirAt utils from pages-service

BPScott:
We've moved some datetime logic into pages-service, so use it.
Some slight tweaks to naming while I'm in there.


programmes-developers:

tatitati:
I like it

PR #29, opened 2017-05-16
5584 Adding date bar

votemike:
I believe I have fulfilled the ticket 5584.  
This involved blindly copying some scss files from the programmes codebase.  
I have added [Carbon](http://carbon.nesbot.com/docs/) in as a way to manipulate dates as I believe it is easier/clearer to use than doing things such as `new \DateInterval('P7Y6M5D')`. If people agree we shouldn't use Carbon, I can take it out 😢 .  
I believe I have simplified the current logic as much as possible without sacrificing functionality. Please let me know if any logic can be refactored or any functionality has disappeared.


programmes-developers:

BjornTwachtmann:
I don't really have time to review this properly. One quick comment, why are we importing a whole new DateTime library, that makes me nervous. Just because we've had major performance problems with date manipulation and Zend_Date on these pages in the past. It may be justified, but I'd rather not do it unless we have to.

votemike:
@BjornTwachtmann Mainly because of the clarity and ease of use. I think it will be a lot more difficult to make mistakes with this rather than manually constructing `DateInterval`s in order to manipulate dates.  
I can remove Carbon if you like.  
What were the problems you had before?

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -1,6 +1,11 @@
 {% extends 'base-ds2013.html.twig' %}
 
+{% block html_classes %}fixed-supported{% endblock %}
+{% block page_classes %}programmes-page text-base b-g-p programmes-page--flush{% endblock %}
+
 {% block body %}
+    {{ ds2013('dateList', date, service ) }}
+
     {% include 'schedules/components/nav.html.twig' %}
```
votemike: What is this? Is this where `{{ ds2013('dateList', date, service ) }}` should actually go?

Tasiobg:
I have been using Carbon previously for a year and never had an issue with it.
I haven't deeply check the carbon code but I think it is only a fancy way to calls the PHP DateTime class, I belive the perfotmance impact of Carbon should be irrelevant but I haven't benchmark it.

On my point of view the important decision here is whether to use standar PHP DateTime or a popular library to make dealing with times more redable/easy. If we want to use carbon we should use it project wide because 2 reasons: 
- To keep code consistent.
- Carbon class is inherited from the PHP DateTime class.

I have no strong opinion, both looks good options to me, I think carbon make code more for human

@BPScott any thought about using [Carbon](http://carbon.nesbot.com/docs/) ? if there is no strong opinions, I would suggest to start a vote.

src/Ds2013/Molecule/DateList/DateListPresenter.php
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    public function getServicePid(): string
+    {
+        return (string) $this->service->getPid();
+    }
+
+    private function constructDate(Carbon $selectedDate, int $offset): stdClass
+    {
+        $holder = new stdClass();
```
BPScott: What do you think about having a presenter class for this rather than being a stdClass. Would be a bit clearer about what you can do with the item

src/Ds2013/Molecule/DateList/DateListPresenter.php
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        return $holder;
+    }
+
+    private function shouldShowLink(int $offset, Carbon $datetime): bool
+    {
+        // if the date is more than 90 DAYS from now, then don't allow a link (page will still exist)
+        return $offset != 0 &&
+            Carbon::now()->addDays(90)->greaterThan($datetime) &&
```
BPScott: Use ApplicationTime::getTime() if you need to create 'now'. This is a central place that allows us to fix the concept of 'now' in unit tests.

Carbon is a bit more fluent than the raw php style, but for this sort of manipulation the raw php is still pretty readable (once you learn what ISO8601 periods are).

`ApplicationTime::getTime()->addInterval(new DateTimeInterval('P90D')) > $dateTime`

src/Ds2013/Molecule/DateList/dateList.html.twig
```diff
@@ -0,0 +1,32 @@
+<nav class="schedule-nav-wrapper fixedsticky">
+    <div id="schedule-daynav" class="g-o-l centi br-box-highlight">
+        <ol class="nav nav--banner date-list {{ dateList.cssExtraClasses }}">
+            {#If two <li> tags have whitespace between them, they render with space between them#}
+            {% spaceless %}
+                {% for item in dateList.items %}
```
BPScott: Might be worth using the per-tag whitespace control to avoid the need for the {% spaceless %} tag: https://twig.sensiolabs.org/doc/2.x/templates.html#whitespace-control

IMO i reckon we should be doing this on pretty much every tag by default anyway

src/Ds2013/Molecule/DateList/dateList.html.twig
```diff
@@ -0,0 +1,32 @@
...
+                    <li class="{{ item.cssClassesFor.li }}">
+                        {% if item.showLink %}
+                            <a href="{{ path('schedules_by_day', {'pid': dateList.servicePid, 'date': item.date.toDateString}) }}" class="{{ item.cssClassesFor.item }}" {{ item.attributesFor.item }}>
+                                {% if item.date.isToday %}
+                                    <span class="date-stack__line1">{{ tr('schedules_today') }}</span>
```
BPScott: Might a ternary be clearer and less repetitive here than a full {% if %} block?

src/Ds2013/Molecule/_date-stack.scss
```diff
@@ -0,0 +1,27 @@
+.date-stack {
```
BPScott: If you've got just a CSS file it's an atom, rather than a molecule.

src/Ds2013/Molecule/_fixed-sticky.scss
```diff
@@ -0,0 +1,37 @@
+/**
```
BPScott: If you've got just a CSS file it's an atom, rather than a molecule.

It looks like browser support for position:fixed is pretty good these dates, we might be able to get rid of this polyfill. See if things work as expected if you remove it.

votemike: position:fixed? Or position:sticky?

src/Ds2013/Molecule/DateList/DateListPresenter.php
votemike: Fine if you're checking greater than. But gets more verbose if you're checking if a date is between two dates etc...

src/Ds2013/Molecule/_fixed-sticky.scss
BPScott: autopiloting, sorry. position:sticky.

votemike:
So do you want Carbon removed?

src/Ds2013/Molecule/DateList/dateList.html.twig
votemike: Seems now that it's in it's own template as part of a presenter, I need both.

src/Ds2013/Molecule/_fixed-sticky.scss
votemike: Doesn't seem to work in IE at all... https://caniuse.com/#search=position%3Asticky

src/Ds2013/Molecule/DateListItem/dateListItem.html.twig
```diff
@@ -0,0 +1,24 @@
+{% spaceless %}
+    <li class="{{ dateListItem.cssClassesForLi }}">
+        {% if dateListItem.shouldShowLink %}
+            <a href="{{ path('schedules_by_day', {'pid': dateListItem.servicePid, 'date': dateListItem.datetime.toDateString}) }}"
+               class="{{ dateListItem.cssClassesForItem }}" {{ dateListItem.attributesForItem }}>
+            <span class="date-stack__line1">
+                {{ dateListItem.datetime.isToday ? tr('schedules_today') : dateListItem.datetime.format('D') }}
```
votemike: How should the date be translated?

src/Ds2013/Molecule/DateListItem/dateListItem.html.twig
```diff
@@ -0,0 +1,24 @@
...
+            </span>
+                <span class="date-stack__line2">
+                {{ dateListItem.datetime.format('d M') }}
```
votemike: How should the month be translated?

src/Ds2013/Molecule/DateList/DateListPresenter.php
```diff
@@ -0,0 +1,54 @@
...
+class DateListPresenter extends Presenter
+{
+    /**
+     * @inheritDoc
+     */
+    protected $options = [
+        'css_classes' => '',
+        'user_timezone' => 'GMT',
+    ];
+    /**
```
tatitati: add a white line between both variables

src/Ds2013/Molecule/DateList/DateListPresenter.php
```diff
@@ -0,0 +1,54 @@
...
+     * @var DateTimeImmutable
+     */
+    private $datetime;
+    /**
```
tatitati: same (white line)

src/Ds2013/Molecule/DateList/DateListPresenter.php
```diff
@@ -0,0 +1,54 @@
...
+    public function __construct(DateTimeImmutable $datetime, Service $service, array $options = [])
+    {
+        parent::__construct($options);
+        $this->datetime = $datetime;
+        $this->service = $service;
+    }
+
+    public function getCssExtraClasses(): string
+    {
+        return $this->options['css_classes'];
```
tatitati: we have a getOption function in the parent Presenter class. We should use it in order to avoid to access to a key that doesn't exist. I think can give us less error prone. Also is redundant to have another getOptions() function in here. Usually we will need to access to one option when need it, but not to all of them

src/Ds2013/Molecule/DateList/DateListPresenter.php
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    public function getDateTime(): DateTimeImmutable
+    {
+        return $this->datetime;
+    }
+
+    public function getOptions(): array
```
tatitati: this function is not necessary I think, we can use getOption() from the parent

src/Ds2013/Molecule/DateListItem/DateListItemPresenter.php
```diff
@@ -0,0 +1,88 @@
...
+
+class DateListItemPresenter extends Presenter
+{
+    /**
+     * @inheritDoc
+     */
+    protected $options = [
+        'user_timezone' => 'GMT',
+    ];
+    /**
```
tatitati: white line

src/Ds2013/Molecule/DateList/DateListPresenter.php
votemike: Can do. Is that our coding standard? If so, I'll set my IDE up to do it automatically...

src/Ds2013/Molecule/DateList/DateListPresenter.php
votemike: But I need to get all the options to pass through to the next presenter.

src/Ds2013/Molecule/DateList/DateListPresenter.php
BPScott: Yep, should have a blank line here.
 
If it's only one annotation then you do it one one line too: `/** @var DateTimeImmutable */`. We do this throughout pages-service.

src/Ds2013/Molecule/DateListItem/DateListItemPresenter.php
```diff
@@ -0,0 +1,88 @@
+<?php
```
BPScott: Am I right in saying this is an implementation detail of the DateList, and thus it shouldn't be constructed on its own outside of the DateList? If so should we move it into Molecule/DateList to make it clear that it's not a stand-alone thing?

src/Ds2013/Molecule/DateListItem/date_list_item.html.twig
```diff
@@ -0,0 +1,24 @@
+{% spaceless %}
+    <li class="{{ date_list_item.cssClassesForLi }}">
+        {% if date_list_item.shouldShowLink %}
```
BPScott: How about this to avoid duplication of all the bits inside the container??

```
 {% if date_list_item.shouldShowLink %}
  <a href="..." ...>
{% else %}
  <span class="{{ date_list_item.cssClassesForItem }}">
{% endif %}

blah

 {% if date_list_item.shouldShowLink %}
  </a>
{% else%}
  </span>
{% endif %}

File: src/Ds2013/PresenterFactory.php [hunk 17]
```diff
@@ -47,6 +51,34 @@ public function setTranslate(Translate $translate): void
         $this->translate = $translate;
     }
 
+    /** Molecules **/
+    public function dateListItemPresenter(
```
BPScott: If this is internal to the DateList and has no value being constructed on its own, we can probably remove it from the factory.

This would mean you wouldn't be able to use `ds2013('dateListItem', ...)` in the date_list template and instead you'd have to construct the Presenter within the DateListPresenter, and call {% include ... %} within the template. Which is a bit uglier but avoids publicising this implementation detail.

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -1,6 +1,11 @@
 {% extends 'base_ds2013.html.twig' %}
 
+{% block html_classes %}fixed-supported{% endblock %}
```
votemike: So..... what needs to be done here? Remove this line and not worry about the sticky? If so, do you want to paste the link to the sticky ticket in a comment?

src/Ds2013/Molecule/DateListItem/date_list_item.html.twig
```diff
@@ -13,7 +13,7 @@
         {% else %}
             <span class="{{ date_list_item.cssClassesForItem }}">
                 <span class="date-stack__line1">
-                    {{ date_list_item.datetime.isToday ? tr('schedules_today') : date_list_item.datetime.format('D') }}
+                    {{ (date_list_item.datetime.format('Y-m-d') == date_list_item.now.format('Y-m-d')) ? tr('schedules_today') : date_list_item.datetime.format('D') }}
```
BPScott: Expose an 'isToday` method on the date list item?

src/Ds2013/Molecule/DateListItem/DateListItemPresenter.php
votemike: Should `DateListItemPresenter` be renamed `ItemPresenter` so that it is `App\Ds2013\Molecule\DateList\ItemPresenter`? Or leave it as `App\Ds2013\Molecule\DateList\DateListItemPresenter`?

src/Ds2013/Molecule/DateListItem/date_list_item.html.twig
votemike: I thought about that. But it just doesn't seem right to me. Seems..... grim.

src/Ds2013/Molecule/DateListItem/DateListItemPresenter.php
```diff
@@ -68,11 +72,16 @@ public function getCssClassesForLi(): string
         ]);
     }
 
-    public function getDatetime(): Carbon
+    public function getDatetime(): DateTimeImmutable
     {
         return $this->datetime;
     }
 
+    public function getNow()
```
tatitati: type hint in return

src/Ds2013/Molecule/DateListItem/date_list_item.html.twig
votemike: How about putting the innards in a separate template to avoid duplication of those bits?

PR #30, opened 2017-05-17
Image presenter

felipeparaujo:
Image presenter, use it to show responsive images which could be lazy loaded through the 'is_lazy_loaded' option.


programmes-developers:

src/Ds2013/Molecule/Image/image.html.twig
```diff
@@ -0,0 +1,12 @@
+{% if image.getOption('is_lazy_loaded') %}
+    <div class="rsp-img lazy-img"
+        data-img-blank="{{ asset("images/blank.png") }}"
+        {{ image.responsiveAttributes|raw }}
+    ></div>
+{% else %}
+    <img src="{{ image.getUrl(256, 144) }}"
+        class="rsp-img"
+        alt=""
```
votemike: Is a blank alt useful?

tests/Ds2013/Molecule/Image/ImagePresenterTest.php
```diff
@@ -0,0 +1,59 @@
...
+        $image = $this->createMockImage();
+        $imagePresenter = new ImagePresenter($image);
+
+        $this->assertEquals(true, $imagePresenter->getOption('is_lazy_loaded'));
+    }
+
+    public function testGetUrl()
+    {
+        $image = $this->createMockImage();
+        $image->expects($this->once())->method('getUrl');
```
BPScott: Do we need to care about expecting that a particular method gets called?

tests/Ds2013/Molecule/Image/ImagePresenterTest.php
```diff
@@ -0,0 +1,59 @@
...
+{
+    public function testOptionDefaults()
+    {
+        $image = $this->createMockImage();
+        $imagePresenter = new ImagePresenter($image);
+
+        $this->assertEquals(true, $imagePresenter->getOption('is_lazy_loaded'));
+    }
+
+    public function testGetUrl()
```
BPScott: I reckon merge all these functions into a single test - you're not changing anything about the setup of the test and you're creating boilerplate. It's not too bad in a Presenter this simple but it'll start to look really unwieldly in the in bigger presenters like Programme.

votemike:
Is this making use of https://github.com/BBC-News/Imager.js/ ?

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -25,5 +25,3 @@
     </ol>
 
 {% endblock %}
-
```
votemike: 👍

src/Ds2013/Molecule/Image/ImagePresenter.php
```diff
@@ -10,30 +10,100 @@ class ImagePresenter extends Presenter
...
+     * @param array $options
+     */
+    public function __construct(Image $image, $sizes, array $options = [])
     {
         parent::__construct($options);
         $this->image = $image;
+        $this->sizes = $sizes;
+    }
+
+    public function getAlt(): ?string
```
BPScott: Alt isn't nullable. default is an empty string.

src/Ds2013/Molecule/Image/ImagePresenter.php
```diff
@@ -10,30 +10,100 @@ class ImagePresenter extends Presenter
...
+        }
+
+        if (is_numeric($this->getOption('srcsets')[0])) {
+            return $this->generateSrcUrl((int) $this->getOption('srcsets')[0]);
+        }
+
+        return null;
+    }
+
+    public function getSrcsets(): ?string
```
BPScott: not nullable, would be empty string.

src/Ds2013/Molecule/Image/image.html.twig
```diff
@@ -0,0 +1,15 @@
+{% if image.getOption('is_lazy_loaded') %}
+    <div class="rsp-img lazy-img"
```
BPScott: Do we want to add in the ratio-box stuff into this so we fix the sizing of the image? Also standardising the class names might be handy - they're in image.scss so the class names should start with image.

`image` and `image--lazy`?

src/Ds2013/Molecule/Image/ImagePresenter.php
```diff
@@ -0,0 +1,100 @@
...
+        'ratio' => null,
+        'alt' => '',
+        'src_width' => null,
+    ];
+
+    /** @var Image */
+    private $image;
+
+    /** @var array|string */
+    private $sizes;
```
votemike: I reckon accepting an array or string in the constructor is fine. But should we manipulate the parameter in the constructor so that a definitive type is stored on the object?

src/Ds2013/Molecule/Image/ImagePresenter.php
```diff
@@ -0,0 +1,100 @@
...
+    {
+        if ($this->getOption('src_width')) {
+            return $this->generateSrcUrl($this->getOption('src_width'));
+        }
+
+        if (is_numeric($this->getOption('srcsets')[0])) {
+            return $this->generateSrcUrl($this->getOption('srcsets')[0]);
+        }
+
+        return '';
```
votemike: Should you not throw an exception?

BPScott:
@votemike this doesn't use imager.js. The old codebase used our own js that dealt with responsive images that did something very similar to imager (it was developed about the same time) however the main thing those libs provided - responsive images - is now baked into browsers as srcset/sizes so we should leverage that rather than use js. We still need some js to do the "only load images when they get near to the viewport" bit of functionality though. 

The vast majority of images in /progs are fluff - they're a nice visual but dont add any value (e.g. the schedule page still tells you what you need if there is no image next to the programme info) thus in those cases showing a low-res image for browsers that don't support srcset is acceptable. Yay progressive enhancement etc.

src/Ds2013/Molecule/Image/ImagePresenter.php
BPScott: $sizes shall now always be a string as you build it in the constructor

src/Ds2013/Molecule/Image/ImagePresenter.php
```diff
@@ -0,0 +1,150 @@
...
+        foreach ($this->getOption('srcsets') as $srcset) {
+            $srcsets[] = $this->buildSrcUrl($srcset) . ' ' . $srcset . 'w';
+        }
+
+        return implode(', ', $srcsets);
+    }
+
+    protected function validateOptions(array $options): void
+    {
+        parent::validateOptions($options);
```
BPScott: The parent doesn't do anything, so no point calling it

src/Ds2013/Molecule/Image/ImagePresenter.php
```diff
@@ -0,0 +1,150 @@
...
+
+        // add the final 100vw in case nothing matched
+        $parts[] = '100vw';
+
+        return implode(', ', $parts);
+    }
+
+    private function buildSrcUrl(?int $width): string
+    {
+        if (is_numeric($this->getOption('ratio'))) {
```
BPScott: We've already validated that  it is numeric, so we just need an not null check

`if ($this->getOption('ratio'))`?

src/Ds2013/Molecule/Image/ImagePresenter.php
felipeparaujo: I put this in because in the future Presenter might have some logic for some option we want all presenters to have. You think that will never be the case?

BPScott: Options are per-presenter, I don't think there is ever a case for an option that would apply across ALL presenters, and thus there would never be a reason to validate.

PR #31, opened 2017-05-17
5511 Adding masthead

votemike:
This is pretty much ready to review, however I think we should wait for the date picker to be merged in first so that it can be incorporated in to this branch.


programmes-developers:

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -1,8 +1,30 @@
 {% extends 'base-ds2013.html.twig' %}
 
+{% block page_classes %}programmes-page text-base b-g-p programmes-page--flush{% endblock %}
+
 {% block body %}
     {% include 'schedules/components/nav.html.twig' %}
-    {% include 'schedules/components/regionpick.html.twig' with { 'servicesInNetwork': servicesInNetwork, 'service': service } only %}
+    {% include 'schedules/components/regionpick.html.twig' with { 'servicesInNetwork': servicesInNetwork, 'service': service, 'twinService': twinService, 'date': date } only %}
```
votemike: Do we need the `with`?

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -1,8 +1,30 @@
...
+
+    {% if not serviceIsTv and not groupedBroadcasts is empty %}
```
votemike: I have a feeling this logic is generic, but not sure which PR should be responsible for implementing it.

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -1,8 +1,30 @@
...
+        <div class="g-c-l">
+            <div class="schedules-jump island micro">
+                <span class="pull--left schedules-jump__title">{{ tr('skip_to') }}: </span>
+                <ul class="schedules-jump__links nav nav--piped">
+                    {#if (isset($this->data->has_on_air) && $this->data->has_on_air):?>#}
```
votemike: Should this be included? Or not? The ticket didn't mention.

app/Resources/views/schedules/components/nav.html.twig
```diff
@@ -0,0 +1,28 @@
...
+{% endif %}
+    <h1>
+        <span class="visually-hidden">
+            <span class="outlet">
+                {{ tr('schedule_title', {'%1':service.name}) }}
+            </span>
+        </span>
+        <time class="date">
+            {#@TODO The old code referenced the broadcast, do I need to do that?#}
+            {#@TODO Make this transloate $this->setBrandingId('br-00435'); http://sandbox.bbc.co.uk:8000/schedules/p00fzl7b/2017-05-16#}
```
votemike: How would we like to do this translation? It looks like it's reasonably simple with Carbon: http://carbon.nesbot.com/docs/#api-localization otherwise I think we'll have to re-implement something like this: https://github.com/bbc/programmes/blob/develop/webapp/php/sharedmodules/programmes/views/helpers/DateTime.php

BPScott: A date presenter perhaps? https://github.com/bbc/rmp-amen/tree/master/lib/src/RMP/Amen/Molecule/Date

votemike: Seems a bit like overkill.....

app/Resources/views/schedules/by_day.html.twig
tatitati: yes, as Twig doc says:

> Included templates have access to the variables of the active context.

Which can be dangerous.

app/Resources/views/schedules/by_day.html.twig
BPScott: Yep put this in

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -1,8 +1,30 @@
...
+    {% if not service_is_tv and not grouped_broadcasts is empty %}
+        <div class="g-c-l">
+            <div class="schedules-jump island micro">
+                <span class="pull--left schedules-jump__title">{{ tr('skip_to') }}: </span>
+                <ul class="schedules-jump__links nav nav--piped">
+                    {#if (isset($this->data->has_on_air) && $this->data->has_on_air):?>#}
+                    {#<li><a href="#on-now"><?=$this->tr('on_air')?></a></li>#}
+                    {#<?endif;#}
+                    {% for period_name, broadcast_in_period in grouped_broadcasts %}
+                        {% if not broadcast_in_period is empty %}
```
BPScott: {% if broadcast_in_period %}

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -1,8 +1,30 @@
 {% extends 'base_ds2013.html.twig' %}
 
+{% block page_classes %}programmes-page text-base b-g-p programmes-page--flush{% endblock %}
+
 {% block body %}
     {% include 'schedules/components/nav.html.twig' %}
-    {% include 'schedules/components/regionpick.html.twig' with { 'services_in_network': services_in_network, 'service': service } only %}
+    {% include 'schedules/components/regionpick.html.twig' with { 'services_in_network': services_in_network, 'service': service, 'twin_service': twin_service, 'date': date } only %}
+
+    {% if not service_is_tv and not grouped_broadcasts is empty %}
```
BPScott: This page shall only be rendered if there are grouped broadcasts so we don't need that part of the check

{% if not service_is_tv %}

app/Resources/views/schedules/components/nav.html.twig
```diff
@@ -0,0 +1,28 @@
+{% if service_is_tv %}
```
BPScott: Use a ternary to avoid this repitition

app/Resources/views/schedules/components/nav.html.twig
BPScott: Better plan might be to register put the localisation logic into rmp-translate, and expose a translate function / filter that calls that  (the same as how the tr() function works

BPScott: Stick with .format for now, and we can fix the localisation in a later PR

app/Resources/views/schedules/components/nav.html.twig
```diff
@@ -0,0 +1,28 @@
...
+    <div class="b-g-p no-margin-vertical islet--vertical br-box-highlight br-highlight-bg-onbg080" id="schedule-header">
+{% endif %}
+    <h1>
+        <span class="visually-hidden">
+            <span class="outlet">
+                {{ tr('schedule_title', {'%1':service.name}) }}
+            </span>
+        </span>
+        <time class="date">
+            {#@TODO The old code referenced the broadcast, do I need to do that?#}
```
BPScott: `@TODO deal with international services always showing UTC rather than using GMT/BST`

We need to come up with a good place to store things like `isInternational` which don't really belong in the Domain models, as they're view specific (see @BjornTwachtmann's plotting around custom mappers).

Lets do the simple UK case for now, and we can deal with world-service funkyness later.

app/Resources/views/schedules/components/nav.html.twig
```diff
@@ -0,0 +1,28 @@
...
+            {#@TODO Make this transloate $this->setBrandingId('br-00435'); http://sandbox.bbc.co.uk:8000/schedules/p00fzl7b/2017-05-16#}
+            {#@TODO Do we want 'l j F Y' or 'l jS F Y' ?#}
+            {{ date.format('l j F Y') }}
+        </time>
+    </h1>
+</div>
+<div class="schedule-nav-wrapper fixedsticky">
+    {% if service_is_tv %}
```
BPScott: Use a tertiary  {{ service_is_tv ? 'g-o-l daynav--tvschedule' : 'g-f-l' }}.
Use {% set %} to whack it into a variable if the line length gets too long

app/Resources/views/schedules/components/nav.html.twig
```diff
@@ -0,0 +1,28 @@
...
+        <div id="schedule-daynav" class="centi br-box-highlight g-o-l daynav--tvschedule">
+    {% else %}
+        <div id="schedule-daynav" class="centi br-box-highlight g-f-l">
+    {% endif %}
+        {#<?$this->partial('/partials/schedules/daynav.spv', $params);?>#}
```
BPScott: `{# @TODO day nav here #}`, rather than copying code from the old codebase that won't work when uncommented.

app/Resources/views/schedules/components/regionpick.html.twig
```diff
@@ -2,14 +2,18 @@
...
+                {{ tr('schedule_title', {'%1':service.name})}}
             </p>
         </div>
 
          <div class="grid 1/2@bpw2 1/2@bpe">
             <div class="bpw-text--right">
-                {% set translationKey = service.getNetwork().getMedium() == 'tv' ? 'schedules_regional' : 'schedules_regional_change' %}
-                <a href="{{ path('schedules_home') }}" class="delta">{{ tr(translationKey, {'%1':service.getNetwork().name }) }}</a>
+                {% if twin_service is empty %}
+                    {% set translationKey = service.getNetwork().getMedium() == 'tv' ? 'schedules_regional' : 'schedules_regional_change' %}
```
BPScott: Use `service_is_tv` rather than comparing again?

app/Resources/views/schedules/components/regionpick.html.twig
```diff
@@ -2,14 +2,18 @@
...
+                    <a href="{{ path('schedules_home') }}" class="delta">{{ tr(translationKey, {'%1':service.getNetwork().name }) }}</a>
```
BPScott: Use the shornames for getters `service.network.name`

src/Ds2013/_ds2013-components.scss
```diff
@@ -48,7 +48,7 @@
 // Core objects
 // @import 'objects/media';
 // @import 'objects/link-complex';
-// @import 'objects/nav';
+@import 'Molecule/nav';
```
BPScott: I think nav is an Atom rather than a Molecule - it's scss only, it has no presenter.

PR #32, opened 2017-05-17
Presenters template paths and variables should be snake_case

BPScott:
The convention for Twig is that template names and variables should be
snake case. This makes it so the Presenters adhere to that rule.


programmes-developers:

PR #36, opened 2017-05-17
[schedules] - home page

tatitati:


programmes-developers:

src/Controller/SchedulesHomeController.php
```diff
@@ -2,12 +2,204 @@
...
+use BBC\ProgrammesPagesService\Domain\Entity\Service;
+use BBC\ProgrammesPagesService\Service\ServicesService;
+use DateTimeImmutable;
+
 class SchedulesHomeController extends BaseController
 {
-    public function __invoke()
+    public function __invoke(ServicesService $servicesService)
     {
+        $this->setBrandingId('br-08368');
```
votemike: What's this? Should this be here?

src/Controller/SchedulesHomeController.php
```diff
@@ -2,12 +2,204 @@
...
 namespace App\Controller;
 
+use BBC\ProgrammesPagesService\Domain\Entity\Service;
+use BBC\ProgrammesPagesService\Service\ServicesService;
+use DateTimeImmutable;
+
 class SchedulesHomeController extends BaseController
 {
-    public function __invoke()
+    public function __invoke(ServicesService $servicesService)
```
votemike: This is a massive method, it should be made more readable by splitting up in to some private methods.

src/Controller/SchedulesHomeController.php
```diff
@@ -2,12 +2,204 @@
...
+            $endDate
+        );
+
+        $result['offset'] = $diffFromStart->days * $pointsPerDay;
+        $result['width'] = $diffForLength->days * $pointsPerDay;
+
+        return $result;
+    }
+
+    private function getDecade(DateTimeImmutable $datetime): float
```
votemike: How long is this system going to be around? Earlier code suggests that the earliest date is in 1922. Which is the 20's. But in 30 months or so, we'll be back in the 20's.

src/Ds2013/Core/Gel/Objects/_link-complex.scss
```diff
@@ -0,0 +1,47 @@
+/**
+ * Link-Complex
+ *
+ * A link where only some of it's children should have underline styles
+ */
+
+// Styling underline using ".programmes-page a" means we need increased specificity
+.link-complex,
+.programmes-page a.link-complex {
+    &,
```
votemike: What does `&` by itself do?

src/Ds2013/Organism/_schedules.scss
```diff
@@ -0,0 +1,83 @@
+.schedules-main {
+    font-size: 1.6em;
+}
+.schedules-main * {
+    box-sizing: border-box;
+}
+.schedules-main img {
```
votemike: Could this be nested inside the `.schedules-main` block?

src/Controller/SchedulesHomeController.php
tatitati: yeah, I know, is forcing it. Branding stuff is not ready yet, so in the home schedules page we are setting the branding that should appear. At some point this will be removed I guess.

src/Controller/SchedulesHomeController.php
```diff
@@ -2,12 +2,212 @@
...
+        $year = (int) $datetime->format('Y');
+        return floor($year / 10) * 10;
+    }
+
+    private function assignServiceToGroup(Service $service): string
+    {
+        $network = $service->getNetwork();
+        $end = $service->getEndDate();
+        if ($end) {
+            $groupId = 'historical';
```
votemike: Instead of assigning to `$groupId`, you could just `return`.

src/Controller/SchedulesHomeController.php
```diff
@@ -2,12 +2,212 @@
...
+
+    private function getDecade(DateTimeImmutable $datetime): float
+    {
+        $year = (int) $datetime->format('Y');
+        return floor($year / 10) * 10;
+    }
+
+    private function assignServiceToGroup(Service $service): string
+    {
+        $network = $service->getNetwork();
```
votemike: This is only used in the else statement. Maybe move it in to that?

tatitati: is also used here (outside the if)
$nid = (string) $network->getNid();

votemike: Where? I don't see it.

src/Controller/SchedulesHomeController.php
```diff
@@ -2,12 +2,205 @@
...
+        $end = $service->getEndDate();
+        if ($end) {
+            return 'historical';
+        } else {
```
votemike: You are returning in the `if` statement. So you don't need the `else`.

src/Controller/SchedulesHomeController.php
```diff
@@ -2,12 +2,205 @@
...
+    private function getDecade(DateTimeImmutable $datetime): float
+    {
+        $year = (int) $datetime->format('Y');
+        return floor($year / 10) * 10;
+    }
+
+    private function assignServiceToGroup(Service $service): string
+    {
+        $network = $service->getNetwork();
+        $end = $service->getEndDate();
```
votemike: If the service has an end date, it's historical? Surely the end date has to be in the past?

src/Controller/SchedulesHomeController.php
tatitati: https://github.com/bbc/programmes-frontend/pull/36/files/ab03f6f4814f641e6f5d87b14d58d76c4021b093#diff-53a4aeac8db95418191f9d96db1c1781R78

votemike: That's not in this method.....

src/Controller/SchedulesHomeController.php
BPScott: The schedules home page should use the Schedule branding. There's no external place to get it from, it should be hardcoded in the controller (which it is here)

src/Controller/SchedulesHomeController.php
BPScott: Love me some early returns for flow control

src/Controller/SchedulesHomeController.php
BPScott: There are pop-up services that appear, broadcast for a few weeks and then disapear. Their start/end dates are explicitly known even before they've started airing. Examples are "radio 5 live olympics extra", which was just on during the 2012 olympics and "radio two 50s" which was a "pop-up" themed station on DAB that lasted for about a week.

votemike: So, does that mean this is wrong? or right?

votemike:
What is this page for? It seems odd.... There is some functionality that doesn't seem very well thought out.  
Also, the styling seems off in places.  
It also feels like there should be navigation to the various groups.

tatitati:
@votemike the ticket is:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5502

I agree, there is some differences in the style of the code because @BPScott made this part like a mini-project a time ago, in a hack session. We took that project because everything was already done. Repo here:
https://github.com/bbc/schedules

votemike:
This seems like new functionality which I cannot compare with the current site. So I would like more information in the ticket if possible. I have **many** questions about this...

app/Resources/views/schedules/home.html.twig
```diff
@@ -1,5 +1,67 @@
 {% extends 'base_ds2013.html.twig' %}
 
+{% block title %}BBC - {{tr('schedules')}}{% endblock %}
+
 {% block body %}
-Schedules Home
+    <div class="schedules-main">
+        <div class="front-page">
+            {% for group in groups %}
```
votemike: There is a lot of logic inside this for loop. More `for`s, `if`s and `else`s. Would it be possible to make this cleaner? Possibly using macros?

https://twig.sensiolabs.org/doc/2.x/tags/macro.html

https://github.com/bbc/programmes-frontend/blob/bbe2ae39604f83d9ce690893789df47447df5aad/src/Ds2013/Molecule/DateList/date_list_item.html.twig#L17

BPScott:
@votemike yep this is new functionality compared to the current site - we don't have a schedules homepage at the moment (we've got a list of schedules on the /programmes homepage, but that's not /schedules). PROGRAMMES-5502 was supposedly building up the skeleton of this page, but we've dug out an old hackday project by me and @djmarland that mocked up an idea for a schedule homepage. As it was done in a day the original is very rough and ready and needs lots of polish.

djmarland:
Hackday code? Spit and polish. Spit and polish.

votemike:
@BPScott I see. How should I go about reviewing this then? Just check the code is OK and the page looks acceptable in the browser?

src/Controller/SchedulesHomeController.php
```diff
@@ -2,12 +2,205 @@
...
+            'title' => 'Other',
+            'networks' => [],
+        ],
+        'historical' => [
+            'title' => 'Historical',
+            'networks' => [],
+        ],
+    ];
+
+    public function __invoke(ServicesService $servicesService)
```
felipeparaujo: This function is very long, could you extract some more functionality to few other functions?

src/Controller/SchedulesHomeController.php
```diff
@@ -2,12 +2,205 @@
...
+                    'services' => [],
+                ];
+            }
+            $networkGroups[$nid]['services'][] = $service;
+        }
+
+        // get earliest broadcast
+        $earliestBroadcastDate = new DateTimeImmutable('1922-01-01T12:00:00Z');
+        // get latest broadcast
+        $latestBroadcastDate = new DateTimeImmutable('2016-12-01T12:00:00Z');
```
felipeparaujo: Why is latest broadcast date hardcoded here?

votemike:
If you rebase with master and run `script/test` I think you'll get some warnings.

BPScott:
This is still WIP, so don't bother review yet please.

programmes-developers:

BPScott:
This is probably ready for review again. I've managed to kill off a bunch of stuff that was unrelated to this PR / was dead code / wasn't used in views.

src/Controller/SchedulesHomeController.php
```diff
@@ -2,12 +2,111 @@
...
+            'Regional Radio' => [],
+            'Local Radio' => [],
+            'Other' => [],
+            'Historical' => [],
+        ];
+
+        $services = $servicesService->getAllInNetworks();
+
+        $earliestBroadcastDate = new DateTimeImmutable('1922-01-01T12:00:00Z');
+        $latestBroadcastDate = new DateTimeImmutable('2019-12-01T12:00:00Z');
```
felipeparaujo: Why are we hardcoding latestBroadcastDate?

BPScott: Good shout. This is kinda screwy as it doesn't work out the date sizing properly anyway. Pushed a fix to make this based on what the current decade is.

app/Resources/views/schedules/home.html.twig
```diff
@@ -1,5 +1,69 @@
...
+                            </div>
+                        {% endfor %}
+                    {% endif %}
+                {% endfor %}
+            </div>
+        {% endfor %}
+    </div>
 {% endblock %}
+
+{% macro decades_bar(decades, decadePercent) %}
```
votemike: Does `decadePercent` need to be passed in? Can't it just find it from the controller?

app/Resources/views/schedules/home.html.twig
```diff
@@ -1,5 +1,69 @@
...
+                        {#NETWORK HAS ONLY ONE SERVICE, SO NETWORK LINK POINT TO THIS SERVICE INSTEAD OF LISTING THEM#}
+                        <div class="sch-service-row">
+                            <a class="deemphasize gamma grid-wrapper" href="{{ path('schedules_by_day', { pid : service.service.pid }) }}">
+                                {{ self.service_name(service.service.name, "gamma sch-network-name") }}
+                                {{ self.service_availability(service) }}
+                            </a>
+                        </div>
+                    {% else %}
+                        <div class="gamma sch-network-name">{{ network.network.name }}</div>
+                        {% for service in network.services %}
```
votemike: This for loop is inside 2 **other** for's. Should a bit of this view be refactored into a macro?

src/Controller/SchedulesHomeController.php
```diff
@@ -2,12 +2,123 @@
...
+        $diffForLength = $service->getStartDate()->diff($endDate);
+        $result['offset'] = $diffFromStart->days * $pointsPerDay;
+        $result['width'] = $diffForLength->days * $pointsPerDay;
+
+        return $result;
+    }
+
+    private function groupKeyForService(Service $service): string
+    {
+        if ($service->getEndDate()) {
```
votemike: What if the end date hasn't happened yet?

src/Controller/SchedulesHomeController.php
```diff
@@ -2,12 +2,123 @@
...
+            return $type;
+        }
+
+        return 'Other';
+    }
+
+    private function getDecade(DateTimeImmutable $datetime): float
+    {
+        $year = (int) $datetime->format('Y');
+        return floor($year / 10) * 10;
```
votemike: Not sure if it's "more" correct, but this could be `intdiv($year, 10) * 10` couldn't it?

src/Controller/SchedulesHomeController.php
```diff
@@ -2,12 +2,123 @@
...
+    }
+
+    private function getFinalMomentInCurrentDecade(): DateTimeImmutable
```
votemike: Just thought I'd mention that Carbon has an `endOfDecade` method..... 😝   
#allHailCarbon

src/Controller/SchedulesHomeController.php
BPScott: Then it goes into whichever bucket, as defined by the network types below

votemike: Does the endDate only get set once the service has ended? Otherwise it'll go in the 'Historical' bucket....

app/Resources/views/schedules/home.html.twig
BPScott: It could but I quite like the idea being explicit with variable dependencies rather than pulling in  global scope.

djmarland: And yet, this method relies on global scope rather than receiving the data as a parameter (https://github.com/bbc/programmes-frontend/blob/programmeitem-presenter/src/Controller/SchedulesHomeController.php#L119-L123) ;-)

(didn't expect the conversation to continue for so long after you tagged me. I should probably un-subscribe, but its fun to watch)

BPScott: ApplicationTime is a special case and you know it you dang contrarian ;)

djmarland: I meant it wasn't passed in from the main part of the controller, which would make it a pure function.
Anyway, I look forward seeing how this all manifests itself into a nice new page. :-)

PR #37, opened 2017-05-17
[schedules] - broadcast

tatitati:
- Renamed function in twig template for broadcast
- Generation of urls by using PATH twig function
- Unit tests


programmes-developers:

src/Ds2013/Organism/Broadcast/BroadcastPresenter.php
```diff
@@ -82,6 +87,13 @@ public function getObjectClassesCss(): string
         ]);
     }
 
+    public function getProgrammeCssClasses()
```
BPScott: the class bit feels superflous. We should make sure we have a convention for naming these css functions.  `CssClasses` or `ClassesCss` feel like they're repeating the same idea, how about just ending the method name with `Classes?` - `getProgrammeClasses` and `getInfoClasses`?

`infoClassesCss` should be updated too.

src/Ds2013/Organism/Broadcast/BroadcastPresenter.php
```diff
@@ -38,6 +38,7 @@ public function getServiceName(): string
 
     public function getStartAt(): DateTimeImmutable
     {
+        // Todo: in programmes we show the timezone for radio.
```
BPScott: No need for this todo, this shall be fixed in the template rather than here.

src/Ds2013/Organism/Broadcast/broadcast.html.twig
```diff
@@ -1,29 +1,32 @@
...
 
-                {% if broadcast.isOnAirNow() %}
+                {% if broadcast.isOnAirNow %}
                     <div class="broadcast__live centi br-box-highlight">
-                        {{ broadcast.getNetworkMedium() == 'tv' ? tr('on_now') : tr('on_air') }}
+                        {{ broadcast.networkMedium == 'tv' ? tr('on_now') : tr('on_air') }}
                     </div>
                 {% endif %}
+        </div>
+        <div class="broadcast__programme grid {{ broadcast.programmeCssClasses }}" rev="publication" >
```
BPScott: trailing space?

tests/Ds2013/Organism/Broadcast/BroadcastPresenterTest.php
```diff
@@ -0,0 +1,103 @@
...
+use BBC\ProgrammesPagesService\Domain\ValueObject\Pid;
+
+class BroadcastPresenterTest extends TestCase
+{
+    /** @var  DateTimeImmutable */
+    private $now;
+
+    public function setUp()
+    {
+        ApplicationTime::setTime(1495094745);//2017-05-18 08:05:45
```
BPScott: `new DateTimeImmutable('2017-05-18 08:05:45')->format('U')` might be more readable

tests/Ds2013/Organism/Broadcast/BroadcastPresenterTest.php
```diff
@@ -0,0 +1,103 @@
...
+
+        $this->now = ApplicationTime::getTime();
+    }
+
+    public function testGetObjectClassesCssWhenOnPast()
+    {
+
+        $start = $this->now->modify('-2 hour');
```
BPScott: Would it be easier if we use explicit dates rather than defining offsets from the existing one?

PR #38, opened 2017-05-17
[schedules] - programme

tatitati:


programmes-developers:

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -28,23 +73,243 @@ public function __construct(
...
     }
 
     public function getSynopsis(): string
     {
-        return $this->programme->getShortSynopsis();
+        $truncateLength = ($this->getOption('truncate')) ? $this->getOption('truncate') * 1.5 : null;
+        return $this->truncateText($this->programme->getShortSynopsis(), intval($truncateLength));
+    }
+
+    public function getImage($x, $y)
```
votemike: Typehints? And maybe more explanatory variable names?

src/Ds2013/Organism/Programme/_programme.scss
```diff
@@ -0,0 +1,433 @@
...
+    text-align: left;
+}
+.programme__overlay--live,
+.programme__overlay--available {
+    min-width: 92.857142857142857%;
+}
+.programme__availability {
+    display: block;
+}
+.programme__availability:hover,
```
votemike: Can these be nested in the block above?

src/Ds2013/Organism/Programme/programme.html.twig
```diff
@@ -1,8 +1,145 @@
-<div class="programme">
-    <h3>{{programme.title}}</h3>
-    {%- if programme.showSynopsis -%}
-        <p>{{programme.synopsis}}</p>
+<div class="{{programme.objectClasses}}" typeof="{{programme.schemaType}}" resource="{{ path('find_by_pid', {'pid': programme.pid}) }}" {{ programme.htmlAttributes }}>
+
+
+    {%- if programme.option('show_image') or ( programme.option('stack_overlay') and programme.option('show_overlay')) -%}
```
votemike: You are mixing an `or` and an `and`. Should you include some parentheses to ensure the correct thing is evaluated first?

src/Ds2013/Organism/Programme/programme.html.twig
```diff
@@ -1,8 +1,145 @@
...
+        <div class="{{ programme.programmeImgClasses }}">
+            {#IMAGE#}
+
+            {#OVERLAY#}
+            {%- if programme.option('show_overlay') -%}
+                <div class="{{ programme.programmeOverlayCss }}">
+                    {%- if programme.isContainer == false -%}
+                        <meta property="description" content="{{ programme.synopsis }}" />
```
votemike: Is meta a real tag?

src/Ds2013/Presenter.php
```diff
@@ -93,6 +96,27 @@ protected function buildCssClasses(array $cssClassTests = []): string
...
+    {
+        $attributes = [];
+        foreach ($htmlAttributes as $attribute => $value) {
+            $attributes[] = $attribute . '="' . htmlentities($value, ENT_COMPAT, 'UTF-8') . '"';
+        }
+
+        return trim(implode(' ', $attributes));
+    }
+
+    protected function truncateText(string $text, ?int $length, $suffix = '...'): string
```
votemike: https://github.com/twigphp/Twig-extensions/blob/master/doc/text.rst ???

src/Ds2013/Organism/Programme/programme.html.twig
tatitati: The most famous is this one:

> <head>
>   <meta charset="UTF-8">

votemike: Of course. Silly me.  
Never seen it used inside the `<body>` thought.

src/Ds2013/Presenter.php
tatitati: I disagree. Our implementation is not only truncating, is also calculating the length in case we don't indicate it and doing some more stuff. But we can discuss if we can simplify this function anyway in the future. For now I would like to keep the original behaviour to reduce differences during the migration

src/Ds2013/Organism/Programme/programme.html.twig
```diff
@@ -1,8 +1,145 @@
...
+                        <meta property="thumbnailUrl" content="{{ programme.image(1920,1080) }}" />
```
votemike: Is `thumbnailUrl` a valid property? Looks like only a few are valid. https://www.w3schools.com/tags/att_meta_name.asp

tatitati: yeah....

src/Ds2013/Organism/Programme/programme.html.twig
BPScott: It's used for specify RDFA schema information for our pages.

votemike: Then it should be in the <head> right?
> The <meta> element declaring the encoding must be inside the <head> element and within the first 1024 bytes of the HTML, as some browsers only look at those bytes before choosing an encoding.

From https://developer.mozilla.org/en/docs/Web/HTML/Element/meta

src/Ds2013/Presenter.php
BPScott: This should be a twig extension imo rather than something intrinsic to a presenter

votemike:
How can I test this?

src/Ds2013/Organism/Programme/programme.html.twig
BPScott: That's the meta element describing the encoding of the document. This meta element is defining the description property of the RDF for the programme that is this element (as started by the div with `typeof` as an attribute). On a single page there will be many programme objects.

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -28,23 +73,242 @@ public function __construct(
...
+    }
+
+    public function isContainer(): bool
+    {
+        return $this->programme instanceof ProgrammeContainer;
+    }
+
+    public function isAvailable(): bool
+    {
+        // todo: dependency of Nitro
```
BPScott: Nope, check `isStreamable().

src/Ds2013/Organism/Programme/programme.html.twig
BPScott: http://schema.org/Episode defines the valid properties for an RDFa Episode. See getSchemaFromProgramme for the other schema types we have beside episode. all have thumbnailUrl

src/Ds2013/Organism/Programme/programme.html.twig
votemike: Is there some documentation I can read on how meta tags work in this way? I can only find info to do with meta tags in `<head>`

votemike:
Should the whole programme item be clickable rather than just the title?

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -4,23 +4,25 @@
...
-    </ol>
+    <div class="b-g-p br-box-page programmes-page">
+        <ol class="list-unstyled g-c-l">
+            {% for period_name, broadcast_in_period in grouped_broadcasts %}
+                <li id="{{ period_name }}" class="schedule-skip-link">
+                    <h2 class="delta text--shout island--horizontal component">{{ period_name }}</h2>
+                    <ol class="highlight-box-wrapper">
+                        {#list of broadcasts in day period#}
+                        {% for broadcast in broadcast_in_period %}
+                            <li {{ broadcast == on_air_broadcast ? 'id=on-now class=schedule-skip-link-on-now' : null }} >
```
votemike: Should this instead be `isOnAir()`?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -28,23 +74,248 @@ public function __construct(
         $this->programme = $programme;
     }
 
-    public function getTitle(): string
+    public function getTitle(): array
```
votemike: I'm still unclear as to what a title is.

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -28,23 +74,248 @@ public function __construct(
...
+        if ($this->getOption('bottom_up') || $this->getTypeFromProgramme($this->programme) == self::CLIP) {
+            array_pop($stackedTitles);
+            return $stackedTitles;
+        }
+
+        array_shift($stackedTitles);
+        return $stackedTitles;
+    }
+
+    public function getNumberCurrentEpisode(): ?int
```
votemike: `getNumberOfCurrentEpisode`?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -28,23 +74,248 @@ public function __construct(
...
     {
-        return $this->programme->getTitle();
+        return $this->programme->getPosition();
+    }
+
+    public function getAmountEpisodes(): ?int
```
votemike: `getAmountOfEpisodes`?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -28,23 +74,248 @@ public function __construct(
...
     public function getSynopsis(): string
     {
         return $this->programme->getShortSynopsis();
     }
 
-    public function showSynopsis(): bool
+    public function getImageUrl(int $xSize, int $ySize): ?string
+    {
+        if ($this->programme->getImage()) {
+            return $this->programme->getImage()->getUrl($xSize, $ySize);
```
votemike: Do we want to run `$this->programme->getImage()` twice? Would it be better to save it in to a variable?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -28,23 +74,248 @@ public function __construct(
...
+        }
+
+        return null;
+    }
+
+    public function getImage(): ?Image
+    {
+        if ($this->programme->getImage()) {
```
votemike: Do we want to run `$this->programme->getImage()` twice? Would it be better to save it in to a variable?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -28,23 +74,248 @@ public function __construct(
...
+
+    public function getSchemaType(): string
+    {
+        return $this->getSchemaFromProgramme($this->programme);
+    }
+
+    public function getObjectClasses(): string
+    {
+        $typeProgramme = $this->getTypeFromProgramme($this->programme);
+        $isClip = $this->getTypeFromProgramme($this->programme) == self::CLIP;
```
votemike: `$isClip = $typeProgramme == self::CLIP;`

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -28,23 +74,248 @@ public function __construct(
...
+    {
+        return $this->programme instanceof ProgrammeContainer;
+    }
+
+    public function isAvailable(): bool
+    {
+        return $this->programme->isStreamable();
+    }
+
+    public function getMediaIcon(): string
```
votemike: I'm sure you could simplify this by returning early.

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -28,23 +74,248 @@ public function __construct(
...
+        return $this->buildCssClasses([
+              'programme__overlay' => true,
+              'programme__overlay--available' => true, //(!$has_live_broadcast && $is_available),
+              'programme__overlay--live' => true, //$simulcast_url
+        ]);
+    }
+
+    public function getTypeFromProgramme(Programme $programme): string
+    {
+        switch ($programme) {
```
votemike: What if the `$programme` is none of the options?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -28,23 +74,248 @@ public function __construct(
...
+        }
+
+        // sort from container to item
+        return array_reverse($stackedTitles);
+    }
+
+    private function getNetworkMedium(): ?string
+    {
+        if ($this->programme->getNetwork()) {
+            return $this->programme->getNetwork()->getMedium();
```
votemike: Do you want to run `$this->programme->getNetwork()` twice?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -28,23 +74,248 @@ public function __construct(
...
+        }
+
+        return null;
+    }
+
+    private function getSchemaFromProgramme(Programme $programme): string
+    {
+        $type = $this->getTypeFromProgramme($programme);
+
+        switch ($type) {
```
votemike: What if the `$type` is none of these options?

src/Ds2013/Organism/Programme/programme.html.twig
```diff
@@ -1,8 +1,152 @@
...
+
+                {#BROADCAST EVENT#}
+                {#TO-DO: broadcasts events will be implemented in the future. Dependency of programme with broadcasts.#}
+
+                {#SIMULCAST#}
+                {#TO-DO: Dependency of programme with broadcasts.#}
+
+                {#AVAILABILITY#}
+                {%- if programme.option('show_child_availability') -%}
+                    {%- if programme.isProgrammeTleo == false or programme.isAvailable == false -%}
```
votemike: Can this `if` be combined with the one above?

src/Twig/DesignSystemPresenterExtension.php
```diff
@@ -60,6 +61,19 @@ public function trWrapper(
         return $this->tr($key, $substitutions, $numPlurals, $domain);
     }
 
+    public function truncateText(string $text, ?int $length = null, string $suffix = '...')
```
votemike: Can you not use a twig filter for this?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -2,17 +2,57 @@
 declare(strict_types = 1);
 namespace App\Ds2013\Organism\Programme;
 
+use App\Ds2013\Exception\UnknownTypeException;
 use App\Ds2013\Presenter;
 use App\Ds2013\TranslatableTrait;
-use BBC\ProgrammesPagesService\Domain\Entity\Programme;
```
votemike: Any reason for moving this out of alphabetical order?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -28,23 +68,255 @@ public function __construct(
...
+            case ($programme instanceof Clip):
+                return self::CLIP;
+            default:
+                throw new UnknownTypeException('Expected a programme but received othery type.');
+        }
+    }
+
+    private function getRecursiveTitles(Programme $programme, array $stackedTitles = []): array
+    {
+        $stackedTitles[] = [
```
votemike: @BPScott Should these be `Title` objects instead of this generic array of which people have to guess the contents?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -28,23 +68,255 @@ public function __construct(
...
+        if ($this->getOption('bottom_up') || $this->getTypeFromProgramme($this->programme) == self::CLIP) {
+            array_pop($stackedTitles);
+            return $stackedTitles;
+        }
+
+        array_shift($stackedTitles);
+        return $stackedTitles;
+    }
+
+    public function getNumberCurrentOfEpisode(): ?int
```
votemike: `getNumberOfCurrentEpisodes`

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -28,23 +68,255 @@ public function __construct(
...
+    }
+
+    public function isAvailable(): bool
+    {
+        return $this->programme->isStreamable();
+    }
+
+    public function getMediaIcon(): string
+    {
+        $programmeMediaType = $this->programme->getMediaType();
```
votemike: Could this be moved below the first `if` statement?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -28,23 +68,255 @@ public function __construct(
...
+        return $stackedTitles;
+    }
+
+    public function getNumberCurrentOfEpisode(): ?int
     {
-        return $this->programme->getTitle();
+        return $this->programme->getPosition();
+    }
+
+    public function getAmountOfEpisodes(): ?int
```
felipeparaujo: Maybe we should standardise our count functions? I think this should be called `getEpisodeCount`. @BPScott @votemike ? Also, shouldn't it return 0 when there's no parent?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -28,23 +68,255 @@ public function __construct(
...
+        if ($img) {
+            return $img->getUrl($xSize, $ySize);
+        }
+
+        return null;
+    }
+
+    public function getImage(): ?Image
+    {
+        $img = $this->programme->getImage();
```
felipeparaujo: Is the check necessary? just return `$this->programmes->getImage()`

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -28,23 +68,255 @@ public function __construct(
...
+            case ($programme instanceof Series):
+                return self::SERIES;
+            case ($programme instanceof Brand):
+                return self::BRAND;
+            case ($programme instanceof Episode):
+                return self::EPISODE;
+            case ($programme instanceof Clip):
+                return self::CLIP;
+            default:
+                throw new UnknownTypeException('Expected a programme but received othery type.');
```
felipeparaujo: Typo in `othery`

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -28,23 +68,255 @@ public function __construct(
...
+    public function programmeOverlayCss(): string
+    {
+        return $this->buildCssClasses([
+              'programme__overlay' => true,
+              'programme__overlay--available' => true, //(!$has_live_broadcast && $is_available),
+              'programme__overlay--live' => true, //$simulcast_url
+        ]);
+    }
+
+    public function getTypeFromProgramme(Programme $programme): string
```
felipeparaujo: `getProgrammeType`?

src/Ds2013/Organism/Programme/ProgrammePresenter.php
```diff
@@ -28,23 +68,255 @@ public function __construct(
...
+            case ($programme instanceof Episode):
+                return self::EPISODE;
+            case ($programme instanceof Clip):
+                return self::CLIP;
+            default:
+                throw new UnknownTypeException('Expected a programme but received othery type.');
+        }
+    }
+
+    private function getRecursiveTitles(Programme $programme, array $stackedTitles = []): array
```
felipeparaujo: Does this function have to be recursive? It's probably not very hard to do this in an iterative fashion.

PR #39, opened 2017-05-19
5577 | schedule error states 

zoranikolic:


programmes-developers:

votemike:
Should the page also be shown if the service isn't found? Or is the exception there fine?

zoranikolic:
@votemike 404/500 service error pages have been moved to new ticket for next release

votemike:
Does the link to the start/end date need to be styled?

votemike:
Also, what happens if `$servicesInNetwork` is falsey?

app/Resources/views/schedules/no_schedule.html.twig
```diff
@@ -0,0 +1,15 @@
+{% extends 'base_ds2013.html.twig' %}
+
+{% block body %}
+
+    <p class="noschedule">
+        {% if date < service.startDate %}
```
votemike: What is a service doesn't have a startDate?

app/Resources/views/schedules/no_schedule.html.twig
```diff
@@ -0,0 +1,15 @@
...
+            {{ tr('schedule_begins_on', {'%1':'<a href="' ~ path('schedules_by_day', {'pid': service.pid, 'date': service.startDate|date('Y-m-d')}) ~ '">' ~ service.startDate|date("l j F Y") ~ '</a>'})|raw }}
+        {% elseif date > service.endDate %}
```
votemike: What is a service doesn't have a endDate?

app/Resources/views/schedules/no_schedule.html.twig
BPScott: If a service does not have a start date it should be considered available, so `date and date < service.startDate`

app/Resources/views/schedules/no_schedule.html.twig
BPScott: If a service does not have a end date it should be considered available, so `date and date > service.endDate`

PR #40, opened 2017-05-19
Fix gulp-sass compilation behaviour

BPScott:
Previously the gulp build would succeed even if there was an error in
sass compilation. This makes sure we exit with a non-zero error code if
there is a sass compilation error. Running watch shall output the error
messages but won't interupt the watch if there was an error.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -2,6 +2,7 @@
 
 const gulp = require('gulp');
 const sass = require('gulp-sass');
+const { gulpSassError } = require('gulp-sass-error');
 const sourcemaps = require('gulp-sourcemaps');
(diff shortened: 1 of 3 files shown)


programmes-developers:

votemike:
Does this definitely fail on fail? Could you require a non-existent SASS file to make sure the build fails?

BPScott:
@votemike: I've ran gulp locally with broken sass and asserted it exits with a non-zero error code. I've also pushed a broken commit to this branch and asserted that the build failed. I've since rebased that commit out of existence, but the build for it was https://ci.rmp.tools.bbc.co.uk/job/pull-request-programmes-frontend/263.

PR #41, opened 2017-05-19
Refactoring schedule templates

votemike:


programmes-developers:

votemike:
It has been found that the top nav is incorrect at the moment. The fixes are here: https://github.com/bbc/programmes-frontend/pull/42

PR #42, opened 2017-05-22
Fixing top nav

votemike:


programmes-developers:

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -1,50 +1,52 @@
 {% extends 'base_ds2013.html.twig' %}
 
-{% block page_classes %}programmes-page text-base b-g-p programmes-page--flush{% endblock %}
+{% block page_classes %}programmes-page text-base programmes-page--flush{% endblock %}
 
 {% block body %}
     {% include 'schedules/components/nav.html.twig' %}
-    {% include 'schedules/components/regionpick.html.twig' with { 'services_in_network': services_in_network, 'service': service, 'service_is_tv': service_is_tv, 'twin_service': twin_service, 'date': date } only %}
+    <div class="b-g-p br-box-page programmes-page">
```
votemike: This element has the "programmes-page" class on it as it does in the live site, however this class is already a class in it's parent in `{% block page_classes %}` which doesn't seem correct. Should this be changed? What to?

PR #43, opened 2017-05-22
Adding CS for Nesting Levels

votemike:
I have noticed some PRs with large functions and deeply nested logic. Rather than lots of comments asking for these to be refactored, I wonder what people think of adding this code sniffer. It should reduce the number of comments about refactoring methods and should also make for more readable code. Please reject if you don't agree.


programmes-developers:

Tasiobg:
I like to keep a low nesting lever by separating bits of logic into different functions when possible so you can understand what is going on at a glance but I'm not 100% sure of the impact of enforcing anything so I can't support or oppose to enforcing.

I support the warning message but not sure about enforcing.

I'm interested in knowing what the other team members thing.

felipeparaujo:
I agree with @Tasiobg, a warning is very useful, but enforcing might be a bit overreaching.

felipeparaujo:
Also, before you merge this, make sure that showing a warning doesn't make the build process fail. Sometimes jenkins interprets warnings as errors.

src/Ds2013/Molecule/Image/ImagePresenter.php
```diff
@@ -83,11 +83,7 @@ protected function validateOptions(array $options): void
         if (!is_array($options['srcsets'])) {
             throw new InvalidOptionException("Option 'srcsets' must be an array");
         } else {
-            foreach ($options['srcsets'] as $srcset) {
```
BPScott: Better refactoring here would be getting rid of the else clause as the throw acts as an early return

```
if (!is_array($options['srcsets'])) {
    throw new InvalidOptionException("Option 'srcsets' must be an array");
}

foreach ($options['srcsets'] as $srcset) {
    // ...
}

votemike:
It seems that just a warning makes Jenkins fail.... Does that mean people don't want this sniff?

felipeparaujo:
@votemike no, only that we'll have to update ci-docker. I believe that adding the -n argument when running phpcs would fix that. See here: https://manned.org/phpcs/a11ee962#head5

Tasiobg:
In case someone is not aware, you can enable "Php code sniffer" and "Mess detector" within IntelliJ so it will warn you as you type. I found it a good practice

BjornTwachtmann:
If you want to emit a warning on this, fine. It absolutely should not break the build. To be honest code sniffers with too many opinions about this kind of thing just piss me off. If I indent code deeply it's because I had a reason to...

Tasiobg:
@BjornTwachtmann @votemike @felipeparaujo @BPScott 
This is an example of mess detector and code sniffer, those warning are quite soft and well integrated on IntelliJ.

screen shot 2017-05-22 at 11 31 42

PR #44, opened 2017-05-22
PROGRAMMES-5518 Redis on production

Tasiobg:
Update config to use the RedisAdapter in production.

The endpoint of the cluster is passed into the configuration based on a Cosmos parameter


programmes-developers:

Tasiobg:
@BPScott @BjornTwachtmann what redis server size should we use for int/test?

BjornTwachtmann:
This looks good to me. Will come over to chat about instance size.

PR #45, opened 2017-05-22
Lazy load images with requireJs and versioning

tatitati:
Lazy load for images in schedules by day.

NOTE :fire: Not possible to test directly because of a dependency with another PR ([schedules] - programme). It needs to be merge in order to have the layout for probrammes in broadcasts


programmes-developers:

programmes-developers:

app/Resources/views/base_ds2013.html.twig
```diff
@@ -19,8 +19,15 @@
         {# TODO: Build Page Title correctly #}
         {# TODO: Twitter Tags, Facebook OpenGraph Tags and Apple touch Icons #}
         {# TODO: Canonical URL #}
-
-        <script>document.documentElement.className += ' has-js';</script>
+        <script>
```
votemike: Do we want inline JS in our templates?

gulpfile.js
```diff
@@ -7,13 +7,46 @@ const sourcemaps = require('gulp-sourcemaps');
...
+});
+
+gulp.task('js', ['js:clean'], function () {
+
+    var config = {
+        // by default load modules from...
+        baseUrl: 'node_modules',
+        // Specific paths for modules in another path.
+        paths: {
+            'jquery': 'jquery/dist/jquery',
```
votemike: Should these all by in a [gulp-config file](http://gulpjs.org/recipes/using-external-config-file.html)?

app/Resources/views/base_ds2013.html.twig
tatitati: well, we only require Requirejs to start the modules. But apart of that we don't need to ingest any logic more about those to use them, requirejs makes that work for us. So I don't expect to see this growing up. 
Someone?, Any suggestion?

src/Resources/js/modules/images.js
```diff
@@ -0,0 +1,109 @@
...
+            if (this.options.appear.onscroll) {
+                $(img).addClass(this.options.lazy_img_css);
+            }
+
+            $(div).replaceWith($(img));
+
+            return img;
+        },
+        addOnScrollHandler : function (context) {
+            var _this = this;
```
votemike: @BPScott @BjornTwachtmann Do we have a standard for doing creating a local version of `this`?

votemike: Or do we want to use `.bind()` instead?

app/Resources/views/base_ds2013.html.twig
```diff
@@ -19,8 +19,15 @@
...
+            requirejs.config({
+                "baseUrl": "/assets/js",
```
BPScott: This won't work when deployed.

config_requirejs.json
```diff
@@ -0,0 +1,17 @@
+{
+    "requirejs": {
+        "baseUrl": "node_modules",
```
BPScott: Presumably most of our modules would come through `../app/Resources/js` so would that be a better default. Then we'd only need to special case for things that come through node_modules. Related why have you put the JS in `src/Resources` when all the scss and images are in `app/Resources`?

config_requirejs.json
```diff
@@ -0,0 +1,17 @@
...
+        "paths": {
+            "jquery": "jquery/dist/jquery",
```
BPScott: We don't want to include jquery in our compiled files, that should be a separate one file, thus the path for this should be `empty:` as in the current rjs config. We should inline jquery version from frameworks as the orb docs suggest: https://navigation.api.bbc.co.uk/docs/requirejs.md

app/Resources/views/base_ds2013.html.twig
```diff
@@ -19,8 +19,16 @@
...
+                "paths": {
+                    'rv-bootstrap': "{{ asset('js/rv-bootstrap.js')|split('.js')|first }}",
+                    'jquery': "https://ajax.googleapis.com/ajax/libs/jquery/1.7.2/jquery.min"
```
tatitati: I know this address is not good at all. I tried all the urls in https://navigation.api.bbc.co.uk/docs/requirejs.md but all of them give 404 responses.

PR #46, opened 2017-05-23
Fixing early morning viewing of TV schedules

votemike:
Fixing bug when viewing a TV schedule between midnight and 6AM.  
Also adding tests.


programmes-developers:

tests/DataFixtures/ORM/BroadcastsFixture.php
```diff
@@ -0,0 +1,89 @@
...
+        string $pid,
+        Version $version,
+        DateTime $startDateTime,
+        DateTime $endDateTime,
+        Service $service,
+        ProgrammeItem $programmeItem
+    ) {
+        $entity = new Broadcast($pid, $version, $startDateTime, $endDateTime);
+        $entity->setService($service);
+        $entity->setProgrammeItem($programmeItem);
```
votemike: Why does a `Broadcast` have a direct relation to a `ProgrammeItem`? They are already linked via the `Version`. This is just duplicating information and increasing the possibility of discrepancies...

src/Controller/SchedulesByDayController.php
```diff
@@ -152,4 +134,36 @@ private function getOnAirBroadcast(array $broadcasts): ?Broadcast
...
+            if (!$startDateTime) {
+                throw $this->createNotFoundException('Invalid date');
+            }
+        } else {
+            // Otherwise use now
+            $today = ApplicationTime::getTime();
+
+            // If a user is viewing the TV schedule between midnight and 6AM, we actually want to display yesterday's schedule.
+            if ($serviceIsTv && (int) $today->format('H') < $tvOffsetHours) {
+                $today = $today->sub(new DateInterval('P1D'));
```
tatitati: this is confuse.....today is actually yesterday....might be possible to change it by "day" or something like that?

tatitati:
I like. Also I tested it

tests/DataFixtures/ORM/BroadcastsFixture.php
felipeparaujo: I believe this is so we can avoid joining to another table. We want to reduce the amount of joins we make as much as we can, and a common use case is getting the programme info for a broadcast. Joining is very costly, as we can see in the rather slow category pages

tests/DataFixtures/ORM/BroadcastsFixture.php
```diff
@@ -0,0 +1,89 @@
...
+        return $entity;
+    }
+
+    private function buildRadioSchedule()
+    {
+        $service = $this->getReference('p00fzl8v'); //bbc_radio_two
+
+        $version = $this->getReference('p4000001');
+        $programmeItem = $this->getReference('p3000001'); //Yesterday Afternoon Episode
+        $this->buildBroadcast('p5000001', $version, new DateTime('2017-05-21 14:00:00'), new DateTime('2017-05-21 14:45:00'), $service, $programmeItem);
```
felipeparaujo: Add a little bit of spacing between the broadcasts to make this a bit more readable. Do the same for the other fixtures.

tests/DataFixtures/ORM/BroadcastsFixture.php
BPScott: `Broadcast` -> `Version` -> `ProgrammeItem` is the canonical relationship
`Bradocast -> `ProgrammeItem` is a shortcut denormalisation.

Joining through an additional table is often quite expensive as Version is a big table. We can see some surprising speed gains by sidestepping it and instead doing things like `SELECT * from broadcast where programme_item_id = {{blah}}` to remove that join.

votemike: The specific join needed seem like a cheap join to me. It would be an inner join, not a left join. I'm just not keen on having two ways to get from one entity to another.

BjornTwachtmann: There are reasons for this, mostly to do with database performance on certain queries. It's pretty robustly handled in faucet, but yeah it can make writing fixtures a bit confusing.

tests/Controller/SchedulesByDayControllerTest.php
```diff
@@ -0,0 +1,75 @@
...
+     * @param null|string $timeNow            The system time, can be null if setting $scheduleDate
+     * @param string $network            The pid of the network
+     * @param null|string $scheduleDate       The date the user is viewing the schedule for, can be null if $timeNow is set
+     * @param string[] $expectedBroadcasts An array of expected broadcast times
+     */
+    public function testScheduleDisplaysCorrectBroadcastsForTime(?string $timeNow, string $network, ?string $scheduleDate, array $expectedBroadcasts)
+    {
+        if (!is_null($timeNow)) {
+            $timestamp = (new DateTimeImmutable($timeNow))->getTimestamp();
+            ApplicationTime::setTime($timestamp);
```
BPScott: `ApplicationTime::setTime(time())`

tests/Controller/SchedulesByDayControllerTest.php
```diff
@@ -0,0 +1,75 @@
...
+    {
+        $this->loadFixtures(["BroadcastsFixture", "NetworksFixture"]);
+
+        $client = static::createClient();
+        $crawler = $client->request('GET', '/schedules/p00fzl6p/2017-03-04');
+
+        $this->assertResponseStatusCode($client, 404);
+    }
+
+    public function scheduleDateTestProvider(): array
```
BPScott: Convention in other projects is to put the dataProvider below the test that calls it, should we do the same here?

tests/DataFixtures/ORM/BroadcastsFixture.php
votemike: Done

BjornTwachtmann:
The logic for the BBC day stuff looks correct. Which is no mean feat. I've got some bits I want to add on top of this, so please merge as soon as you're ready.

tests/Controller/SchedulesByDayControllerTest.php
```diff
@@ -0,0 +1,75 @@
...
+use BBC\ProgrammesPagesService\Domain\ApplicationTime;
+use DateTimeImmutable;
+use Tests\App\BaseWebTestCase;
+
+/**
+ * @covers App\Controller\SchedulesByDayController
+ */
+class SchedulesByDayControllerTest extends BaseWebTestCase
+{
+    public function testScheduleForToday()
```
BPScott: We should remove/rename `testScheduleForToday` and `testScheduleForDate` as they're testing out invalid sids and inactive dates.

(we should probably move those error-state tests to be below the "sucessful" tests

src/Controller/SchedulesByDayController.php
```diff
@@ -152,4 +134,36 @@ private function getOnAirBroadcast(array $broadcasts): ?Broadcast
 
         return null;
     }
+
+    private function getStartAndEndTimes(bool $serviceIsTv, ?string $date): array
+    {
+        $tvOffsetHours = 6;
+        if ($date) {
+            // Try and create a date from the one provided
+            $startDateTime = DateTimeImmutable::createFromFormat('Y-m-d|', $date);
```
BPScott: Whats the | for?

src/Controller/SchedulesByDayController.php
```diff
@@ -152,4 +134,36 @@ private function getOnAirBroadcast(array $broadcasts): ?Broadcast
 
         return null;
     }
+
+    private function getStartAndEndTimes(bool $serviceIsTv, ?string $date): array
```
BPScott: It might be handy for future-us to drop a little docblock in here that talks about why this is needed - TV vs Radio days being different and we want to show you the day that the "currently broadcasting thing".

tests/Controller/SchedulesByDayControllerTest.php
votemike: What? Why?

src/Controller/SchedulesByDayController.php
votemike: `|` means `00:00:00`

BPScott: Cool, hadn't came across that before

tests/Controller/SchedulesByDayControllerTest.php
BPScott: just a bit shorter than ApplicationTime::setTime((new DateTimeImmutable($timeNow))->getTimestamp());

votemike: But `ApplicationTime::setTime(time())` doesn't take in to account `$timeNow`....

BPScott: ah yes I'm being blind. sorry ignore me.

votemike: I shall

PR #47, opened 2017-05-23
Fixing broadcast periods

votemike:
This should stop TV programmes between midnight and 6AM being shown at the top of the page and instead shown at the bottom.


tatitati:
I like

PR #49, opened 2017-05-24
Schedules controller tidy

BjornTwachtmann:
For some reason this incredibly simple PR turned into the most complex thing that has ever happened. 

All this actually does:
Filter service dates at the DB level, not in code.
Move service isTV/isRadio logic to the domain model layer.
Move some shit around.


programmes-developers:

PR #50, opened 2017-05-24
Redis cache dev

tatitati:
The idea is to have running redis only in dev.


programmes-developers:

PR #51, opened 2017-05-25
Lazy load for images

tatitati:


programmes-developers:

src/Twig/AssetsExtension.php
```diff
@@ -0,0 +1,39 @@
...
+        $this->container = $container;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getFunctions()
+    {
+        return [
+            new Twig_Function('assetJs', [$this, 'assetJs']),
```
BPScott: Twig convention is to name things in snake_case rather than camelCase, so `asset_js`

gulpfile.js
```diff
@@ -7,13 +7,49 @@ const sourcemaps = require('gulp-sourcemaps');
...
 var throwError = true;
+
+gulp.task('js:clean', function () {
+    return del([staticPathDist + '/js']);
+});
+
+gulp.task('js', ['js:clean'], function () {
+
+    var config = {
+        "requirejs": {
```
BPScott: No point having this key, you can remove this level of nesting then reference `requirejsOptimize(config)`

gulpfile.js
```diff
@@ -7,13 +7,49 @@ const sourcemaps = require('gulp-sourcemaps');
...
+            },
+            "map": {
+                "*": {
+                    "jquery": "jquery-1.9"
+                }
+            }
+        }
+    };
+
+    return gulp.src([staticPathSrc + '/js/**/*.js', nodeModulesPath + '/jquery.appear/**/*.js'])
```
BPScott: We don't want to create an output file for every file, only the ones that we'll be trying to load on the front-end. For instance "images" gets pulled into "rv-bootstrap", we'd never try and load "images" directly on the front end so there is no point creating an output file for it.

gulpfile.js
```diff
@@ -7,13 +7,50 @@ const sourcemaps = require('gulp-sourcemaps');
...
+    return del([staticPathDist + '/js']);
+});
+
+gulp.task('js', ['js:clean'], function () {
+
+    var modulesToOptimize = [
+        staticPathSrc + '/js/**/rv-bootstrap.js'
+    ];
+
+    var config = {
```
votemike: Is this ever changed? Should it be `const` instead of `var`?

app/Resources/js/images.js
```diff
@@ -0,0 +1,150 @@
+/*
+This module lazy load images and convert them in case that they are <div> elements
+to the proper image. This lazy-loading is triggered by the scroll.
+ */
+define(['jquery-1.9', 'jquery.appear'], function ($) {
```
votemike: Why jQuery 1.9? It's 4/5 years old! Why not jQuery 1.12 (a year old)? or even better jQuery 3.2.

app/Resources/js/images.js
```diff
@@ -0,0 +1,150 @@
...
+            var _this = this;
+            var hiddenImages = $('.' + this.options.lazy_img_css, context);
+
+            hiddenImages.each(function(index, element){
+                if (_this.isCloseToAppear(element)) {
+                    _this.displayImageOnViewport(element);
+                }
+            })
+        },
+        isCloseToAppear: function(element) {
```
votemike: I don't understand this function name.

app/Resources/js/images.js
```diff
@@ -0,0 +1,150 @@
...
+            var _this = this;
```
votemike: Why do you need this line?

app/Resources/views/base_ds2013.html.twig
```diff
@@ -19,8 +19,22 @@
         {# TODO: Build Page Title correctly #}
         {# TODO: Twitter Tags, Facebook OpenGraph Tags and Apple touch Icons #}
         {# TODO: Canonical URL #}
-
         <script>document.documentElement.className += ' has-js';</script>
+        <script>
```
votemike: Can this not go in the script tag above? Why do we need multiple script tags?

src/Twig/AssetsExtension.php
```diff
@@ -0,0 +1,39 @@
...
+            new Twig_Function('asset_js', [$this, 'assetJs']),
+        ];
+    }
+
+    public function assetJs($manifestKey): ?string
+    {
+        // get url based on manifest and s3
+        $path = $this->container->get('assets.packages')->getUrl($manifestKey);
+
+        if (preg_match('/(.*)(\.js)$/', $path, $matches)) {
```
votemike: Is this preg match as simple as it an be?

src/Twig/AssetsExtension.php
```diff
@@ -0,0 +1,39 @@
...
+     * {@inheritdoc}
+     */
+    public function getFunctions()
+    {
+        return [
+            new Twig_Function('asset_js', [$this, 'assetJs']),
+        ];
+    }
+
+    public function assetJs($manifestKey): ?string
```
votemike: It is not instantly clear what this method is doing. Could you simplify the method or name it in a clearer way?

src/Twig/AssetsExtension.php
```diff
@@ -0,0 +1,32 @@
...
+     * {@inheritdoc}
+     */
+    public function getFunctions()
+    {
+        return [
+            new Twig_Function('asset_js', [$this, 'assetJs']),
+        ];
+    }
+
+    public function assetJs(string $path): string
```
votemike: It's still not clear to me what this method does. I think it needs a clearer name or a docblock. If reading the code where it is called, I wouldn't know if it:

- Opens a JS file and returns the JS contents
- Returns a bool if the file extension is JS.
- Returns a bool if the files content is JS.

votemike: Would `SplFileInfo::getBasename()` be better?

app/Resources/js/images.js
BPScott: This is a hangover from the PAL days, where we all had a shared platform, and we pulled in modules from other BBC teams. Everyone ended up centering around 1.9, because the alternative would be multiple different versions of jQuery on a page, which is grim from a performance perspective. I remember the annoying midpoint as people migrated their things from 1.7 to 1.9, bad times :(

However I had a chat with @Jayflux last night, and new /radio bits are moving away from that shared model. Currently the RadioNav uses jquery 1.9, but we could jab them to prioritise the ticket to remove their jQuery dependencies and they'd be up for doing it. Other than that, there's nothing in the ORB that requires jQuery, and we can make it a requirement of 3rd party modules we have to embed (e.g. bump, snippets) that they shouldn't rely on a global jQuery (TODO check if bump currently depends on jQuery, I don't think it does).

I was initially conservative (hence saying to go for 1.9) but you're right, that leaves us standing in the past. If we're breaking free of that legacy then we should go for the most recent version of use to us. It looks like 1.x is no longer maintained aside from security issues and if we don't need IE6-9 support (TODO confirm this with stats) we should go with 3.2. In either case, pulling that from npm would be nicer than the frameworks sharedmodules space.

BPScott: Swinging back again after more research: [bump uses 1.9](https://confluence.dev.bbc.co.uk/display/mp/Embedding+and+controlling+a+Standard+Media+Player). Both iPlayer and playspace use bump for media playout, and thus 1.9. Playspace doesn't use jQuery directly. iPlayer meanwhile do use jQuery - they have a standard 'jquery' require map which points to the 1.9 file, and an alias of 'jquery-1.9' that maps to 'jquery'. This means that in their own code they reference 'jquery' (making the upgrade path easy) while avoiding duplicate loading of 1.9.

I don't think we want to break away from that legacy just yet unfortunately. 1.9 is going to crop up often enough that having to load two versions of jquery on pages with media playout shall be obnoxious. I do however like iPlayer's approach of naming things, that'll mean we can easily jump to a newer version by changing one alias rather than having to update many references in our js files.

app/Resources/views/base_ds2013.html.twig
BPScott: Hany to make it obvious that these are two distinct things

app/Resources/views/base_ds2013.html.twig
```diff
@@ -19,8 +19,22 @@
...
+            requirejs.config({
+                "paths": {
+                    'jquery-1.9': "https://static.bbci.co.uk/frameworks/jquery/0.4.1/sharedmodules/jquery-1.9.1",
```
votemike: You have changed some reference from jquery-1.9 to jquery. Should that also happen here?

src/Twig/AssetsExtension.php
```diff
@@ -0,0 +1,36 @@
...
+        ];
+    }
+
+    /**
+     * Given a js asset path, return it in a format suitable for inclusion in
+     * the require config (i.e. with the ".js" extension removed)
+     */
+    public function assetJs(string $path): string
+    {
+        return preg_replace('/\.js$/', '', $this->packages->getUrl($path, null));
```
votemike: `preg_replace` seems overkill. Can you not use `rtrim` of something?

votemike:
What's the bast way to test this locally?

app/Resources/views/base_ds2013.html.twig
tatitati: In this case no. The map declare something like this:
"if something needs jquery, inject jquery-1.9, if something needs jquery-1.9, go ahead"

PR #52, opened 2017-05-25
Adding BST support

votemike:
This goes part way to solving the BST problem. One of the things that doesn't work yet is the time period grouping, but ignore that for the moment.  

Now:  
When dates go in to Doctrine with a timezone, they are converted to UTC.  
When date strings come out of doctrine, they are converted to DateTimes with the UTC timezone.  
Twig automatically formats all times to be Europe/London.  

How do we want to deal with timezones in PHP (such as working out the period of the day a broadcast is in)? On a case-by-case basis?


programmes-developers:

src/Types/UtcDateTimeType.php
```diff
@@ -0,0 +1,46 @@
...
+use Doctrine\DBAL\Types\DateTimeType;
+
+class UTCDateTimeType extends DateTimeType
+{
+    static private $utc;
+
+    public function convertToDatabaseValue($value, AbstractPlatform $platform)
+    {
+        if ($value instanceof \DateTime) {
+            $value->setTimezone(new \DateTimeZone('UTC')); //@TODO can I change this to the interface and just assign this to a new variable as below?
```
BjornTwachtmann: is it better to create a new object here than to call setTimezone on the existing one? I'm somewhat concerned this will have side effects on the calling function that might not be expected.

src/Types/UtcDateTimeType.php
```diff
@@ -0,0 +1,46 @@
...
+    public function convertToPHPValue($value, AbstractPlatform $platform)
+    {
+        if (null === $value || $value instanceof \DateTimeInterface) {
+            return $value;
+        }
+
+        $converted = \DateTime::createFromFormat(
+            $platform->getDateTimeFormatString(),
+            $value,
+            self::$utc ? self::$utc : self::$utc = new \DateTimeZone('UTC')
```
BjornTwachtmann: what's the purpose of the static variable? Just to save creating a new object?

app/config/config.yml
```diff
@@ -34,6 +34,8 @@ framework:
 
 # Twig Configuration
 twig:
+    date:
+        timezone: Europe/London
```
BjornTwachtmann: There are pages where we don't want the output to be in BST. Though this is a sensible default. Is there a way to override from the controller?

BjornTwachtmann:
I think this is the right approach. I would move UtcDateTimeType.php to Pages Service so that anything can use it (this will not just be an issue in frontend), and add instructions on how to enable it to the Readme.md stating that this is optional.

src/Types/UtcDateTimeType.php
votemike: In the same way that `DateTimeImmutable` creates a new object when we change the timezone. Then we can reduce the amount of code we write.

src/Types/UtcDateTimeType.php
votemike: That's the only thing I could think it might be for. If returning a bucket load of date from the DB. I just copy/pasted it from the docs.

app/config/config.yml
votemike: I'm not sure if we can override it in the controllers. But should be able to pass in the timezone we actually want in the date filter in twig.

felipeparaujo:
test this please

src/Controller/SchedulesByDayController.php
```diff
@@ -148,16 +149,18 @@ private function getOnAirBroadcast(array $broadcasts): ?Broadcast
...
-            $startDateTime = DateTimeImmutable::createFromFormat('Y-m-d|', $date);
+            $startDateTime = DateTimeImmutable::createFromFormat('Y-m-d|', $date, $londonTimezone);
 
             if (!$startDateTime) {
                 throw $this->createNotFoundException('Invalid date');
             }
         } else {
             // Otherwise use now
             $dateTime = ApplicationTime::getTime();
+            $dateTime = $dateTime->setTimezone($londonTimezone); //@TODO or do we want the timezone specified in setTime()?
```
BPScott: Letting you specify an optional timezone arg to getTime sounds handy. Would Europe/London be the default that we need in most cases?

votemike: I think it would be..... Wouldn't it?

PR #53, opened 2017-05-25
Redis to allow to run phpunit and composer from outside

tatitati:


programmes-developers:

PR #54, opened 2017-05-26
Class attribute building

BPScott:
Defining classes and attributes in templates makes more sense as then they're much nearer to where they're used. Having to flit between two files to find the css classes that are used in a given case is annoying.

So thus define classes in the presenters.

### DIFF
--- a/src/Ds2013/Molecule/DateList/DateListItemPresenter.php
+++ b/src/Ds2013/Molecule/DateList/DateListItemPresenter.php
@@ -45,30 +45,14 @@ class DateListItemPresenter extends Presenter
         return $this->buildHtmlAttributes(['data-href-add-utcoffset' => 'true']);
     }
 
-    public function getCssClassesForItem(): string
-    {
-        $shouldShowLink = $this->shouldShowLink();
-        return $this->buildCssClasses([
-            'br-page-bg-onbg--hover br-page-linkhover-ontext--hover' => $shouldShowLink,
-            'br-box-page' => ($this->offset == 0),
-            'text--faded' => ($this->offset != 0 && !$shouldShowLink),
-        ]);
-    }
-
-    public function getCssClassesForLi(): string
-    {
-        return $this->buildCssClasses([
-            'date-list__page' => true,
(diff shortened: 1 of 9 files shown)


programmes-developers:

src/Ds2013/Molecule/DateList/DateListItemPresenter.php
```diff
@@ -37,55 +37,36 @@ public function __construct(DateTimeImmutable $datetime, Service $service, int $
...
+        return $this->offset;
     }
 
     public function getServicePid(): string
     {
         return (string) $this->service->getPid();
     }
 
-    public function shouldShowLink(): bool
+    public function isLink(): bool
```
votemike: What is `isLink()`?

src/Ds2013/Molecule/DateList/DateListItemPresenter.php
```diff
@@ -37,55 +37,36 @@ public function __construct(DateTimeImmutable $datetime, Service $service, int $
...
     }
+
+    public function isGmt(): bool
+    {
+        return $this->options['user_timezone'] == 'GMT';
+    }
+
+    public function isToday(): bool
+    {
+        return ApplicationTime::getTime()->format('Y-m-d') == $this->datetime->format('Y-m-d');
```
votemike: This doesn't seem to be comparing timezones. What if each time being compared has a different timezone?

src/Ds2013/Molecule/DateList/DateListItemPresenter.php
BPScott: If the current date list item is a link

src/Ds2013/Molecule/DateList/DateListItemPresenter.php
BPScott: Good point. This is the same logic as what was in here previously. The diff is moving the order of the methods so the `is*` methods sit together. Fixing this comparison is out of scope of this PR imo.

votemike: We can save it for Chronos to deal with 😄

PR #55, opened 2017-05-31
Adding redis extension requirement to composer

votemike:
This should throw a clearer messages when using composer without the redis extension installed.


programmes-developers:

BjornTwachtmann:
Sounds sensible to me

PR #56, opened 2017-05-31
Basic refactor of schedules page to page presenter model. Needs tests

BjornTwachtmann:
Opionions? I thought moving view logic out of the Controller would make things
a) A fuckload easier to test
b) Easier to read

Do you agree?


programmes-developers:

src/Controller/SchedulesByDayController.php
```diff
@@ -27,126 +28,39 @@ public function __invoke(Pid $pid, ?string $date, ServicesService $servicesServi
...
-        $twinService = null;
-        if (count($servicesInNetwork) == 2) {
-            // If there are two services, find the "other" service
-            $otherServices = array_filter($servicesInNetwork, function (Service $sisterService) use ($service) {
-                return ($service->getSid() !== $sisterService->getSid());
-            });
-            $twinService = reset($otherServices);
-        }
+        // Get broadcasts in relevant period
+        $broadcasts = $broadcastService->findByServiceAndDateRange($service->getSid(), $startDateTime, $endDateTime);
```
BjornTwachtmann: Yeah, fair point. It was easier to construct the presenter object that way. Was more looking for comments on the core concept than the implementation details at this stage, but this would be worth fixing.

src/Ds2013/Page/SchedulesByDay/SchedulesByDayPagePresenter.php
```diff
@@ -0,0 +1,162 @@
...
+     */
+    private $endDate;
+    /** @var Broadcast[] */
+    private $broadcasts;
+    /**
+     * @var array
+     */
+    private $servicesInNetwork;
+
+    /** @var Broadcast */
```
votemike: `@var Broadcast|bool`?

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -24,13 +24,13 @@
...
         <ol class="list-unstyled g-c-l">
-            {% for period_name, broadcast_in_period in grouped_broadcasts %}
+            {% for period_name, broadcast_in_period in pagePresenter.BroadcastsGroupedByPeriodOfDay %}
                 <li id="{{ period_name }}" class="schedule-skip-link">
                     <h2 class="delta text--shout island--horizontal component">{{ period_name }}</h2>
                     <ol class="highlight-box-wrapper">
                         {#list of broadcasts in day period#}
                         {% for broadcast in broadcast_in_period %}
-                            <li {{ broadcast == on_air_broadcast ? 'id=on-now class=schedule-skip-link-on-now' : null }} >
+                            <li {{ broadcast == pagePresenter.onAirBroadcast ? 'id=on-now class=schedule-skip-link-on-now' : null }} >
```
BjornTwachtmann: I've been considering refactoring this logic into the broadcast domain model or presenter. Not quite sure why it's done that way, but haven't gotten round to that yet.

votemike: I think it should get done soonish. This way of finding out if the broadcast is on now is inefficient. If we leave it too long, we'll forget about it.

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -1,7 +1,7 @@
 {% extends 'schedules/base.twig' %}
 
 {% block content %}
-    {% if not service.isTv %}
+    {% if not pagePresenter.service.isTv %}
```
votemike: Is it worth having sub-templates for radio and tv, so that we don't need to check everywhere inside templates if a page should display like a tv or a radio schedule?

BPScott:
As discussed IRL yesterday (but putting here so others can see): Nice, lets get this green and merge it in. We can add tests later as having this languish in PR means nobody else can work on by-day without causing rebasing headaches.

Agreed that extracting this out makes things easier to test. Yay for skinny controllers. I feel that we should keep  templates in `app/resources/views` as the entry point that controllers load, rather than a the controllers loading any of the presenter templates (which this PR does as of the 'change everything again' commit). I'd like to preserve the notion that "presenters are for parts of a page, while `app/resources/views` templates are for whole pages", and this helps make that separation clearer.

I'm a little wary of the page presenters becoming god objects - containing all state for a page, when I think we should be mostly composing a page out of several smaller presenters. I'd be interested to see if we could extract some of this out, but that's a task for a separate PR, this is a really good step in the right direction.

I'm not totally convinced by instantiating the presenters within the controller. If this sort of thing is possible to split into smaller presenters, it might be easier to instantiate them from the templates using `ds2013()`, but that'll become clearer when we try that.

For clarity for others, this is a process for particularly complex page constructs - not all controllers would need page presenters.

programmes-developers:

app/Resources/views/schedules/no_schedule.html.twig
```diff
@@ -2,10 +2,10 @@
 
 {% block content %}
     <p class="noschedule">
-        {% if service.startDate and date < service.startDate %}
-            {{ tr('schedule_begins_on', {'%1':'<a href="' ~ path('schedules_by_day', {'pid': service.pid, 'date': service.startDate|date('Y-m-d')}) ~ '">' ~ service.startDate|date("l j F Y") ~ '</a>'})|raw }}
-        {% elseif service.endDate and date > service.endDate %}
-            {{ tr('schedule_ended_on',{'%1':'<a href="' ~ path('schedules_by_day', {'pid': service.pid, 'date': service.endDate|date('Y-m-d')}) ~ '">' ~ service.endDate|date("l j F Y") ~ '</a>'})|raw }}
+        {% if page_presenter.service.startDate and page_presenter.startDate < page_presenter.service.startDate %}
```
Tasiobg: 2 questions.

I havent checked how those dates looks like but I would like to confirm if it is safe to use comparison operators on twig dates ?

I can't understand why "page_presenter.service.startDate" is compared against the same object, can someone explain me if this is right ?

BjornTwachtmann: it's comaring page_presenter.startDate and page_presenter.service.startDate, which are different things, although, yeah, this could be clearer. They're just PHP DateTimes, it should be safe

PR #57, opened 2017-05-31
Tidy broadcast

BPScott:


programmes-developers:

src/Ds2013/Organism/Broadcast/BroadcastPresenter.php
```diff
@@ -36,43 +36,34 @@ public function getServiceName(): string
...
         return $this->broadcast->getProgrammeItem();
     }
 
-    public function getMessageIsNow(): string
+    public function getOnAirMessage(): string
     {
-        if ($this->getNetworkMedium() == 'tv') {
-            return 'on_now';
-        }
+        return $this->broadcast->getService()->isTv() ? 'on_now' : 'on_air';
```
votemike: Might be overkill.... but is it worth having subclasses for TV and Radio instead of doing this check here?

BPScott: Overkill in this case i reckon. It'd require creating a new presenter, an abstract presenter, and then doing logic within the factory to choose which Presenter to instantiate.

PR #58, opened 2017-05-31
Throw helpful error if you reference an unknown presenter option

BPScott:
Move the Exception class to the Ds2013 namespace as we're not expecting
having lots of different exceptions.


programmes-developers:

PR #60, opened 2017-06-01
Schedules splitting

BPScott:
I've had a play with extracting the services list into its own presenter, and along the way I've tidied up the region picker presenter too.

At this point there are things in the schedules page presenter that are never used within its own template, but are used in the base.html.twig and other templates in app/Resources. Personally I think this is odd, but we can have that discussion when I'm back in the office. Changes in this should be non-controversial, as it's mostly fixing and tidying things rather than changing how things get passed around

### DIFF
--- a/app/Resources/views/schedules/base.twig
+++ b/app/Resources/views/schedules/base.twig
@@ -25,6 +25,12 @@
 
     <div class="b-g-p br-box-page island--vertical">
         {{ ds2013('regionPick', page_presenter.service, app.request.attributes.get('date'), page_presenter.servicesInNetwork) }}
+
         {% block content %}{% endblock %}
+
+        {# Only show services list if there are more than two #}
+        {% if page_presenter.servicesInNetwork|length > 2 %}
+            {{ ds2013('networkServicesList', page_presenter.service, app.request.attributes.get('date'), page_presenter.servicesInNetwork) }}
+        {% endif %}
     </div>
 {% endblock %}
--- a/app/Resources/views/schedules/base.twig
+++ b/app/Resources/views/schedules/base.twig
@@ -16,13 +16,15 @@
             </time>
         </h1>
     </div>
+
     <nav class="schedule-nav-wrapper fixedsticky">
         <div id="schedule-daynav" class="centi br-box-highlight {{ page_presenter.service.isTv ? 'g-o-l daynav--tvschedule' : 'g-f-l' }}">
             {{ ds2013('dateList', page_presenter.startDate, page_presenter.service ) }}
         </div>
     </nav>
+
     <div class="b-g-p br-box-page island--vertical">
-        {{ ds2013('regionPick', page_presenter.service, page_presenter.startDate, page_presenter.servicesInNetwork) }}
+        {{ ds2013('regionPick', page_presenter.service, app.request.attributes.get('date'), page_presenter.servicesInNetwork) }}
         {% block content %}{% endblock %}
     </div>
 {% endblock %}
--- a/src/Ds2013/Page/Schedules/RegionPick/RegionPickPresenter.php
+++ b/src/Ds2013/Page/Schedules/RegionPick/RegionPickPresenter.php
@@ -4,20 +4,19 @@ namespace App\Ds2013\Page\Schedules\RegionPick;
 
 use App\Ds2013\Presenter;
 use BBC\ProgrammesPagesService\Domain\Entity\Service;
-use DateTimeImmutable;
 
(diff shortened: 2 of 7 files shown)


programmes-developers:

app/Resources/views/schedules/base.twig
```diff
@@ -16,13 +16,21 @@
...
+
     <div class="b-g-p br-box-page island--vertical">
-        {{ ds2013('regionPick', page_presenter.service, page_presenter.startDate, page_presenter.servicesInNetwork) }}
+        {{ ds2013('regionPick', page_presenter.service, app.request.attributes.get('date'), page_presenter.servicesInNetwork) }}
+
         {% block content %}{% endblock %}
+
+        {# Only show services list if there are more than two #}
+        {% if page_presenter.servicesInNetwork|length > 2 %}
+            {{ ds2013('networkServicesList', page_presenter.service, app.request.attributes.get('date'), page_presenter.servicesInNetwork) }}
```
votemike: Is `app.request.attributes.get('date')` the date param from the URL?

src/Ds2013/Page/Schedules/RegionPick/RegionPickPresenter.php
```diff
@@ -4,52 +4,72 @@
...
         $this->service = $service;
         $this->date = $date;
         $this->servicesInNetwork = $servicesInNetwork;
     }
 
-    public function getService(): Service
+    public function hasRegionPicker(): bool
     {
-        return $this->service;
+        return count($this->servicesInNetwork) > 1;
```
votemike: Is this the case. What if there are only 2 services? The base service and the HD service. Do we still want the picker?

src/Ds2013/Page/Schedules/RegionPick/RegionPickPresenter.php
```diff
@@ -4,52 +4,72 @@
...
+        return $twin ? $twin->getName() : $this->service->getNetwork()->getName();
+    }
+
+    public function getDate(): ?string
     {
         return $this->date;
     }
 
-    public function getServicesInNetwork(): array
+    public function getTwinServicePid(): string
```
votemike: Is this really necessary? Can you not just use `getTwinService()`

votemike:
Does any of this need tests?

app/Resources/views/schedules/base.twig
BPScott: Yep

votemike: Do we want to be grabbing/reading request data from views? Seems like a bit of a code smell to me.

src/Ds2013/PresenterFactory.php
```diff
@@ -9,6 +9,7 @@
 use App\Ds2013\Organism\Programme\ProgrammePresenter;
 use App\Ds2013\Page\Schedules\ByDayPage\SchedulesByDayPagePresenter;
 use App\Ds2013\Page\Schedules\RegionPick\RegionPickPresenter;
+use App\Ds2013\Page\Schedules\NetworkServicesList\NetworkServicesListPresenter;
```
votemike: This should be up one line. Also is `App\Ds2013\Molecule\DateList\DateListItemPresenter;` used any more?

app/Resources/views/schedules/base.twig
BPScott: It's reading information about the current route parameters to build a new route, which feels like an acceptable use-case to me. The alternative is having to pass around all the route params from the controller into the view, which has the potential to get clunky.

src/Ds2013/Page/Schedules/RegionPick/RegionPickPresenter.php
BPScott: Yep. Radio 4 lets you toggle between FM and LW services: http://www.bbc.co.uk/radio4/programmes/schedules

votemike: This method name is confusing for me.... Could you maybe add a docblock to make it clear that this is the bit at the top?

BPScott: Yeah, RegionPick isn't not the best name for this (it's the title block and the link to the services list at the bottom of the page), but I didn't want to change too much all in one go.

src/Ds2013/PresenterFactory.php
BPScott: Yep, that can be removed as it is now "internal" to the date list item presenter and not exposed in the factory.

PR #61, opened 2017-06-01
PROGRAMMES-5525 Cache resilience

Tasiobg:
This PR is to implement a mechanims to prevent downtime when Redis Cluster is not working.

If our application can't connect to a Redis Cluster we would like to not use it for the rest of the request. As a solution what I have implemented is a Redis Cluster factory class which wrap the connection into a "try" and if something goes wrong it will catch the "RedisClusterException" exception and return a NullAdapter, this way the rest of the request will be served with zero downtime and totally transparent for the user.

Notes:
- "cache.adapter.redis" only accept RedisAdapter instances an will fail with NullAdapter, as a solution I'm using "cache.adapter.psr6" which basically is a proxy adapter wich accepts any AdapterInterface implementation
- Not sure where could be best place to leave the RedisClusterFactory.php file, feel free to suggest better place
- I had a look at "cache.adapter.psr6" class (ProxyAdapter.php) and also did a commandline bencharch and I can't see any performance penalitation

@BjornTwachtmann suggested to create a ticket (minor) to investigate how to use "cache.adapter.redis" instead of "cache.adapter.psr6" so we can avoid the ProxyAdapter.php. On my point of view "cache.adapter.psr6" is good enough and not sure how posibble is to implement something like that withot having a negative impact, any thoughts about it?


programmes-developers:

BjornTwachtmann:
I think this looks sensible. The RedisClusterFactory class needs somewhere to live though, not the root folder. Possibly just create a "redis" folder or something?

src/Redis/RedisClusterFactory.php
```diff
@@ -0,0 +1,41 @@
+<?php
+//declare(strict_types = 1); strict_types doesnt't work with RedisCluster yet
```
votemike: Remove this line?

src/Redis/RedisClusterFactory.php
```diff
@@ -0,0 +1,41 @@
...
+namespace App\Redis;
+
+use RedisCluster;
+use RedisClusterException;
+use \Symfony\Component\Cache\Adapter\RedisAdapter;
```
votemike: You don't need this first slash.

src/Redis/RedisClusterFactory.php
```diff
@@ -0,0 +1,41 @@
...
+     * @param array $redisEndpoint
+     * @param LoggerInterface $logger
+     * @return AdapterInterface
+     */
+    public static function createRedisCluster(array $redisEndpoint, LoggerInterface $logger): AdapterInterface
+    {
+        try {
+            $redisClusterInstance = new RedisCluster(null, $redisEndpoint);
+            // always distribute readonly commands between masters and slaves, at random
+            $redisClusterInstance->setOption(
```
votemike: Should these options be set here? Or in the Yaml config?

src/Redis/RedisClusterFactory.php
```diff
@@ -0,0 +1,41 @@
...
+                RedisCluster::OPT_SLAVE_FAILOVER,
+                RedisCluster::FAILOVER_DISTRIBUTE
+            );
+            $cacheAdapter = new RedisAdapter($redisClusterInstance);
+        } catch (RedisClusterException $e) {
+            // if redis cluster fail, log the error and use a null adapter as a cache provider
+            $logger->error("Redis Cluster Error: " . $e->getMessage() . '. Using NullAdapter as a fallback cache adapter');
```
votemike: Do we need to log the stack trace anywhere?

src/Redis/RedisClusterFactory.php
Tasiobg: The comment is to prevent it from being added, I wans't sure if was a good idea but the error generated by strict_types is not easy to relate.
I'll remove the comment

src/Redis/RedisClusterFactory.php
Tasiobg: I just run optimeze imports, thanks!!

src/Redis/RedisClusterFactory.php
Tasiobg: On my point of view this is not something we would like to change like other options so I think is better to keep it inside the factory

src/Redis/RedisClusterFactory.php
Tasiobg: There are few reasons why I have consider this option good enough.
- An average of 300 request a second can flood the logs and affect the performance
- We are in an early stage of dealing with log, this may change on a future
- Connexion is only made at one place on the code and the redis client should provide the issue message, so far I have seen no help from a stack trace when the issue is related to redis.

src/Redis/RedisClusterFactory.php
BPScott: strict_types doesn't cascade down to classes referenced by this file, it's on a file-by-file basis, so I think if this is throwing an error it is because there is a bad type in this file, rather than something downstream.

src/Redis/RedisClusterFactory.php
BPScott: We'd never want to change them, so no point having to inject them via config.

app/config/services.yml
```diff
@@ -42,15 +42,12 @@ services:
...
         arguments:
-            - ~
             - ['%redis_endpoint%']
-        calls:
-        # By default, RedisCluster will only send commands to master nodes, this option distribute readonly commands between masters and slaves
-        # RedisCluster::OPT_SLAVE_FAILOVER  = 5
-        # RedisCluster::FAILOVER_DISTRIBUTE = 2
-            - { method: 'setOption', arguments: [5, 2] }
+        tags:
+            - { name: 'monolog.logger', channel: 'main' }
```
BPScott: `app` is the default channel name.  Alternatively `cache` might be a more relevant channel that is already used for other bits

To see examples of channel names check a log file, they contain mentions of 'app.INFO', 'cache.WARNING' etc, that bit before the dot - 'app' and 'cache' are the channels. Alternativly check `app/console debug:container` and look for services that start with 'monolog.logger.'

Tasiobg:
@BjornTwachtmann 
I have set the timeout to 100ms, I think if the redis connexion expend more than that this mean something is wrong an our app sholdn't wait more time, instead it should fall back to the null adapter.
Let me know if any objections.

BjornTwachtmann:
I would set the timeout to at least 2-300 ms . I know it shouldn't take that long, but 100 seems very low.

src/Redis/RedisClusterFactory.php
BjornTwachtmann: For the love of jesus, do not dump out pointless stack traces 300x a second. If this thing is blowing up then then it's almost certain the redis server is down and a stack trace will reveal very little.

Tasiobg:
@BjornTwachtmann thanks, I'll update it to 200ms

PR #63, opened 2017-06-05
Making date bar sticky

votemike:
Did not bother with the JS polyfill as @BPScott said not too. See [his comment](https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5590?focusedCommentId=3445365&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-3445365).

DO NOT MERGE while I test that {{ top: 0 }} doesn't break IE and EDGE.


programmes-developers:

votemike:
Implementing the sticky bar in this way leaves a 56px gap when using the period jump-links. Is that acceptable? Or should I re-implement the JS version of this?

BPScott:
A realisation: Would it be worth using [`@supports`](https://developer.mozilla.org/en/docs/Web/CSS/@supports) to wrap the sticky nav logic, so that we don't end up applying a bunch css to browsers that don't support position:sticky?

BjornTwachtmann:
I've tested it locally, seems to work pretty well. Lack of IE support might be an issue for some people, but we can re-visit it if they complain. 

I would say merge it.

src/Ds2013/Core/Modules/Schedules/_fixed-nav.scss
```diff
@@ -5,15 +5,21 @@
...
+@supports (position: sticky) {
     $skiplink-pushdown: 56px;
 
     .schedule-nav-wrapper {
         top: 0;
         width: 100%;
         z-index: 1000;
+
+        position: -webkit-sticky;
+        position: -moz-sticky;
```
BPScott: Taking a look at https://caniuse.com/#search=sticky it seems the only vendor prefixes that was ever used is -webkit, we can remove all the other prefixed versions

src/Ds2013/Molecule/DateList/_date-list.scss
```diff
@@ -115,13 +115,6 @@
             }
         }
     }
-
-    // Don't pull upwards when fixed to the top of the page
-    .fixed-supported {
```
BPScott: Would deleting break the positioning for tv schedules? It seems like this was reset for a reason

votemike: It doesn't seem to have. What about it is different to radio?

PR #64, opened 2017-06-05
More shuffling

BPScott:
Inching towards a cohesive idea of how to lay things out  I think.
I'm not overly sold on the idea of Page Presenters. I think presenters should be for a partial slice of a page rather than the whole thing, otherwise we end up with large objects that end up sprawling. I think focusing Presenters on particular things shall help encapsulate them better.


programmes-developers:

BjornTwachtmann:
I honestly preferred the model of a page template and a page presenter. If we split things this way we need to create and call off to a separate presenter/template combo any time there's any significant view logic, which seems unnecessary. 

That said this will work fine too if everyone wants to go this way...

src/Controller/SchedulesByDayController.php
```diff
@@ -35,24 +33,34 @@ public function __invoke(Pid $pid, ?string $date, ServicesService $servicesServi
...
-            );
-        }
+        // If there are no broadcasts, then the status code should be 404, so
+        // that search engines do not index thousands of empty pages
+        $response = new Response('', $broadcasts ? 200 : 404);
+        return $this->renderWithChrome('schedules/by_day.html.twig', $viewData, $response);
+    }
 
-        return $this->renderWithChrome('schedules/by_day.html.twig', ['page_presenter' => $pageViewModel]);
+    private function viewData(
```
votemike: Where we have methods that accept an array, can the method also have a docblock so that we can see what the array is made up of?

votemike: Same with return types of array if possible.

src/Controller/SchedulesByDayController.php
```diff
@@ -35,24 +33,34 @@ public function __invoke(Pid $pid, ?string $date, ServicesService $servicesServi
...
-        if (!$broadcasts) {
-            return $this->renderWithChrome(
-                'schedules/no_schedule.html.twig',
-                ['page_presenter' => $pageViewModel],
-                new Response('', Response::HTTP_NOT_FOUND)
-            );
-        }
+        // If there are no broadcasts, then the status code should be 404, so
+        // that search engines do not index thousands of empty pages
+        $response = new Response('', $broadcasts ? 200 : 404);
```
votemike: Why not 204 instead of 404?

src/Controller/SchedulesByDayController.php
```diff
@@ -35,24 +33,34 @@ public function __invoke(Pid $pid, ?string $date, ServicesService $servicesServi
...
+        return $this->renderWithChrome('schedules/by_day.html.twig', $viewData, $response);
```
votemike: I feel that if a page is being rendered, it should always be a 200. It feels odd to me to send a template AND a response.

src/Controller/SchedulesByDayController.php
BPScott: 2xx class codes are considered successes, and thus would be added to search engines. We don't want to create search results for days that don't have schedules, like schedules/p00fzl6h/1895-01-01 etc

votemike: Is a [noindex meta tag](https://support.google.com/webmasters/answer/93710?hl=en) not the way to do that properly?

BPScott:
> I honestly preferred the model of a page template and a page presenter. If we split things this way we need to create and call off to a separate presenter/template combo any time there's any significant view logic, which seems unnecessary.

I think it depends upon the sort of page we're rendering, for pages that are more intertwined a single page presenter is better and we should certainly go down that route. The schedule pages however can be hived off into multiple smaller presenters that are stacked. Though looking at this some of those smaller presenters become so trival the almost fade into nothing, and they can be merged into the main presenter. This was made a load clearer when I refactored the base/extends stuff out of existence. Let me play again, with the single page presenter approach, I think it'll be cleaner now we've canned the base/extends bits.


Our original concern was "it's hard to tell what variables are used by what, and if it's particularly sprawly it become hard to be confident about identifying stuff that can be removed/refactored".
Presenters are a way of encapsulating that logic, and reducing mental load by isolating features.
I'm wary of single page presenters that get so big we end up back at the same problem.

Tasiobg:
@BPScott 
There are some know concept about data transformation layer which may be of use, I have used it only for a JSON API and worked fine.
the next link has some info and example of the transformation layer which is what I found interesting
http://fractal.thephpleague.com/
if there is any interest or any question, let me know and we can talk.

src/Controller/SchedulesByDayController.php
BjornTwachtmann: I think a 404 works better here. This is esentially an "error this page doesn't exist" with some helpful links back to pages that do exist. Okay, those links look a lot like our actual schedule pages, but I don't see that as problematic.

BPScott:
Blaaaarge!

We split regionList and NetworkServicesList into separate presenters because they needed to be rendered rendered independently of the central bit of the page. Refactoring to not need the separate no_schedule template (and instead handle that as part of the page presenter) means we can roll those other two presenters back into the main page presenter.

The page presenter is a little chunky but there's not much complex logic in there, so that's acceptable.

Having the page presenter template reference something in app/Resources is kinda rubbish but I'm not sure where that can go, if we want that logic if we do the month/year views. Perhaps I should copy that into the page presenter and we can deal with where a common area can be later.

src/Ds2013/Page/Schedules/ByDayPage/schedules_by_day_page.html.twig
```diff
@@ -20,27 +45,56 @@
...
+                            {{ ds2013('broadcast', broadcast, {
+                                'highlight_box_classes' : 'highlight-box--list'
+                            }) }}
+                        </li>
+                    {% endfor %}
+                </ol>
+            </li>
+        {% endfor %}
+    </ol>
+{% else %}
```
votemike: Do we not want this logic in the controller instead of a template?

PR #65, opened 2017-06-06
Adding error pages

votemike:
In order to test, visit http://sandbox.bbc.co.uk/_error/404 and http://sandbox.bbc.co.uk/_error/500


programmes-developers:

app/Resources/TwigBundle/views/Exception/error.html.twig
```diff
@@ -0,0 +1,16 @@
+{% extends 'base_ds2013.html.twig' %}
+
+{% block title %}{{ tr('page_not_found_title') }}{% endblock %}
```
BPScott: Is the error messaging different for 500s? "Internal server error" or something along those lines.

See also the lines below.

src/Controller/ExceptionController.php
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public function __invoke(Request $request, FlattenException $exception, DebugLoggerInterface $logger = null)
+    {
+        $currentContent = $this->getAndCleanOutputBuffering($request->headers->get('X-Php-Ob-Level', -1));
+        $showException = $request->attributes->get('showException', $this->debug); // As opposed to an additional parameter, this maintains BC
+
+        $code = $exception->getStatusCode();
+
+        $branding = $this->brandingClient->getContent('br-00001');
```
BPScott: When doing local dev, when we see the expection page there's no point making the calls to branding and orbit, these can probably wrapped in an `if (!$showException){}` to avoid those extra http calls.

app/Resources/TwigBundle/views/Exception/error.html.twig
votemike: Just checked. Yes. Both lines are different for the 500 error... Although in programmes they are not translated. What would you like me to do? Copy them across un-translated? Or update the po files?

src/Controller/ExceptionController.php
votemike: Done

app/Resources/TwigBundle/views/Exception/error.html.twig
BPScott: Untranslated raw text is fine, as we don't localise the exception pages.

PR #66, opened 2017-06-08
Broadcast day

BPScott:
Extract Broadcast Day logic into its own value object

Given a DateTime, BroadcastDay will tell you the start/end times of the
broadcast day that point in time fits within.

### DIFF
--- a/src/Controller/SchedulesByDayController.php
+++ b/src/Controller/SchedulesByDayController.php
@@ -62,12 +62,12 @@ class SchedulesByDayController extends BaseController
 
     private function viewData(
         Service $service,
-        Chronos $startDateTime,
+        Chronos $broadcastDayStart,
         SchedulesByDayPagePresenter $pagePresenter
     ): array {
         return [
             'service' => $service,
-            'start_date' => $startDateTime,
+            'broadcast_day_start' => $broadcastDayStart,
             'page_presenter' => $pagePresenter,
         ];
     }
--- a/src/Controller/SchedulesByDayController.php
+++ b/src/Controller/SchedulesByDayController.php
@@ -4,6 +4,7 @@ namespace App\Controller;
 
 use App\Ds2013\Page\Schedules\ByDayPage\SchedulesByDayPagePresenter;
 use App\Ds2013\PresenterFactory;
+use App\ValueObject\BroadcastDay;
 use BBC\ProgrammesPagesService\Domain\ApplicationTime;
 use BBC\ProgrammesPagesService\Domain\Entity\Broadcast;
 use BBC\ProgrammesPagesService\Domain\Entity\Service;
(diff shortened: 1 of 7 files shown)


programmes-developers:

src/Controller/SchedulesByDayController.php
```diff
@@ -23,34 +24,46 @@ public function __invoke(
...
+            $broadcastDay->start(),
             $broadcasts,
             $date,
             $servicesInNetwork
         );
 
         $viewData = $this->viewData(
             $service,
-            $startDateTime,
+            $broadcastDay->start(),
```
votemike: I realise this is just a change from before, but why does `$broadcastDay->start()` need to go into the page present and the view data (which already contains `$broadcastDay->start()` as part of the page presenter)

BjornTwachtmann:
Looks reasonable. That class may start looking less reasonable when we add the rest of the crazy-arse logic too it, but at least it's in one place.

src/Controller/SchedulesByDayController.php
```diff
@@ -62,55 +75,24 @@ public function __invoke(
...
-            $startDateTime = $startDateTime->addHours($tvOffsetHours);
-            $scheduleHours = '24';
-        }
-
-        $endDateTime = $startDateTime->addHours($scheduleHours);
-
-        return [$startDateTime, $endDateTime];
+        // If a date has been provided, use the broadcast date for midday on
+        // the given date
+        return Chronos::createFromFormat('Y-m-d H:i:s|', $dateString . '12:00:00', 'Europe/London');
```
votemike: You don't need the `|`

src/ValueObject/BroadcastDay.php
```diff
@@ -0,0 +1,67 @@
...
+ */
+class BroadcastDay
+{
+    /** @var Chronos */
+    private $start;
+
+    /** @var Chronos */
+    private $end;
+
+    public function __construct(Chronos $date, string $networkMedium)
```
votemike: It's a datetime, not a date.

src/ValueObject/BroadcastDay.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        if (!in_array($networkMedium, [NetworkMediumEnum::RADIO, NetworkMediumEnum::TV, NetworkMediumEnum::UNKNOWN])) {
```
votemike: Should this not use your new `validValues()` method?

src/ValueObject/BroadcastDay.php
```diff
@@ -0,0 +1,67 @@
...
+                'Called new BroadcastDay() with an invalid networkMedium. Expected one of %s but got "%s"',
+                '"' . implode('", "', [NetworkMediumEnum::RADIO, NetworkMediumEnum::TV, NetworkMediumEnum::UNKNOWN]) . '"',
+                $networkMedium
+            ));
+        }
+
+        if ($networkMedium == NetworkMediumEnum::TV) {
+            // If the time is before 6am, then the broadcast day begins at 6am
+            // on the previous day
+            if ($date->lt($date->setTime(6, 0, 0))) {
```
votemike: `$date->setTime(6, 0, 0)` will create a new Chronos object won't it? Why not just check `$data->hour < 6`?

src/ValueObject/BroadcastDay.php
```diff
@@ -0,0 +1,67 @@
...
+                $date = $date->addDays(-1);
```
votemike: I don't know if `$date->subDays(1);` would be more readable?

src/Controller/SchedulesByDayController.php
BPScott: I think `by_day.html.twig` (and other templates within `app/Resources`) should only reference the top level template variables that have been passed into the template, rather than reaching into the presenters that have been passed in.

I'm wary of some potential refactoring where someone removes the need for the start date within the Presenter, and assumes "these getters will only be used within the template for this presenter" (which is a reasonable assumption imo). At which point the page blows up because the `by_day.html.twig` was referencing that getter too.

A bit of repetition to reduce potential for unexpected surprises.

src/Controller/SchedulesByDayController.php
```diff
@@ -23,34 +24,46 @@ public function __invoke(
...
 
         $broadcasts = [];
 
-        if ($service->isActiveAt($startDateTime)) {
+        if ($service->isActiveAt($broadcastDay->start())) {
             // Get broadcasts in relevant period
-            $broadcasts = $broadcastService->findByServiceAndDateRange($service->getSid(), $startDateTime, $endDateTime);
+            $broadcasts = $broadcastService->findByServiceAndDateRange(
+                $service->getSid(),
+                $broadcastDay->start(),
```
votemike: Maybe just pass in the `$broadcastDay`?

votemike: Actually, `findByServiceAndDateRange()` is a pages service thing isn't it, so that's not a great idea.

src/ValueObject/BroadcastDay.php
BPScott: Yes it should (and doing this here was what galvanised me to make that PR). I'm too lazy to cut a release at the moment though as I've going to be adding some more stuff to pages-service relating to BroadcastGaps. I'll cut the release and update this to use the validValues() once my BroadcastGap stuff goes into pages-service.

src/Controller/SchedulesByDayController.php
BPScott: There may be other cases where we want a list of broadcasts between as pair of datetimes that aren't the BroadcastDay start/end times. It's useful to keep the Services functions fairly generic so they can handle multiple cases.

PR #67, opened 2017-06-08
Add example for how to Freeze "now" for testing

BPScott:
We don't need to do this in the general case, as the first call to ApplicationTime::getTime() shall fixture the time to "now".


programmes-developers:

PR #68, opened 2017-06-08
Add Broadcast gaps when there are spaces in schedules

BPScott:
To preview: https://sandbox.bbc.co.uk/schedules/p00fzl7h/2013-03-15

### DIFF
--- a/src/Ds2013/Organism/Broadcast/BroadcastPresenter.php
+++ b/src/Ds2013/Organism/Broadcast/BroadcastPresenter.php
@@ -4,7 +4,9 @@ namespace App\Ds2013\Organism\Broadcast;
 
 use App\Ds2013\Presenter;
 use BBC\ProgrammesPagesService\Domain\Entity\Broadcast;
+use BBC\ProgrammesPagesService\Domain\Entity\BroadcastGap;
 use Cake\Chronos\Chronos;
+use InvalidArgumentException;
 
 class BroadcastPresenter extends Presenter
 {
@@ -16,15 +18,24 @@ class BroadcastPresenter extends Presenter
         'steal_blocklink' => true,
     ];
 
-    /** @var Broadcast */
+    /** @var Broadcast|BroadcastGap */
     private $broadcast;
 
     private $now;
 
     public function __construct(
-        Broadcast $broadcast,
+        $broadcast,
         array $options = []
     ) {
+        if (!($broadcast instanceof Broadcast || $broadcast instanceof BroadcastGap)) {
+            throw new InvalidArgumentException(sprintf(
+                'Expected $broadcast to be an instance of "%s" or "%s". Found instance of "%s"',
+                Broadcast::CLASS,
+                BroadcastGap::CLASS,
+                (is_object($broadcast) ? get_class($broadcast) : gettype($broadcast))
+            ));
+        }
+
         parent::__construct($options);
         $this->broadcast = $broadcast;
         $this->now = Chronos::now('Europe/London');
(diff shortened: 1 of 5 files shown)


programmes-developers:

src/Ds2013/Organism/Broadcast/BroadcastPresenter.php
```diff
@@ -16,15 +18,24 @@ class BroadcastPresenter extends Presenter
...
     private $broadcast;
 
     private $now;
 
     public function __construct(
-        Broadcast $broadcast,
+        $broadcast,
         array $options = []
     ) {
+        if (!($broadcast instanceof Broadcast || $broadcast instanceof BroadcastGap)) {
```
votemike: Would it be worth making them both extend an interface or base class?

src/Ds2013/Organism/Broadcast/broadcast.html.twig
```diff
@@ -29,20 +29,28 @@
...
         </div>
         <div class="broadcast__programme grid {{ programme_classes }}" rev="publication">
-            <div class="grid__inner">
-
-            </div>
+            {% if broadcast.programmeItem %}
+                {# TODO Render Programme Presenter #}
+            {% else %}
+                <p class="text--subtle centi" data-timezone="true" content="{{ broadcast.endAt|date('c') }}">
+                    {{ tr('programmes_resume_at', {'%1': '<span class="timezone--time">' ~ broadcast.endAt|date('H:i') ~ '</span>'})|raw }}
```
votemike: Do we have to send HTML though to a translator?

src/Ds2013/Organism/Broadcast/BroadcastPresenter.php
BPScott: Had a chat with @BjornTwachtmann about this yesterday, we came to the conclusion that there's not much point as later we have to do checks on "is it this class or is it that class", which misses the point of the interface (the point being that you don't shouldn't need any further typechecks). Rather than have a not-that-useful interface, he convinced me to go down this route instead.

src/Ds2013/Organism/Broadcast/broadcast.html.twig
BPScott: It's a bit grim but yes. We want to wrap the time within a span (for the timezone fiddling JS that we've not ported across yet). The time is a placeholder within the translation string, because word ordering may change between languages.

BjornTwachtmann:
All looks pretty sensible to me.

src/Ds2013/Organism/Broadcast/broadcast.html.twig
```diff
@@ -29,20 +29,28 @@
 
             {#end time and zone#}
             <meta property="endDate" datatype="xsd:dateTime" content="{{ broadcast.endAt|date('c')}}" />
-            <span property="publishedOn" typeof="BroadcastService" resource="{{ path('schedules_by_day', {'pid' : broadcast.servicePid }) }}>">
+            <span property="publishedOn" typeof="BroadcastService" resource="{{ path('schedules_by_day', {'pid' : broadcast.servicePid }) }}">
                 <meta property="name" content="{{ broadcast.serviceName }}" />
             </span>
 
-            {% if broadcast.isOnAirNow %}
+            {% if not broadcast.programmeItem %}
```
votemike: `if not broadcast.programmeItem` isn't massively clear at first glance. Would `instanceof` be better?

tatitati:
looks good 👍

src/Ds2013/Organism/Broadcast/broadcast.html.twig
BPScott: Can't do instanceof in a twig template.

BPScott: You can't flick the ordering of the clauses around either as we always want to show this "off-air" message at the off air time

PR #69, opened 2017-06-09
Adding Meta info

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5634


programmes-developers:

src/ValueObject/MetaContext.php
```diff
@@ -0,0 +1,85 @@
...
+
+    /** @var string */
+    private $socialDescription = '';
+
+    public function __construct($context)
+    {
+        if ($context instanceof Programme) {
+            $this->description = $context->getShortSynopsis();
+            $this->image = $context->getParent()->getImage();
+            $this->isRadio = $context->getMasterBrand()->getNetwork()->isRadio();
```
BjornTwachtmann: I think network (and possibly masterbrand) are nullable, I've actually added isTv() and isRadio() methods to pages service as part of the programme-object work, see https://github.com/bbc/programmes-pages-service/compare/programme-tv-radio

src/ValueObject/MetaContext.php
```diff
@@ -0,0 +1,85 @@
...
+            $this->keywords = $context->getTitle();
+            $this->titlePrefix = $context->getMasterBrand()->getName();
+        } elseif ($context instanceof Service) {
+            $this->image = $context->getNetwork()->getImage();
+            $this->isRadio = $context->getNetwork()->isRadio();
```
BjornTwachtmann: I think there should be isTv and isRadio on the service object

BjornTwachtmann:
Added a couple of minor comments, but I like it, looks like a good approach to take with this.

votemike:
Could a pull request be created for that programmes-pages-service work so that it can be merged in to master and a tag created?

app/Resources/views/base_ds2013.html.twig
```diff
@@ -1,10 +1,26 @@
...
+        <meta name="keywords" content="{{ metaContext.keywords }}">
+        <meta name="twitter:card" content="summary_large_image">
+        <meta property="og:site_name" content="BBC">
+        <meta property="og:title" content="{{ block('title') }}">
+        <meta property="og:type" content="website">
+        <link rel="canonical" href="{{ canonical }}">
+        <meta property="og:url" content="{{ canonical }}">
+        <meta property="og:description" content="{{ metaContext.socialDescription ?: metaContext.description }}">
+        <meta property="og:image" content="{{ metaContext.image.getUrl(1200, 675) }}">
+        <meta name="msapplication-square150x150logo" content="{{ metaContext.image.getUrl(272, 272) }}">
```
votemike: @BPScott Why is the name 150x150, yet we are retrieving a 272x272 image URL?

src/Controller/SchedulesByDayController.php
```diff
@@ -29,7 +29,9 @@ public function __invoke(
             throw $this->createNotFoundException('Invalid date');
         }
 
-        $service = $servicesService->findByPidFull($pid);
+        $this->fixCanonicalUrl();
```
votemike: @BPScott I believe the way I have written `getCanonicalUrl()` doesn't include the query string parameters anyway.  
Should I update `getCanonicalUrl()` so that it does include the params? Or shall I just can rid of the code to "fix" the URL?

Tasiobg:
I haven't test it but code structure looks good

votemike:
Do we want any more data from http://ogp.me/ ?  
I was thinking we should have access to...

- og:image:type
- og:image:width
- og:image:height
- music.radio_station
- video.episode and it's sub-information

Also, do we want the ability to specify [multiple images](http://ogp.me/#array)? Such as for both Episode AND brand?

And then other more specific ones:

- fb:app_id

Also, if no date is passed in, should the title be different, otherwise, if a schedule page without a date is shared to FB, the title will be incorrect within 24 hours.

src/Controller/FindByPidController.php
```diff
@@ -14,6 +14,7 @@ public function __invoke(Pid $pid, ProgrammesService $programmesService)
         // TODO lookup programme or group in one request
         $programme = $programmesService->findByPidFull($pid);
         if ($programme) {
+            $this->context = $programme;
```
BPScott: For brandingId, I've kept the property as private but provided a protected setter to deal with passing it into the base controller. We should be consistent, do you think the setter approach is worth it or is setting properties fine?

src/ValueObject/MetaContext.php
```diff
@@ -0,0 +1,92 @@
...
+    private $isRadio = false;
+
+    /** @var string */
+    private $keywords = '';
+
+    /** @var string */
+    private $titlePrefix = '';
+
+    /** @var string */
+    private $socialDescription = '';
```
BPScott: There doesn't seem to be any way to set this at the moment?

src/ValueObject/MetaContext.php
```diff
@@ -0,0 +1,92 @@
...
+            );
+        }
+    }
+
+    public function description(): string
+    {
+        return $this->description;
+    }
+
+    public function image(): Image
```
BPScott: Thoughts on using having `->imageUrl($size)` rather than exposing the whole domain object?

src/ValueObject/MetaContext.php
```diff
@@ -0,0 +1,92 @@
...
+    {
+        if ($context instanceof Programme) {
+            $this->description = $context->getShortSynopsis();
+            $this->isRadio = $context->isRadio();
+            $this->keywords = $context->getTitle();
+            if ($context->getParent()) {
+                $this->image = $context->getParent()->getImage();
+            }
+            if ($context->getMasterBrand()) {
+                $this->titlePrefix = $context->getMasterBrand()->getName();
```
BPScott: Use the network name rather than the masterbrand. This avoids showing "BBC One Scotland" in the cases where the masterbrand is `bbc_one_scotland` (yay River City)

src/ValueObject/MetaContext.php
```diff
@@ -0,0 +1,92 @@
...
+    /** @var string */
+    private $socialDescription = '';
+
+    public function __construct($context)
+    {
+        if ($context instanceof Programme) {
+            $this->description = $context->getShortSynopsis();
+            $this->isRadio = $context->isRadio();
+            $this->keywords = $context->getTitle();
+            if ($context->getParent()) {
```
BPScott: No need to go to the parent. Use $context->getImage().

Consider film cases such as The Incredibles, that don't have a parent programme

src/ValueObject/MetaContext.php
votemike: Ahh yes.... What should it be?

src/ValueObject/MetaContext.php
votemike: I'm sure I did have it just as `$context->getImage()` but that wasn't correct for some reason. Maybe I'm mistaken, will change it.

src/ValueObject/MetaContext.php
votemike: Not keen. As soon as we do that, we'll need some extra info from the Image. Then we'll end up with a load of accessors in related models.

app/Resources/views/base_ds2013.html.twig
votemike: What we we like to do about the others? Do you have a list of the ones you'd like me to add?

src/ValueObject/MetaContext.php
BPScott: Cool, happy with that rational :)

src/ValueObject/MetaContext.php
BPScott: Looking at where this is used currently and when it differs from the meta description (search `social_description` and `blqDescription` in the progs codebase) - it seems the only time social description is different to the meta description is when you're viewing a gallery image, which seems like an oversight/bug.

In all the other cases we can pass in some sort of new context - Group / Image / Article / Profile and set the description based off that. These domain objects don't exist in the new world yet so they can be ignored for now.

Get rid of the socialDescription value, and always set the og:description tag to MetaContext->description.

PR #70, opened 2017-06-12
Presenter Template tests

BPScott:
A framework for creating tests of presenter templates for when we want
to test the logic inside them. Most of the time we should be able to
test the Presenter directly, but there are some cases where we want to
assert on the output of view.

Test broadcast to see some examples.

### DIFF
--- a/tests/BaseTemplateTestCase.php
+++ b/tests/BaseTemplateTestCase.php
@@ -1,0 +1,61 @@
+<?php
+declare(strict_types = 1);
+namespace Tests\App;
+
+use App\Ds2013\Presenter;
+use Symfony\Component\DomCrawler\Crawler;
+use Twig_Environment;
+use PHPUnit\Framework\TestCase;
+
+abstract class BaseTemplateTestCase extends TestCase
+{
+    /** @var Twig_Environment */
+    static private $twig;
+
+    public static function setUpBeforeClass()
+    {
+        if (self::$twig === null) {
+            self::$twig = TwigEnvironmentProvider::twig();
+        }
+    }
+
+    /**
+     * Get a Dom Crawler populated with a given Twig template.
+     *
+     * @param string $name    The template name
+     * @param array  $context An array of parameters to pass to the template
+     *
+     * @return Crawler The Dom Crawler populated with the twig template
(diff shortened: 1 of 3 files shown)


programmes-developers:

tests/BaseTemplateTestCase.php
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    protected function presenterCrawler(Presenter $presenter): Crawler
+    {
+        return $this->crawler($presenter->getTemplatePath(), [
+            $presenter->getTemplateVariableName() => $presenter,
+        ]);
+    }
+
+    protected function assertSchemaOrgItem($expected, $node)
```
votemike: What is the type of `$node`?

tests/TwigEnvironmentProvider.php
```diff
@@ -0,0 +1,84 @@
...
+        if (self::$twig === null) {
+            self::$twig = self::buildTwig();
+        }
+
+        return self::$twig;
+    }
+
+    private static function buildTwig(): Twig_Environment
+    {
+        $loader = new Twig_Loader_Filesystem(__DIR__ . '/../app/Resources');
```
votemike: Are these paths not defined by a variable somewhere?

BPScott: Nothing that we've got access to it a plain-old unit test.

Much of this config already exists in the container config, but building the container is expensive (for a unittest) and does a bunch of other unneeded stuff, so I would like to avoid it to keep things as fast as is reasonable

BjornTwachtmann:
Christ on a cracker, that TwigEnvironmentProvider is a bit hairy. 

I think it's good to have these, it's probably better to make light use of them rather than heavy use though. These kinds of tests are useful, but can be brittle and hard to modify.

I'm fine with merging this

tests/BaseTemplateTestCase.php
BPScott: Crawler, Have added typehints for these functions

BPScott:
Agreed on making light use of these and relying on boring old tests of the Presenter side where possible.

PR #71, opened 2017-06-14
Call the method for getting a localised date 'localDate'

BPScott:
Previously it was dateFormat, but that does a really bad job at making
it clear that the function returns a localised date (and thus takes the format in a different style to the date() filter).


programmes-developers:

PR #73, opened 2017-06-28
Adjustments to the schedules home page

BPScott:
* Only show services that have a start date (thus hiding some of the
  weirder 'freeview green button' services)
* Simplify logic for service rows by using flexbox
* Move the 'decades bar' onto the same line as the group title
* Only show duration bars if flexbox is supported


programmes-developers:

BjornTwachtmann:
Looks good to me

PR #76, opened 2017-07-03
[assets build] - Autoprefixer added

tatitati:
This PR includes:
- autoprefixer with this configuration:
  ->  find compatibility for the **last 3 versions of browsers**
  -> **cascade = false** (it refers to layout in the resulting css, something just visual for the developer)
  -> **remove: false**  (removes old, unnecessary prefixes from your CSS, is false to keep very specific options in our css code, i.e. https://github.com/bbc/programmes-frontend/blob/master/src/Ds2013/Core/Vendor/_normalize.scss#L333 ).

- deleted css for Opera (-o), Microsoft (-ms), Firefox (-moz), Chrome (-webkit)


programmes-developers:

src/Ds2013/Core/Generic/_mixins.scss
```diff
@@ -43,23 +43,13 @@
 }
 
 @mixin transform($transform) {
```
BPScott: Now that we're using autoprefixer there is no point in having these mixins, as they specify a single property. Remove where we're using these mixins and replace them with the transform property.

src/Ds2013/Core/Vendor/_normalize.scss
```diff
@@ -67,8 +67,6 @@ audio:not([controls]) {
 
 html {
     font-size: 100%; /* 1 */
-    -webkit-text-size-adjust: 100%; /* 2 */
```
BPScott: Normalise is a vendored library. I'd be very wary of removing anything from it, as it isn't our code. Can you revert these changes.

gulpfile.js
```diff
@@ -65,6 +66,9 @@ gulp.task('sass', ['sass:clean'], function() {
             'node_modules'
         ]}).on('error', gulpSassError(throwError)))
         .pipe(sourcemaps.write())
+        .pipe(autoprefixer({
```
BPScott: Would doing this transform after writing the sourcemaps mean that all our sourcemap info is wrong?

BPScott:
For the record, here's what "last 3 browsers" means: http://browserl.ist/?q=Last+3+versions

src/Ds2013/Core/Generic/_mixins.scss
```diff
@@ -42,24 +42,10 @@
     zoom: 1;
 }
 
-@mixin transform($transform) {
```
tatitati: Deleted the whole mix as is not being used yet.

PR #78, opened 2017-07-04
Change to disable db cache on int for debugging porpose

tatitati:


programmes-developers:

PR #79, opened 2017-07-05
Ensure all social icons appear in the header

BPScott:
* Rename canonical to canonical_url because nouns are useful
* Rename metaContext to meta_context because we follow the twig
  convention of snake_case variables names.

### DIFF
--- a/app/Resources/views/base_ds2013.html.twig
+++ b/app/Resources/views/base_ds2013.html.twig
@@ -1,23 +1,39 @@
 <!DOCTYPE html>
-<html class="{{ branding.orbitThemeClasses|raw }} {% block html_classes %}{% endblock %}" lang="{{ branding.options.language|default('en') }}">
+<html class="{{ branding.orbitThemeClasses|raw }} {% block html_classes %}{% endblock %}" lang="{{ branding.language|default('en') }}">
     <head>
         <meta charset="UTF-8">
         <title>{% block title %}{% endblock %}</title>
-        <link rel="icon" href="https://www.bbc.co.uk/favicon.ico" type="image/x-icon"/>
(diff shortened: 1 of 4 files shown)


programmes-developers:

app/Resources/views/base_ds2013.html.twig
```diff
@@ -1,23 +1,39 @@
...
+<html class="{{ branding.orbitThemeClasses|raw }} {% block html_classes %}{% endblock %}" lang="{{ branding.language|default('en') }}">
     <head>
         <meta charset="UTF-8">
         <title>{% block title %}{% endblock %}</title>
-        <link rel="icon" href="https://www.bbc.co.uk/favicon.ico" type="image/x-icon"/>
-        <meta name="theme-color" content="{{ branding.colours.highlight.bg|default('#2B2B2B') }}"/>
-        <meta name="description" content="{{ metaContext.description }}">
-        <meta name="twitter:card" content="summary_large_image">
+        <link rel="icon" href="https://www.bbc.co.uk/favicon.ico" type="image/x-icon">
+        <link rel="canonical" href="{{ canonical_url }}">
```
BjornTwachtmann: Is there a reason canonical_url isn't part of meta_context?

BPScott: The canonical url is set based upon the current page route, rather than being built based upon if the current page is rendering a programme/service.

We could pass in the current route into where we [construct MetaContext in the BaseController](https://github.com/bbc/programmes-frontend/blob/7be77a1f83bc0a6db7239f80088acba561ee1a06/src/Controller/BaseController.php#L67-L69), instead of passing two things into the template.

PR #80, opened 2017-07-06
Translate provider

BjornTwachtmann:
Make language for translations settable at runtime.


programmes-developers:

app/config/services.yml
```diff
@@ -80,9 +80,8 @@ services:
               debug: '%kernel.debug%'
               basepath: '%kernel.project_dir%/app/Resources/translations'
 
-    RMP\Translate\Translate:
-        factory: ['@RMP\Translate\TranslateFactory', 'create']
-        arguments: ['%app.default_locale%']
+    RMP\Translate\TranslateProvider:
+        public: true
```
BPScott: I don't think this needs to be public

src/Controller/BaseController.php
```diff
@@ -24,7 +24,7 @@ public static function getSubscribedServices()
             BrandingClient::class,
             OrbitClient::class,
             DesignSystemPresenterExtension::class,
```
BPScott: Don't need to depend on the twig extension if we're no longer referencing it

BjornTwachtmann:
R.e. not passing down into the consuming classes. I don't think half-arsing this is a good idea. We don't know where in the application lifecycle we'll be changing the language and what will have been instantiated by then. What if somebody instantiates a presenter before the basecontroller sets the language? I'd really rather not have to do all of this again.

PR #81, opened 2017-07-06
Branding

felipeparaujo:


programmes-developers:

BPScott:
I think CI is failing due to a problem introduced in Symfony 3.3.3, I've updated to symfony 3.3.4 on master so rebase and it should work

PR #82, opened 2017-07-07
Fallback to a default branding, instead of 500ing

BPScott:
Use a default branding of br-00001 or br-00002 if you end up requesting
a branding ID that does not exist.

I've also worked out how to not use $_GET for the preview parameter -
get the current request from the request_stack that comes from the
container.


programmes-developers:

BjornTwachtmann:
looks good to me

PR #83, opened 2017-07-11
PROGRAMMES-5655 - Config symfony to work behind a STM, load balancer or a reverse proxy

Tasiobg:
When accessing the application through a reverse proxy (e.g. the STMs, Varnish, and ELBs) the client host, client IP, port and protocol is reported as how that proxy accessed the application rather than how the end user accessed the application.

In order to build URLs we need to know the client host, client ip, port and protocol of the end user rather than the intermediate proxy. To do this the STM and ELB both set a set of headers, containing the client information that the application can then consume. The list of headers that the STM provides is available at https://confluence.dev.bbc.co.uk/display/ta/Request+and+response+HTTP+headers+-+standardised+format+for+BBC+traffic+management+services.

Symfony can read X-Forwarded-* headers, [which are a de-facto standard](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto) as per https://symfony.com/doc/current/request/load_balancer_reverse_proxy.html

Unfortunately the headers that Symfony consumes and Varnish sets do not line up, so we have to do some transformation to set the headers that Symfony understands.

As a result of this fix, `$request->getPort()` should return 80 or 443 when accessing the APP through the STM even though the nginx server is listening on port 7443.


programmes-developers:

Tasiobg:
Also added missing schema on "routing_3rd_party.yml", these routes are only HTTP

BPScott:
The `iplayer_play` route should also be http only, but that causes some places where it is used as a `resource` tag to be generated incorrectly (it ends up being `http:/www.bbc.co.ukhttp:/www.bbc.co.uk/iplayer...` as the template is assuming a relative path, but that can wait for a different PR.

web/app.php
```diff
@@ -8,6 +8,27 @@
...
+}
+if (isset($_SERVER['HTTP_X_BBC_EDGE_SCHEME'])) {
+    $_SERVER['HTTP_X_FORWARDED_PROTO'] = $_SERVER['HTTP_X_BBC_EDGE_SCHEME'];
+    $_SERVER['HTTP_X_FORWARDED_PORT']  = $_SERVER['HTTP_X_FORWARDED_PROTO'] == 'https' ? 443 : 80;
+} else {
+    $_SERVER['HTTP_X_FORWARDED_PORT'] = $_SERVER['REQUEST_SCHEME'] == 'https' ? 443 : 80;
+}
+
+// Covers our subnets in dev and live, this could be more specific but that requires maths
+Request::setTrustedProxies(['10.48.0.0/16', '10.49.0.0/16'], Request::HEADER_X_FORWARDED_ALL);
```
BjornTwachtmann: Where are these magic IP address range coming from? If it's cosmos then these go in cosmos config. I'm happy to generalise this to 10.0.0.0/8 as a magic number. But not this.

BPScott: These come from CIDR blocks of our AWS account's subnets (as defined in the core-infra stack), they're still broader than what they could be, but are still a lot more specific than the 10.0.0.0/8 block

PR #84, opened 2017-07-11
Lazysizes library added

tatitati:
**To test lazyload behaviour:**
Just scroll the page and see in dev tools the requests for network/images

**To test responsive sizes:**
With different zooms reload the page. Check that for 'big-zoom out', images are requested with lower resolution than for 'big-zoom in'.
For instance:
With big zoom-out: https://ichef.bbci.co.uk/images/ic/320x180/p057qg1r.jpg
With big zoom-in: https://ichef.bbci.co.uk/images/ic/480x270/p057qg1r.jpg


programmes-developers:

src/Ds2013/Molecule/Image/image.html.twig
```diff
@@ -1,5 +1,5 @@
 {%- if image.option('is_lazy_loaded') -%}
-    <div class="image image--lazy" data-src="{{ image.src }}" data-srcset="{{ image.srcsets }}" data-sizes="{{ image.sizes }}" data-alt="{{ image.option('alt') }}" data-img-blank="{{ asset("images/blank.png") }}"></div>
+    <img class="image lazyload" data-src="{{ image.src }}" data-srcset="{{ image.srcsets }}" data-sizes="auto" data-alt="{{ image.option('alt') }}" data-img-blank="{{ asset("images/blank.png") }}"></img>
```
tatitati: Now data-sizes="auto". Should we update the backend?

gulpfile.js
```diff
@@ -32,13 +32,7 @@ gulp.task('js', ['js:clean'], function () {
...
         "paths": {
             "jquery-1.9": "empty:",
-            "jquery.appear": "../../../node_modules/jquery.appear/jquery.appear"
-        },
-        "shim": {
-            "jquery.appear": {
-                "deps": ["jquery"],
-                "exports": "jquery.appear"
-            }
+            "lazysizes": "../../../node_modules/lazysizes/lazysizes-umd"
```
votemike: Don't we have requireJS or something similar so that we don't have to put these paths in to the node_modules directory?

tatitati: Well, requirejs try to find modules in the way: <name module>.js
In this case the lazysizes modules is coming in two flavours: AMD and non AMD. 
The AMD version is the lazysizes-umd, so in this case must be pointed.

package.json
```diff
@@ -12,12 +12,14 @@
...
-        "jquery.appear": "^1.0.1",
         "sass-mq": "^3.3.2"
     },
     "scripts": {
         "gulp": "gulp default",
         "gulp-ci": "gulp distribution",
         "watch": "gulp watch"
+    },
+    "dependencies": {
+        "lazysizes": "^3.0.0"
```
BPScott: This should be a dev dependency

PR #86, opened 2017-07-12
Link to the http version of iplayer episode pages

BPScott:
Ensure we always use absolute URLs for image and title links in the
programme object as they may link to iplayer which is http only (for now)

### DIFF
--- a/src/Ds2013/Organism/Programme/SubPresenters/ProgrammeTitlePresenter.php
+++ b/src/Ds2013/Organism/Programme/SubPresenters/ProgrammeTitlePresenter.php
@@ -49,7 +49,7 @@ class ProgrammeTitlePresenter extends ProgrammePresenterBase
     public function getTitleLinkUrl(): string
     {
         // Link to iplayer/podcasts will be added here later
-        return $this->router->generate('find_by_pid', ['pid' => $this->programme->getPid()]);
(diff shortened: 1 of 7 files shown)


programmes-developers:

src/Ds2013/Organism/Programme/SubPresenters/ProgrammeTitlePresenter.php
```diff
@@ -49,7 +49,7 @@ public function __construct(
     public function getTitleLinkUrl(): string
     {
         // Link to iplayer/podcasts will be added here later
-        return $this->router->generate('find_by_pid', ['pid' => $this->programme->getPid()]);
+        return $this->router->generate('find_by_pid', ['pid' => $this->programme->getPid()], UrlGeneratorInterface::ABSOLUTE_URL);
```
BjornTwachtmann: Do we need to specify a schema for the find_by_pid route. Could get a bit complex with splitting that route I guess.

BjornTwachtmann:
looks good to me.

src/Ds2013/Organism/Programme/SubPresenters/ProgrammeTitlePresenter.php
BPScott: I think we'd be able to have two named routes, one without a schema and one called "find_by_pid_external" that does have a schema, but then we'd have to write logic to chose which of those we want, and we're going to forget to update that.

The long-term fix is to move the PAL-served stuff over to https.

PR #87, opened 2017-07-13
Moar subpresenters

BjornTwachtmann:
You probably want to view https://github.com/bbc/programmes-frontend/commit/a7690700a901041888417251e7b6fc012cc40493 to understand this PR. I've renamed BroadcastProgramme*Presenter to CollapsedBroadcastProgramme*Presenter which makes the diff rather confusing.


programmes-developers:

PR #88, opened 2017-07-14
PROGRAMMES-5682 - Add headers by default

Tasiobg:
Add response headers
- Update the BaseController to instanciate and use the response object so we can add headers in any controllet using something like `$this->response()->headers->set('header', 'value');`
- Created ResponseSubscriber, this allows us to add required headers to all responses.
- Created unitest assert to check if controllers have the required headers


programmes-developers:

tests/BaseWebTestCase.php
```diff
@@ -24,6 +24,20 @@ public function assertRedirectTo($client, $code, $expectedLocation)
         $this->assertEquals($expectedLocation, $client->getResponse()->headers->get('location'));
     }
 
+    public function assertHasRequiredResponseHeaders($client, $contentLanguage = null, $cacheControl = 'max-age=120, public')
```
BPScott: What are we going to change (and thus need to test) more often - the contentLanguage or the cacheControl? I reckon the item that is less likely to change should be towards the end of the list so that we don't have to always specify it.

I spot that you're currently doing ` $this->assertHasRequiredResponseHeaders($client, null, 'no-cache, private');` in the status controller test. Switching these arguments around would mean you can write  $this->assertHasRequiredResponseHeaders($client, 'no-cache, private');` which might be a bit simpler.

src/Controller/BaseController.php
```diff
@@ -73,7 +83,12 @@ protected function setContext($context)
         }
     }
 
-    protected function renderWithChrome(string $view, array $parameters = [], Response $response = null)
+    protected function response(): Response
```
felipeparaujo: why not getResponse?

src/Controller/BaseController.php
```diff
@@ -40,6 +42,14 @@ public static function getSubscribedServices()
         ]);
     }
 
+    public function __construct()
+    {
+        $this->response = new Response();
+        // It's required to set Cache-Control header before sending the response otherwise symfony will create
+        // the follow headers "Cache-Control: no-cache, private" by default
```
BPScott: > It is required to set the cache-control header when creating the response object otherwise Symfony will create set its value to "no-cache, private" by default

reads a little better IMO

src/EventSubscriber/ResponseSubscriber.php
```diff
@@ -0,0 +1,44 @@
...
+
+        // Always add the following varys to whatever is currently there
+        $response->setVary(['X-CDN', 'X-BBC-Edge-Scheme'], false);
+
+        $response->headers->set('X-UA-Compatible', 'IE=edge');
+
+        $languageCode = $this->translateProvider->getTranslate()->translate('language_code');
+        $response->headers->set('Content-Language', $languageCode);
+
+        $response->headers->set('X-Webapp', 'programmes-frontend');
```
BPScott: It might be useful to put in some comments about what these headers are for as I'll have forgotten by the next time I look at this code

src/EventSubscriber/ResponseSubscriber.php
```diff
@@ -0,0 +1,44 @@
...
+
+        $response->headers->set('X-Cache-Control', 'stale-while-revalidate=30');
```
felipeparaujo: A little bit of doc on what these headers do and what they're for would be good

src/Controller/BaseController.php
BjornTwachtmann: Why a function at all for that matter? Why not just make $this->response protected?

BPScott: BaseController also has `logger()` and `request()`. For when we need bits out of the container (they're private atm, but can be bumped up to protected if we need them).

It seemed a little inconsistent to access some things through properties and others through methods.

BjornTwachtmann: *shrugs*, fair enough then

BjornTwachtmann:
Looks good. Should do what we need.

PR #90, opened 2017-07-17
5653 monthly and yearly schedules

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5653


programmes-developers:

app/Resources/views/schedules/by_month.html.twig
```diff
@@ -0,0 +1,30 @@
...
+{% endblock %}
+
+{%- block dateNav -%}
+    <div id="schedule-monthnav" class="centi br-box-highlight g-f-l">
+        {{ ds2013('dateList', first_of_month, service, {'format': 'month'}) }}
+    </div>
+{% endblock %}
+
+{%- block schedule -%}
+    {% if service.startDate and service.startDate > first_of_month.endOfMonth %}
```
votemike: The year view never seemed to do checks like this. So I haven't implemented them in this PR. Speak now if I should....

src/Controller/SchedulesByMonthController.php
```diff
@@ -0,0 +1,39 @@
...
+        }
+
+        $this->setContext($service);
+
+        $firstOfMonth = Date::createFromFormat('Y-m', $date, ApplicationTime::getLocalTimeZone())->firstOfMonth();
+        $viewData = ['first_of_month' => $firstOfMonth, 'service' => $service,];
+
+        // If there are no broadcasts, then the status code should be 404, so
+        // that search engines do not index thousands of empty pages
+        $response = new Response('', true ? 200 : 404);
```
votemike: @BPScott Do we want to query a whole month for broadcasts?

src/Controller/SchedulesByYearController.php
```diff
@@ -0,0 +1,39 @@
...
+        }
+
+        $this->setContext($service);
+
+        $startOfYear = Date::createFromFormat('Y', $year, ApplicationTime::getLocalTimeZone())->firstOfYear();
+        $viewData = ['start_of_year' => $startOfYear, 'service' => $service,];
+
+        // If there are no broadcasts, then the status code should be 404, so
+        // that search engines do not index thousands of empty pages
+        $response = new Response('', true ? 200 : 404);
```
votemike: @BPScott Do we want to query a whole month for broadcasts?

src/Ds2013/Molecule/DateList/DateListPresenter.php
```diff
@@ -3,38 +3,53 @@
...
     /** @inheritDoc */
     protected $options = [
         'css_classes' => '',
         'user_timezone' => 'GMT',
+        'format' => 'day',
     ];
 
-    /** @var Chronos */
+    /** @var string */
+    private $dateListItemPresenter;
```
Tasiobg: Is this a string or a DateListItemPresenter?

src/Controller/SchedulesByMonthController.php
BPScott: Nah, no point doing a query for broadcasts on that wide a range. I think it should throw a 404 if it outside the startDate - endDate range for the service though.

Set the status code on `$this->response()` ([like this](https://github.com/bbc/programmes-frontend/blob/63ff14be3f975b3e34571f44300db5bfe8887ae2/src/Controller/SchedulesByDayController.php#L83-L85)) rather than creating a new response.

src/Controller/SchedulesByYearController.php
BPScott: Nah, no point doing a query for broadcasts on that wide a range. I think it should throw a 404 if it outside the startDate - endDate range for the service though.

Set the status code on `$this->response()` ([like this](https://github.com/bbc/programmes-frontend/blob/63ff14be3f975b3e34571f44300db5bfe8887ae2/src/Controller/SchedulesByDayController.php#L83-L85)) rather than creating a new response.

app/Resources/views/schedules/by_month.html.twig
BPScott: I reckon keeping it consistent would be handy, if we've got this sort of check in by month and year, then we should do by year too

src/Controller/SchedulesByMonthController.php
```diff
@@ -0,0 +1,38 @@
...
+use Cake\Chronos\Date;
+use Symfony\Component\HttpFoundation\Response;
+
+class SchedulesByMonthController extends BaseController
+{
+    public function __invoke(
+        Pid $pid,
+        string $date,
+        ServicesService $servicesService,
+        BroadcastsService $broadcastService,
```
BPScott: Don't need broadcastService or collapsedBroadcastsService

src/Controller/SchedulesByYearController.php
```diff
@@ -0,0 +1,38 @@
...
+use Cake\Chronos\Date;
+use Symfony\Component\HttpFoundation\Response;
+
+class SchedulesByYearController extends BaseController
+{
+    public function __invoke(
+        Pid $pid,
+        string $year,
+        ServicesService $servicesService,
+        BroadcastsService $broadcastService,
```
BPScott: Don't need broadcastService or collapsedBroadcastsService

BjornTwachtmann:
I think this PR should add the "Calendar" and "This year" links at the top right of the page, see http://www.bbc.co.uk/radio1/programmes/schedules . Unless I've gratuitously missed those?

src/Ds2013/Molecule/DateList/day_date_list_item.html.twig
```diff
@@ -0,0 +1,13 @@
+{% set date_list_item = day_date_list_item %}
+{% set item_attributes = date_list_item.isGmt ? 'data-href-add-utcoffset="beans"' %}
```
BjornTwachtmann: beans? are they magic?

BjornTwachtmann:
Also, all uses of someDate|date('...') should be replaced by someDate|local_date('...') to ensure correct timezone is used. There's quite a few of these but should be quick.

src/Ds2013/Molecule/DateList/date_list_item.html.twig
```diff
@@ -15,28 +13,16 @@
...
     }) %}
 
-    {% set item_attributes = date_list_item.isGmt ? 'data-href-add-utcoffset="beans"' %}
-
     <li class="{{ li_classes }}">
         {% if date_list_item.isLink %}
-            <a href="{{ path('schedules_by_day', {'pid': date_list_item.servicePid, 'date': date_list_item.datetime|local_date('Y-m-d')}) }}"
-               class="{{ item_classes }}" {{ item_attributes|raw }}>
-                {{ self.dateLines(date_list_item.datetime) }}
+            <a href="{{ path(date_list_item.linkName, date_list_item.linkParams) }}"
```
BjornTwachtmann: I think it would make more sense to pass the router into the presenter (this is already done in the programme presenter) and have a getLink() function. This is a bit confusing.

src/Ds2013/Molecule/DateList/month_date_list_item.html.twig
```diff
@@ -0,0 +1,11 @@
+{% set date_list_item = month_date_list_item %}
```
BjornTwachtmann: You actualy don't need this. Override the variable name in the presenter class. See https://github.com/bbc/programmes-frontend/blob/master/src/Ds2013/Organism/Programme/CollapsedBroadcastSubPresenters/CollapsedBroadcastProgrammeBodyPresenter.php#L74

src/Ds2013/Molecule/DateList/day_date_list_item.html.twig
votemike: ¯\_(ツ)_/¯ Code Monkey just copy/paste

https://github.com/bbc/programmes-frontend/blob/f0d575e76ebf1311ae111088ac52774fd3447f90/src/Ds2013/Molecule/DateList/date_list_item.html.twig#L18

votemike:
As far as I was aware, the month and year views were supposed to be scrapped, I wasn't involved in any conversations about bringing it back. @BPScott should the "Calendar" and "This year" links be reinstated as @BjornTwachtmann mentioned?

app/Resources/views/schedules/by_base.html.twig
```diff
@@ -0,0 +1,30 @@
+{% extends 'base_ds2013.html.twig' %}
+
+{%- block title -%}
+    {{ meta_context.titlePrefix }} - {{ tr('schedules') }}, {{ block('date') }}
+{%- endblock -%}
+
+{% block page_classes %}programmes-page text-base programmes-page--flush{% endblock %}
+
+{% block body %}
+    <div class="b-g-p no-margin-vertical islet--vertical br-box-highlight br-highlight-bg-onbg080{{ service.isTv ? ' visually-hidden' }}" id="schedule-header">
```
BPScott: Just had a chat with Cathal - the plan is to remove some of the visual differences between TV and Radio - this bar is one of them. Show this bar at all times rather than hiding it for TV

app/Resources/views/schedules/by_base.html.twig
```diff
@@ -0,0 +1,30 @@
...
+        <h1>
+            <span class="visually-hidden">
+                <span class="outlet">
+                    {{ tr('schedule_title', {'%1': service.name}) }}
+                </span>
+            </span>
+            <time class="date">
+                {{ block('date') }}
```
BPScott: Idiomatic twig is to use block tags rather than the function when specifying block that can be overridden by child templates, so `{% block date %}{% endblock %}`

see https://github.com/bbc/programmes-frontend/blob/master/app/Resources/views/base_ds2013.html.twig#L71 for an example

app/Resources/views/schedules/by_base.html.twig
```diff
@@ -0,0 +1,30 @@
...
+            </time>
+        </h1>
+    </div>
+
+    <nav class="schedule-nav-wrapper">
+        {{ block('dateNav') }}
```
BPScott: Twig naming convention suggests using snake_case for variable names. It doesn't explicitly say anything about block names, but i reckon they should fall under the same bracket.

How about `{% block date_nav %}{% endblock %}`?

app/Resources/views/schedules/by_base.html.twig
```diff
@@ -0,0 +1,30 @@
...
+    </nav>
+
+    <div class="b-g-p br-box-page island--vertical">
+        {{ block('schedule') }}
```
BPScott: {% block schedule %}{% endblock %}

app/Resources/views/schedules/by_month.html.twig
```diff
@@ -0,0 +1,30 @@
...
+        {% set url = path('schedules_by_day', {pid: service.pid, date: service.startDate|local_date('Y-m-d')}) %}
+        <p class="noschedule">{{ tr('schedule_begins_on', {'%1': '<a href="' ~ url ~ '">' ~ service.startDate|local_date_intl('EEEE d MMMM yyyy') ~ '</a>'})|raw }}</p>
+    {% elseif service.endDate and service.endDate < first_of_month.startOfMonth %}
+        {% set url = path('schedules_by_day', {pid: service.pid, date: service.endDate|local_date('Y-m-d')}) %}
+        <p class="noschedule">{{ tr('schedule_ended_on', {'%1': '<a href="' ~ url ~ '">' ~ service.endDate|local_date_intl('EEEE d MMMM yyyy') ~ '</a>'})|raw }}</p>
+    {% else %}
+        <p>{{ tr('schedules_calendar_note') }}</p>
+        <div class="grid-wrapper">
+            <div class="grid 1/2@bpw 1/3@bpw2 1/3@bpe">
+                <div class="grid__inner">
```
BPScott: Don't need the `grid__inner` div in the new world. Previously this was used because having to support IE7 meant we couldn't use box-sizing, now we can, which means this div is not needed in the updated grid system

app/Resources/views/schedules/by_year.html.twig
```diff
@@ -0,0 +1,32 @@
...
+    <p class="noschedule">{{ tr('schedule_begins_on', {'%1': '<a href="' ~ url ~ '">' ~ service.startDate|local_date_intl('EEEE d MMMM yyyy') ~ '</a>'})|raw }}</p>
+    {% elseif service.endDate and service.endDate < start_of_year %}
+        {% set url = path('schedules_by_day', {pid: service.pid, date: service.endDate|local_date('Y-m-d')}) %}
+    <p class="noschedule">{{ tr('schedule_ended_on', {'%1': '<a href="' ~ url ~ '">' ~ service.endDate|local_date_intl('EEEE d MMMM yyyy') ~ '</a>'})|raw }}</p>
+    {% else %}
+        <div class="grid-wrapper">
+            {%- for offset in 0..11 -%}
+                {% set first_of_month = start_of_year.addMonths(offset) %}
+                <div class="grid 1/2@bpw 1/3@bpw2 1/3@bpe">
+                    <div class="grid__inner">
```
BPScott: The `.grid__inner` is now unneeded (see prior comment)

src/Ds2013/Molecule/Calendar/_calendar-table.scss
```diff
@@ -0,0 +1,26 @@
...
+        text-transform: uppercase;
+    }
+    td {
+        width: percentage(1/7);
+        border-width: 1px;
+        border-style: solid;
+    }
+}
+
+@include mq-range ('bpw') {
```
BPScott: No need for the space between the mixin name and the args `mq-range('bpw')`

src/Ds2013/Molecule/DateList/DateListItemPresenter.php
```diff
@@ -6,33 +6,36 @@
 use App\Ds2013\Presenter;
 use BBC\ProgrammesPagesService\Domain\Entity\Service;
 use Cake\Chronos\Chronos;
+use Cake\Chronos\ChronosInterface;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 
-class DateListItemPresenter extends Presenter
+abstract class DateListItemPresenter extends Presenter
```
BPScott: In PagesService we prefixed the name of abstract classes with `Abstract` ([per Symfony's naming conventions](https://symfony.com/doc/current/contributing/code/standards.html#naming-conventions)). Is that a convention we want to carry on with?

votemike: I'm not a fan of it. But I don't feel strongly enough to go against it. I feel it just makes the name longer to type without adding any actual value.  
I'll change it.

app/Resources/views/schedules/by_base.html.twig
votemike: Left as is so that the template acts as an abstract class and REQUIRES the date block be defined.

BPScott:
Yeah, put the Calendar / This Year links in. I'm not a big fan of sending traffic to these pages but thats what we do currently

votemike:
Have added the Calendar and This Year links in. This Year was a list item, but I changed it to a paragraph and changed some classes, I hope that's OK.

votemike:
@BPScott I have added some tests. Not sure how to test templates though.

djmarland:
Being nosey and saw these coming back. I guess that means the URLs should be back to 
`yyyy/mm/dd` rather than `yyyy-mm-dd` then?

votemike:
Sounds sensible to me....

BPScott:
@djmarland let it go already ;)

TBH I quite like jumping straight up a level rather than trawling through the month/year views.

PR #91, opened 2017-07-17
X-Content-Type-Options and X-Frame-Options

Tasiobg:
PROGRAMMES-5712
Added 2 new default headers
- X-Content-Type-Options
- X-Frame-Options

Added into the BaseController construct so we can unset the header when needed.


programmes-developers:

BjornTwachtmann:
looks good to me

PR #92, opened 2017-07-18
Adding BACK TO TOP links

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5702


programmes-developers:

src/Ds2013/Page/Schedules/ByDayPage/schedules_by_day_page.html.twig
```diff
@@ -48,8 +47,15 @@
 {% if schedules_by_day_page.hasBroadcasts %}
     <ol class="list-unstyled g-c-l">
         {% set onairshown = false %}
+        {% set showBackToTop = false %}
         {% for period_name, broadcast_in_period in schedules_by_day_page.BroadcastsGroupedByPeriodOfDay %}
             <li id="{{ period_name }}" class="schedule-skip-link">
+                {% if showBackToTop %}
+                    <a class="pull--right micro link-complex"
+                       href="#top">{{ gelicon('basics', 'to-top', 'centi', 'leading') }}<span
```
BPScott: Where's the id="top" element that you're linking to?

votemike: Seems to magically be there, possibly injected by the branding tool? `<span class="radionav-drawer__title top b-g-m">Stations</span>`

votemike: Actually, that's a class isn't it. In which case it looks like the LIVE site doesn't have an element with the id top either.  
How would you like to proceed?

votemike: Change it to just `#`?

BPScott: See if there's a container div somewhere that is suitable for having that ID

PR #94, opened 2017-07-18
Fixing bug with timezones

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5707


programmes-developers:

BPScott:
Nice. Should we have a controller test that checks the messages on these start/end days?

votemike:
That would mean testing the view. Right? How does one test the view?

votemike:
Added tests

PR #95, opened 2017-07-19
Add BrandingPlaceHolderResolver

BPScott:
This deals with resolving branding placeholders:

`<!--BRANDING_PLACEHOLDER_TITLE-->` is replaced with the programem title
`<!--BRANDING_PLACEHOLDER_NAV-->` is replaced with a default navigation

`<!--BRANDING_PLACEHOLDER_SPONSOR-->` is currently replaced with an empty
string, as we do not yet have anything we need to pass into it.

Currently this only supports having Programmes as the context, this
shall need to deal with Groups once we have Group domain models.

See /programmes/b08y249y for an example of this in action.

### DIFF
--- a/src/Branding/BrandingPlaceholderResolver.php
+++ b/src/Branding/BrandingPlaceholderResolver.php
@@ -1,0 +1,175 @@
+<?php
+declare(strict_types = 1);
+namespace App\Branding;
+
+use App\Translate\TranslateProvider;
+use BBC\BrandingClient\Branding;
+use BBC\ProgrammesPagesService\Domain\Entity\Episode;
+use BBC\ProgrammesPagesService\Domain\Entity\Programme;
+use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeContainer;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+
+/**
+ * Consider "I'm Sorry I Haven't a Clue" - it is a Radio 4 programme. Radio
+ * programmes tend not to have bespoke per-brand theming, instead they inherit
+ * their theme from their Service (in this case Radio 4). This however poses
+ * a small problem - if we display the Service theme, how shall we show the
+ * correct per-programme Title and Navigation?
+ *
+ * This is solved by the Branding Tool outputting templates that contain
+ * placeholder sections such as  <!--BRANDING_PLACEHOLDER_TITLE--> and
+ * <!--BRANDING_PLACEHOLDER_NAV--> when there is not programme set.
+ *
+ * This takes a Branding instance that contains those placeholders and a
+ * "context" and replaces those placeholders with a title and default
+ * navigation, based up the contents of the context.
+ */
+class BrandingPlaceholderResolver
+{
+    private const PLACEHOLDER_TITLE = '<!--BRANDING_PLACEHOLDER_TITLE-->';
+    private const PLACEHOLDER_NAV = '<!--BRANDING_PLACEHOLDER_NAV-->';
+    private const PLACEHOLDER_SPONSOR = '<!--BRANDING_PLACEHOLDER_SPONSOR-->';
+
+    /** @var UrlGeneratorInterface */
+    private $router;
+
+    /** @var TranslateProvider */
+    private $translateProvider;
+
+    public function __construct(
+        UrlGeneratorInterface $router,
+        TranslateProvider $translateProvider
+    ) {
+        $this->router = $router;
+        $this->translateProvider = $translateProvider;
+    }
+
+    public function resolve(Branding $branding, $context): Branding
+    {
+        // If the context is not a Programme or Group then don't attempt to resolve
+        if (!($context instanceof Programme /*|| $context instanceof Group*/)) {
+            return $branding;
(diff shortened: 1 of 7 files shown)


programmes-developers:

src/Branding/BrandingPlaceholderResolver.php
```diff
@@ -0,0 +1,164 @@
...
+            $branding->getBodyLast(),
+            $branding->getColours(),
+            $branding->getOptions()
+        );
+    }
+
+    private function resolvePlaceholders(Branding $branding, $context, string $haystack)
+    {
+        // Title
+        if (strpos($haystack, self::PLACEHOLDER_TITLE) !== 0) {
```
votemike: Why do you need to do this `strpos` check?

src/Branding/BrandingPlaceholderResolver.php
```diff
@@ -0,0 +1,164 @@
...
+        $navItems = [];
+
+        $hasEpisodes = false;
+        $hasClips = false;
+        $hasGalleries = false;
+
+        // Home link is always present
+        $navItems[] = $branding->buildNavItem(
+            $translate->translate('home'),
+            '/programmes/' . $tleo->getPid(),
```
votemike: Why is `/programmes/` hard coded and not using routing?

src/Branding/BrandingPlaceholderResolver.php
```diff
@@ -0,0 +1,164 @@
...
+                'nav_galleries'
+            );
+        }
+
+        return implode('', $navItems);
+    }
+
+    private function buildSponsor($context): string
+    {
+        // TODO
```
votemike: Link to ticket to do this?

src/Branding/BrandingPlaceholderResolver.php
BPScott: These placeholders may not be present in the returned branding (e.g. Doctor Who has a context set in its branding so it has the title in the header rather than a placeholder https://branding.files.bbci.co.uk/branding/live/projects/br-09041.json)

I want to avoid spending unneeded effort working out what the replacements should be if the placeholder isn't there

src/Branding/BrandingPlaceholderResolver.php
BPScott: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5723 was the ticket to do all of this, I've left a comment on it stating this needs to be finished. I'm thinking we can leave that ticket where it is, and finish it + test this all in a few sprints time.

src/Branding/BrandingPlaceholderResolver.php
BPScott: Currently pure lazyness as i wasn't initially sure about where this should live / if I'd need any dependencies. It turns out I've ended up needing to DI the TranslateProvider into the constructor so adding the router would be simple too. Do you reckon it's worth it?

src/Branding/BrandingPlaceholderResolver.php
BjornTwachtmann: I agree this strpos check is redundant. If self::PLACEHOLDER_TITLE is not present in the input, the input should simply be returned unmodified. A single call to str_replace should be as efficient as a single call to strpos. Unless I'm missing something.

src/Branding/BrandingPlaceholderResolver.php
```diff
@@ -0,0 +1,164 @@
...
+                $haystack
+            );
+        }
+
+        return $haystack;
+    }
+
+    private function buildTitle($context): string
+    {
+        return sprintf(
```
BjornTwachtmann: Should we be building URLs like this or using the router?

src/Branding/BrandingPlaceholderResolver.php
BjornTwachtmann: I would add the router. Consistency is good here, and fishing around in a bunch of text if we ever need to change the routing would be annoying.

BPScott: 👍 :

src/Branding/BrandingPlaceholderResolver.php
BPScott: The call to `str_replace` is cheap, it's the effort of working out `$this->buildTitle()` (which for title is easy, but for buildNav is a bit more effort) that I want to avoid.

I don't think there's some other way that says "only work out the the value that gets injected into the string if it finds a replacement that needs to occur"

src/Branding/BrandingPlaceholderResolver.php
```diff
@@ -0,0 +1,171 @@
...
+            '<a href="/programmes/%s">%s</a>',
```
BjornTwachtmann: This URL is wrong, no?

BPScott: Whoops / fixed

src/Branding/BrandingPlaceholderResolver.php
votemike: Maybe put a comment to say that?

PR #96, opened 2017-07-20
Remove tomorrow bar if service is not active

votemike:


programmes-developers:

PR #97, opened 2017-07-20
Removing period headings from International pages

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5703

I haven't touched the `<li>` with the class `schedule-skip-link`. This means there is a small gap between periods, but I think the gap is preferable to over-complicating the template logic.


programmes-developers:

PR #98, opened 2017-07-20
Add additional prod_int and prod_test environments

BPScott:
These environments shall be used on INT and TEST cosmos respectively.
LIVE cosmos shall use the `prod` environment.

Update the bake script to setup these environments. The cache warmup uses the right env, and the environment value in the front controller is overwritten.

Let's talk about naming for this lot tomorrow as it's rather contentious.


programmes-developers:

PR #99, opened 2017-07-21
Adding channels nav to schedule page

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5698

Do we want to add in the HTML5 `<nav>` tag anywhere in this PR?


programmes-developers:

src/Ds2013/Core/Modules/Schedules/_tv-channels.scss
```diff
@@ -0,0 +1,169 @@
...
+}
+
+.tv-channels__listitem {
+    vertical-align: top;
+}
+
+.tv-channels__item {
+    width: 96px;
+    height: 41px;
+    background-image: image-url('../images/logos/services_channel_sprite.png');
```
votemike: Seems a bit weird I have to do `../`, but I couldn't get it to work any other way.

src/Ds2013/Core/Modules/Schedules/_tv-channels.scss
```diff
@@ -0,0 +1,169 @@
...
+
+.tv-channels__item--bbc_one {
+    background-position: 0 0;
+}
+
+.tv-channels__item--bbc_two {
+    background-position: 0 -41px;
+}
+
+.tv-channels__item--bbc_three {
```
votemike: Do we need this any more?

src/Ds2013/Core/Modules/Schedules/_tv-channels.scss
```diff
@@ -0,0 +1,169 @@
...
+}
+
+.tv-channels__item--bbc_two.tv-channels__item--active,
+.tv-channels__item--bbc_two:active,
+.tv-channels__item--bbc_two:focus,
+.tv-channels__item--bbc_two:hover {
+    background-position: -96px -41px;
+}
+
+.tv-channels__item--bbc_three.tv-channels__item--active,
```
votemike: Do we need this any more?

src/Ds2013/Core/Modules/Schedules/_tv-channels.scss
Tasiobg: This is because "image-url" is custom function defined inside _functions.scss, you could fix this by replacing this variable `$image-url-path: "/assets/images/" !default;` and then use `background-image: image-url('logos/services_channel_sprite.png');` 

This requires test to make sure s3 deployment works fine.

On the other hand I'm not sure if "image-url" is better aproach than using normal css like: `background-image: url('/assets/images/logos/services_channel_sprite.png');`

src/Ds2013/Core/Modules/Schedules/_tv-channels.scss
BPScott: Something for Cathal to decide

src/Controller/SchedulesByDayController.php
```diff
@@ -42,6 +43,15 @@ public function __invoke(
             $broadcastDay->start()
         );
 
+        $otherServices = [];
+
+        if ($service->isTv()) {
+            $otherServices = $servicesService->findRelated(
```
BPScott: We've got a known list of NIDs we need to show, which isn't going to change, do we need to make a query for them?

BPScott: Ah, ignore me, we need the defaultService for each of those networks.

votemike:
Do we want to update the image to remove BBC Three?

src/Ds2013/Core/Modules/Schedules/_tv-channels.scss
votemike: @BPScott any thoughts?

BPScott: This is going to break on live as we rename filenames to include their URL hash (through gulp-rev). You'll need to do some work in the gulp file to work out how to deal with references to hashed assets within the sass file.

A brief google suggests https://github.com/jamesknelson/gulp-rev-replace might be relevant

votemike:
I wonder if people think this would be useful to add to our project: https://github.com/JeffreyWay/laravel-mix/blob/master/docs/installation.md#stand-alone-project

It runs on top of Webpack and can do things like pre-process/minify SASS and JS, version files, replace versioned file names, automatically refresh the browser upon changes, show notifications on you Mac for successful/failed builds

src/Ds2013/Core/Modules/Schedules/_tv-channels.scss
votemike: Added gulp-rev-css-url, but still doesn't solve the problem of the`../`. What would you like done about that?

votemike:
On a date when BBC Three is on air, the number of images is too wide for the page. What should we do about this?

src/Ds2013/Core/Modules/Schedules/_tv-channels.scss
votemike: Turns out we need this anyway.

programmes-developers:

BPScott:
> I wonder if people think this would be useful to add to our project: https://github.com/JeffreyWay/laravel-mix/blob/master/docs/installation.md#stand-alone-project

A move to webpack would be nice as it is the new shiny. I've been keeping an eye on Symfony Encore which does a pretty similar thing: https://symfony.com/blog/introducing-webpack-encore-for-asset-management. I'd be tempted to lean towards that as we know it'll play nice with other symfony bits

votemike:
Haha
> Encore is inspired by Webpacker and Mix.
But either Mix or Encore would be useful!

src/Ds2013/Core/Modules/Schedules/_tv-channels.scss
BPScott: I think we should strive to stay away from relative links, instead we should always use some kind of utility function such as image-url('logos/services_channel_sprite.png') that anchors all image references to a particular base path rather than having to know the relative location between the css file and the image file every time

BPScott:
I think the move to Webpack-based things is comparatively low priority at the moment, our Gulp config works without much pain and we've got bigger fish to fry

src/Ds2013/Core/Modules/Schedules/_tv-channels.scss
votemike: Are you saying I should update the `image-url` function?

BPScott: Yep

PR #100, opened 2017-07-21
Use start of day as condition to show end date of service in schedule page

felipeparaujo:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5708


programmes-developers:

tatitati:
good for me

BjornTwachtmann:
Looks like a sensible solution, though it is 5PM on a friday and I would probably say that to any code that didn't actually involve eval() on user input.

PR #101, opened 2017-07-24
Mechanism for FindByPid controllers to have dedicated classes

BPScott:
This its better than having to jam them all into a single class (which
would get huuuuuge)


programmes-developers:

src/Controller/FindByPid/BrandController.php
```diff
@@ -0,0 +1,67 @@
...
+use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeContainer;
+use BBC\ProgrammesPagesService\Service\ProgrammesService;
+
+/**
+ * Brand Page
+ *
+ * Strictly speaking this isn't for Brand entities, it's for any
+ * ProgrammeContainer that appears at the top of a programme hierarchy. A Series
+ * may appear at the top of the hierarchy in addition to Brands. Technically
+ * this should be a Top-level Editorial Container (TLEC) Controller, but good
```
votemike: I vote we rename it TLECPage

src/Controller/FindByPid/BrandController.php
```diff
@@ -0,0 +1,67 @@
...
+ * luck getting anybody to call it that.
+ *
+ * This controller is not instantiated by Symfony's ControllerResolver (through
+ * creating a route that resolves to this controller). Instead the
+ * FindByPidController creates and calls this Controller when it has a page it
+ * wishes to render. Thus we any classes we require from the container need to
```
votemike: "Thus we any"?

src/Controller/FindByPid/BrandController.php
```diff
@@ -0,0 +1,67 @@
...
+     * A programme should get the TV brand page in the following cases:
+     * The programmes's network is not international
+     * AND The programme's network is TV, OR bbc_news or bbc_sport
+     *
+     * bbc_news and bbc_sport are special cases as PIPS does not consider them
+     * as TV, however they predominately show video content.
+     */
+    private function isTVBrandPage(ProgrammeContainer $programme)
+    {
+        $network = $programme->getNetwork();
```
votemike: How can a programme be international or not? Surely a programme can be shown on multiple networks? Wouldn't it be a broadcast that is international or not?

src/Controller/FindByPid/README.md
```diff
@@ -0,0 +1,23 @@
...
+Version, Segment or SegmentEvent. It is possible to store all of these
+controller actions within a single class, but then that class becomes gargantuan
+and thus finding things within it becomes a pain. It would be preferable to
+split these controllers up into multiple classes - one for the TV brand page,
+one for the radio brand page, one for the episode page etc. However working out
+what controller we want requires a database query so we are unable to do at the
+routing level - where Symfony dispatches to controllers.
+
+The solution to this is have a front-door controller (FindByPidController),
+whose job is look up the entity type, and then defer to another controller class
```
votemike: Why is the `FindByPidController` not just the parent of all the other controllers?

src/Controller/FindByPid/BrandController.php
BPScott: Whoops.

src/Controller/FindByPid/BrandController.php
BPScott: A programme has a "commissioning entity" - it's masterbrand - which is who paid for a programme to be made. Normally this correlates to the service that shows it.

In this case we determine if a programme is international or not based on if the its master brand (i.e. who paid for the programme) is international or not.

src/Controller/FindByPid/README.md
BPScott: How would that help?

votemike: We could use the controller in the normal Symfony way couldn't we?

BPScott: No, because you need to know what controller to route to. Which requires a DB query, which you can't do at the "the routing works out what controller to send to" level

programmes-developers:

BPScott:
Closing in favour of #105

PR #102, opened 2017-07-24
Fix linking radio player over http

felipeparaujo:


programmes-developers:

BjornTwachtmann:
Code looks good to me.

src/Ds2013/Helpers/LiveBroadcastHelper.php
```diff
@@ -153,19 +155,35 @@ class LiveBroadcastHelper
...
+
     public function simulcastUrl(CollapsedBroadcast $collapsedBroadcast, ?Service $preferredService = null): string
     {
         $servicesBySid = $this->getServicesBySid($collapsedBroadcast);
         if ($preferredService) {
             $preferredServiceId = (string) $preferredService->getSid();
             if (isset($servicesBySid[$preferredServiceId]) && isset(self::LIVE_SERVICE_URLS[$preferredServiceId])) {
-                return self::LIVE_SERVICE_URLS[$preferredServiceId];
+                return $this->router->generate(
+                    self::LIVE_SERVICE_URLS[$preferredServiceId][0],
```
votemike: Should these be keyed instead of referencing them as 0 and 1?

src/Ds2013/Helpers/LiveBroadcastHelper.php
```diff
@@ -6,145 +6,147 @@
...
-        'bbc_one_north_west'                    => '/iplayer/live/bbcone?area=north_west',
-        'bbc_one_oxford'                        => '/iplayer/live/bbcone?area=oxford',
-        'bbc_one_south'                         => '/iplayer/live/bbcone?area=south',
-        'bbc_one_south_east'                    => '/iplayer/live/bbcone?area=south_east',
-        'bbc_one_south_west'                    => '/iplayer/live/bbcone?area=south_west',
-        'bbc_one_west'                          => '/iplayer/live/bbcone?area=west',
-        'bbc_one_west_midlands'                 => '/iplayer/live/bbcone?area=west_midlands',
-        'bbc_one_yorks'                         => '/iplayer/live/bbcone?area=yorks',
-        'bbc_two_england'                       => '/iplayer/live/bbctwo',
+        //sid                                   => [route, [...arguments]],
```
votemike: Does all of these really need to be manually written out? Can it not be gleaned from the DB?

src/Ds2013/Helpers/LiveBroadcastHelper.php
felipeparaujo: They were at first, but the list got really noisy with honestly useless data, so I removed it

src/Ds2013/Helpers/LiveBroadcastHelper.php
BPScott: This list doesn't come from PIPS, annoyingly it's a manually curated list. We could stick that in into Faucet's generation of a service but it'd still be a hardcoded list, just living somewhere else.

src/Ds2013/Helpers/LiveBroadcastHelper.php
votemike: Fair enough

PR #104, opened 2017-07-25
Housekeeping

votemike:
* Updating composer packages

* Updating yarn packages

* Sorting use statements


programmes-developers:

PR #105, opened 2017-07-26
Mechanism for dedicated FindByPid Controllers

BPScott:
This is a lot better than cramming them all into a single class

This one uses event hooks rather than a single entry point class. This is neat as it means that the requests controller is "ContainerController", "EpisodeController", etc rather than always being "FindByPidController"


programmes-developers:

src/Controller/FindByPid/TlecController.php
```diff
@@ -0,0 +1,27 @@
...
+ * We tend to call this "the brand page", but both Brands and Series are both
+ * ProgrammeContainers that may appear at the top of the programme hierarchy.
+ */
+class TlecController extends BaseController
+{
+    public function __invoke(ProgrammeContainer $programme, ProgrammesService $programmesService)
+    {
+        $this->setContext($programme);
+
+        return $this->renderWithChrome('find_by_pid/brand.html.twig', [
```
votemike: `find_by_pid/tlec.html.twig`?

src/EventSubscriber/FindByPidRouterSubscriber.php
```diff
@@ -0,0 +1,124 @@
...
+use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeContainer;
+use BBC\ProgrammesPagesService\Domain\ValueObject\Pid;
+use BBC\ProgrammesPagesService\Service\ServiceFactory;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+/**
+ * A single route /programmes/:pid can serve one of several page types depending
```
votemike: Not convinced this block is required

src/EventSubscriber/FindByPidRouterSubscriber.php
```diff
@@ -0,0 +1,124 @@
...
+ * preferable to split these controllers up into multiple classes - one for the
+ * brand page, one for the episode page, one for the clip page etc.
+ *
+ * This is an event that spots requests to the find_by_pid route, makes a DB
+ * query to work out what Controller should be used for a given PID and sets
+ * the _controller property to the relevant controller name. We also store the
+ * entity in the request attributes so that the controller can access it
+ * directly instead of having to make another DB request for the entity.
+ *
+ * This happens after Routing has taken place so we have a request with a
```
votemike: Not convinced this block is required

src/EventSubscriber/FindByPidRouterSubscriber.php
```diff
@@ -0,0 +1,124 @@
...
+ */
+class FindByPidRouterSubscriber implements EventSubscriberInterface
+{
+    /** @var ServiceFactory */
+    private $serviceFactory;
+
+    public static function getSubscribedEvents()
+    {
+        return [
+            // This needs to happen after the RouterSubscriber, which has a
```
votemike: What is "This"?

src/EventSubscriber/FindByPidRouterSubscriber.php
```diff
@@ -0,0 +1,124 @@
...
+
+        // Attempt to find a Programme
+        // TODO attempt to find a group in the same query
+        $programme = $this->serviceFactory->getProgrammesService()->findByPidFull($pid);
+        if ($programme) {
+            $request->attributes->set('programme', $programme);
+
+            if ($programme instanceof ProgrammeContainer) {
+                if (!$programme->getParent()) {
+                    $request->attributes->set('_controller', 'App\\Controller\\FindByPid\\TlecController');
```
votemike: I'd still prefer `TlecController::class` than this magic string. Happy to see what other people think.

src/EventSubscriber/FindByPidRouterSubscriber.php
```diff
@@ -0,0 +1,124 @@
...
+                return;
+            }
+
+            if ($programme instanceof Clip) {
+                $request->attributes->set('_controller', 'App\\Controller\\FindByPid\\DefaultController');
+                return;
+            }
+
+            // Otherwise something has gone very wrong
+            throw new NotFoundHttpException(sprintf('The item with PID "%s" was of an unknown type', $pid));
```
votemike: `The programme with PID "%s" was of an unknown type` maybe?

BjornTwachtmann:
Nice, looks a lot simpler than the last time I saw it.

src/EventSubscriber/FindByPidRouterSubscriber.php
BPScott: Have reworded the comment to give more context

src/EventSubscriber/FindByPidRouterSubscriber.php
BPScott: whoops, last minute renaming inconsistencies

src/Controller/FindByPid/TlecController.php
BPScott: :+1: Whoops, last minute renaming inconsistencies

src/EventSubscriber/FindByPidRouterSubscriber.php
BPScott: I think it's useful to provide a bit of context about what we're trying to solve and how we solve it. I wonder if others understand what is happening/why without this text.

felipeparaujo: i'm always in favour of exposition

src/EventSubscriber/FindByPidRouterSubscriber.php
```diff
@@ -0,0 +1,125 @@
...
+declare(strict_types = 1);
+namespace App\EventSubscriber;
+
+use BBC\ProgrammesPagesService\Domain\Entity\Clip;
+use BBC\ProgrammesPagesService\Domain\Entity\Episode;
+use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeContainer;
+use BBC\ProgrammesPagesService\Domain\ValueObject\Pid;
+use BBC\ProgrammesPagesService\Service\ServiceFactory;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
```
votemike: Exception should come after Event

PR #106, opened 2017-07-26
Add FindByPid VersionController

BPScott:
Visting /programmes/:versionPid now redirects you to the episode that the
version belongs to

### DIFF
--- a/src/Controller/FindByPid/VersionController.php
+++ b/src/Controller/FindByPid/VersionController.php
@@ -1,0 +1,20 @@
+<?php
+declare(strict_types = 1);
+namespace App\Controller\FindByPid;
+
+use App\Controller\BaseController;
+use BBC\ProgrammesPagesService\Domain\Entity\Version;
+
+/**
+ * Version Page
+ *
+ * Redirect to the Episode that the Version belongs to
+ */
+class VersionController extends BaseController
+{
+    public function __invoke(Version $version)
+    {
+        $episodePid = (string) $version->getProgrammeItem()->getPid();
+        return $this->redirectToRoute('find_by_pid', ['pid' => $episodePid], 303);
+    }
+}
--- a/tests/EventSubcriber/FindByPidRouterSubscriberTest.php
+++ b/tests/EventSubcriber/FindByPidRouterSubscriberTest.php
@@ -6,7 +6,10 @@ use App\EventSubscriber\FindByPidRouterSubscriber;
 use BBC\ProgrammesPagesService\Domain\Entity\Brand;
 use BBC\ProgrammesPagesService\Domain\Entity\Clip;
 use BBC\ProgrammesPagesService\Domain\Entity\Episode;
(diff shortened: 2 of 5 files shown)


programmes-developers:

src/Controller/FindByPid/VersionController.php
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Redirect to the Episode that the version belongs to
+ */
+class VersionController extends BaseController
+{
+    public function __invoke(Version $version)
+    {
+        $episodePid = (string) $version->getProgrammeItem()->getPid();
+        $response = $this->redirectToRoute('find_by_pid', [ 'pid' => $episodePid], 303);
+        $response->setPublic()->setMaxAge(120);
```
votemike: Why only 2 minutes?

src/Controller/FindByPid/VersionController.php
```diff
@@ -0,0 +1,23 @@
...
+use BBC\ProgrammesPagesService\Domain\Entity\Version;
+
+/**
+ * Version Page
+ *
+ * Redirect to the Episode that the version belongs to
+ */
+class VersionController extends BaseController
+{
+    public function __invoke(Version $version)
```
votemike: Return typehint?

tests/EventSubcriber/FindByPidRouterSubscriberTest.php
```diff
@@ -54,8 +57,10 @@ public function coreEntityDataProvider()
             [$series, null, null, 'App\Controller\FindByPid\DefaultController'],
             [$this->createMock(Episode::class), null, null, 'App\Controller\FindByPid\DefaultController'],
             [$this->createMock(Clip::class), null, null, 'App\Controller\FindByPid\DefaultController'],
+            // TODO add checks for Groups
```
votemike: Groups exist don't they? Why is this a TODO?

BPScott: Shall be done once I have a coreEntitiesService that I can use to make a query, and get data from :)

src/Controller/FindByPid/VersionController.php
BPScott: We haven't bothered setting them on any of our other Controllers.

src/Controller/FindByPid/VersionController.php
BPScott: Standard cache lifetime of our pages. This is the duration of time that a page is cached in Varnish too.

Admittedly it's heinously that a Version shall change the programme it is attached to, but it is useful for that to be reflected reasonably quickly.

src/Controller/FindByPid/VersionController.php
votemike: Great time to start 👍

BPScott: I don't think it's really worth it tbh, it's an extra use statement and a typehint that I don't think adds any meaningful value

PR #107, opened 2017-07-27
Add Controllers for all FindByPid routes

BPScott:
The controllers don't do anything useful at the moment, but this means
that the FindByPidRouterSubscriber is complete and shouldn't need
changing at a later date.

I've not bothered with automated tests for these routes as they don't do anything useful (aside from exist so we're not spawning loads of TODOs).


programmes-developers:

BPScott:
I'm also burying the lede: This gives us "find a programme or group" in a single DB query, which is kinda a big deal as APS/Clifton and Nitro were never capable of this :D

BjornTwachtmann:
This will technically add a couple of DB queries to any 404s we have to serve from frontend. That said stuff that 404s is going to hit the PAL anyway probably. Looks good.

BPScott:
If the findByPidProxy fails to find a PID in the core_entity table we'd direct traffic to the PAL as that may be able to serve a Version or Segment. Once we implement both of those in frontend we'd be able to switch to start sending "not found" to frontend rather than PAL

PR #108, opened 2017-07-27
Metrics again

tatitati:


programmes-developers:

app/config/config_test.yml
```diff
@@ -31,3 +31,7 @@ doctrine:
 
 liip_functional_test:
     cache_sqlite_db: true
+
+services:
+    App\Metrics\Cache\MetricCacheInterface:
+        class: App\Metrics\Cache\MetricCacheNull
```
BPScott: Rather than redefine the service, set the class value to be a parameter, like what `app.orbit_client.class` and `app.branding_client.class` does

app/config/services.yml
```diff
@@ -85,6 +87,36 @@ services:
             - { name: 'kernel.event_subscriber' }
             - { name: 'monolog.logger', channel: 'app_access' }
 
+    App\Metrics\MetricsManager:
```
BPScott: I can't see why this needs to be public? Could be `App\Metrics\MetricsManager: ~`

app/config/services.yml
```diff
@@ -85,6 +87,36 @@ services:
...
+        public: true
+
+    App\Metrics\MetricsMiddleware:
+        tags:
+            - { name: 'csa_guzzle.middleware', alias: 'metrics_middleware', pritority: 90 }
```
BPScott: I'm fairly sure `pritority` isn't a valid tag. Why does this need an alias?

app/config/services.yml
```diff
@@ -85,6 +87,36 @@ services:
...
+
+    RMP\CloudwatchMonitoring\MonitoringHandler:
+        class: RMP\CloudwatchMonitoring\MonitoringHandler
```
BPScott: You don't need to specify the class if it is the same as the service name, this line can be removed

app/config/services.yml
```diff
@@ -85,6 +87,36 @@ services:
...
+        arguments:
+            $client: '@Aws\CloudWatch\CloudWatchClient'
```
BPScott: This service is the only thing in the DI container that matches this interface, thus Symfony can infer this value automatically, thus this line can be removed.

app/config/services.yml
```diff
@@ -85,6 +87,36 @@ services:
...
+            $namespace: 'frontend'
+            $env: '%cosmos_environment%'
+
+    Aws\CloudWatch\CloudWatchClient:
+        class: Aws\CloudWatch\CloudWatchClient
```
BPScott: You don't need to specify the class if it is the same as the service name, this line can be removed

app/config/services.yml
```diff
@@ -85,6 +87,36 @@ services:
...
+        arguments:
+            - region: '%aws_s3_region%'
+              version: '2010-08-01'
+              http:
+                timeout: '%frontend.aws.http_timeout%'
+                connect_timeout: '%frontend.aws.http_connect_timeout%'
+
+
+    App\Metrics\Backend\MetricBackendInterface:
+        class: App\Metrics\Backend\CloudWatchMetricBackend
```
BPScott: I think this should be `alias` rather than `class`, that's how we do this for `BBC\ProgrammesPagesService\Cache\CacheInterface`.

app/config/services.yml
```diff
@@ -85,6 +87,36 @@ services:
...
+
+    App\Metrics\Cache\MetricCacheInterface:
+        class: App\Metrics\Cache\MetricCacheApcu
```
BPScott: I think this should be `alias` rather than `class`, that's how we do this for `BBC\ProgrammesPagesService\Cache\CacheInterface`.

src/EventSubscriber/MonitoringSubscriber.php
```diff
@@ -45,38 +50,26 @@ public function terminateEnd(KernelEvent $event)
...
-        }
-
-        // Strip off the common preamble for the sake of readability
-        $controllerAction = str_replace('App\\Controller\\', '', $controllerAction);
-
-        // Skip if it is the status controller
-        // This gets pinged every 15 seconds by the ELB and we don't need that noise
-        if ($controllerAction == 'StatusController::showAction') {
+        $request = $event->getRequest();
+        $routeName = $request->get('_route');
```
BPScott: I think we should name this lot based on controller name ('_controller') then we can tell the difference between The Tlec page vs  the Episode page vs the Clip page etc

src/EventSubscriber/MonitoringSubscriber.php
```diff
@@ -45,38 +50,26 @@ public function terminateEnd(KernelEvent $event)
...
+        $this->metricsManager->send();
     }
 
     private function getControllerPeriod()
     {
         foreach ($this->stopwatch->getSections() as $section) {
             $event = $section->getEvents()[self::REQUEST_TIMER] ?? null;
             if ($event) {
-                return $event->getDuration();
+                return (int) round($event->getDuration());
```
BPScott: I think with some new stopwatch stuff we can get sub-millisecond precision, do we want to deliberately make our timings coarser?

app/config/services.yml
tatitati: Just tested it with alias, but failed. Is interesting, but because we didn't define App\Metrics\Backend\CloudWatchMetricBackend as a service we cannot refer it by using alias. 💭

src/EventSubscriber/MonitoringSubscriber.php
```diff
@@ -40,43 +45,32 @@ public function requestStart(KernelEvent $event)
 
     public function terminateEnd(KernelEvent $event)
     {
-        $this->logRequestTime($event);
+        $this->metrics($event);
     }
 
-    private function logRequestTime(KernelEvent $event)
+    private function metrics(KernelEvent $event)
```
votemike: `metrics()` doesn't make it clear to me what this method is doing. Sending metrics? Receiving metrics? Displaying metrics? Logging metrics?  
Could you come up with a clearer name, or put some docblocks on the method.

src/Metrics/Backend/CloudWatchMetricBackend.php
```diff
@@ -0,0 +1,44 @@
...
+    /** @var LoggerInterface */
+    private $logger;
+
+    public function __construct(MonitoringHandler $monitoringHandler, LoggerInterface $logger)
+    {
+        $this->monitoringHandler = $monitoringHandler;
+        $this->logger = $logger;
+    }
+
+    public function sendMetrics(array $metrics): void
```
votemike: Maybe add a docblock with `/** @param ProgrammesMetricInterface[] $metrics */`

src/Metrics/Backend/LogMetricBackend.php
```diff
@@ -0,0 +1,32 @@
...
+    {
+        $this->logger = $logger;
+    }
+
+    public function sendMetrics(array $metrics): void
+    {
+        /** @var ProgrammesMetricInterface $metric */
+        foreach ($metrics as $metric) {
+            foreach ($metric->getMetricData() as $metricDatum) {
+                $logString = join(',', [$metricDatum['MetricName'], $metricDatum['Value'], $metricDatum['Unit']]);
```
votemike: Do we use `join`? or `implode`?

src/Metrics/Cache/MetricCacheApcu.php
```diff
@@ -0,0 +1,217 @@
...
+declare(strict_types = 1);
+
+namespace App\Metrics\Cache;
+
+use App\Metrics\ProgrammesMetrics\ProgrammesMetricInterface;
+
+class MetricCacheApcu implements MetricCacheInterface
+{
+    /** apcu cache key prefix */
+    public const PREFIX = 'PRFE_METRICS';
```
votemike: What is PRFE?

src/Metrics/Cache/MetricCacheApcu.php
```diff
@@ -0,0 +1,217 @@
...
+                $apcKeyValues[$apcKey] += $metricValue;
+            }
+        }
+
+        foreach ($apcKeyValues as $key => $value) {
+            apcu_inc($key, $value);
+        }
+    }
+
+    public function getAndClearReadyToSendMetrics(callable $getAllPossibleMetrics): array
```
votemike: What's in the returned array?

src/Metrics/Cache/MetricCacheApcu.php
```diff
@@ -0,0 +1,217 @@
...
+     *
+     * Once bucket_no_send expires, the contents of bucket "A" will be retrieved,
+     * cleared, and eventually sent to the backend.
+     */
+    private function rotateBucketIfNeeded(): void
+    {
+        $currentBucketKey = $this->makeApcKey('internal', '', 'bucket_index');
+        $bucketTimeoutKey = $this->makeApcKey('internal', '', 'bucket_timeout');
+
+        if ($this->tryToLockOnKey($bucketTimeoutKey, self::SEND_INTERVAL)) {
```
votemike: Maybe reverse this check and return early to avoid so much code indentation?

src/Metrics/MetricsManager.php
```diff
@@ -0,0 +1,132 @@
...
+                    $this->validRouteControllers[$controllerName] = $controllerName;
+                }
+            }
+        }
+        return $this->validRouteControllers;
+    }
+
+    private function controllerName(string $controllerClass): ?string
+    {
+        if (strpos($controllerClass, 'App\\Controller') === 0) {
```
votemike: Would this solution be better? I guess this method will be called a lot.  
https://stackoverflow.com/a/25472778/4009201

src/Metrics/MetricsMiddleware.php
```diff
@@ -0,0 +1,76 @@
...
+        return function (RequestInterface $request, array $options) use ($handler) {
+            $this->logger->info("HTTP Request:" . $request->getUri());
+
+            $options['on_stats'] = function (TransferStats $stats) use (&$responseTime) {
+                $uri = $stats->getEffectiveUri();
+                $apiName = $this->getApiNameFromUrl($uri);
+                if (!$apiName) {
+                    return;
+                }
+                $responseTime = (int) round($stats->getTransferTime() * 1000);
```
votemike: Why does this need rounding? And why is it cast to an int?

src/Metrics/ProgrammesMetrics/ApiResponseMetric.php
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * Accepts the format returned by getCacheKeyValuePairs, but sets
+     * the values in the class from values returned from the cache.
+     */
+    public function setValuesFromCacheKeyValuePairs(array $keyValuePairs): void
+    {
+        $this->count = $keyValuePairs[$this->cacheKey()];
+    }
+
+    public function getMetricData(): array
```
votemike: Is this data in a specific format for AWS? Should it be abstracted in case we decide to move to a different metric system?

src/Metrics/MetricsManager.php
tatitati: No, we don't want to know the class name, just if is a call to a controller.

src/Metrics/MetricsMiddleware.php
tatitati: get TransferTime return times in seconds, which can have decimals. We are sending this to AWS in miliseconds. So far everything is good. 
But now the problem: APCu can increase only rounded numbers.

src/Metrics/Cache/MetricCacheApcu.php
tatitati: I guess we tried to write PREF instead of PRFE 🤔

src/Metrics/ProgrammesMetrics/ApiResponseMetric.php
BjornTwachtmann: I think this mess is complex enough already. Otherwise we need to introduce some kind of mapping from a *metric to a AwsData object. Which I seriously can't be arsed with. If we need to move to representing metrics differently we can add an extra function or modify these ones

app/config/services.yml
```diff
@@ -2,6 +2,10 @@ parameters:
     app.default_locale: 'en_GB'
     app.orbit_client.class: BBC\BrandingClient\OrbitClient
     app.branding_client.class: BBC\BrandingClient\BrandingClient
+    app.metric_cache.null.class: App\Metrics\Cache\MetricCacheNull
```
BjornTwachtmann: Rather than have a null class and an apcu class shouldn't we just have a metric_cache.class that's override per environment?

src/Metrics/Backend/LogMetricBackend.php
BjornTwachtmann: should probably be implode

src/Metrics/MetricsMiddleware.php
```diff
@@ -0,0 +1,76 @@
...
+        return function (RequestInterface $request, array $options) use ($handler) {
+            $this->logger->info("HTTP Request:" . $request->getUri());
+
+            $options['on_stats'] = function (TransferStats $stats) use (&$responseTime) {
+                $uri = $stats->getEffectiveUri();
+                $apiName = $this->getApiNameFromUrl($uri);
+                if (!$apiName) {
+                    return;
+                }
+                $responseTime = round($stats->getTransferTime() * 1000);
```
BjornTwachtmann: This needs to be cast to an int. round returns a float.

programmes-developers:

programmes-developers:

app/config/services.yml
BPScott: Yes, exactly.

src/Metrics/Backend/LogMetricBackend.php
```diff
@@ -0,0 +1,32 @@
...
+    private $logger;
+
+    public function __construct(LoggerInterface $logger)
+    {
+        $this->logger = $logger;
+    }
+
+    public function sendMetrics(array $metrics): void
+    {
+        /** @var ProgrammesMetricInterface $metric */
```
votemike: Can't you avoid this docblock by putting a docblock on the method saying what is in the array?

src/Metrics/Backend/NullMetricBackend.php
```diff
@@ -0,0 +1,12 @@
+<?php
+declare(strict_types = 1);
+
+namespace App\Metrics\Backend;
+
+class NullMetricBackend implements MetricBackendInterface
+{
+    public function sendMetrics(array $metrics): void
+    {
+        return;
```
votemike: Is this needed?

src/Metrics/Backend/CloudWatchMetricBackend.php
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * @param ProgrammesMetricInterface[] $metrics
+     */
+    public function sendMetrics(array $metrics): void
+    {
+        /** @var ProgrammesMetricInterface $metric */
+        foreach ($metrics as $metric) {
+            foreach ($metric->getMetricData() as $metricDatum) {
+                $this->monitoringHandler->putMetricData(
```
BPScott: Considering we're putting a bunch of metric data onto the list and then sending them all immediately, is there actually any value in using MonitoringHandler? 

It seems like we'd be able to just an instance of CloudWatchClient directly then we'd have to implement a [sendMetrics()](https://github.com/bbc/rmp-monitoringhandler/blob/master/src/MonitoringHandler.php#L128-L144) function in this class that loops through `$metric->getMetricData()` and makes a bunch of http requests.

That'd save us needing that additional library that we're using basically nothing of.

src/EventSubscriber/MonitoringSubscriber.php
```diff
@@ -40,43 +45,32 @@ public function requestStart(KernelEvent $event)
...
-        }
-
-        // Strip off the common preamble for the sake of readability
-        $controllerAction = str_replace('App\\Controller\\', '', $controllerAction);
-
-        // Skip if it is the status controller
-        // This gets pinged every 15 seconds by the ELB and we don't need that noise
-        if ($controllerAction == 'StatusController::showAction') {
+        $request = $event->getRequest();
+        $controllerName = $request->get('_controller');
```
BjornTwachtmann: We actually kind of want this to run for the status controller route, as on INT/TEST that will mean that statistics are sent continuously despite them not getting hit very often. There is somethign in the metrics manager that filters to routes that have controllers in \App\Controller, which will exclude things like profiling and redirects.

src/Metrics/Backend/CloudWatchMetricBackend.php
BjornTwachtmann: MonitoringHandler already implements bundling this stuff up and sending it in parallel HTTP requests. I could re-implement that, but why would I? What's your objection the library.

BjornTwachtmann: *to the library

src/Metrics/Backend/LogMetricBackend.php
BjornTwachtmann: Yeah, good point.

BjornTwachtmann:
Approved by the squirrel army.

PR #109, opened 2017-07-28
Add ContextEntityByPidValueResolver

BPScott:
This automatically looks up an entity based upon the pid value of the
route and the typehinting on the Controller, to save you having to
request the item through the service in every controller and then throw
a 404 if the entity does not exist.

The typehinting thing is particularly powerful as you can request an
instance of ProgrammeContainer in your controller and this will throw a
404 if the pid exists but is an Episode. This is useful for Pages like
/programmes/:pid/series that is only valid for Brands and Series.

### DIFF
--- a/app/config/services.yml
+++ b/app/config/services.yml
@@ -29,11 +29,15 @@ services:
         arguments:
             $debug: '%kernel.debug%'
 
-    # This needs to be triggered before the any existing default Resolvers,
-    # the largest of which has a priority of 100
+    ### Argument Value Resolvers
+
+    App\ArgumentResolver\ContextEntityByPidValueResolver:
+        tags: [{name: 'controller.argument_value_resolver', priority: 0 }]
+
+    # This needs to be triggered before the RequestAttributeValueResolver,
+    # which has a priority of 100
     App\ArgumentResolver\IdentifierValueResolver:
         tags: [{name: 'controller.argument_value_resolver', priority: 125 }]
-
 
     ### Caches
 
--- a/src/ArgumentResolver/ContextEntityByPidValueResolver.php
+++ b/src/ArgumentResolver/ContextEntityByPidValueResolver.php
@@ -1,0 +1,102 @@
+<?php
+
+namespace App\ArgumentResolver;
+
+use BBC\ProgrammesPagesService\Domain\Entity\Brand;
+use BBC\ProgrammesPagesService\Domain\Entity\Episode;
+use BBC\ProgrammesPagesService\Domain\Entity\Clip;
+use BBC\ProgrammesPagesService\Domain\Entity\Collection;
+use BBC\ProgrammesPagesService\Domain\Entity\CoreEntity;
+use BBC\ProgrammesPagesService\Domain\Entity\Franchise;
+use BBC\ProgrammesPagesService\Domain\Entity\Gallery;
+use BBC\ProgrammesPagesService\Domain\Entity\Programme;
+use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeContainer;
+use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeItem;
+use BBC\ProgrammesPagesService\Domain\Entity\Group;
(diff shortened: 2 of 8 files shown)


programmes-developers:

src/ArgumentResolver/ContextEntityByPidValueResolver.php
```diff
@@ -0,0 +1,82 @@
...
+
+        if (is_a($type, CoreEntity::class, true)) {
+            // Attempt to look up the CoreEntity. Filter to only return the
+            // type of entity requested
+
+            // Conveniently the filter to pass into findByPidFull is the short
+            // class name of the entity - found by removing the namespace
+            $entity = $this->serviceFactory->getCoreEntitiesService()->findByPidFull(
+                $pid,
+                str_replace('BBC\\ProgrammesPagesService\\Domain\\Entity\\', '', $type)
```
votemike: Maybe this instead? https://stackoverflow.com/a/25472778/4009201

src/ArgumentResolver/ContextEntityByPidValueResolver.php
```diff
@@ -0,0 +1,82 @@
...
+    ];
+
+    private $serviceFactory;
+
+    public function __construct(ServiceFactory $serviceFactory)
+    {
+        $this->serviceFactory = $serviceFactory;
+    }
+
+    public function supports(Request $request, ArgumentMetadata $argument)
```
tatitati: just a tiny thing, type hints for return values in function supports and function resolve.

app/config/services.yml
```diff
@@ -29,12 +29,16 @@ services:
         arguments:
             $debug: '%kernel.debug%'
 
-    # This needs to be triggered before the any existing default Resolvers,
-    # the largest of which has a priority of 100
+    ### Argument Value Resolvers
+
+    App\ArgumentResolver\DomainEntityByPidValueResolver:
```
BjornTwachtmann: I think maybe you forgot to rename this? That class doesn't exist as far as I can see.

BPScott: Gah yes, also need to pull in the latest pages-service

src/ArgumentResolver/ContextEntityByPidValueResolver.php
```diff
@@ -0,0 +1,82 @@
...
+    {
+        if (!$request->attributes->has('pid') || $argument->isVariadic()) {
+            return false;
+        }
+
+        foreach (self::SUPPORTED_CLASSES as $class) {
+            if (is_a($argument->getType(), $class, true)) {
```
BjornTwachtmann: This is... Less than optimal from an execution perspective. Every request will run is_a on every argument in the controller for everything in the constant here. Is there an obvious way to optimise this?

BPScott: The headache here is subclasses. `is_a($argument->getType(), 'BBC\ProgrammesPagesService\Domain\Entity\Brand', true)` will return true. I could extend SUPPORTED_CLASSES to include all subtypes of CoreEntity and then this  loop becomes `in_array($argument->getType(), self::SUPPORTED_CLASSES)`.

src/ArgumentResolver/ContextEntityByPidValueResolver.php
BPScott: Done a little perf test, Reflection does seem to be faster than str_replace

votemike: @tatitati I think you are also getting the shortname in one of your PRs. Was it in frontend? Would it be worth making a helper for this?

BjornTwachtmann: I'm really not convinced by the argument for using reflection here. Creating the same reflectionClass 30k times in a loop is likely to be optimised by the compiler/language. Creating a load of different ones less so. In any case the actual overhead is minimal and I would avoid reflection unless it is strictly necessary.

PR #110, opened 2017-07-31
Redirect to programme options location if required

BPScott:
Add ProgrammeOptionsRedirectHttpException that returns a redirect to the
given location.

Throw that exception when we get an entity in the
ContextEntityByPidValueResolver and FindByPidRouterSubscriber.

This is done in both places as if the FindByPidRouterSubscriber adds a
domain object to the request attributes then the
ContextEntityByPidValueResolver is not fired, because the
RequestAttributeValueResolver resolves the controller arguments first.

### DIFF
--- a/src/ArgumentResolver/ContextEntityByPidValueResolver.php
+++ b/src/ArgumentResolver/ContextEntityByPidValueResolver.php
@@ -2,6 +2,7 @@
 declare(strict_types = 1);
 namespace App\ArgumentResolver;
 
+use App\Exception\ProgrammeOptionsRedirectHttpException;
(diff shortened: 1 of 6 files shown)


programmes-developers:

BPScott:
The arguments for the exception / how to check if a page is a redirect could be made a bit nicer by rejigging how the redirect values are stored in the options, but that's a bit of tech debt that I'm ok with taking on for now, as it will require a bit of thought about what we want that format to be and shall require rejigging stuff in Faucet.

programmes-developers:

BPScott:
Build failure also seems to be on master, and is unrelated to this change.

PR #111, opened 2017-08-01
Branding / Orbit should store results in Redis

BPScott:
Previously they were using a local filesystem cache.
Use the same cache pool as service queries.


programmes-developers:

PR #113, opened 2017-08-02
Fixing skip links

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5774

If testing locally, don't forget to run gulp.


programmes-developers:

BjornTwachtmann:
This is the same fix from v2 right? Looks good.

votemike:
It is.

PR #116, opened 2017-08-02
Adding social tags for Brand Page

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5739

I believe this is the only change that is needed. The amazing way we set up Context/Meta_Context means that we hardly have to do anything in order to add the meta tags.


programmes-developers:

BPScott:
When the context is a CoreEntity, I think the titlePrefix should contain the network name and programme title (including all its hierarchy), otherwise we're going to have to specify the programme title every time.

e.g. http://www.bbc.co.uk/programmes/b006q2x0/series has the title "BBC One - Doctor Who - Series". The titlePrefix in the MetaContext would be "BBC One - Doctor Who" and the title block would be`{{ meta_context.titlePrefix }} - {{ tr('series') }}`.

e.g. http://www.bbc.co.uk/programmes/p00t7hbb/episodes has the title "BBC One - Doctor Who, Season 25, The Greatest Show in the Galaxy - Next on". The titlePrefix in the MetaContext would be "BBC One - Doctor Who, Season 25, The Greatest Show in the Galaxy" and the title block would be`{{ meta_context.titlePrefix }} - {{ tr('next_on') }}`.


For this TLEO page (and the series, episode, clip etc pages) the title block would be `{{ meta_context.titlePrefix }}`

BPScott:
Admittedly we don't need the full ancestry right now, but we might as well do it right once rather than have to change it later.

src/ValueObject/MetaContext.php
```diff
@@ -33,9 +33,7 @@ public function __construct($context = null, string $canonicalUrl = '')
             $this->description = $context->getShortSynopsis();
             $this->image = $context->getImage();
             $this->isRadio = $context->isRadio();
-            if ($context->getNetwork()) {
-                $this->titlePrefix = $context->getNetwork()->getName();
-            }
+            $this->setCoreEntityTitlePrefix($context);
```
BPScott: I'm not a big fan of these setter functions as it's not immediately clear what the value ends up changing as it's a side-effect rather than the return value. I think a having`setCoreEntityTitlePrefix` be a pure function and having this line be `$this->titlePrefix = $this->coreEntityTitlePrefix($context);` would be clearer

tests/ValueObject/MetaContextTest.php
```diff
@@ -30,18 +30,21 @@ public function testWithProgrammeContext()
...
+        $network->method('getName')->willReturn('Network name');
 
         $programme = $this->createMock(Programme::class);
+        $programme->method('getTleo')->willReturnSelf();
+        $programme->method('getTitle')->willReturn('Programme Title');
         $programme->method('getNetwork')->willReturn($network);
         $programme->method('getImage')->willReturn($image);
         $programme->method('getShortSynopsis')->willReturn('A short synopsis');
         $programme->method('isRadio')->willReturn(true);
+        $programme->method('getAncestry')->willReturn([$programme]);
```
BPScott: How about if there were multiple CoreEntities in the ancestry?

votemike: You want an additional test for that?

PR #118, opened 2017-08-03
PROGRAMMES-5648 Stats tracking

Tasiobg:
This PR adds iStats tracking for all pages.

Two new classes have been created AnalyticsCounterName and AnalyticsLabels, those classes are called each time we render a page with "renderWithChrome" and attach the parameters to the Orbit library.

This is a copy of V2 behaviour when posibble, see the ticket comments to find the V2 classes.

Orbit client sent iStats data via an AJAX request to "sa.bbc.co.uk".

The requirement for this ticket is to implement it only for schedules pages but this code already  covers future V3 pages because logic mimics V2.

If anyone prefer a walk through the code, please feel free to request it.


programmes-developers:

src/Controller/SchedulesByDayController.php
```diff
@@ -81,7 +83,61 @@ public function __invoke(
...
-        return $this->renderWithChrome('schedules/by_day.html.twig', $viewData);
+
+        $istatsExtraLabels = $this->getIstatsExtraLabels($date);
+
+        return $this->renderWithChrome('schedules/by_day.html.twig', $viewData, $istatsExtraLabels);
+    }
+
+    private function getIstatsExtraLabels($date): array
+    {
+        if (!isset($date)) {
```
votemike: Could you reverse this check? Otherwise the else means not not set.

src/Controller/SchedulesByDayController.php
```diff
@@ -81,7 +83,61 @@ public function __invoke(
...
+     * Examples returned: "-5", "0", "+3"
+     *
+     * @param ChronosInterface $urlDate
+     * @return string
+     */
+    private function getScheduleOffset(ChronosInterface $urlDate): string
+    {
+        $diffInDays = (string) $urlDate->diffInDays();
+        if ($urlDate->isPast()) {
+            $diffInDays = '-' . $diffInDays;
```
votemike: You could just return here instead of setting a variable.

src/Controller/SchedulesByDayController.php
```diff
@@ -81,7 +83,61 @@ public function __invoke(
...
+            $diffInDays = '+' . $diffInDays;
+        }
+        return $diffInDays;
+    }
+
+    private function getScheduleContext(ChronosInterface $urlDate): string
+    {
+        if ($urlDate->isPast()) {
+            return 'past';
+        } elseif ($urlDate->isToday()) {
```
votemike: Should you be checking `isToday()` first?

src/Controller/SchedulesByDayController.php
```diff
@@ -81,7 +83,61 @@ public function __invoke(
...
+            return 'today';
+        } else {
+            return 'future';
+        }
+    }
+
+    private function getScheduleCurrentFortnight(ChronosInterface $urlDate): string
+    {
+        if ($urlDate->diffInDays() > 7) {
```
votemike: Convert to a ternary?

src/ValueObject/AnalyticsLabels.php
```diff
@@ -0,0 +1,228 @@
...
+     *
+     * @param $context
+     * @return string
+     */
+    private function getParentTitlesRecursively($context): string
+    {
+        if (isset($context)) {
+            // concatenate title with the parents titles
+            $titlesConcatenation = $this->getParentTitlesRecursively($context->getParent());
+            if (!empty($titlesConcatenation)) {
```
votemike: Reverse this check?

src/ValueObject/AnalyticsLabels.php
```diff
@@ -0,0 +1,228 @@
...
+            }
+        } else {
+            return '';
+        }
+    }
+
+    /**
+     * Returns The PIDs of the different genres the programme belongs to, concatenated
+     *
+     * @param $context
```
votemike: What is $context?

src/ValueObject/AnalyticsLabels.php
```diff
@@ -0,0 +1,228 @@
...
+     * @return string
+     */
+    private function getGenreLabel($context): string
+    {
+        $genres = $context->getGenres($context);
+        $genreLabel = '';
+        foreach ($genres as $genre) {
```
votemike: Is this not just an implode?

src/ValueObject/AnalyticsLabels.php
```diff
@@ -0,0 +1,228 @@
...
+        if (method_exists($context, 'getMediaType')) {
+            $mediaType = $context->getMediaType();
+            if (!empty($mediaType)) {
+                $mediaType = ($context->getMediaType() == 'audio') ? 'audio' : 'video';
+            } else {
+                $mediaType = 'null';
+            }
+            return $mediaType . '_null_2';
+        } else {
+            return 'null_null_2';
```
votemike: What's this magic string?

src/ValueObject/AnalyticsLabels.php
```diff
@@ -0,0 +1,228 @@
...
+    {
+        if (method_exists($context, 'getMediaType')) {
+            $mediaType = $context->getMediaType();
+            if (!empty($mediaType)) {
+                $mediaType = ($context->getMediaType() == 'audio') ? 'audio' : 'video';
+            } else {
+                $mediaType = 'null';
+            }
+            return $mediaType . '_null_2';
+        } else {
```
votemike: You are returning from the `if`, so no need for the `else`

src/Controller/SchedulesByDayController.php
Tasiobg: Could do but it looks to me like personal coding style, not sure what is best, open to see what other people think

src/Controller/SchedulesByDayController.php
Tasiobg: Open to see what other people think, not sure what is best, to me this order is easy to read.

src/ValueObject/AnalyticsLabels.php
Tasiobg: Some $context don't have getMediaType so it should go to the else

src/Controller/BaseController.php
```diff
@@ -92,8 +96,9 @@ protected function response(): Response
         return $this->response;
     }
 
-    protected function renderWithChrome(string $view, array $parameters = [])
+    protected function renderWithChrome(string $view, array $parameters = [], array $istatsExtraLabels = [])
```
BPScott: If we keep finding new things we need to add, this method signature will end up getting very long.

What do you think to having a `setIstatsLabels()` function that sets a private property that we can then reference (in the same way we've got `setContext()`)

src/Controller/SchedulesByDayController.php
BPScott: I like the early return pattern, we use those in most other places in the code to avoid `else` statements

src/Controller/SchedulesByDayController.php
BPScott: could be `if ($date)`

src/ValueObject/AnalyticsLabels.php
Tasiobg: This logic is copied from V2 (Labels.php:43), this is a RecEng label, when the $context has no media should be set to 'null_null_2'

src/Controller/SchedulesByDayController.php
```diff
@@ -81,7 +83,61 @@ public function __invoke(
...
+    }
+
+    private function getIstatsExtraLabels($date): array
+    {
+        if (!isset($date)) {
+            $urlDate = ApplicationTime::getLocalTime();
+        } else {
+            $urlDate = Date::createFromFormat('Y-m-d', $date, ApplicationTime::getLocalTimeZone());
+        }
+        $labels['schedule_offset'] = $this->getScheduleOffset($urlDate);
```
BPScott: Does PHP complain here because you're referencing a previously undefined variable? Might be more readable to return an array immediatly:

```php
return [
  'schedule_offset' => $this->getScheduleOffset($urlDate),
  'schedule_context' => $this->getScheduleContext($urlDate),
  'schedule_current_fortnight' => $this->getScheduleCurrentFortnight($urlDate),
];
```

src/Controller/SchedulesByDayController.php
BPScott: Also will this this susceptible to the same fencing issues that @votemike  has been battling with?

src/Controller/SchedulesByDayController.php
```diff
@@ -81,7 +83,61 @@ public function __invoke(
...
+        $labels['schedule_context'] = $this->getScheduleContext($urlDate);
+        $labels['schedule_current_fortnight'] = $this->getScheduleCurrentFortnight($urlDate);
+        return $labels;
+    }
+
+    /**
+     * Returns a string representation of days between the current schedules page and today.
+     * Examples returned: "-5", "0", "+3"
+     *
+     * @param ChronosInterface $urlDate
```
BPScott: Don't need these `@param` and `@return` docblocks as they don't add any new information. The function signature already tells you this info.

src/Controller/SchedulesByDayController.php
BPScott: If you're doing an early return you can do another if block rather than an elseif

src/ValueObject/AnalyticsCounterName.php
```diff
@@ -0,0 +1,155 @@
...
+
+    private function getProgrammeCounterNamePart(Programme $context, string $relativePath): string
+    {
+        // Example: programmes.doctor_who.brand.b006q2x0.page
+        $partialString = $this->getParentTitlesRecursively($context);
+        $partialString .= '.' . $this->getProgrammeType($context);
+        $partialString .= '.' . $context->getPid();
+
+        // add the rest of the URL to the counter name
+        $restOfUrl = str_replace('/programmes/', '', $relativePath);
```
BPScott: str_replace takes an array of values to search/replace which would be more efficient than calling it 3 times

src/ValueObject/AnalyticsLabels.php
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    /**
+     * DUPLICATED FUNCTION, CAN I MOVE IT INSIDE PROGRAMMES CLASS?
+     *
+     * @param Programme $context
+     * @return string
+     * @throws Exception
+     */
+    private function getProgrammeType(Programme $context)
```
BPScott: This function is cropping up quite a lot. Moving it into CoreEntity would be useful.

I think it can be made quite a bit shorter too:

```php
return strtolower((new \ReflectionClass($this))->getShortName());
```

tests/Controller/SchedulesByMonthControllerTest.php
```diff
@@ -52,4 +52,25 @@ public function testServiceIsNotFound()
...
+    {
+        $this->loadFixtures(["NetworksAndServicesFixture"]);
+
+        $client = static::createClient();
+        $url = '/schedules/p00rfdrb/' . $scheduleDate;
+
+        $crawler = $client->request('GET', $url);
+
+        $labels = $this->extractIstatsLabels($crawler);
+        $this->assertEquals('programmes', $labels['app_name']);
```
BPScott: Is it worthwhile checking the standard labels in the schedule tests or should we only bother with the labels that get added by the schedule page?

It looks like we're checking those standard labels in 3 different tests atm.

tests/ValueObject/AnalyticsLabelsTest.php
```diff
@@ -0,0 +1,115 @@
+<?php
+declare(strict_types = 1);
+namespace Tests\App\Controller;
```
BPScott: Wrong namespace

tests/ValueObject/AnalyticsLabelsTest.php
```diff
@@ -0,0 +1,115 @@
...
+use BBC\ProgrammesPagesService\Domain\Entity\MasterBrand;
+use BBC\ProgrammesPagesService\Domain\Entity\Network;
+use BBC\ProgrammesPagesService\Domain\Entity\Service;
+use BBC\ProgrammesPagesService\Domain\ValueObject\Mid;
+use BBC\ProgrammesPagesService\Domain\ValueObject\Nid;
+use BBC\ProgrammesPagesService\Domain\ValueObject\Pid;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @covers App\ValueObject\AnalyticsLabels
```
BPScott: No need for the covers annotation in simple unit tests.

We use them for controller tests because the controller tests are integration tests that run the whole code base. This simple unit test doesn't touch anything outside AnalyticsLabels so we don't need to explicitly limit the test's coverage stats

tests/ValueObject/CounterNameTest.php
```diff
@@ -0,0 +1,211 @@
+<?php
+declare(strict_types = 1);
+namespace Tests\App\Controller;
```
BPScott: wrong namespace

tests/ValueObject/CounterNameTest.php
```diff
@@ -0,0 +1,211 @@
...
+use BBC\ProgrammesPagesService\Domain\Entity\Service;
+use BBC\ProgrammesPagesService\Domain\Entity\Unfetched\UnfetchedProgrammeItem;
+use BBC\ProgrammesPagesService\Domain\ValueObject\Pid;
+use BBC\ProgrammesPagesService\Domain\ValueObject\Sid;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Based on V2 BBC_Programmes_Helper_Nedstat_Test
+ *
+ * @covers App\ValueObject\AnalyticsCounterName
```
BPScott: no need for the covers annotation, see above

src/ValueObject/AnalyticsLabels.php
```diff
@@ -0,0 +1,228 @@
...
+    private $orbLabels = [];
+
+    public function __construct($context, string $controllerName, string $appVersion, array $extraLabels)
+    {
+        $this->labels = [
+            'app_name'           => 'programmes',
+            'prod_name'          => 'programmes',
+            // In V2 progs_page_type is "controller + action" but in V3 is only the controller name.
+            // A regex is used on iStats page to map V2 and V3 values
+            'progs_page_type'    => $controllerName,
```
BjornTwachtmann: The comment says that progs_page_type is being mapped to V2 values, but I can't see that happening anywhere. Am I missing something? I think we probably need to do this.

src/ValueObject/AnalyticsLabels.php
```diff
@@ -0,0 +1,228 @@
...
+            'bbc_radio_scotland_music_extra' => 'iplayerradio-scotland',
+        ];
+
+        if (($context instanceof CoreEntity || $context instanceof Service) && !empty($context->getNetwork())) {
+            $nid = (string) $context->getNetwork()->getNid();
+            if ($nid === 'bbc_music') {
+                $this->labels['bbc_site'] = 'music';
+            } elseif ($nid === 'bbc_arts') {
+                $this->labels['bbc_site'] = 'arts';
+            } elseif ($context->getNetwork()->getMedium() === 'tv') {
```
BjornTwachtmann: Both coreEntity and Service have an isTV() method.

src/ValueObject/AnalyticsLabels.php
```diff
@@ -0,0 +1,228 @@
...
+                $this->labels['bbc_site'] = 'tvandiplayer';
+            } elseif (isset($specialCases[$nid])) {
+                $this->labels['bbc_site'] = $specialCases[$nid];
+            } elseif ($context->getNetwork()->getMedium() === 'radio') {
```
BjornTwachtmann: there's an isRadio method too

src/ValueObject/AnalyticsLabels.php
votemike: I mean that you could have
```
if (method_exists($context, 'getMediaType')) {
  $mediaType = $context->getMediaType();
  if (!empty($mediaType)) {
    $mediaType = ($context->getMediaType() == 'audio') ? 'audio' : 'video';
  } else {
    $mediaType = 'null';
  }
  return $mediaType . '_null_2';
}
return 'null_null_2';
```
It's not a massive problem, it just gets rid of an `else`.  
Although it might be nice to check `!method_exists()` as you could then get rid of a level of indentation. Like:
```
if (!method_exists($context, 'getMediaType')) {
  return 'null_null_2';
}
$mediaType = $context->getMediaType();
$mediaType = 'null';
if (!empty($mediaType)) {
  $mediaType = ($context->getMediaType() == 'audio') ? 'audio' : 'video';
}
return $mediaType . '_null_2';
```

src/ValueObject/AnalyticsLabels.php
```diff
@@ -0,0 +1,228 @@
...
+        }
+        return $genreLabel;
+    }
+
+    private function getMediaTypeLabel($context): string
+    {
+        if (method_exists($context, 'getMediaType')) {
+            $mediaType = $context->getMediaType();
+            if (!empty($mediaType)) {
+                $mediaType = ($context->getMediaType() == 'audio') ? 'audio' : 'video';
```
votemike: What are the possible values of `$context->getMediaType()`? Is EVERYTHING that's not audio, video?

src/Controller/SchedulesByDayController.php
votemike: I mean for correctness. Not readability.  
If the $urlDate is 2017-01-01 06:00:00 and the time now is 2017-01-01 06:00:01. Then 'past' will be returned. When it's actually today. Right?

src/Controller/SchedulesByDayController.php
```diff
@@ -81,7 +83,57 @@ public function __invoke(
...
             $this->response()->setStatusCode(404);
         }
-        return $this->renderWithChrome('schedules/by_day.html.twig', $viewData);
+
+        $istatsExtraLabels = $this->getIstatsExtraLabels($date);
+
+        return $this->renderWithChrome('schedules/by_day.html.twig', $viewData, $istatsExtraLabels);
+    }
+
+    private function getIstatsExtraLabels($date): array
```
votemike: What is `$date`? A `string`? A `Date`? A `DateTime`? A `Chronos`?

src/ValueObject/AnalyticsLabels.php
BPScott: https://github.com/bbc/programmes-pages-service/blob/2cc1f01356af6fa58153af0a74e5acf8e6d6440c/src/Domain/Enumeration/MediaTypeEnum.php

src/ValueObject/AnalyticsLabels.php
BPScott: Instead of `method_exists($context, 'getMediaType')` use `$context instanceof ProgrammeItem`

Soooo...

```php
$mediaType = 'null';

if ($context instanceof ProgrammeItem) {
  $mediaType = ($context->getMediaType() == 'audio') ? 'audio' : 'video';
}

return $mediaType . '_null_2';
```

src/ValueObject/AnalyticsLabels.php
votemike: And if it is unknown it should be "video"?

BPScott: Seems like an odd choice yeah. There's 217870 episodes with an unknown media type (no clips). I guess it is for backfilled content which never had any on-demand asset (i.e. something you can play out).

Let's just use the value `$context->getMediaType()` gives us rather than doing any logic.

src/Controller/SchedulesByDayController.php
Tasiobg: Is that what you mean guys?

```
$diffInDays = (string) $urlDate->diffInDays();
if ($urlDate->isPast()) {
    return '-' . $diffInDays;
} elseif ($urlDate->isFuture()) {
    return '+' . $diffInDays;
}
return $diffInDays;
```

votemike:
```
$diffInDays = (string) $urlDate->diffInDays();
if ($urlDate->isPast()) {
    return '-' . $diffInDays;
}
if ($urlDate->isFuture()) {
    return '+' . $diffInDays;
}
return $diffInDays;
```

votemike: Also, not sure this code works as you think it will anyway. Are there tests for it?

Tasiobg: Yes, test name is "testSchedulesIstatsLabels"

Thanks for the example. Why are we avoiding "elseif" statement or in other words, why is this example better?

src/Controller/SchedulesByDayController.php
```diff
@@ -81,7 +83,57 @@ public function __invoke(
...
+    {
+        if (isset($date)) {
+            $urlDate = Date::createFromFormat('Y-m-d', $date, ApplicationTime::getLocalTimeZone());
+        } else {
+            $urlDate = ApplicationTime::getLocalTime();
```
votemike: If this is the case you know the values below will always be 0, today and true don't you? So there should be not need to do the logic below if you end up in this if statement?

src/ValueObject/AnalyticsLabels.php
Tasiobg: $genre is not a value, is a class and can't be converted to string, something like this is not working `$genreLabel = implode(', ', $genres); ` assuming this is what you mean, please  correct me if I'm wrong

votemike: I have no idea what I was thinking when I wrote this.... haha.

votemike: Maybe something like
```
$ids = array_map("getId", $context->getGenres($context));
return implode(', ', $ids);
```
Would be shorter/tidier?

src/ValueObject/AnalyticsLabels.php
```diff
@@ -0,0 +1,228 @@
...
+
+class AnalyticsLabels
+{
+    /** @var array */
+    private $labels = [];
+
+    /** @var array */
+    private $orbLabels = [];
+
+    public function __construct($context, string $controllerName, string $appVersion, array $extraLabels)
```
votemike: What is a $context?

src/ValueObject/AnalyticsLabels.php
```diff
@@ -0,0 +1,228 @@
...
+
+    /**
+     * Returns The PIDs of the different genres the programme belongs to, concatenated
+     *
+     * @param $context
+     * @return string
+     */
+    private function getGenreLabel($context): string
+    {
+        $genres = $context->getGenres($context);
```
votemike: Don't need to pass in `$context`

src/ValueObject/AnalyticsLabels.php
```diff
@@ -0,0 +1,228 @@
...
+use BBC\ProgrammesPagesService\Domain\Entity\Episode;
+use BBC\ProgrammesPagesService\Domain\Entity\Franchise;
+use BBC\ProgrammesPagesService\Domain\Entity\Gallery;
+use BBC\ProgrammesPagesService\Domain\Entity\Programme;
+use BBC\ProgrammesPagesService\Domain\Entity\Season;
+use BBC\ProgrammesPagesService\Domain\Entity\Series;
+use BBC\ProgrammesPagesService\Domain\Entity\Service;
+use Exception;
+
+class AnalyticsLabels
```
votemike: Should this be abstract and then have different classes for the different type of context?

src/ValueObject/AnalyticsCounterName.php
```diff
@@ -0,0 +1,155 @@
...
+        // Suffix
+        $this->counterName .= '.page';
+    }
+
+    public function __toString(): string
+    {
+        return $this->counterName;
+    }
+
+    private function getServiceCounterNamePart($context, string $relativePath): string
```
votemike: `$context` is always a Service here. Right?

src/ValueObject/AnalyticsCounterName.php
```diff
@@ -0,0 +1,155 @@
...
+
+        // add the rest of the URL to the counter name
+        $restOfUrl = str_replace('/programmes/', '', $relativePath);
+        $restOfUrl = str_replace($context->getPid(), '', $restOfUrl);
+        $restOfUrl = str_replace('/', '.', $restOfUrl);
+        $partialString .= $restOfUrl;
+        return $partialString;
+    }
+
+    private function getProgrammeType(Programme $context)
```
votemike: Can you not just return `str_lower(get_class())`?

src/ValueObject/AnalyticsCounterName.php
```diff
@@ -0,0 +1,155 @@
...
+     * @param $context
+     * @return string
+     */
+    private function getParentTitlesRecursively($context): string
+    {
+        if (isset($context)) {
+            // concatenate title with the parents titles
+            $titlesConcatenation = $this->getParentTitlesRecursively($context->getParent());
+            return $titlesConcatenation . '.' . $this->replaceDisallowedCharacters($context->getTitle());
+        } else {
```
votemike: Can just `return '';`. No need to nest it in an `else`

src/ValueObject/AnalyticsCounterName.php
```diff
@@ -0,0 +1,155 @@
...
+use BBC\ProgrammesPagesService\Domain\Entity\Episode;
+use BBC\ProgrammesPagesService\Domain\Entity\Franchise;
+use BBC\ProgrammesPagesService\Domain\Entity\Gallery;
+use BBC\ProgrammesPagesService\Domain\Entity\Programme;
+use BBC\ProgrammesPagesService\Domain\Entity\Season;
+use BBC\ProgrammesPagesService\Domain\Entity\Series;
+use BBC\ProgrammesPagesService\Domain\Entity\Service;
+use Exception;
+
+class AnalyticsCounterName
```
votemike: Should this be abstract?

src/Controller/SchedulesByDayController.php
Tasiobg: yes, this will do the same
```
        if (isset($date)) {
            $urlDate = Date::createFromFormat('Y-m-d', $date, ApplicationTime::getLocalTimeZone());
            return [
                'schedule_offset' => $this->getScheduleOffset($urlDate),
                'schedule_context' => $this->getScheduleContext($urlDate),
                'schedule_current_fortnight' => $this->getScheduleCurrentFortnight($urlDate),
            ];
        } else {
            // today
            return [
                'schedule_offset' => '0',
                'schedule_context' => 'today',
                'schedule_current_fortnight' => 'true',
            ];
        }
```
 what do you think @BPScott abot doing it like that?

src/ValueObject/AnalyticsLabels.php
Tasiobg: I was going to benchmark the option you mention because I haven't used array_map much but looks like the "getId" is not a valid callback becuause the function doesn't exist in the current class, its exist inside @context so I get this error: array_map() expects parameter 1 to be a valid callback, function 'getId' not found

Should be fine the way it is now

src/Controller/SchedulesByDayController.php
```diff
@@ -81,7 +83,57 @@ public function __invoke(
...
+
+        $istatsExtraLabels = $this->getIstatsExtraLabels($date);
+
+        return $this->renderWithChrome('schedules/by_day.html.twig', $viewData, $istatsExtraLabels);
+    }
+
+    private function getIstatsExtraLabels($date): array
+    {
+        if (isset($date)) {
+            $urlDate = Date::createFromFormat('Y-m-d', $date, ApplicationTime::getLocalTimeZone());
```
votemike: I think this should by 'Y-m-d|' so that we are always measuring from midnight. Otherwise we may get confusing statistics. Unless we are supposed to be measuring by broadcast day, which will require more logic.  
@BPScott any views?

src/ValueObject/AnalyticsLabels.php
BPScott: I prefer building an array and imploding it rather than lots of string concatenation. That way you don't need to have that if conditional.

I don't mind if that array is built with a foreach loop or using functional programming. Personally I prefer the functional approach where possible as it fees a bit shorter:

```php
$ids = implode(', ', array_map(function($g) {
    return $g->getId();
}, $context->getGenres()));
```

src/ValueObject/AnalyticsLabels.php
Tasiobg: I can do it like that assuming there is no concern about not validating the class name.

src/ValueObject/AnalyticsLabels.php
Tasiobg: anyone else can share their views on it?

tests/Controller/SchedulesByMonthControllerTest.php
Tasiobg: My aproach was to test as much as posibble from the controller request. Do you think I should remove the standard labels from the schedules test?

src/ValueObject/AnalyticsLabels.php
BPScott: What would the different classes be?

I don't think it's worth it, as then you have to do logic about which version of Analytics labels you instantiate, which feels like overhead.

src/ValueObject/AnalyticsCounterName.php
BPScott: See reasoning in AnalyticsLabels, it's essentially the same discussion.

src/Controller/BaseController.php
Tasiobg: Good idea

src/Controller/SchedulesByDayController.php
BPScott: I ~think~ this might need to be done from the start of the broadcast day, otherwise if you hit the page at 3am it'll give you the wrong offsets.

Hitting the a TV schedule /2017-01-04 at 2017-01-05 03:00:00 whould show you the schedule for 2017-01-04 06:00 till 2017-01-05 06:00. Which should give you a stats 'offset' of 0 as you're still on "today's schedule". However if we use midnight as the timing point then it'll think you're a day off.

tests/Controller/SchedulesByMonthControllerTest.php
BPScott: I think we should test the standard labels in some other controller test. Otherwise if we wanted to be consistent we'd end up testing all the standard labels in every controller test where the page has additional labels, which seems like unneeded duplication.

Tasiobg: ok, I'm removing "app_name" and "prod_name" asserts, should I remove any other?

composer.json
```diff
@@ -25,7 +25,7 @@
         "ext-redis": "*",
         "bbc-rmp/cloudwatch-monitoringhandler": "^1.0",
         "bbc-rmp/translate": "^1.8",
-        "bbc/branding-client": "^2.1.0",
+        "bbc/branding-client": "^2.2.0",
```
votemike: Should the composer.lock file not also have changed?

src/Controller/BaseController.php
```diff
@@ -150,6 +166,16 @@ protected function redirect($url, $status = 302): RedirectResponse
         return new RedirectResponse($url, $status, $headers);
     }
 
+    protected function setIstatsExtraLabels(array $labels): void
```
votemike: An array of strings? And array of Label objects? An array of arrays?

src/Controller/HomeController.php
```diff
@@ -10,6 +10,7 @@ class HomeController extends BaseController
 {
     public function __invoke(EntityManagerInterface $em, NetworksService $networksService)
     {
+        $this->setIstatsProgsPageType('home_index');
```
votemike: Does this need to be done in every controller? Would it be best for a method in the base constructor to call an abstract method to ensure all controllers set this?

src/Controller/SchedulesByDayController.php
```diff
@@ -80,9 +83,63 @@ public function __invoke(
...
+     * Returns a string representation of days between the current schedules page and today.
+     * Examples returned: "-5", "0", "+3"
+     */
+    private function getScheduleOffset(ChronosInterface $urlDate): string
+    {
+        $diffInDays = (string) $urlDate->diffInDays();
+        if ($urlDate->isPast()) {
+            return '-' . $diffInDays;
+        }
+        if ($urlDate->isFuture()) {
```
votemike: `diffInDays()`, `isPast()`, `isFuture()` and `isToday()` all make new date instances. It would be best to avoid this. I have an idea on how to refactor... but too long to write. Come and grab me when you're ready and we can discuss.

src/ValueObject/AnalyticsLabels.php
```diff
@@ -0,0 +1,182 @@
...
+        $this->setProgrammeLabels($context);
+        $this->setBbcSiteLabel($context);
+        $this->setEventMasterBrand($context);
+
+        if (!empty($extraLabels)) {
+            $this->labels = array_merge($this->labels, $extraLabels);
+        }
+    }
+
+    public function orbLabels(): array
```
votemike: An array of what? Arrays?

src/ValueObject/AnalyticsLabels.php
```diff
@@ -0,0 +1,182 @@
...
+namespace App\ValueObject;
+
+use BBC\ProgrammesPagesService\Domain\Entity\CoreEntity;
+use BBC\ProgrammesPagesService\Domain\Entity\Programme;
+use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeItem;
+use BBC\ProgrammesPagesService\Domain\Entity\Service;
+
+class AnalyticsLabels
+{
+    /** @var array */
```
votemike: An array of what?

src/ValueObject/AnalyticsLabels.php
```diff
@@ -0,0 +1,182 @@
...
+            $this->labels['rec_v'] = '2'; // The version of RecEng being used
+            $this->labels['rec_app_id'] = 'programmes'; // The app id of the client calling RecEng
+            $this->labels['rec_p'] = $this->getMediaTypeLabel($context);
+        }
+    }
+
+    /**
+     * This function uses recursion to climb up the parents tree and return a concatenation of the titles
+     */
+    private function getParentTitlesRecursively($context): string
```
votemike: This method could be refactored to reduce indentation.

src/ValueObject/AnalyticsLabels.php
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    /**
+     * Returns The PIDs of the different genres the programme belongs to, concatenated
+     */
+    private function getGenreLabel($context): string
+    {
+        $genres = $context->getGenres();
+        $genreLabel = '';
+        foreach ($genres as $genre) {
```
votemike: I thought this was going to be refactored?

composer.json
Tasiobg: composer.lock got "bbc/branding-client" updated yesterday with your branch (Housekeeping) and this is rebased on master.

src/Controller/BaseController.php
```diff
@@ -150,6 +166,19 @@ protected function redirect($url, $status = 302): RedirectResponse
         return new RedirectResponse($url, $status, $headers);
     }
 
+    /**
+     * @param array $labels associative array. example: ['schedule_offset' => '-3', 'schedule_context' => 'past']
```
votemike: Would you be able to change `@param array` to `@param mixed[]`? Keep the explanation too though.

src/Controller/HomeController.php
Tasiobg: That's one of the implementations I did yesterday, after a review with @BPScott I have been suggested to implement it the way it is now because new non existing values on V2 should not manually set this value, instead they should use the controller name.

Tasiobg: This has already been discussed

src/Controller/SchedulesByDayController.php
```diff
@@ -80,9 +82,65 @@ public function __invoke(
...
+    }
+
+    /**
+     * Returns a string representation of days between the current schedules page and today.
+     * Examples returned: "-5", "0", "+3"
+     */
+    private function getScheduleOffset(int $diffInDays): string
+    {
+        // past
+        if ($diffInDays > 0) {
```
votemike: Isn't this the wrong way around? Shouldn't it be `if ($diffInDays < 0) {`? Or am I getting confused with what `diffInDays()` returns?

Tasiobg: example returned valuews for this code
```
$urlDate = Date::createFromFormat('Y-m-d', $date, ApplicationTime::getLocalTimeZone());
$diffInDays = $urlDate->diffInDays(null, false);
var_dump($date, $diffInDays); exit;
```
'2017-08-01', 9
'2017-09-01', -22

votemike: Weird way for them to choose to do that.

src/Controller/BaseController.php
```diff
@@ -103,12 +112,19 @@ protected function renderWithChrome(string $view, array $parameters = [])
 
         $translateProvider->setLocale($locale);
 
+        if (!isset($this->istatsProgsPageType)) {
```
BPScott: We know this is always set so we can use the shorter syntax `if ($this->istatsProgsPageType)`

src/Controller/HomeController.php
BPScott: An abstract method would mean that we can only ever have one controller method per class. Which admittedly I like, but I don't think we should totally close the door on having many `*Action` functions in a single class.


(I suggested exactly the same thing before realising that corner-case)

src/ValueObject/AnalyticsLabels.php
```diff
@@ -0,0 +1,176 @@
...
+            'bbc_radio_five_live_sports_extra' => 'iplayerradio-5livesportsextra',
+            'bbc_6music' => 'iplayerradio-6music',
+            'bbc_asian_network' => 'iplayerradio-asiannetwork',
+            'bbc_radio_cymru' => 'iplayerradio-radiocymru',
+            'bbc_radio_cymru_mwy' => 'iplayerradio-radiocymru',
+            'bbc_radio_scotland' => 'iplayerradio-scotland',
+            'bbc_radio_scotland_music_extra' => 'iplayerradio-scotland',
+        ];
+
+        if (($context instanceof CoreEntity || $context instanceof Service) && !empty($context->getNetwork())) {
```
BPScott: You can use `!$context->getNetwork()` instead of `!empty($context->getNetwork())`

src/ValueObject/AnalyticsLabels.php
```diff
@@ -0,0 +1,176 @@
...
+            } elseif ($context->getNetwork()->isRadio()) {
+                $this->labels['bbc_site'] = 'iplayerradio';
+            }
+        }
+    }
+
+    private function setEventMasterBrand($context): void
+    {
+        if ($context instanceof CoreEntity && !empty($context->getMasterBrand())) {
+            $this->labels['event_master_brand'] = (string) $context->getMasterBrand()->getMid();
```
BPScott: I think even for core entities we want the network NID instead of the masterbrand MID.

Find a programme that has the masterbrand bbc_one_scotland, and see what it does on the current site - is this label bbc_one_scotland or bbc_one?

Tasiobg:
This is the original @BPScott message

> Hitting the a TV schedule /2017-01-04 at 2017-01-05 03:00:00 whould show you the schedule for 2017-01-04 06:00 till 2017-01-05 06:00. Which should give you a stats 'offset' of 0 as you're still on "today's schedule". However if we use midnight as the timing point then it'll think you're a day off.

getIstatsExtraLabels() will always return same values if there is not date in the URL, I think we are fine, please feel free to provide more feedback if anyone think I could be wrong, I may not be full aware of the concept of a broadcast day.

votemike:
OK. If you get the tests working, I'll have a look and see if I think it's fine as it is or if it needs to be changed.

src/ValueObject/AnalyticsLabels.php
Tasiobg: Labe is "bbc_one_scotland" for this 2 pids "b0080wsx" and "b00w4gnm", same value as in this implementation. I have created this method with @BjornTwachtmann help, may be he coul provide more info if he remembers the logic

votemike:
This test fails. I believe it should work.
```
public function testGetThisWorking()
    {
        $this->loadFixtures(["BroadcastsFixture"]);

        $timeViewing = new Chronos('2017-01-05 03:00:00');
        ApplicationTime::setTime($timeViewing->timestamp);
        Date::setTestNow($timeViewing); //Move this line to inside ApplicationTime::setTime()
        $scheduleViewing = $timeViewing->subDay()->setTime(11, 0, 0);

        $client = static::createClient();
        $url = '/schedules/p00fzl6p'; // TV I believe
        $url .= '/' . $scheduleViewing->format('Y-m-d');
        $crawler = $client->request('GET', $url);

        $labels = $this->extractIstatsLabels($crawler);
        $this->assertEquals('0', $labels['schedule_offset']);
        $this->assertEquals('today', $labels['schedule_context']);
        $this->assertEquals('true', $labels['schedule_current_fortnight']);
    }
```

src/ValueObject/AnalyticsLabels.php
Tasiobg: Do you mean this?
```
if (($context instanceof CoreEntity || $context instanceof Service) && $context->getNetwork()) { 
```
Otherwise it is having the opposite impact

src/Controller/SchedulesByDayController.php
```diff
@@ -80,9 +82,72 @@ public function __invoke(
...
+            $urlDate = Date::createFromFormat('Y-m-d', $date, $tz);
+            $diffInDays = Date::now($tz)->diffInDays($urlDate, false);
+            return [
+                'schedule_offset' => $this->getScheduleOffset($diffInDays),
+                'schedule_context' => $this->getScheduleContext($diffInDays),
+                'schedule_current_fortnight' => $this->getScheduleCurrentFortnight($diffInDays),
+            ];
+        }
+
+        // for TV if there is no date in the URL and is before 6:00 we are displaying yesterday schedule page.
```
Tasiobg: @BPScott Can you provide feedback about the use of $broadcastStartedYesterday in this function? feel free to suggest a better wording for the comments.

tests/Controller/SchedulesByDayControllerTest.php
```diff
@@ -50,6 +50,60 @@ public function scheduleDateTestProvider(): array
...
+     * @dataProvider scheduleDateIstatsTestProvider
+     * @param string $network           The pid of the network
+     * @param null|string $scheduleDate      The date the user is viewing the schedule for, can be null if $timeNow is set
+     * @param string $bbcSite           iStats label
+     * @param string $eventMasterBrand  iStats label
+     * @param string $scheduleOffset    iStats label
+     * @param string $scheduleContext   iStats label
+     * @param string $scheduleFortnight iStats label
+     */
+    public function testSchedulesIstatsLabels(
```
Tasiobg: @BPScott 
This test fails if runned between 00:00 and 5:59 because today TV broadcast returns different values.
Should I use `Chronos::setTestNow` to fake the date and test both cases or do you have a different point of view?

votemike: You should test both before and after 6AM

tests/Controller/SchedulesByDayControllerTest.php
```diff
@@ -50,6 +50,96 @@ public function scheduleDateTestProvider(): array
...
+            'radio-tomorrow'  => ['p00fzl8v', $dateTomorrow, 'iplayerradio-radio2', 'bbc_radio_two', '+1', 'future', 'true'],
+            'radio-yesterday' => ['p00fzl8v', $dateYesterday, 'iplayerradio-radio2', 'bbc_radio_two', '-1', 'past', 'true'],
+            'tv-with-date'    => ['p00fzl6p', $dateTwentyDaysAgo, 'tvandiplayer', 'bbc_one', '-20', 'past', 'false'],
+        ];
+    }
+
+    /**
+     * BroadcastDay start at 6:00, this method test schedules page before and after 6:00
+     */
+    public function testSchedulesIstatsLabelsAtDifferentDayTime()
```
votemike: Have you also got a test for a radio schedule. @BPScott how should the radio stuff work?

Tasiobg: My understanding and test points that radio is not afected by this behaviour, only TV is displaying yesterday date if the today schedule is not started yet.

src/ValueObject/AnalyticsCounterName.php
```diff
@@ -0,0 +1,108 @@
...
+        return $this->counterName;
+    }
+
+    private function getServiceCounterNamePart(Service $context, string $relativePath): string
+    {
+        // Example: programmes.schedules.bbc_one_cambridge.2017.07.17.page
+        $pid = $context->getPid();
+        $sid = $context->getSid();
+        $partialString = str_replace($pid, $sid, $relativePath);
+        $partialString = str_replace('/', '.', $partialString);
```
votemike: `$partialString = str_replace(['/', '-'], '.', $partialString);`?

Tasiobg: thanks

tests/Controller/SchedulesByDayControllerTest.php
Tasiobg: Test added for both cases

votemike: Where abouts?

Tasiobg: SchedulesByDayControllerTest -> testSchedulesIstatsLabelsAtDifferentDayTime

src/Controller/SchedulesByDayController.php
BPScott: I'm not familiar enough  with how the existing code works off the top of my head, so I can't offer any insight, sorry.

`$urlDate = Date::createFromFormat('Y-m-d', $date, $tz);` will cause some date math fencing headaches though. Let's leave it for now though as this PR has gone on long enough :) we can create a new issue to look into the stats labels for schedule dates and confirm they're working as expected.

src/ValueObject/AnalyticsLabels.php
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    /**
+     * This function uses recursion to climb up the parents tree and return a concatenation of the titles
+     */
+    private function getParentTitlesRecursively(?Programme $context): string
+    {
+        if (isset($context)) {
+            // concatenate title with the parents titles
+            $titlesConcatenation = $this->getParentTitlesRecursively($context->getParent());
```
BPScott: There is a $context->getAncestry() that will return current item and all it's parents. Using that might be a bit more readable than recursion

src/ValueObject/AnalyticsLabels.php
BPScott: Ok that's cool. I wasn't sure off the top off my head. Thank you for checking.

tests/Controller/SchedulesByDayControllerTest.php
BPScott: I ~think~ radio should be unaffected as their days start at 00:00:00, but a test to prove that is the case would be handy (again we can finish off the schedule day stats at a later date, let's not bother keeping this open any longer)

src/ValueObject/AnalyticsLabels.php
Tasiobg: I didn't know about getAncestry method, recursion is not easy to read, thanks!!

PR #120, opened 2017-08-04
Add base DsAmen CSS, JS, Presenter and factory

BjornTwachtmann:


programmes-developers:

app/Resources/views/base_ds2013.html.twig
```diff
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+<0!DOCTYPE html>
```
felipeparaujo: What's this 0 for?

src/DsAmen/Core/Vendor/_mq.scss
```diff
@@ -0,0 +1,280 @@
+@charset "UTF-8"; // Fixes an issue where Ruby locale is not set properly
```
BPScott: We already pull in MQ through yarn dependencies, so we don't need to include it within here.

(Previously we included our vendors deps because Amen was pulled in through compser not npm).

See https://github.com/bbc/programmes-frontend/blob/master/src/Ds2013/_ds2013-components.scss#L15 for an example of how we include it

src/DsAmen/Core/Settings/_mixins.scss
```diff
@@ -0,0 +1,105 @@
+// flexbox
```
BjornTwachtmann: Hmm. There's various bits of magic here that deal with fallback options (inline-block fallback for flexbox, some magical stuff for IE10, etc. I will remove the ones that are purely vendor prefixes though

src/DsAmen/Core/Settings/_colors.scss
```diff
@@ -0,0 +1,152 @@
+//
```
BjornTwachtmann: Some of the variables are used in various places. I've commented the unused ones just so they're there for reference

PR #122, opened 2017-08-07
5732 map

votemike:
Framework for Media Availability Panel https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5732

Only the grid layout is part of this PR, not the contents of the columns.


programmes-developers:

src/Controller/FindByPid/TlecController.php
```diff
@@ -16,12 +18,23 @@
  */
 class TlecController extends BaseController
 {
-    public function __invoke(ProgrammeContainer $programme, ProgrammesService $programmesService)
+    public function __invoke(ProgrammeContainer $programme, ProgrammesService $programmesService, CollapsedBroadcastsService $collapsedBroadcastsService)
     {
         $this->setContext($programme);
 
+        $hasComingSoon = $programme->getOption('coming_soon') || $programme->getOption('comingsoon_textonly');
+        $programmeHasEpisodes = $programme->getAggregatedEpisodesCount() > 0; //@TODO is this correct?
```
votemike: Is this the equivalent of `$this->_controller->context->programme->has_episodes`?

src/DsAmen/Organism/Map/MapPresenter.php
```diff
@@ -0,0 +1,157 @@
...
+
+    /** @var string */
+    private $rightGridClasses = '1/2@gel3b';
+
+    /** @var bool */
+    private $showMiniMap = false;
+
+    public function __construct(ProgrammeContainer $programme, int $upcomingEpisodesCount, ?CollapsedBroadcast $mostRecentBroadcast, array $options = [])
+    {
+        //@TODO What about things that aren't using the 2016 page, such as Being Human http://www.bbc.co.uk/programmes/b01dzwlw
```
votemike: @BPScott What are we doing about pages using the old styling?

src/DsAmen/Organism/Map/MapPresenter.php
```diff
@@ -0,0 +1,157 @@
...
+            $this->rightColumns[] = 'OnDemand';
+            $this->rightColumns[] = 'ComingSoon';
+        } elseif ($this->isWorldNews()) {
+            $this->rightColumns[] = 'LastOn';
+            $this->rightColumns[] = 'Tx';
+        } elseif ($this->mustShowTxColumn($hasComingSoon, $upcomingEpisodesCount, $mostRecentBroadcast)) {
+            $this->mustShowTxColumn = true;
+
+            // Minimap has a different set of widths
+            if ($this->programme->getOption('showMinimap')) { //@TODO showMinimap used to be a function, do I need to recreate that?
```
votemike: @BPScott How should this check work? The old method seemed confusing and convoluted.

src/DsAmen/Organism/Map/MapPresenter.php
```diff
@@ -0,0 +1,157 @@
...
+
+    public function getProgramme(): Programme
+    {
+        return $this->programme;
+    }
+
+    public function getPromo()
+    {
+        /* iVote */
+        //$iVotePanel = $this->_getVote($this->context->programme); //@TODO
```
votemike: Will iVote be done as part of a separate ticket?

src/DsAmen/Organism/Map/map.html.twig
```diff
@@ -0,0 +1,33 @@
+{% set left_classes = build_css_classes({
```
votemike: Should we both with these two build_css_classes or just add them on to $this->leftGridClasses etc... in the Presenter?

src/Controller/FindByPid/TlecController.php
BPScott: Yes, it is

src/DsAmen/Organism/Map/MapPresenter.php
BPScott: We shall be flicking the flagpole that forces all TV TLECs to use the 2016 brand page within the next month or so.
The plan is that once we launch this for TV TLECs then we can update it to add features that radio request, at which point we can sent all TLECs to frontend.

Thus we can build assuming all pages shall use the 2016 style.

src/DsAmen/Organism/Map/MapPresenter.php
BPScott: We would want to force the minimap view (i.e. no images in the MAP) when the page is in vote priority view, so either that needs to be a function in the `Options` object, or have a function in this class that does logic based on the showMiniMap and isVotePriority options.

The name for this option is `brand_2016_layout_use_minimap`, from https://github.com/bbc/programmes-pages-service/blob/master/src/Mapper/ProgrammesDbToDomain/OptionsMapper.php

src/DsAmen/Organism/Map/MapPresenter.php
BPScott: Yes. We currently have no idea how the voting module shall be provided to us, as it is out for tender atm.

src/DsAmen/Organism/Map/map.html.twig
BPScott: If there's nothing conditional (i.e. only add this class is a particular condition is met), there's not much point using build_css_classes. Stick them in the presenter or include them when building out the div (`<div class="grid__item map__item {{map.LeftGridClasses}}">` stylee)

src/DsAmen/Organism/Map/MapPresenter.php
votemike: Should I just leave out any logic to do with voting? Or leave it in but comment it out. Or what? We wouldn't want the logic to get lost.

src/Controller/FindByPid/TlecController.php
```diff
@@ -16,12 +20,21 @@
...
 class TlecController extends BaseController
 {
-    public function __invoke(ProgrammeContainer $programme, ProgrammesService $programmesService)
+    public function __invoke(Request $request, ProgrammeContainer $programme, ProgrammesService $programmesService, CollapsedBroadcastsService $collapsedBroadcastsService)
     {
         $this->setContext($programme);
 
+        $upcomingEpisodesCount = $collapsedBroadcastsService->countUpcomingByProgramme($programme);
+        $mostRecentBroadcast = null;
+        if (empty($upcomingEpisodesCount)) {
```
BPScott: if (!$upcomingEpisodesCount)

src/DsAmen/Organism/Map/MapPresenter.php
```diff
@@ -0,0 +1,176 @@
...
+    private $leftGridClasses = '1/2@gel3b';
+
+    /** @var bool */
+    private $mustShowTxColumn = false;
+
+    /** @var Programme */
+    private $programme;
+
+    /** @var Presenter[] */
+    private $rightColumns;
```
BPScott: initialise to empty array? `private $rightColumns = [];`

I'm not sure what happens you do `$var[] = 1` when `$var` is null but should probably avoid it.

src/DsAmen/Organism/Map/MapPresenter.php
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    public function showMap(): bool
+    {
+        return $this->showMap;
+    }
+
+    private function isWorldNews(): bool
+    {
+        $masterbrand = $this->programme->getMasterBrand();
```
BPScott: This should be based on the network rather than the masterbrand.

Can be `$this->programme->getNetwork()->isWorldNewsInternational()`?

votemike: There is no isWorldNewsInternational() method...

PR #124, opened 2017-08-08
Update phpcs with symfony 3 coding standard

tatitati:
This new phpcs follow the same rules than before (the new library is an extension of the symfony2 with small differences):
- now is not allowed anymore multiple white lines
- output in color and displaying only errors (no more warning with low severity)
- now is not allowed anymore to have a file without one last empty line

----

To check exactly the rules that are applied now with the new library check:
https://github.com/Endouble/Symfony3-custom-coding-standard/blob/master/Symfony3Custom/ruleset.xml


programmes-developers:

phpcs.xml.dist
```diff
@@ -1,63 +1,25 @@
...
-    <rule ref="Generic.Metrics.NestingLevel">
-        <properties>
-            <property name="nestingLevel" value="3" />
-        </properties>
+    <!-- COMMAND LINE OPTIONS -->
+    <arg value="np"/>
+    <arg name="colors"/>
+    <arg name="extensions" value="php"/>
+    <file>./src</file>
+    <file>./tests</file>
```
felipeparaujo: why remove the rest of the excludes? we probably don't want to analyse what's on those folders, especially vendor

tatitati: well, we can exclude all the folders that we don't want to analyze, or we can specify the folder we want to analyze (is less text)

composer.json
```diff
@@ -41,12 +42,12 @@
...
     "require-dev": {
         "doctrine/doctrine-fixtures-bundle": "^2.3",
-        "escapestudios/symfony2-coding-standard": "^2.10",
         "liip/functional-test-bundle": "^1.7",
         "phpstan/phpstan": "^0.8.0",
         "phpunit/phpunit": "^6.1",
         "squizlabs/php_codesniffer": "^2.9",
-        "symfony/phpunit-bridge": "^3.3.0"
+        "symfony/phpunit-bridge": "^3.3.0",
+        "endouble/symfony3-custom-coding-standard": "^2.10"
```
votemike: sort-packages is set to true, how did you manage to get a package out of alphabetical order?

composer.json
```diff
@@ -17,7 +17,8 @@
         { "type": "vcs", "url": "git@github.com:bbc/DoctrineExtensions.git" },
         { "type": "vcs", "url": "git@github.com:bbc/programmes-pages-service.git" },
         { "type": "vcs", "url": "git@github.com:bbc/rmp-translate.git" },
-        { "type": "vcs", "url": "git@github.com:bbc/rmp-monitoringhandler.git" }
+        { "type": "vcs", "url": "git@github.com:bbc/rmp-monitoringhandler.git" },
+        { "type": "vcs", "url": "git@github.com:Endouble/Symfony3-custom-coding-standard"}
```
felipeparaujo: this package is on packagist, so you don't need to add the repository here

felipeparaujo: https://packagist.org/packages/endouble/symfony3-custom-coding-standard

tatitati: ouch! (I will remember for the future, thanks)

composer.json
tatitati: I guess without running composer require.

votemike: Could you move it? Otherwise the next time someone uses composer require, this will move even though it has nothing to do with the next persons PR.

phpcs.xml.dist
felipeparaujo: ah, fair enough

phpcs.xml.dist
```diff
@@ -1,63 +1,29 @@
...
+        <!-- We don't care if parameter names all line up as if one changes then the diff will be noisy -->
+        <exclude name="Symfony3Custom.Commenting.FunctionComment.SpacingAfterParamType"/>
+        <exclude name="Symfony3Custom.Commenting.FunctionComment.SpacingAfterParamName"/>
+        <exclude name="Symfony3Custom.Commenting.FunctionComment.MissingParamTag"/>
 
         <!-- We don't care about requiring a blank line before a return -->
-        <exclude name="Symfony2.Formatting.BlankLineBeforeReturn"/>
-
-        <!-- We don't care about abstract class names having to be prefixed with "Abstract" -->
-        <exclude name="Symfony2.NamingConventions.ValidClassName.InvalidAbstractName"/>
```
BPScott: Is there a reason why this was removed?

phpcs.xml.dist
```diff
@@ -1,63 +1,29 @@
...
-        <!-- We don't care if parameter names all line up as if one changes then
-            the diff will be noisy -->
-        <exclude name="Symfony2.Commenting.FunctionComment.SpacingAfterParamType"/>
-
-        <!-- This doesn't enforce what we want it to.
-        See https://github.com/escapestudios/Symfony2-coding-standard/issues/27 -->
-        <exclude name="PEAR.Functions.FunctionCallSignature.Indent"/>
-    </rule>
-
-    <!-- Concatenation should be surrounded by a single space -->
```
BPScott: Is there a reason why this was removed?

phpcs.xml.dist
```diff
@@ -1,63 +1,29 @@
...
-    <rule ref="Squiz.Strings.ConcatenationSpacing">
-        <properties>
-           <property name="spacing" value="1"/>
-           <property name="ignoreNewlines" value="true"/>
-        </properties>
-    </rule>
-
-    <rule ref="Generic.Metrics.NestingLevel">
```
BPScott: Is there a reason why this was removed?

phpcs.xml.dist
```diff
@@ -1,63 +1,29 @@
 <?xml version="1.0"?>
-<ruleset name="Symfony-Programmes">
+<ruleset name="programmes-frontend-phpcs">
```
BPScott: Is there much point adding "phpcs" to the end of a PHPCS ruleset? Surely it's always going to be a php-cs ruleset. Also Ideally this file should be copy-pastable into new projects without needing to change anything, and having to change the name frontend might be annoying.

phpcs.xml.dist
tatitati: Yes. In the readme.md of the new component there is a section saying:
```
Customizations:
......
In Sniff/NamingConventions/ValidClassNameSniff
   - remove the abstract class name rule
```
Is this section:
https://github.com/Endouble/Symfony3-custom-coding-standard#customizations

phpcs.xml.dist
tatitati: Similar
```
Customizations:
....
Changed the concatenation spacing rule, for readability, to require 1 space around concatenation dot, instead of no spaces as the Symfony standard requires.
```

BPScott:
Ok that looks good, though we're still getting failures, fix them and this can be merged.

PR #126, opened 2017-08-08
Rdfa tags

felipeparaujo:


programmes-developers:

BjornTwachtmann:
Looks sensible to me.

src/ValueObject/MetaContext.php
```diff
@@ -25,6 +29,12 @@ class MetaContext
     /** @var string */
     private $canonicalUrl = '';
 
+    /** @var string */
+    private $schemaType = '';
+
+    /** @var string */
+    private $vocabulary = 'http://schema.org/';
```
votemike: Does this need to be a property?

src/ValueObject/MetaContext.php
```diff
@@ -83,6 +90,16 @@ public function titlePrefix(): string
         return $this->titlePrefix;
     }
 
+    public function getRdfaAttributes(): string
+    {
+        $tags = 'vocab=' . $this->vocabulary;
+        if (!empty($this->schemaType)) {
+            $tags .= ' typeof=' . $this->schemaType;
```
votemike: Tempted to build an array an implode on a space. But not super important.

src/ValueObject/MetaContext.php
```diff
@@ -98,4 +115,27 @@ private function coreEntityTitlePrefix(CoreEntity $coreEntity): string
...
+
+        if ($entity instanceof Brand) {
+            return $type . 'Series';
+        }
+
+        if ($entity instanceof Clip) {
+            return $type . 'Clip';
+        }
+
+        return '';
```
votemike: throw an Exception???

src/ValueObject/MetaContext.php
felipeparaujo: Not really, I just want to make clear which vocab we're using

src/ValueObject/MetaContext.php
felipeparaujo: Why? it's just something we haven't mapped yet. In that case, we just don't add the typeof tag.

src/ValueObject/MetaContext.php
felipeparaujo: I'm changing this, but it's going to look cleaner without the implode

src/ValueObject/MetaContext.php
```diff
@@ -83,6 +87,15 @@ public function titlePrefix(): string
         return $this->titlePrefix;
     }
 
+    public function getRdfaAttributes(): string
+    {
+        if (empty($this->schemaType)) {
```
BPScott: `if (!$this->schemaType)`

PR #128, opened 2017-08-08
Initial work into working out how adverts do their thing

BPScott:
* Enable blocking ORB only when ads should be served.
* Only inject the ad block when ads should be served.
* Add a _cloudlabs/adverts route to test ads

### DIFF
--- a/app/Resources/views/base_ds2013.html.twig
+++ b/app/Resources/views/base_ds2013.html.twig
@@ -1,3 +1,5 @@
+{%- set hasAdverts = adverts_set_blocks(meta_context.showAdverts and block('adsToDisplay') is defined ? block('adsToDisplay') : '') -%}
+
 <!DOCTYPE html>
 <html class="{{ branding.orbitThemeClasses|raw }} {% block html_classes %}{% endblock %}" lang="{{ tr('language_code') }}">
     <head>
@@ -36,6 +38,12 @@
             <meta name="apple-itunes-app" content="app-id=560458506, app-argument={{ meta_context.canonicalUrl }}?bbc_referrer={{ meta_context.canonicalUrl|url_encode }}">
         {% endif %}
 
+        {#
+            bbcdotcom adverts requires synchronous fig loading.
+            See https://navigation.api.bbc.co.uk/docs/fig.md#using-a-blocking-fig
+            Remove this once Ads work without the blocking fig - GNLADAPPS-38
+        #}
+        {% if hasAdverts %}<script>window.orb_fig_blocking = true;</script>{% endif %}
         {{ orb.head|raw }}
         {{ branding.head|raw }}
 
@@ -62,6 +70,7 @@
                 }
             });
         </script>
+        {{ adverts_head() }}
     </head>
     <body {{ meta_context.rdfaAttributes }}>
         {{ orb.bodyfirst|raw }}
--- a/app/Resources/views/cloud_labs/adverts.html.twig
+++ b/app/Resources/views/cloud_labs/adverts.html.twig
@@ -1,0 +1,16 @@
+{% extends 'base_ds2013.html.twig' %}
+
+{% block title %}Adverts Test{% endblock %}
+{% block adsToDisplay %}leaderboard mpu{% endblock %}
+
+{% block body %}
+    <div class="br-box-page">
+        <p>You'll need to trick the FIG into thinking you're outside the UK for these adverts to appear</p>
+
+        <h1>LEADERBOARD</h1>
+        {{ advert('leaderboard', [1,2,3,4]) }}
+
+        <h1>MPU</h1>
+        {{ advert('mpu', [1,2,3,4]) }}
+    </div>
+{% endblock %}
--- a/src/Twig/BbcDotComExtension.php
+++ b/src/Twig/BbcDotComExtension.php
@@ -1,0 +1,98 @@
+<?php
+declare(strict_types = 1);
+namespace App\Twig;
+
+use Twig_Extension;
+use Twig_Function;
+
+class BbcDotComExtension extends Twig_Extension
+{
+     /** @var string */
+    private $advertBlocks = '';
+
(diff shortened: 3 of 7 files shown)


programmes-developers:

app/Resources/views/base_ds2013.html.twig
```diff
@@ -36,6 +38,12 @@
             <meta name="apple-itunes-app" content="app-id=560458506, app-argument={{ meta_context.canonicalUrl }}?bbc_referrer={{ meta_context.canonicalUrl|url_encode }}">
         {% endif %}
 
+        {#
+            bbcdotcom adverts requires synchronous fig loading.
+            See https://navigation.api.bbc.co.uk/docs/fig.md#using-a-blocking-fig
+            Remove this once Ads work without the blocking fig - GNLADAPPS-38
+        #}
+        {% if hasAdverts %}<script>window.orb_fig_blocking = true;</script>{% endif %}
```
votemike: Would it be better to always set this?

app/Resources/views/cloud_labs/adverts.html.twig
```diff
@@ -0,0 +1,16 @@
+{% extends 'base_ds2013.html.twig' %}
+
+{% block title %}Adverts Test{% endblock %}
+{% block adsToDisplay %}leaderboard mpu{% endblock %}
+
+{% block body %}
+    <div class="br-box-page">
+        <p>You'll need to trick the FIG into thinking you're outside the UK for these adverts to appear</p>
```
votemike: Allow it to be toggled with a $_GET param?

src/Twig/BbcDotComExtension.php
```diff
@@ -0,0 +1,87 @@
...
+            new Twig_Function('adverts_head', [$this, 'advertsHead'], ['is_safe' => ['html']]),
+        ];
+    }
+
+    /**
+     * Display an Advert
+     */
+    public function advert(
+        string $advertId,
+        array $groups,
```
votemike: What is $groups?

src/Twig/BbcDotComExtension.php
```diff
@@ -0,0 +1,87 @@
...
+        $ad = sprintf($template, $divId, $advertId, $advertGroups);
+
+        if ($containerClass) {
+            $ad = '<div class="' . $containerClass . '">' . $ad . '</div>';
+        }
+
+        return $ad;
+    }
+
+    public function advertsConfig(string $advertBlocks): bool
```
votemike: This is a weird name for a method that returns a bool.

src/Twig/BbcDotComExtension.php
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public function advertsHead(): string
+    {
+        if (!$this->advertBlocks) {
+            return '';
+        }
+
+        $template = <<<HEAD
+<script>
```
votemike: Why is this not all in a twig template?

src/ValueObject/MetaContext.php
```diff
@@ -42,6 +45,10 @@ public function __construct($context = null, string $canonicalUrl = '')
             $this->isRadio = $context->isRadio();
             $this->titlePrefix = $this->coreEntityTitlePrefix($context);
             $this->schemaType = $this->getSchemaTypeEquivalent($context);
+
+            if ($context->getNetwork()->isInternational()) {
```
votemike: Network might be null. No?

src/ValueObject/MetaContext.php
```diff
@@ -42,6 +45,10 @@ public function __construct($context = null, string $canonicalUrl = '')
...
+                $this->showAdverts = true;
```
votemike: `$this->showAdverts = $context->getNetwork()->isInternational()`?

app/Resources/views/base_ds2013.html.twig
BPScott: Making it the fig loading blocking has a negative performance impact (as the rest of the page waits for this bit of JS to load before loading the rest of the page, rather than loading the JS in the background) and is deprecated, so I'd rather only make it blocking when we need to

app/Resources/views/cloud_labs/adverts.html.twig
BPScott: Doesn't work that way unfortunately, we don't have access to override that value as it is set in other people's code and we don't have the ability to override it on our side.

The trick is to change your proxy settings to look like you're in the US. (There is another admin panel override toggle thing but I couldn't get that to work)

src/Twig/BbcDotComExtension.php
BPScott: It is. 

I need a way to check if adverts are enabled to set the synchronous fig loading. Adding an extra twig helper just for that seemed annoying, so making this initialisation function pull double-duty for now and return if there should be adverts or not based on the config it got passed in seemed like the simplest way to do that.

`adverts_set_blocks()` perhaps? At least then "it returns a boolean that reflects if the blocks were set or not" makes a little more sense

src/Twig/BbcDotComExtension.php
BPScott: It requires a little bit of state - we shouldn't show any advert blocks or initialisation if we don't want to show adverts (we only show adverts on certain pages belonging to programmes on international networks). Placing the HTML in here keeps all the stuff relating to adverts all in one place rather than spreading it about a few places.

src/Twig/BbcDotComExtension.php
votemike: Something like that sounds better.

src/ValueObject/MetaContext.php
BPScott: Yep, good spot.

src/Twig/BbcDotComExtension.php
BPScott: Done.

PR #130, opened 2017-08-09
Fixing unwanted newline in date block

felipeparaujo:


programmes-developers:

tatitati:
simple 👍

app/Resources/views/schedules/by_base.html.twig
```diff
@@ -1,8 +1,6 @@
 {% extends 'base_ds2013.html.twig' %}
 
-{%- block title -%}
-    {{ meta_context.titlePrefix }} - {{ tr('schedules') }}, {{ block('date') }}
-{%- endblock -%}
+{% block title %}{{ meta_context.titlePrefix }} - {{ tr('schedules') }}, {{ block('date') }}{% endblock %}
```
votemike: Did `{%-` not make it spaceless?

felipeparaujo: yeah it did, but this block is so short we could have it in one line, at which point we don't need the `{%-` anymore

BjornTwachtmann:
looks sensible

PR #131, opened 2017-08-09
Controller clips and galleries for brandpage

tatitati:
First PR for "Build Brand Page controller skeleton" ticket.
This PR includes clips/galleries data from db


programmes-developers:

PR #132, opened 2017-08-09
5738 programmeinfo

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5738


src/DsAmen/Core/Generic/_moreless.scss
```diff
@@ -0,0 +1,111 @@
+.ml__status {
```
votemike: Are we happy with where this SASS file lives?

src/DsAmen/Molecule/Image/ImagePresenter.php
```diff
@@ -0,0 +1,146 @@
...
+use App\DsAmen\Presenter;
+use BBC\ProgrammesPagesService\Domain\Entity\Image;
+use InvalidArgumentException;
+
+class ImagePresenter extends Presenter
+{
+    /** @inheritdoc */
+    protected $options = [
+        'is_lazy_loaded' => true,
+        'srcsets' => [80, 160, 320, 480, 640, 768, 896, 1008],
```
votemike: Should these numbers be constants of some sort?

programmes-developers:

src/Controller/FindByPid/TlecController.php
```diff
@@ -44,10 +46,13 @@ public function __invoke(
...
         if ($upcomingEpisodesCount === 0) {
-            $mostRecentBroadcast = $collapsedBroadcastsService->findPastByProgramme($programme, 1)[0];
+            $pastBroadcasts = $collapsedBroadcastsService->findPastByProgramme($programme, 1);
+            if (!empty($pastBroadcasts)) {
+                $mostRecentBroadcast = $pastBroadcasts[0];
+            }
         }
 
-        $mapPresenter = new MapPresenter($request, $programme, $upcomingEpisodesCount, $mostRecentBroadcast);
+        $mapPresenter = new MapPresenter($request, $helperFactory, $programme, $upcomingEpisodesCount, $mostRecentBroadcast);
```
BjornTwachtmann: We should probably be creating the map presenter using the presenter factory rather than passing the helper factory into the controller

src/DsAmen/Core/Generic/_moreless.scss
BjornTwachtmann: seems fine

src/Controller/FindByPid/TlecController.php
votemike: So pass the PresenterFactory in to the `__invoke` method?

src/DsAmen/Molecule/Image/_image.scss
```diff
@@ -0,0 +1,9 @@
+.image {
```
BjornTwachtmann: I don't think this CSS file needs to be here, the .img-rsp etc already exists in Atom/images

src/DsAmen/Core/Generic/_moreless.scss
BPScott: `DsAmen/Molecule/_moreless.scss`

One day I'll write up the logic for what goes where, sorry.

src/DsAmen/Molecule/Image/image.html.twig
```diff
@@ -0,0 +1 @@
+<img data-molecule="responsive-image" class="img-rsp" src="{{ image.src }}" srcset="{{ image.srcsets }}" sizes="{{ image.sizes }}" alt="{{ image.option('alt') }}" />
```
BPScott: Should sizes be "auto" same as the Ds2013 image?

src/DsAmen/Organism/Map/MapPresenter.php
```diff
@@ -51,11 +55,12 @@ class MapPresenter extends Presenter
     /** @var Request */
     private $request;
 
-    public function __construct(Request $request, ProgrammeContainer $programme, int $upcomingEpisodesCount, ?CollapsedBroadcast $mostRecentBroadcast, array $options = [])
+    public function __construct(Request $request, HelperFactory $helperFactory, ProgrammeContainer $programme, int $upcomingEpisodesCount, ?CollapsedBroadcast $mostRecentBroadcast, array $options = [])
```
BPScott: This is getting pretty long, time to split onto multiple lines?

src/DsAmen/Organism/Map/MapPresenter.php
```diff
@@ -105,12 +113,14 @@ public function getProgramme(): Programme
...
-            'coming_soon_takeover' => $this->comingSoonTakeover,
-            'world_news' => $this->isWorldNews(),
-            'must_show_tx_column' => $this->mustShowTxColumn,
-            'show_mini_map' => $this->showMiniMap,
-        ]);
+        return new ProgrammeInfoPresenter(
+            $this->helperFactory->getProseToParagraphsHelper(),
+            $this->programme,
+            [
+                '3column' => $this->comingSoonTakeover || $this->isWorldNews() || $this->mustShowTxColumn,
```
BPScott: Do we want these options to be prefixed with verbs so they answer questions? "is_three_column" perhaps?

You've already got that style with "show_mini_map" instead of "mini_map".

src/DsAmen/Organism/Map/SubPresenter/OnDemandPresenter.php
```diff
@@ -19,7 +19,7 @@ public function __construct(Programme $programme, array $options = [])
         $this->programme = $programme;
 
         if ($this->getOption('must_show_tx_column')) {
-            $this->class = '1/2@gel1b';
+            $this->class = '1/2@gel1b'; //@TODO figure out the difference between this and 1/1
```
BPScott: 1/1 shall trigger mean full width at all screen sizes. 1/2@gel1b shall make the div that has this be half width when the screen is wide than the "gel1b" size (which is 320px, as defined at https://github.com/bbc/programmes-frontend/blob/5f19f265ea4af6ae95f70b74e6ed13734d1dd69a/src/DsAmen/Core/_breakpoints.scss#L4)

tests/DsShared/Helpers/ProseToParagraphsHelperTest.php
```diff
@@ -0,0 +1,75 @@
...
+    private $mockText;
+
+    public function setUp()
+    {
+        $mockTranslate = $this->createMock(Translate::class);
+        $mockTranslateProvider = $this->createMock(TranslateProvider::class);
+        $mockTranslateProvider->method('getTranslate')->willReturn($mockTranslate);
+        $this->helper = new ProseToParagraphsHelper($mockTranslateProvider);
+        $space = ' ';
+        $this->mockText = "Lorem ipsum dolor sit
```
BPScott: Huh, i didn't realise you could do multiline strings with "", I always used the HEREDOC syntax.

src/DsAmen/Organism/Map/SubPresenter/OnDemandPresenter.php
votemike: But we couldn't demonstrate that yesterday.

src/DsAmen/Molecule/Image/image.html.twig
votemike: Should they?

BPScott: I think we should keep it consistent yeah.

src/DsAmen/Organism/Map/SubPresenter/ProgrammeInfoPresenter.php
```diff
@@ -31,7 +31,7 @@ public function __construct(ProseToParagraphsHelper $ptpHelper, Programme $progr
         $this->ptpHelper = $ptpHelper;
         $this->showMiniMap = $this->getOption('show_mini_map');
 
-        if ($this->getOption('3column')) {
+        if ($this->getOption('is_three_column')) {
```
BPScott: It looks like this presenter is missing the default values for these options

votemike: Done. But, do we want these as options? Or passed through as proper parameters?

PR #134, opened 2017-08-09
Cache the creation of IntlDateFormatter in the TranslateableTrait

BPScott:
Creating IntlDateFormatters are expensive and we don't need a whole new
one every time, instead we can just update the format.

This should save about 8ms on a schedule page.


programmes-developers:

BjornTwachtmann:
lgtm

PR #135, opened 2017-08-10
Rdfa attributes rework

felipeparaujo:


BjornTwachtmann:
Looks good, merge it.

src/ValueObject/MetaContext.php
```diff
@@ -75,11 +68,21 @@ public function canonicalUrl(): string
         return $this->canonicalUrl;
     }
 
+    public function context(): mixed
```
votemike: Didn't know you could use `: mixed`

src/ValueObject/MetaContext.php
```diff
@@ -29,22 +25,19 @@ class MetaContext
...
-    /** @var string */
-    private $schemaType = '';
-
     /** @var bool */
     private $showAdverts = false;
 
     public function __construct($context = null, string $canonicalUrl = '')
     {
         $this->canonicalUrl = $canonicalUrl;
+        $this->context = $context;
```
votemike: Do you need to initialise this property?

src/ValueObject/MetaContext.php
felipeparaujo: It's useless having a mixed return type tho, it's probably the default value

src/ValueObject/MetaContext.php
felipeparaujo: yes, I need that to get the schema type

votemike: I mean do you need:
```
/** @var CoreEntity|Network|Etc... */
private $context;
```

felipeparaujo: oh yeah forgot about that lol

src/ValueObject/MetaContext.php
BPScott: I don't think it's got much value

PR #136, opened 2017-08-10
Added service blacklist for schedule homepage

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5697


programmes-developers:

src/Controller/SchedulesHomeController.php
```diff
@@ -36,13 +36,19 @@ public function __invoke(ServicesService $servicesService)
 
         $decades = range($earliestDecade, $latestDecade, 10);
 
+        $blacklist = ['p00v5fbq', 'p00qvyk4', 'p00qvyjs'];
```
BPScott: The list on PROGRAMMES-5792 had a few more in. I reckon add them all to this list, we can do the smarter "hide the service if it has no broadcasts" thing at a later date, it's unlikely that any of those services will gain broadcasts any time soon.

src/Controller/SchedulesHomeController.php
```diff
@@ -36,13 +36,35 @@ public function __invoke(ServicesService $servicesService)
 
         $decades = range($earliestDecade, $latestDecade, 10);
 
+        // https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5792
+        $blacklist = [
```
felipeparaujo: Should this be an attribute?

votemike: Like a class property? Is there an advantage to that?

felipeparaujo: Honestly don't know. If you think this is fine, then I'm cool with it.

BjornTwachtmann:
looks fine to me

PR #137, opened 2017-08-11
Shared images

votemike:
Moving images from Ds2013 to DsShared


programmes-developers:

src/Ds2013/Presenter.php
```diff
@@ -1,118 +1,17 @@
...
-
-        // Trim the class name from the word 'Presenter'
-        $presenterName = substr($shortClassName, 0, strpos($shortClassName, 'Presenter'));
-
-        return strtolower(preg_replace(
-            ["/([A-Z]+)/", "/_([A-Z]+)([A-Z][a-z])/"],
-            ["_$1", "_$1_$2"],
-            lcfirst($presenterName)
-        ));
+        return '2013';
```
votemike: BEST FILE CHANGE EVAAAA

app/Resources/views/base_ds_amen.html.twig
```diff
@@ -49,9 +49,11 @@
 
         <!--[if (gt IE 9)|!(IE)]><!-->
         <link rel="stylesheet" href="{{ asset('css/dsAmen.css') }}" />
+        <link rel="stylesheet" href="{{ asset('css/dsShared.css') }}" />
```
BPScott: I don't think there is any value in having a separate stylesheet, instead items from the DsShared namespace can be `@import`ed into the dsAmen / ds2013 component files as needed

src/Ds2013/_ds2013-components.scss
```diff
@@ -73,7 +73,6 @@
 @import 'Molecule/Calendar/calendar-table';
 // @import 'objects/contact-detail';
 @import 'Molecule/highlight-box';
-@import 'Molecule/Image/image';
```
BPScott: Change this to `@import ../dsShared/Molecule/Image/image`, or whatever the path to the shared image is, so we can avoid having to load two stylesheets on the browser-side

src/DsShared/PresenterFactory.php
```diff
@@ -0,0 +1,46 @@
...
+    /** @var TranslateProvider */
+    private $translateProvider;
+
+    /** @var UrlGeneratorInterface */
+    private $router;
+
+    /** @var HelperFactory */
+    private $helperFactory;
+
+    public function __construct(TranslateProvider $translateProvider, UrlGeneratorInterface $router, HelperFactory $helperFactory)
```
BPScott: Is it worth injecting these if we don't need them yet?

src/Twig/DesignSystemPresenterExtension.php
```diff
@@ -18,39 +21,42 @@ class DesignSystemPresenterExtension extends Twig_Extension
...
-                'is_safe' => ['html'],
-                'is_variadic' => true,
-                'needs_environment' => true,
-            ]),
-            new Twig_Function('ds2013_presenter', [$this, 'ds2013Presenter'], [
-                'is_safe' => ['html'],
-                'needs_environment' => true,
-            ]),
-            new Twig_Function('ds_amen', [$this, 'dsAmen'], [
+        $systems = [
```
BPScott: Not convinced by dynamically building this array, I find it less readable. It's always going to end up being the same set of values and PHP7 has some [nice little optimisations around immutable arrays](https://blog.blackfire.io/php-7-performance-improvements-immutable-arrays.html)

votemike: OK. Want me to change it back?

src/Ds2013/_ds2013-components.scss
votemike: Done

BjornTwachtmann:
Looks fine to me. Yeah, it's probably best to move this stuff into a base class if we're going to have 3 sets of the same presenters.

PR #139, opened 2017-08-14
Fun with twig

BjornTwachtmann:
So... Twig is faster if we write function calls out in full. Not much faster, but enough for this to be worth it.


programmes-developers:

BPScott:
I for one welcome our new parenthesis based overlords.

We should really write up a style/formatting guide somewhere

PR #140, opened 2017-08-14
Speed up base presenter by caching templatePath and snakeCasePresente…

BjornTwachtmann:
…rName statically


programmes-developers:

BPScott:
How does this interact with the cases where we override the path / presentername? It's overriding the functions, but perhaps there is a nicer way to set those if we are dealing with cached properties?

Can we avoid the need for having a cache arrayson the base class by having static` `templateName` and `templateVariableName` properties per-class) using `static::` and late static binding to reference them? http://php.net/manual/en/language.oop5.late-static-bindings.php

BjornTwachtmann:
Just for reference, in anwser to bens comment (I anwsered these in person already). 1. The functions are overriden to return hardcoded strings, so this doesn't affect those at all.

2. Not really, that would mean we needed to create 2 static variables on every child of BasePresenter. This is less painful.

PR #141, opened 2017-08-15
5800 synopsis

votemike:
I have created this for https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5800  
It's ready enough for people to see if this is the direction I should take to address 5800. Which includes new logic.
**The Old Way:**
Used to take the longest synopsis and display it truncated with a toggle to show more.
**The New Way:**
Displays the longest synopsis if it is short enough to display.
Otherwise it takes the medium synopsis if short enough otherwise the short synopsis and displays that with the toggle which completely hides _that_ synopsis and displays the longest synopsis.

Please check out the branch, run `yarn gulp` and try this locally to see what you think. Maybe use the pids `p04z0yc2` and `b0647v1v` although keep in mind that the plan will be to use this other places than just the brand page. If it turns out this isn't suitable for episode pages, we should be able to revert this commit.


programmes-developers:

src/DsAmen/Molecule/Synopsis/SynopsisPresenter.php
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    private function sanitiseText(string $text): string
+    {
+        $text = strip_tags($text);
+        $text = htmlspecialchars($text, ENT_COMPAT, 'UTF-8');
+        $html = nl2br($text);
+        $html = trim(preg_replace('/\s+/', ' ', $html)); // remove left over new lines
+        $html = str_replace('/> ', '/>', $html); // remove space between tags
+        $html = preg_replace('/(<br \/>){2,}/', '</p><p>', $html); // replace multi <br /> with <p>
```
votemike: @BjornTwachtmann @BPScott part of the ticket asked to move HTML out of the PHP and in to the views. Do you think it would be better to return an array of paragraphs from this method that can then be foreached in the template so that we can instead add the `<p>` tags there?

votemike: I guess we then wouldn't need the raw filter in twig either.

BjornTwachtmann:
Looks good. Weren't you going to remove the "show less" stuff? Or did I get that wrong?

src/DsAmen/Molecule/Synopsis/SynopsisPresenter.php
```diff
@@ -0,0 +1,77 @@
...
+    private function countWithoutTags(string $text): int
+    {
+        return strlen(strip_tags($text));
+    }
+
+    /**
+     * @param string $text
+     * @return string[] An array of strings. Each string denotes a paragraph. Strings may include <br/> tags
+     */
+    private function sanitiseText(string $text): array
```
BPScott: What do you think to moving the "sanitize" stuff into the Synopses ValueObject? `getSanitisedShortSynopsis()` etc?

votemike:
@BjornTwachtmann Not sure where we landed on that decision. I can remove it if everyone thinks it's a good idea.

src/DsAmen/Molecule/Synopsis/SynopsisPresenter.php
votemike: I think no. As it includes `<br/>` tags which I don't think should come from the composer package.

votemike: Unless you want to move everything apart from the nl2br part in to the value object.

BjornTwachtmann: I wouldn't move this into pages service. If anything it might want its own helper at some point, but I think it's fine here for now.

src/DsAmen/Molecule/Synopsis/_synopsis.scss
```diff
@@ -0,0 +1,36 @@
+//$synopsis-toggle-fade-in-time: 10s;
+//$synopsis-toggle-fade-out-time: 3s;
```
tatitati: Should we delete these comments?

votemike: I'm tempted to leave them in for easy future debugging. They're in the SASS and never make it in to the compiled CSS.  
What do others think?

src/DsAmen/Molecule/Synopsis/SynopsisPresenter.php
BPScott: Ok groovy :)

src/DsAmen/Molecule/Synopsis/_synopsis.scss
BPScott: I reckon remove, it's trivial to change the fade times. Leaving this sort of stuff in without a "uncomment me to debug" message shall end in confusion.

tatitati: but I see that they are pretty much like the two below, so remember these options to debug should be easy, they are there.

PR #142, opened 2017-08-15
Caching router

BjornTwachtmann:
This shaves about 2ms average off the render time for the schedules page. Not sure if it's worth it. Discuss.


programmes-developers:

votemike:
Does Symfony's inbuilt route caching not cache well enough?

votemike:
Inside the `generate()` method I see the comment `// the Route has a cache of its own and is not recompiled as long as it does not get modified`. Does that not mean it's already cached?

BjornTwachtmann:
It caches the list of routes, but not the generated URL or the complex logic used to do that

BPScott:
I wonder if we could reduce the number of calls to generate a url in some other way - is this because somewhere we're calling generate multiple times for the same url and we should be caching it in a presenter?

BjornTwachtmann:
We could cache in the presenter, however there are multiple places where we're also generating the same URL from the view. 

We could sack this off and make sure the presenters are caching generated URLs, that would help a bit. Currently this reduces number of URLs generated from ~170 to ~100 on schedules_by_day.

votemike:
But this sounds as though the generated route is cached. https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Routing/Generator/UrlGenerator.php#L133

programmes-developers:

programmes-developers:

programmes-developers:

programmes-developers:

BPScott:
Gonna close this, as the indirection introduced isn't worth the performance gain.

PR #145, opened 2017-08-17
promotions-controller

tatitati:


programmes-developers:

src/Controller/FindByPid/TlecController.php
```diff
@@ -58,6 +65,8 @@ public function __invoke(
             'clips' => $clips,
             'galleries' => $galleries,
             'mapPresenter' => $mapPresenter,
+            'regularPromotions' => $promotions['regular'],
```
BPScott: The list of promos shall be displayed in one long list, so there's not much point passing both arrays in separately. Check current site to see if the regular or super promos go first.

Also have you confirmed that the promos shall be output in the correct order?

src/Controller/FindByPid/TlecController.php
```diff
@@ -42,6 +44,11 @@ public function __invoke(
             $galleries = $aggregationService->findDescendantGalleries($programme, 4);
         }
 
+        $promotions = $promotionService->findActivePromotionsByEntity($programme);
+        /**
+         * @todo: get related links for each promo
```
BPScott: This shall be done as part of the service level, so this is the wrong place for this todo statement.

src/Controller/FindByPid/TlecController.php
tatitati: Yeah. I can confirm that the order is the weight. No the type of promotion.
is_descendants | weight
0	|    1
....
1	|    8
0	|    9
0	|    10
1	|    11
0	|    12
1	|    13
0	|    14
0	|    15

programmes-developers:

PR #147, opened 2017-08-22
Add Amen media object

BPScott:
Needed as this is a base for the programme object.


programmes-developers:

PR #148, opened 2017-08-22
Make BasePresenter->getUniqueId() public

BPScott:
We'll need to use this within templates so don't hide it away.
Has the side-effect of simplifying testing this function.


programmes-developers:

PR #149, opened 2017-08-22
Date list presenter improvements

Tasiobg:
Some performance improvements.
- isLink() now gets a property value and this value is only set while calling the constructor
- Moved Chronos instance for "greaterThan" to the contructor to avoid instanciating the Chronos object multiple times
- Updated unitest


programmes-developers:

src/Ds2013/Molecule/DateList/AbstractDateListItemPresenter.php
```diff
@@ -34,7 +39,27 @@ public function __construct(UrlGeneratorInterface $router, ChronosInterface $dat
 
     abstract public function getLink(): string;
 
-    abstract public function isLink(): bool;
+    public function setIsLink(ChronosInterface $greaterThan) : void
```
votemike: This always accepts a `Chronos` doesn't it? Never a `Date`. So typehint `Chronos` instead?

src/Ds2013/Molecule/DateList/YearDateListItemPresenter.php
```diff
@@ -3,17 +3,20 @@
...
 use Cake\Chronos\ChronosInterface;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 
 class YearDateListItemPresenter extends AbstractDateListItemPresenter
 {
-    public function __construct(UrlGeneratorInterface $router, ChronosInterface $datetime, Service $service, int $offset, array $options = [])
+    public function __construct(UrlGeneratorInterface $router, ChronosInterface $datetime, Service $service, int $offset, Chronos $greaterThan, array $options = [])
     {
         parent::__construct($router, $datetime->addYears($offset), $service, $offset, $options);
+        $this->setIsLink($greaterThan);
```
votemike: `$this->setIsLink($greaterThan);` appears in all the children of the abstract class straight after the constructor. Maybe move that call to the abstract constructor? And then set the method as private?

tests/Ds2013/PresenterFactoryTest.php
```diff
@@ -81,6 +81,9 @@ public function testMoleculeDateList()
         $now = Chronos::now();
         $mockService = $this->createMock(Service::class);
 
+        // fake the time to make DateListPresenter item possible to be equal
```
votemike: This comment doesn't quite make sense to me.

src/Ds2013/Molecule/DateList/DayDateListItemPresenter.php
```diff
@@ -31,19 +34,8 @@ public function isGmt(): bool
...
-        }
-
-        $network = $this->service->getNetwork();
-        if (is_null($network)) {
-            return false;
-        }
-
-        $broadcastYear = new BroadcastDay($this->datetime, $network->getMedium());
-        return $broadcastYear->serviceIsActiveInThisPeriod($this->service);
+        return new BroadcastDay($this->datetime, $network->getMedium());
```
votemike: It feels a bit weird to me to be calling `getMedium()` on `$network` here. Do you think it might be more correct to either pass the medium through to the `getBroadcastPeriod()` method. Or not pass anything though and instead call `$this->service->getNetwork()->getMedium()`?

src/Ds2013/Molecule/DateList/AbstractDateListItemPresenter.php
```diff
@@ -34,7 +39,27 @@ public function __construct(UrlGeneratorInterface $router, ChronosInterface $dat
...
+    {
+        // if the date is more than $greaterThan from now, then don't allow a link (page will still exist)
+        if ($this->offset === 0 || $this->datetime->gte($greaterThan)) {
+            $this->isLink = false;
```
Tasiobg: On my point of view this is easy to read but If someone else share your view I'll change it

votemike: Fair enough. @BPScott @felipeparaujo @tatitati any views?

tests/Ds2013/PresenterFactoryTest.php
Tasiobg: Test was failing because the objects were not equal, this happen because "greaterThan" contains the microseconds of when the object was created.

Let me know if you know a better solution

votemike: You need to choose a time to set, otherwise it uses null. Which is no good.  
Maybe I'm misunderstanding you.

src/Ds2013/Molecule/DateList/DayDateListItemPresenter.php
Tasiobg: you right, I like the first option

src/Ds2013/Molecule/DateList/AbstractDateListItemPresenter.php
```diff
@@ -22,19 +25,43 @@
     /** @var Service */
     protected $service;
 
-    public function __construct(UrlGeneratorInterface $router, ChronosInterface $datetime, Service $service, int $offset, array $options = [])
+    /** @var  bool */
+    private $isLink;
+
+    public function __construct(UrlGeneratorInterface $router, ChronosInterface $datetime, Service $service, int $offset, Chronos $greaterThan, array $options = [])
```
BPScott: `$greaterThan` isn't very clear as a variable name. `$unavailableAfterDate`? or something like that?

src/Ds2013/Molecule/DateList/AbstractDateListItemPresenter.php
```diff
@@ -22,19 +25,43 @@
...
     {
         parent::__construct($options);
 
         $this->datetime = $datetime;
         $this->service = $service;
         $this->offset = $offset;
         $this->router = $router;
+        $this->setIsLink($greaterThan);
```
BPScott: Thought for discussion: I'm not a big fan of these private setters, I think it makes finding out what property is set unclear. In a few other places that need to do this sort of thing (notably ImagePresenter and it's used a lot in Fixture building), I've made the "setIsLink" function return a value rather than do any property manipulation directly.  What are peoples thoughts on this:

```php
public function __construct(/* ... */)
{
    // ...
    $this->isLink = $this->buildIsLink($greaterThan);
}

private function buildIsLink(Chronos $greaterThan): bool
{
    // if the date is more than $greaterThan from now, then don't allow a link (page will still exist)
    if ($this->offset === 0 || $this->datetime->gte($greaterThan)) {
        return false;
    }

    $network = $this->service->getNetwork();
    if (!$network) {
        return false;
    }

    return $this->getBroadcastPeriod($network->getMedium())->serviceIsActiveInThisPeriod($this->service);
}
```

IMO I find this a bit nicer as we keep all the property manipulation in the constructor, the logic ends up being a bit shorter as the early returns give values rather than just exiting and the typehint on the `buildIsLink` function is a bit more useful than `void`

tests/Ds2013/PresenterFactoryTest.php
Tasiobg: If you don't set anything it would be set to now, for instance
```
Chronos::setTestNow();
print_r((new Chronos())->format('Y-m-d H:i:s'));
```
This will print the exact time when setTestNow() was called. Does it make sense? not sure if I'm also understanding what you mean.

This could be more clear `Chronos::setTestNow(Chronos::now());`

src/Ds2013/Molecule/DateList/AbstractDateListItemPresenter.php
tatitati: yeah, "greaterThanDate"?

src/Ds2013/Molecule/DateList/AbstractDateListItemPresenter.php
```diff
@@ -50,4 +57,25 @@ public function getTemplateVariableName(): string
     {
         return 'date_list_item';
     }
+
+    abstract public function getLink(): string;
+
+    abstract protected function getBroadcastPeriod(string $medium): BroadcastPeriod;
+
+    private function setIsLink(Chronos $greaterThan) : void
```
tatitati: ```$greaterThan)<NO SPACE>: void```
I will figure out how to solve these problems with phpcs....surely is a custom rule (need to be coded, seems that is not included with phpcs)

tatitati:
seems ok to me

src/Ds2013/Molecule/DateList/AbstractDateListItemPresenter.php
```diff
@@ -50,4 +57,23 @@ public function getTemplateVariableName(): string
     {
         return 'date_list_item';
     }
+
+    abstract public function getLink(): string;
```
votemike: Why are the abstract methods half way down the file? Do we have a standard for where these go? I prefer all abstract methods immediately after the constructor.

Tasiobg: I do too but phpcs complains if 
```
abstract protected function getBroadcastPeriod(string $medium): BroadcastPeriod;
``` 
is above public methods.

Any suggestion?

votemike: Update phpcs

src/Ds2013/Molecule/DateList/AbstractDateListItemPresenter.php
BPScott: greaterThanDate suffers from the same problem, it doesn't describe what the date represents. In this case "we will not link to pages after this date"

src/Ds2013/Molecule/DateList/AbstractDateListItemPresenter.php
BPScott: The PHPCS rule is public methods, then protected, then private.

programmes-developers:

src/Ds2013/Molecule/DateList/AbstractDateListItemPresenter.php
votemike: Could we have the link to the ticket to look in to updating PHPCS?

PR #150, opened 2017-08-23
Allowing partial templates to be rendered

votemike:
I thought this would be best as a separate PR.  
Now we can add `?no_chrome` to the end of a URL to only render the body of DS2013 template instead of all of it.

Replicates http://www.bbc.co.uk/worldserviceradio/programmes/schedules.inc?view=schedule-offset&utcoffset=-07%3A00

Part of https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5672


programmes-developers:

Tasiobg:
looks good

src/Controller/BaseController.php
```diff
@@ -115,22 +114,31 @@ protected function renderWithChrome(string $view, array $parameters = [])
...
-        ]);
-
         $parameters = array_merge([
-            'orb' => $orb,
             'branding' => $branding,
-            'meta_context' => new MetaContext($this->context, $this->getCanonicalUrl()),
+            'with_chrome' => false,
         ], $parameters);
 
+        if (!$this->request()->query->has('no_chrome')) {
```
votemike: Do we want all this nested within an `if` statement? Or shall I reverse the check and return early?

Tasiobg: I can't think about an elegant way to solve this but I suggest to add a coment above the `if` to help understanding that this bit is for rendering partial templates

app/Resources/views/schedules/by_base.html.twig
```diff
@@ -1,4 +1,4 @@
-{% extends 'base_ds2013.html.twig' %}
+{% extends with_chrome is defined and with_chrome ? 'base_ds2013.html.twig' : 'base_ds2013_no_chrome.html.twig' %}
```
BPScott: This seems to be always defined now so the check can be removed

src/Controller/BaseController.php
```diff
@@ -115,22 +114,31 @@ protected function renderWithChrome(string $view, array $parameters = [])
...
-            'skipLinkTarget' => 'programmes-content',
-            'analyticsCounterName' => (string) new AnalyticsCounterName($this->context, $this->request()->getPathInfo()),
-            'analyticsLabels' => (new AnalyticsLabels($this->context, $this->istatsProgsPageType, $appVersion, $this->istatsExtraLabels))->orbLabels(),
-        ]);
-
         $parameters = array_merge([
-            'orb' => $orb,
             'branding' => $branding,
-            'meta_context' => new MetaContext($this->context, $this->getCanonicalUrl()),
+            'with_chrome' => false,
```
BPScott: Set this based on the query param here to save having to set it again later?


What do you reckon to `has_chrome` as a name? 'has' is the verb that tends to be used elsewhere for these sorts of checks? `hasResponse` `hasSchemaType` `hasImage` etc

src/Controller/BaseController.php
tatitati: well, is not a big deal this if. Total this is just part of another PR's. If necessary we can change them in the next PR's

src/Controller/BaseController.php
votemike: I prefer "with". As the chrome has not happened yet; you are asking the template to render serve up with the Chrome. Rather than the template already being rendered and seeing if it "has" the Chrome. Whereas with `hasSchemaType`, the schema type has already been decided upon.
@felipeparaujo @Tasiobg @tatitati any views?

PR #151, opened 2017-08-24
PROGRAMMES-5738 Making programme info image non-lazyload

votemike:
Fix for images as part of PROGRAMMES-5738


programmes-developers:

PR #152, opened 2017-08-24
PROGRAMMES-5672 Restoring period headings

votemike:
Period headings still seem to be on international pages on live. I'm not sure why we were getting rid of them.

http://www.bbc.co.uk/worldserviceradio/programmes/schedules#morning


programmes-developers:

BPScott:
Current logic on LIVE [looks like TV international pages don't display the period headings](
https://github.com/bbc/programmes/blob/3dde1c1e0b98b26b14d0e6e7072c3f6134aa9f4b/webapp/php/sharedmodules/programmes/views/schedules/_day.spv#L42). See http://www.bbc.co.uk/worldnews/programmes/schedules/africa.

That sounds like an odd distinction though. Let's confirm with Cathal that we can show the day period titles for everyone.

votemike:
Cathal says everyone should now be allowed to see the period headings.

BPScott:
Cool, delete away!

PR #154, opened 2017-08-24
Programmes 5672 international timezones

votemike:
PR for https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5672

If you'd like to try this from a different timezone open Chrome with the command `TZ='US/Pacific' open -na "Google Chrome" --args "--user-data-dir=$HOME/chrome-profile"`

FYI world service online is p00fzl9p


programmes-developers:

app/Resources/js/rv-bootstrap.js
```diff
@@ -1,5 +1,11 @@
-define(['jquery', 'respimg', 'lazysizes'],function($){
+define(['jquery', 'lazyload', 'respimg', 'lazysizes'],function($, Lazyload){
```
tatitati: lazyload and lazysizes is starting to be confuse. I think we should change the name of the module. RIght now I think that we have two lazy load modules doing the same thing.

app/Resources/js/timezone-notification.js
```diff
@@ -0,0 +1,197 @@
...
+            cssSelectors: {
+                startDate: '[data-timezone]',
+                offsetReload: '[data-utcoffset-replace]',
+                offsetRewrite: '[data-href-add-utcoffset]',
+                timezone: 'timezone',
+                timeNote: '.timezone-gmt'
+            },
+            showGmtOffset: false
+        },
+        setOptions: function (options) {
```
tatitati: we can make more clear the interface of this module. RIght now is a dump of public methods, but we can do some of the private, and understand better its entry points. 

To create private methods inside a module just change (for instance for setOptions()) this:
```javascript
         setOptions: function(......
```
for this:
```javascript
    var _setOptions = function(.....
```
The var only exist in the scope in which the function is defined, no in which is executed (lexical scope). So in this case you can be sure that it cannot be executed from outside. The underscore is just to make more visual private methods in javascript.

I would do this with the methods that you think should be private

app/Resources/js/rv-bootstrap.js
votemike: I'm not entirely clear on what each of them do or what better names would be. My code actually just does an ajax request... there's nothing lazy about it.  
@BPScott any thoughts?

app/Resources/js/timezone-notification.js
votemike: This sounds like something we should agree on for all Javascript files. We should come up with a good standard and go back through all the current JS and apply the standard to those too.  
Do we want to agree on the standards before this PR is merged? Or shall we have a meeting at some point in the next few days?

app/Resources/js/lazyload.js
```diff
@@ -0,0 +1,138 @@
...
+            for (var i = 0; (content = contents[i]) != null; i++) {
+                if (this.isOverThreshold(content) || this.alwaysLoadContent(content)) {
+                    this.loadRequireHints(content);
+                    if (!this.isDelayed(content)) {
+                        this.loadAjaxContent(content);
+                    }
+                }
+            }
+        },
+        addOnScrollHandler: function (context) {
```
tatitati: addOnScrollHandler (), 
getLazyContents()
This is a lazy load images......my guess is that your module just need to use jquery.appear
Some some functions can be just deleted or tweaked to avoid the duplication of functionality

app/Resources/js/timezone-notification.js
tatitati: True. We don't have any standard about this. For now I feel that our modules are a little "second class citizen", so we should enforce some standards on modules. 
But I would try to put this module properly, maybe we can use it to compare later the result in the meeting. Surely is not the best module (there are actually three ways of creating AMD modules), but can be a good starting point

app/Resources/js/rv-bootstrap.js
tatitati: lazysizes is a lazy loader for images (third party completely) that also manage the sizes for them, making them responsive and using srcset. It makes quite a few things in order to improve the performance of a page with images.

app/Resources/js/lazyload.js
```diff
@@ -0,0 +1,138 @@
...
+                dataType: 'html',
+                success: function (data) {
+                    var html = data;
+
+                    if (addOuterDiv) {
+                        html = '<div>' + data + '</div>';
+                    }
+                    var loaded_content = $($.parseHTML(html, document, true)).addClass(_this.options.lazy_css_state.loading);
+                    content.replaceWith(loaded_content);
+                    setTimeout(function () {
```
votemike: @BPScott any idea why this needs a timeout? Can we bin it?

app/Resources/js/timezone-notification.js
BPScott: Agree that the JS isn't great, it's a lift and shift from the old world. @tatitati to do want to raise a ticket to spend some time investigating how best to tidy our JS, perhaps looking into using webpack to build and writing ES6 style modules perhaps?

app/Resources/js/rv-bootstrap.js
BPScott: lazylad defers loading of content in the page - instead of loading content immediately, it is instead [transcluded](https://en.wikipedia.org/wiki/Transclusion) through an ajax request. Read http://www.bbc.co.uk/blogs/internet/entries/5c236ed9-5895-366b-8f26-f46961274b04 for an oooold blog post that talks about where what the lazyload is commonly used for under the "Content Driven" heading.

I can't think of a better name thank lazyload.

app/Resources/js/timezone-notification.js
votemike: And maybe enforce good code with Typescript 😄

BPScott: +1 to investigating typescript

app/Resources/views/schedules/by_base.html.twig
```diff
@@ -5,25 +5,31 @@
 {% block page_classes %}programmes-page text-base programmes-page--flush{% endblock %}
 
 {% block body %}
-    <div class="b-g-p no-margin-vertical islet--vertical br-box-highlight br-highlight-bg-onbg080" id="schedule-header">
-        {%- block header_links -%}{%- endblock -%}
-        <h1>
-            <span class="visually-hidden">
-                <span class="outlet">
-                    {{ tr('schedule_title', {'%1': service.getName()}) }}
+    {% if schedule_reload is defined and schedule_reload %}
```
BPScott: It looks like this is always defined

src/Ds2013/Molecule/_lazyload.scss
```diff
@@ -0,0 +1,74 @@
...
+        position: absolute;
+        top: 0;
+        bottom: 0;
+        left: 0;
+        right: 0;
+        pointer-events: none;
+    }
+}
+
+// @TODO make this work on non-white backgrounds and support positioning
```
Tasiobg: I think it is intentionally but in case it isn't, there is a TODO here

votemike: Yeah, this existed in the PAL code. I just copied it over.

app/Resources/views/schedules/by_base.html.twig
votemike: I don't see the problem with a `defined` check. What do you think?

BPScott: ok, fine by me :)

PR #155, opened 2017-08-25
Cache 404s for 60 seconds

BPScott:
This avoids repeated requests to non-existant content always hitting the
PAL.

Also enure the ResponseSubscriber only runs on the master request to
avoid duplicating Vary headers when serving exceptions.

Fixes PROGRAMMES-5834

### DIFF
--- a/src/Controller/ExceptionController.php
+++ b/src/Controller/ExceptionController.php
@@ -52,6 +52,21 @@ class ExceptionController extends BaseExceptionController
             ]);
         }
 
+        $headers = [
+            'Content-Type' => $request->getMimeType($request->getRequestFormat()) ?: 'text/html',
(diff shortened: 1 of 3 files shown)


programmes-developers:

Tasiobg:
Looks good

src/Controller/ExceptionController.php
```diff
@@ -52,6 +52,21 @@ public function __invoke(Request $request, FlattenException $exception, DebugLog
             ]);
         }
 
+        $headers = [
+            'Content-Type' => $request->getMimeType($request->getRequestFormat()) ?: 'text/html',
```
BPScott: This comes from Symfony's ExceptionController: https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/TwigBundle/Controller/ExceptionController.php#L75

PR #156, opened 2017-08-29
PROGRAMMES-5761 Moving sibling services to a presenter

votemike:
The sibling service presenter is also needed in the week view. So refactoring it to be reuseable.


programmes-developers:

BjornTwachtmann:
looks fine to me

PR #158, opened 2017-08-30
Programmes 5761 by week page

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5761

Go to a URL such as http://sandbox.bbc.co.uk/schedules/p00fzl7j/2017-w35 to test


programmes-developers:

app/Resources/views/schedules/by_week.html.twig
```diff
@@ -0,0 +1,71 @@
...
+        </ul>
+    {%- endif -%}
+{%- endblock -%}
+
+{%- block schedule -%}
+    <div class="g-f-l br-box-page">
+        {{ ds2013_presenter(page_presenter) }}
+    </div>
+    <script>
+        require(['jquery-1.9'], function($) {
```
votemike: This seems grim for many ways. Is there anything else we can do instead?

src/Controller/SchedulesByWeekController.php
```diff
@@ -0,0 +1,85 @@
...
+use App\ValueObject\BroadcastWeek;
+use BBC\ProgrammesPagesService\Domain\Entity\Service;
+use BBC\ProgrammesPagesService\Service\BroadcastsService;
+use BBC\ProgrammesPagesService\Service\ServicesService;
+
+class SchedulesByWeekController extends BaseController
+{
+    public function __invoke(Service $service, string $date, ServicesService $servicesService, BroadcastsService $broadcastService)
+    {
+//        $this->setIstatsProgsPageType('schedules_month'); @TODO do I need this?
```
votemike: @BPScott Do I need this line?

src/Controller/SchedulesByWeekController.php
```diff
@@ -0,0 +1,85 @@
...
+        $this->setContext($service);
+
+        $broadcastWeek = new BroadcastWeek($date);
+
+        $servicesInNetwork = $servicesService->findAllInNetworkActiveOn(
+            $service->getNetwork()->getNid(),
+            $broadcastWeek->start() //@TODO change to be sensible active on date
```
votemike: @BPScott Do you think that checking for services that are active at the start of the week is good enough?

src/Controller/SchedulesByWeekController.php
```diff
@@ -0,0 +1,85 @@
...
+
+        $broadcasts = [];
+
+        if ($broadcastWeek->serviceIsActiveInThisPeriod($service)) {
+            // Get broadcasts in relevant period
+            $broadcasts = $broadcastService->findByServiceAndDateRange(
+                $service->getSid(),
+                $broadcastWeek->start(),
+                $broadcastWeek->end(),
+                1000
```
votemike: Will 1000 be enough?

src/Ds2013/Page/Schedules/ByWeekPage/TimeSlotItemPresenter.php
```diff
@@ -0,0 +1,74 @@
...
+     * @return Broadcast[]|null
+     */
+    public function getGroupedBroadcasts(): ?array
+    {
+        return $this->groupedBroadcasts[$this->dateTime->format('Y-m-d')][$this->hour] ?? null;
+    }
+
+    public function getBroadcastItem(Broadcast $broadcast)
+    {
+        //@TODO should this just come from the twig template/factory?
```
votemike: @BPScott Should this come from the presenter of through the twig template and the template factory? When do we choose which option?

tests/DataFixtures/PagesService/OptionsFixture.php
```diff
@@ -81,7 +81,7 @@ public static function radioFour()
                 'navigation_links' => [
                     [
                         'title' => 'Schedule',
-                        'url' => '/radio4extra/programmes/schedules/this_week/grid/ataglance',
+                        'url' => '/radio4extra/programmes/schedules/this_week/grid/ataglance', //@TODO what is this?
```
votemike: What is this? Can we get rid of it? Can I get rid of it? Or should a tester?

src/Controller/SchedulesByWeekController.php
BjornTwachtmann: this should be schedules_week to keep analytics output in line with V2

BjornTwachtmann: and yeah, it's needed

src/Controller/SchedulesByWeekController.php
BjornTwachtmann: I dunno if the start or end of the week is better here. Probably the end, we're more likely to get complaints about not showing a schedule that's starting than ending one too early. It's not worth the hastle of doing a date range search here I think.

src/Controller/SchedulesByWeekController.php
BjornTwachtmann: Can we run this query without a limit? I've found it's ill advised to make assumptions about the kinds of things editorial might do with the data.

app/Resources/views/schedules/by_week.html.twig
BjornTwachtmann: What the hell does this even do?

src/Controller/SchedulesByWeekController.php
votemike: This is just for the "other" services. So e.g. London, Scotland, Midlands services.... etc...

app/Resources/views/schedules/by_week.html.twig
votemike: If you look at the page on a small screen it does some toggling.

src/Controller/SchedulesByWeekController.php
```diff
@@ -0,0 +1,85 @@
...
+    {
+        return (!$service->getStartDate() || $service->getStartDate() <= $broadcastWeek->end()) && (!$service->getEndDate() || $broadcastWeek->start() < $service->getEndDate());
+    }
+
+    /**
+     * @param Service $mainService
+     * @param Service[] $servicesInNetwork
+     * @return Service|null
+     */
+    private function twinService(Service $mainService, array $servicesInNetwork): ?Service
```
BjornTwachtmann: In the by_day page this is in the page presenter. Worth moving this into a presenter? Or is it needed here?

votemike: That's because in the by_day view, the logic is in the main body of the page (which the presenter presents) but in the by_week page, the logic is needed in the header of the page.

src/Controller/SchedulesByWeekController.php
BPScott: $broadcastService::NO_LIMIT (which is a const that evaluates to null)

tests/DataFixtures/PagesService/OptionsFixture.php
BPScott: kill the ataglance concept, it's doesn't seem to be any different to the regular week view

votemike: Can I get rid of it? Or do the testers have to remove it?

src/ValueObject/BroadcastWeek.php
```diff
@@ -0,0 +1,27 @@
...
+ * @see BroadcastDay
+ */
+class BroadcastWeek extends BroadcastPeriod
+{
+    public function __construct(string $date)
+    {
+        $dateParts = explode('-', $date);
+        $year = (int) $dateParts[0];
+        $week = (int) str_replace('w', '', $dateParts[1]);
+        $blankChronos = Chronos::now(ApplicationTime::getLocalTimeZone())->startOfDay(); // Today at midnight
```
BjornTwachtmann: Can we use ApplicationTime::getTime() instead of Chronos::now() please, it returns a Chronos object and it makes sure the same thing is used everywhere

src/Controller/SchedulesByWeekController.php
BjornTwachtmann: fair enough

src/Controller/SchedulesByWeekController.php
BjornTwachtmann: Yeah, it's unlikely to be much of an issue either way

app/Resources/views/schedules/by_week.html.twig
BjornTwachtmann: Ugh. You could refactor to put the small screen content in a div that's hidden by media queries or something I suppose. There's probably a reason this exists though.

BjornTwachtmann:
One small bug: 
http://sandbox.bbc.co.uk/schedules/p00fzl7j - the "This week" link has a utcoffset paramter, which should only happen on international pages.

votemike:
@BjornTwachtmann I didn't put the UTC offset there. It looks like Javascript it doing it. Is it a bug in https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5672? Or should `data-href-add-utcoffset` be set to `$programme->isInternational()`?

BjornTwachtmann:
Gut instinct would be it's a bug in https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5672 . That JS should only be added on an international page.

app/Resources/views/schedules/by_day.html.twig
```diff
@@ -1,6 +1,6 @@
 {% extends 'schedules/by_base.html.twig' %}
 
-{% block date %}{{ broadcast_day_start|local_date_intl('eeee d MMMM Y') }}{% endblock %}
+{% block date %}{{ broadcast_day_start|local_date_intl('eeee d MMMM y') }}{% endblock %}
```
BPScott: This looks like it changes the page title to have a two diget year, that doesn't seem right?

src/Controller/SchedulesByWeekController.php
BPScott: Not got a strong opinion either way. Agree with @BjornTwachtmann's point about complains at the start are more likely than complaints at the end.

src/Ds2013/Core/Modules/Schedules/_week-guide.scss
```diff
@@ -0,0 +1,137 @@
...
+    .weeknav--tvschedule {
+        margin-top: -40px;
+    }
+
+    .week-guide__table {
+        min-width: 767px;
+    }
+}
+
+@if (not $ie) {
```
BPScott: Can we give this to IE people now?

app/Resources/views/schedules/by_day.html.twig
votemike: It doesn't show two digits for me..... Can you show me?

src/Ds2013/Core/Modules/Schedules/_week-guide.scss
votemike: What do you mean?  
Are you asking me to delete the `@if (not $ie) {` part?

src/Ds2013/Page/Schedules/ByWeekPage/TimeSlotItemPresenter.php
```diff
@@ -0,0 +1,74 @@
...
+     * @return Broadcast[]|null
+     */
+    public function getGroupedBroadcasts(): ?array
+    {
+        return $this->groupedBroadcasts[$this->dateTime->format('Y-m-d')][$this->hour] ?? null;
+    }
+
+    public function getBroadcastItem(Broadcast $broadcast): BroadcastPresenter
+    {
+        //@TODO should this just come from the twig template/factory?
```
votemike: @BPScott when do we create presenters in another presenter? And when do we create them from inside a template?

BPScott: Haven't worked out a hard and fast rule for that yet. I think we differ atm because if the options to pass into the child presenter become complex it may be easier to construct it in the php rather than the template.

votemike: OK. Will leave this as is at the moment then.

app/Resources/views/schedules/by_day.html.twig
Tasiobg: I think it works fine but is a bit inconsistent becase PHP date functions set a 2 digits year when you use the lower case 'y'. This is using ICU date format which is different

src/Ds2013/Core/Modules/Schedules/_week-guide.scss
BPScott: yeah, is the reason why we're excluding ie still valid? (i.e. can we stop wrapping this in the if clause)

votemike: Other places are using `@include mq` without being wrapped in it, so I guess we can bin it.

app/Resources/views/schedules/by_day.html.twig
BPScott: Ah, I see. I didn't make that PHP vs ICU connection. Thanks.

(for future Ben in ICU [Y is the "week counting" year](http://www.icu-project.org/apiref/icu4c/classSimpleDateFormat.html#details). We should check to see if we're accidentally using Y when we should be using y in other views - I think i see it in by_month.html.twig

votemike: I believe I checked when I found this out and changed everywhere it needed to be changed. Do you have a link to the line you think you found?

BPScott: https://github.com/bbc/programmes-frontend/blob/934053846be684beb42d17b270188183d3297a7b/app/Resources/views/schedules/by_month.html.twig#L3

votemike: O yeah, Not sure how I missed that.... Will change it.

PR #159, opened 2017-08-31
Rmp comscore

BjornTwachtmann:
The infinite joy of comscore.


programmes-developers:

tests/DataFixtures/PagesService/OptionsFixture.php
```diff
@@ -127,6 +127,95 @@ public static function radioFour()
...
+                'comments_episodes_enabled' => false,
+                'comments_episodes_id' => '',
+                'playlister_popularity_enabled' => 'no',
+                'bbc_site' => '',
+                'recipes_enabled' => '',
+                'live_stream_id' => '',
+                'live_stream_heading' => '',
+                'control-17' => '',
+                'projectId' => 'progs-worldservice',
+                'navigation_links' => [
```
BPScott: I think we could blank out the navigation_links and promoted_programmes options to make the fixture a bit shorter without affecting anything but that's not a big deal

gulpfile.js
```diff
@@ -37,9 +38,13 @@ gulp.task('js', ['js:clean'], function () {
...
             "respimg": "../../../node_modules/lazysizes/plugins/respimg/ls.respimg",
             "lazysizes": "../../../node_modules/lazysizes/lazysizes-umd",
-            "eqjs": "../../../node_modules/eq.js/build/eq"
+            "eqjs": "../../../node_modules/eq.js/build/eq",
+            "comscorews" : "../../../vendor/bbc-rmp/comscore/js-modules/comscorews",
+            "rmpcomscore/base" : "../../../vendor/bbc-rmp/comscore/js-modules/base",
+            "orb/cookies": "empty:"
         },
         // @TODO do not optimise on sandbox
+        //"optimize": "none",
```
votemike: What's this?

BjornTwachtmann: It's how you tell it not to minify your JS.

src/ValueObject/ComscoreAnalyticsLabels.php
```diff
@@ -0,0 +1,52 @@
...
+
+namespace App\ValueObject;
+
+use BBC\ProgrammesPagesService\Domain\Entity\CoreEntity;
+use BBC\ProgrammesPagesService\Domain\Entity\Service;
+use RMP\Comscore\Comscore;
+
+class ComscoreAnalyticsLabels
+{
+    private $comscore;
```
votemike: What's a Comscore?

src/ValueObject/ComscoreAnalyticsLabels.php
```diff
@@ -0,0 +1,52 @@
...
+
+    public function getComscore(): ?Comscore
+    {
+        return $this->comscore;
+    }
+
+    private function setComscore(CosmosInfo $cosmosInfo, IstatsAnalyticsLabels $istatsAnalyticsLabels, string $canonicalUrl)
+    {
+        $labels = $istatsAnalyticsLabels->getLabels();
+        $data = array(
```
votemike: Does the PHPCS allow longform array notation?

BjornTwachtmann: Apparently it does. I should probably read code more thoroughly before I paste it though :)

src/ValueObject/ComscoreAnalyticsLabels.php
BjornTwachtmann: The BBC Worldwide analytics system

gulpfile.js
votemike: I mean, why is it here if it is commented out?

src/ValueObject/ComscoreAnalyticsLabels.php
BPScott: We ain't in php5.3 world anymore :o

gulpfile.js
BjornTwachtmann: So that I remember this is how to disable minification and eventually set up something to not do this by default on the sandbox.

votemike: Move the config to a separate file. Then load the config based on env?

BjornTwachtmann: If that would work, please do so. I have spent enough time swearing at Gulp/Require for this week but it would be useful.

votemike: Created a ticket https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5845

PR #162, opened 2017-09-01
PROGRAMMES-5847 Making sure localtime is only run on international pages

votemike:
Should fix https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5847


programmes-developers:

PR #163, opened 2017-09-04
PROGRAMMES-5828 Status check page

Tasiobg:
PROGRAMMES-5828

Same we have in clifton StatusController.php but adapted to programmes-frontend

At the moment unitest is using a deprecated function, this will be improved as a part of a change Bjorn is working on


programmes-developers:

PR #165, opened 2017-09-05
PROGRAMMES-5778 Getting PHPStan up to level 4

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5778

Build is currently failing due to https://github.com/phpstan/phpstan/issues/484 Shall we wait for the issue to be resolved or shall we downgrade PHPStan and remove the memory limit.


programmes-developers:

votemike:
Test Me Please

votemike:
Test This Please

src/Ds2013/Page/Schedules/ByDayPage/SchedulesByDayPagePresenter.php
```diff
@@ -33,7 +33,7 @@ class SchedulesByDayPagePresenter extends Presenter
     /** @var Service */
     private $twinService;
 
-    /** @var Broadcast */
+    /** @var Broadcast|false|null */
```
tatitati: Is really necessary that Broadcast type?

tests/ArgumentResolver/ContextEntityByPidValueResolverTest.php
```diff
@@ -22,17 +23,18 @@ class ContextEntityByPidValueResolverTest extends TestCase
     /** @var ArgumentResolver */
     private $resolver;
 
-    /** @var CoreEntitiesService */
+    /** @var CoreEntitiesService|PHPUnit_Framework_MockObject_MockObject */
```
tatitati: I added a new rule in one of my next PR's to avoid this. We should abstract this kind of stuff. The rule for PHPStan is this:
``` - '#Property Tests.*does not accept PHPUnit_Framework_MockObject_MockObject\.#' ```

File: tests/ValueObject/IstatsAnalyticsLabelsTest.php [hunk 3]
```diff
@@ -17,7 +17,7 @@ class IstatsAnalyticsLabelsTest extends TestCase
 {
     public function testService()
     {
-        $context = $this->ServiceFactory('bbc_one', 'tv');
+        $context = $this->serviceFactory('bbc_one', 'tv');
```
tatitati: :ok_hand:

src/Ds2013/Page/Schedules/ByDayPage/SchedulesByDayPagePresenter.php
votemike: As far as I can tell

tests/ArgumentResolver/ContextEntityByPidValueResolverTest.php
votemike: I quite like this change, because then my IDE works correctly. I'm not sure the ignore rule is a great idea.

tatitati: but in this case we will have that type param everywhere in the tests....

@BPScott ?

src/Ds2013/Page/Schedules/ByDayPage/SchedulesByDayPagePresenter.php
BPScott: Taking a look at https://github.com/bbc/programmes-frontend/blob/4647889a1ee71093b28b24aef3071f65ce4f0b6e/src/Ds2013/Page/Schedules/ByDayPage/SchedulesByDayPagePresenter.php#L159

Broadcast is a lazy field.
false means "we've not yet computed a value for this"
null means "we computed a value for it and it is empty"
Broadcast means "we computed a value for it and it is this broadcast"

We could probably get rid of the false hint by computing this in the constructor and then the getter can return that computed value.

tatitati: Thanks ben, very useful

tests/ArgumentResolver/ContextEntityByPidValueResolverTest.php
votemike: Yes. But the property **is** of that type.

tatitati: Yeah, but for example, you don't need to know any property or method from that PHPUnit_Framework_MockObject_MockObject object even if your IDE know work. In tests you will work with CoreEntitiesService methods and attributes. So the IDE is not gonna give you any additional data you need for testing. Just my thought

src/Ds2013/Page/Schedules/ByDayPage/SchedulesByDayPagePresenter.php
votemike: Have merged this PR so that Felipe can use PHPStan on his branch. If you think this is worth changing, feel free to raise a ticket.

tests/ArgumentResolver/ContextEntityByPidValueResolverTest.php
votemike: Have merged this PR so that Felipe can use PHPStan on his branch. If you think this is worth changing, feel free to raise a ticket.

tatitati: yeah, no worries. We have different preferences on this, sorry!, jeje, let's see what the rest of the people say. I will be happy to adapt to any decission.

BPScott: Having as few ignored rules as possible sounds like the best plan - PHPStan is advising us and we should strive to listen to it. Helping in IDE autocompletion is a handy win (for everyone apart from me as the token SublimeText luddite here :p).

I'm with Mike with this one. It's appears in quite a few places, but it's an easy thing to fix.

votemike: But then the IDE suggests PHPUnit method too. Such as `expects()`

tatitati: ok then. Thanks people!

PR #167, opened 2017-09-05
PROGRAMMES-5855 Fixing no schedule message on by_week page

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5855

Fixed bug, but also refactored. Please check the refactoring is correct.


programmes-developers:

BPScott:
I am rapidly coming to the conclusion that The Atom/Molecule/Organism/Page layout structure, while good for pure CSS bits, is a bad fit for general purpose components. Sticking everything into Molecule/Organism is going to get pretty ugly pretty quickly and it's going to become hard to understand what components are related to each other.

For these small components, that aren't reused across multiple page types I'm leaning towards sticking them in the Page/Schedule folder to make it clear that they are only used on those pages.

While I'm at it I wonder if that Page folder is useful at all and instead if we should get rid of it to have Ds*/Schedule anyway

votemike:
@BPScott want to create a meeting or ticket to discuss changing the project layout? Or are these changes you want done as part of this PR?

BPScott:
The thing that prompted that rant was thinking about if we should move the NoSchedule presenter into the Page/Schedules folder rather than Molecule in this PR. How does doing that in this ticket sound, then we can work out what we're doing on a broader scale at a later point.

votemike:
Test this please

votemike:
Test this please

PR #168, opened 2017-09-05
PROGRAMMES-5729 - Amen Programme Object

felipeparaujo:
**TESTS ARE COMING IN A LATER PR.**

I want to get this out of the door quickly to stop blocking everyone


programmes-developers:

src/DsAmen/Organism/Programme/ProgrammePresenter.php
```diff
@@ -3,22 +3,165 @@
...
+            $this->programme,
+            $this->subPresenterOptions('body_options')
+        );
+    }
+
+    public function getProgrammeCtaPresenter(): ?ProgrammeCtaPresenter
     {
-        return $this->programme;
+        if ($this->programme instanceof ProgrammeItem && $this->programme->isStreamable()) {
+            $this->ctaPresenter = new ProgrammeCtaPresenter(
```
votemike: Why is this being set as a class property?

src/DsAmen/Organism/Programme/SubPresenters/ProgrammeBodyPresenter.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        parent::__construct($options);
+        $this->programme = $programme;
+
+        if ($this->programme instanceof ProgrammeItem && $this->programme->getReleaseDate()) {
+            $this->releaseDate = new Date((string) $this->programme->getReleaseDate());
+        }
+    }
+
+    public function getReleaseDate(): Date
```
votemike: This might return null

src/DsAmen/Organism/Programme/ProgrammePresenter.php
felipeparaujo: Because it's used in line 119 and line 136. I don't want to instantiate it twice, as this is the same thing for both.

votemike:
I have a feeling the PHPStan will complain about a few bits in this. Are we happy if I merge https://github.com/bbc/programmes-frontend/pull/165 before this?

felipeparaujo:
@votemike yeah, merge it. As soon as you do, i'll rebase and run level 4 on this

src/DsAmen/Organism/Programme/ProgrammePresenter.php
votemike: What is stopping it from being instantiated twice? It never looks like there is any check to see if it is already set.

src/DsAmen/Organism/Programme/SubPresenters/programme_cta.html.twig
```diff
@@ -0,0 +1,14 @@
+<div class="icon-link {{ programme_cta.getOption('cta_class') }}">
+    <a class="link-complex" href="{{ programme_cta.getPlayerUrl() }}"
+       data-linktrack="{{ programme_cta.getLinkLocationPrefix() }}calltoaction">
+        <div class="icon icon-cta icon--{{ programme_cta.getMediaIconName() }}">
+            {{ gelicon('audio-visual', programme_cta.getMediaIconName(), 'programme') }}
```
BjornTwachtmann: where is the gelicon--programme class defined? Am I being an idiot?

felipeparaujo: That went into master in a previous pr. it's defined here: https://github.com/bbc/programmes-frontend/blob/amen-programme/src/DsAmen/Atom/_gelicon.scss#L26-L30

BjornTwachtmann:
Structurally this looks very sound. I imagine there will be stuff to add as we work our way through the brand page. Will need a bunch of tests at some point as well, but happy for this to be merged for now.

src/DsAmen/Organism/Programme/ProgrammePresenter.php
felipeparaujo: Damn, you're right. Fixing it.

src/DsAmen/Organism/Programme/SubPresenters/programme_cta.html.twig
BjornTwachtmann: ah, so yes then :)

felipeparaujo:
test this please

felipeparaujo:
test this please

PR #172, opened 2017-09-06
Programmes 5805 - fixture db switching and setup

BjornTwachtmann:


src/EventSubscriber/FindByPidRouterSubscriber.php
```diff
@@ -39,8 +40,8 @@
  */
 class FindByPidRouterSubscriber implements EventSubscriberInterface
 {
-    /** @var ServiceFactory */
-    private $serviceFactory;
+    /** @var ContainerInterface */
+    private $container;
```
BPScott: Rather than injecting the whole container it might be worth looking into the ServiceSubscriberInterface concept - https://github.com/symfony/symfony/pull/21708


Have this class extend ServiceSubscriberInterface and implement a getSubscribedServices() which describes the services you want. This shall then give you container that contains only the services you care about, making it clear what this class depends upon rather than "hoo boy it could be anything from the app container"

programmes-developers:

src/Fixture/ScenarioManagement/ScenarioNameManager.php
```diff
@@ -0,0 +1,33 @@
...
+    private $cosmosInfo;
+
+    public function __construct(CosmosInfo $cosmosInfo)
+    {
+        $this->cosmosInfo = $cosmosInfo;
+    }
+
+    public function scenarioIsActive()
+    {
+        if ($this->isLowerEnvironment() && !empty($_GET['__scenario'])) {
```
votemike: Should we be referencing `$_GET` directly? Not through the request or something? This sort of thing always rings alarm bells for me.

src/Fixture/ScenarioManagement/ScenarioNameManager.php
```diff
@@ -0,0 +1,33 @@
...
+            return true;
+        }
+        return false;
+    }
+
+    private function isLowerEnvironment()
+    {
+        return in_array($this->cosmosInfo->getAppEnvironment(), ['sandbox', 'int', 'test']);
```
votemike: What are the other environments we have?

src/Fixture/ScenarioManagement/ScenarioNameManager.php
BjornTwachtmann: Ideally not. However at the point this gets called the request object sometimes hasn't been instantiated yet. This is the only way to ensure things work 100% of the time.

src/Fixture/ScenarioManagement/ScenarioNameManager.php
BjornTwachtmann: live is the only other one

votemike: Would `$this->cosmosInfo->getAppEnvironment() !== 'live'` be clearer?

BjornTwachtmann: I always prefer a whitelist, in case something ends up not being set or set to something weird.

votemike: Instead of checking what the environment is, could something be set in the config file?

BjornTwachtmann: The environment comes from the config file anyway. I think it's better to be safe here, rather than risking some minor DI change letting this run on the live environment.

src/EventSubscriber/FindByPidRouterSubscriber.php
BjornTwachtmann: Done

PR #173, opened 2017-09-06
PROGRAMMES-5768 - Refactor gelicon so it doesn't prefix classes with 'gelicon--'

felipeparaujo:
Prefixing every class with gelicon-- stops us from using classes like br-secondary-text-ontext, which is used for example in the clips list in the brand page.


programmes-developers:

src/Twig/GelIconExtension.php
```diff
@@ -45,7 +45,7 @@ public function gelIcon(string $set, string $icon, ...$extraClasses)
         $this->addIcon($set, $icon);
         $classes = 'gelicon';
         foreach ($extraClasses as $extraClass) {
-            $classes .= " gelicon--$extraClass";
```
BPScott: If this is now going to be a boring list of extra classnames could we pass it in as a string to save on having to do the building stuff, then you can wouldn't need the variadic function

PR #174, opened 2017-09-06
PROGRAMMES-5767 Add Promotion object & show on brand page

BPScott:
Add Amen Promotion object
Request promotions in TLEC controller
First pass at rendering promos on brand page

### DIFF
--- a/app/Resources/views/find_by_pid/tlec.html.twig
+++ b/app/Resources/views/find_by_pid/tlec.html.twig
@@ -30,8 +30,22 @@
             {# TODO Promos / Galleries list #}
             {%- if promotions or clips or galleries -%}
                 <ul class="grid">
+                    {%- for promotion in promotions -%}
+                        <li class="grid__item 1/2@gel3 1/3@gel3b 1/4@gel5">
+                            {# TODO Update image sizes for double width first promo #}
+                            {{- ds_amen('promotion', promotion, {
+                                media_variant: 'media--column media--card',
+                                cta_class: 'br-box-secondary',
+                                branding_name: 'secondary',
+                                link_location_prefix: 'promo_' ~ loop.index ~ '_',
+                                show_synopsis: false,
+                                img_sizes: { 600: 1/2, 768: 1/3, 1008: '315px', 1280: '300px' },
+                            }) -}}
+                        </li>
+                    {%- endfor -%}
+
                     {%- for clip in clips -%}
-                        <li class="1/4@gel5 1/3@gel3b 1/2@gel3 1/1 grid__item">
+                        <li class="grid__item 1/2@gel3 1/3@gel3b 1/4@gel5">
                             {{ ds_amen('programme', clip, {
                                 'context_programme': programme,
                                 'branding_name': 'secondary',
--- a/src/DsAmen/Organism/Promotion/PromotionPresenter.php
+++ b/src/DsAmen/Organism/Promotion/PromotionPresenter.php
@@ -1,0 +1,182 @@
+<?php
+declare(strict_types = 1);
+
+namespace App\DsAmen\Organism\Promotion;
+
+use App\DsAmen\Presenter;
+use App\Exception\InvalidOptionException;
+use BBC\ProgrammesPagesService\Domain\Entity\CoreEntity;
+use BBC\ProgrammesPagesService\Domain\Entity\Episode;
+use BBC\ProgrammesPagesService\Domain\Entity\Image;
+use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeItem;
+use BBC\ProgrammesPagesService\Domain\Entity\Promotion;
+use BBC\ProgrammesPagesService\Domain\Entity\RelatedLink;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+
+class PromotionPresenter extends Presenter
+{
+    /** @var UrlGeneratorInterface */
+    private $router;
+
+    /** @var Promotion */
+    private $promotion;
+
+    /** @var string[] */
+    private $actionIcon = [];
+
+    /** @var int */
+    private $duration = 0;
+
+    /** @var array */
+    protected $options = [
+        // display options
+        'show_image' => true,
+        'show_synopsis' => true,
+        'show_related_links' => true,
+
+        // classes & elements
+        'h_tag' => 'h4',
+        'title_size' => 'gel-pica-bold',
+        'img_sizes' => [],
+        'media_variant' => 'media--column media--card',
+        'cta_class' => 'icon-cta--dark',
+        'media_panel_class' => '1/1',
+        'branding_name' => 'subtle',
+        'link_location_prefix' => 'promotionobject_',
+    ];
+
+    public function __construct(
+        UrlGeneratorInterface $router,
+        Promotion $promotion,
+        array $options = []
+    ) {
+        parent::__construct($options);
+        $this->router = $router;
+        $this->promotion = $promotion;
+
+        // Build ActionIcon
+        $promotedEntity = $this->promotion->getPromotedEntity();
+        if ($promotedEntity instanceof ProgrammeItem && $promotedEntity->isStreamable()) {
+            $this->actionIcon = ['set' => 'audio-visual', 'icon' => 'play'];
+        } elseif ($this->isExternalLink($this->getUrl())) {
+            $this->actionIcon = ['set' => 'basics', 'icon' => 'external'];
+        }
+
+        // Build Duration - only set for Clips and not-TV Episodes
+        if ($promotedEntity instanceof ProgrammeItem &&
+            !($promotedEntity instanceof Episode && $promotedEntity->isTv())
+        ) {
+            $this->duration = $promotedEntity->getDuration();
+        }
+    }
+
+    public function getTitle(): string
+    {
+        return $this->promotion->getTitle();
+    }
+
+    public function getUrl(): string
+    {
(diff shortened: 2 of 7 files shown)


src/DsAmen/Organism/Promotion/PromotionPresenter.php
```diff
@@ -0,0 +1,187 @@
...
+    private $router;
+
+    /** @var Promotion */
+    private $promotion;
+
+    /** @var string[] */
+    private $actionIcon;
+
+    /** @var int */
+    private $duration;
```
tatitati: just wondering if we know the unit of this duration. ms?, secs?

tatitati: just wondering if we know the unit of this duration. ms?, secs?, minutes?

app/Resources/views/find_by_pid/tlec.html.twig
```diff
@@ -28,6 +28,23 @@
...
+            {%- if promotions or clips or galleries -%}
+                <ul class="grid">
+                {%- for promotion in promotions -%}
+                    <li class="grid__item 1/2@gel3 1/3@gel3b 1/4@gel5">
+                        {# TODO Update image sizes for double width first promo #}
+                        {{- ds_amen('promotion', promotion, {
+                            media_variant: 'media--column media--card',
+                            cta_class: 'br-box-secondary',
+                            branding_name: 'secondary',
+                            link_location_prefix: 'promo_' ~ loop.index ~ '_',
```
votemike: Where is loop coming from?

app/Resources/views/find_by_pid/tlec.html.twig
```diff
@@ -28,6 +28,23 @@
             {# TODO Recipes block #}
 
             {# TODO Promos / Clips / Galleries list #}
+            {%- if promotions or clips or galleries -%}
+                <ul class="grid">
+                {%- for promotion in promotions -%}
+                    <li class="grid__item 1/2@gel3 1/3@gel3b 1/4@gel5">
+                        {# TODO Update image sizes for double width first promo #}
+                        {{- ds_amen('promotion', promotion, {
```
votemike: This is a lot of options to have in a template. Would it be better to construct the presenter in the controller?

app/Resources/views/find_by_pid/tlec.html.twig
```diff
@@ -28,6 +28,23 @@
             {# TODO Recipes block #}
 
             {# TODO Promos / Clips / Galleries list #}
+            {%- if promotions or clips or galleries -%}
```
votemike: Why clips or galleries? I can't see them being used.

src/DsAmen/Organism/Promotion/PromotionPresenter.php
```diff
@@ -0,0 +1,187 @@
...
+
+class PromotionPresenter extends Presenter
+{
+    /** @var UrlGeneratorInterface */
+    private $router;
+
+    /** @var Promotion */
+    private $promotion;
+
+    /** @var string[] */
```
votemike: Does PHPStan complain about this being null to start with?

src/DsAmen/Organism/Promotion/PromotionPresenter.php
```diff
@@ -0,0 +1,187 @@
...
+
+        return $this->promotion->getRelatedLinks();
+    }
+
+    /**
+     * @return string[]
+     */
+    public function getActionIcon(): array
+    {
+        if ($this->actionIcon === null) {
```
votemike: Set in constructor instead?

src/DsAmen/Organism/Promotion/PromotionPresenter.php
```diff
@@ -0,0 +1,187 @@
...
+                $this->duration = 0;
+            }
+        }
+
+        return $this->duration;
+    }
+
+    public function getBrandingBoxClass(): string
+    {
+        if (!empty($this->options['branding_name'])) {
```
votemike: Personal preference. Reverse this and similar checks?

src/DsAmen/Organism/Promotion/PromotionPresenter.php
BPScott: Seconds. We don't do any transformation from the value we get back from PIPS.

tests/DsAmen/Organism/Promotion/PromotionPresenterTest.php
```diff
@@ -0,0 +1,145 @@
...
+class PromotionPresenterTest extends TestCase
+{
+    private $mockRouter;
+
+    public function setUp()
+    {
+        $this->mockRouter = $this->createMock(UrlGeneratorInterface::class);
+    }
+
+    public function testPromotion()
```
tatitati: can be more descriptive?, all this is about promotions. Maybe ```testPromotionReturnDefaultValues()``` ?

app/Resources/views/find_by_pid/tlec.html.twig
BPScott: https://twig.symfony.com/doc/2.x/tags/for.html#the-loop-variable

app/Resources/views/find_by_pid/tlec.html.twig
BPScott: To save people having to edit this line when we do add in the clips and galleries to this list. Possibly getting ahead of myself though :)

app/Resources/views/find_by_pid/tlec.html.twig
votemike: YLSNED

tests/DsAmen/Organism/Promotion/PromotionPresenterTest.php
```diff
@@ -0,0 +1,145 @@
...
+        $this->assertSame('short synopsis', $presenter->getSynopsis());
+        $this->assertSame($relatedLinks, $presenter->getRelatedLinks());
+        $this->assertSame(0, $presenter->getDuration());
+        $this->assertSame('br-box-subtle', $presenter->getBrandingBoxClass());
+        $this->assertSame('br-subtle-text-ontext', $presenter->getTextBrandingClass());
+    }
+
+    public function testPromotionOfProgrammeItem()
+    {
+        $this->mockRouter->method('generate')
```
tatitati: The problem mixing expectations with assertions is that one doesn't know when is really expected this function be called. Is it on presenter->getImage(), presenter->getDuration(), any other?

app/Resources/views/find_by_pid/tlec.html.twig
BPScott: On the existing brand page we create presenters in the controller and it ended up being a very ugly mess (as we never made a view helper for generating a promotion). 

These options are regarding how things look visually so IMO they belong in the view

src/DsAmen/Organism/Promotion/PromotionPresenter.php
BPScott: It does not.

tests/DsAmen/Organism/Promotion/PromotionPresenterTest.php
BPScott: This is used when you call getUrl(). I'm not convinced it's worth creating a new test case just for that single function.

tests/DsAmen/Organism/Promotion/PromotionPresenterTest.php
```diff
@@ -0,0 +1,145 @@
...
+        $this->assertSame('/programmes/b0000001', $presenter->getUrl());
+        $this->assertSame($image, $presenter->getImage());
+        // TV Episodes don't show a duration
+        $this->assertSame(0, $presenter->getDuration());
+        $this->assertSame(['set' => 'audio-visual', 'icon' => 'play'], $presenter->getActionIcon());
+    }
+
+    public function testPromotionOfImage()
+    {
+        $promotedEntity = $this->createMock(Image::class);
```
tatitati: usually dummy objects are used to complete the signature of a function and carry on with the execution. Cause of that I wouldn't create it in a separate variable. That makes me think that is important enough to assert something later agains this variable, keeping it in a separate variable.

BPScott: I do assert on the value. Scroll down a few lines:

```php
 $this->assertSame($image, $presenter->getImage());
```

src/DsAmen/Organism/Promotion/PromotionPresenter.php
tatitati: simple :smile:

tests/DsAmen/Organism/Promotion/PromotionPresenterTest.php
tatitati: I'm fool. Sorry!

programmes-developers:

programmes-developers:

src/DsAmen/Organism/Promotion/PromotionPresenter.php
BPScott: Usually i'd be with you, but in these cases I'm going for "when the data is absent early return the fallback" I think this reads a little better that mixing up if the fallback or the main case is within that first if statement.

src/DsAmen/Organism/Promotion/PromotionPresenter.php
BPScott: Done. We should decide and write up a convention for this sort of ruling.

PR #175, opened 2017-09-06
PROGRAMMES-5866 Adding in vanity URLs

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5866


src/Controller/SchedulesVanityRedirectController.php
```diff
@@ -0,0 +1,43 @@
...
+                $time = $time->addWeek();
+            } elseif ($vanity === 'last_week') {
+                $time = $time->subWeek();
+            }
+
+            $url = $router->generate('schedules_by_week', ['pid' => $service->getPid(), 'date' => $time->format('Y-\\wW')]);
+            return $this->redirect($url);
+        }
+
+        throw new Exception('Vanity URL ' . $vanity . ' not recognised');
```
votemike: Is a bog standard Exception the correct thing to use here?

programmes-developers:

src/Controller/SchedulesVanityRedirectController.php
felipeparaujo: RouteNotFoundException probably fits in better

tests/Controller/SchedulesVanityRedirectControllerTest.php
```diff
@@ -0,0 +1,48 @@
...
+    {
+        ApplicationTime::setTime((new Chronos('2017-09-27 12:00:00'))->getTimestamp());
+        $this->loadFixtures(["NetworksAndServicesFixture"]);
+
+        $client = static::createClient();
+        $url = '/schedules/p00rfdrb/' . $vanity;
+
+        $client->request('GET', $url);
+
+        $this->assertRedirectTo($client, 302, '/schedules/p00rfdrb/' . $urlSuffix);
```
tatitati: for any date is redirected?, or depending on the date?

votemike: It depends on the vanity string.

src/Controller/SchedulesVanityRedirectController.php
```diff
@@ -0,0 +1,42 @@
...
+        $time = ApplicationTime::getLocalTime();
+
+        if (in_array($vanity, ['today', 'tomorrow', 'yesterday'])) {
+            if ($vanity === 'tomorrow') {
+                $time = $time->tomorrow();
+            } elseif ($vanity === 'yesterday') {
+                $time = $time->yesterday();
+            }
+
+            $url = $router->generate('schedules_by_day', ['pid' => $service->getPid(), 'date' => $time->toDateString()]);
```
BPScott: Controllers have a `redirectToRoute()` function that you can use instead of having to inject a $router.

```
return $this->redirectToRoute('schedules_by_day', ['pid' => $service->getPid(), 'date' => $time->toDateString()];
```

https://symfony.com/doc/current/controller.html#generating-urls

src/Controller/SchedulesVanityRedirectController.php
```diff
@@ -0,0 +1,42 @@
...
+                $time = $time->addWeek();
+            } elseif ($vanity === 'last_week') {
+                $time = $time->subWeek();
+            }
+
+            $url = $router->generate('schedules_by_week', ['pid' => $service->getPid(), 'date' => $time->format('Y-\\wW')]);
+            return $this->redirect($url);
+        }
+
+        throw new RouteNotFoundException('Vanity URL ' . $vanity . ' not recognised');
```
BPScott: I'd be tempted to use the Controller's `$this->createNotFoundException()` method (which throws a NotFoundHttpException) rather than using the Router exception - "The route was found but we chose not to handle it in the controller" seems like an odd time to throw `RouteNotFoundException`

PR #177, opened 2017-09-08
PROGRAMMES-5873 fix when first week is part of the previous year

tatitati:


src/ValueObject/BroadcastWeek.php
```diff
@@ -14,12 +14,12 @@ class BroadcastWeek extends BroadcastPeriod
     public function __construct(string $date)
     {
         $dateParts = explode('-', $date);
-        $year = (int) $dateParts[0];
+        $specifiedYear = (int) $dateParts[0];
```
felipeparaujo: why rename everything to specified year?

tatitati: I was thinking that $this->start->year !== year can be a little confuse. I can restore it

felipeparaujo: yeah, i don't see a good reason to change this

BjornTwachtmann:
This should fix the specific issue. That said, I think the schedules controller needs to be doing better validation than "I'm going to chuck whatever shite the user throws me into a DateTime and hope it sticks". 

Week and year should both be validated to contain sensible values before throwing them into Chronos. 

I'm also worried that https://www.test.bbc.co.uk/schedules/p00fzl8y/2009-w01 is just showing a blank page and not a 404/500 error page? Can you have a look what's going on there?

programmes-developers:

votemike:
What was this for?

PR #179, opened 2017-09-11
PROGRAMMES-5875 Fixing spacing

votemike:
Bodge to fixing magic empty div that exists in live.
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5875


programmes-developers:

BjornTwachtmann:
I'm assuming this is what the existing page does?

votemike:
Yeah. There's some mystery empty div.

BjornTwachtmann:
Fair enough. Merge it then I guess.

BPScott:
Of those classes, programmes-page is the only one that provides any vertical padding. It should probably be applied to a wrapping div instead though.

votemike:
@BPScott The gap isn't as big if I put it on the wrapping div... Is that OK?

PR #182, opened 2017-09-11
PROGRAMMES-5881 Validate dates in controllers

tatitati:


src/Controller/SchedulesByDayController.php
```diff
@@ -233,4 +237,20 @@ private function getOtherNetworks(Service $service, NetworksService $networksSer
...
         });
     }
+
+    private function isValidDate(?string $date): bool
+    {
+        if (is_null($date)) {
+            return true;
+        }
+
+        $dateTime = Chronos::createFromFormat('Y/m/d', $date);
```
votemike: Where is `$dateTime` used?

src/Controller/SchedulesByMonthController.php
```diff
@@ -29,4 +33,15 @@ private function serviceIsActiveDuringMonth(Service $service, Date $firstOfMonth
     {
         return (!$service->getStartDate() || $service->getStartDate() <= $firstOfMonth->endOfMonth()) && (!$service->getEndDate() || $firstOfMonth < $service->getEndDate());
     }
+
+    private function isValidMonth(string $date): bool
+    {
+        list($year, $month) = explode('/', $date);
```
votemike: Do we want to validate anything about the year? That it's an int? That it has 4 digits?

src/Controller/SchedulesByMonthController.php
```diff
@@ -29,4 +33,15 @@ private function serviceIsActiveDuringMonth(Service $service, Date $firstOfMonth
...
+
+        if (in_array($month, range(1,12))) {
```
votemike: `return in_array($month, range(1,12))`?

src/Controller/SchedulesByWeekController.php
```diff
@@ -176,4 +180,20 @@ private function twinService(Service $mainService, array $servicesInNetwork): ?S
 
         return reset($otherServices);
     }
+
+    private function isValidWeek($date): bool
+    {
+        $dateParts = explode('/', $date);
```
votemike: You could `list` as you did above

src/Controller/SchedulesByWeekController.php
```diff
@@ -176,4 +180,20 @@ private function twinService(Service $mainService, array $servicesInNetwork): ?S
...
+        $year = (int) $dateParts[0];
+        $week = (int) str_replace('w', '', $dateParts[1]);
+        $blankChronos = ApplicationTime::getLocalTime()->startOfDay(); // Today at midnight
+        $start = $blankChronos->setISODate($year, $week, 1); // Midnight on Monday
+
+        // validate format
+        if ($start->year > $year || !in_array($week, range(1, 53))) {
```
votemike: Maybe a little explanation as to why 53 weeks are allowed?

src/Controller/SchedulesByDayController.php
BjornTwachtmann: I don't like this. We should decide what a valid date looks like for our application, not Chronos.

src/Controller/SchedulesByMonthController.php
```diff
@@ -29,4 +33,15 @@ private function serviceIsActiveDuringMonth(Service $service, Date $firstOfMonth
     {
         return (!$service->getStartDate() || $service->getStartDate() <= $firstOfMonth->endOfMonth()) && (!$service->getEndDate() || $firstOfMonth < $service->getEndDate());
     }
+
+    private function isValidMonth(string $date): bool
```
BjornTwachtmann: Why only valid month? What about valid year?

src/Controller/SchedulesByWeekController.php
```diff
@@ -176,4 +180,20 @@ private function twinService(Service $mainService, array $servicesInNetwork): ?S
 
         return reset($otherServices);
     }
+
+    private function isValidWeek($date): bool
+    {
+        $dateParts = explode('/', $date);
+        $year = (int) $dateParts[0];
+        $week = (int) str_replace('w', '', $dateParts[1]);
+        $blankChronos = ApplicationTime::getLocalTime()->startOfDay(); // Today at midnight
```
BjornTwachtmann: Again. I don't think this is the best approach. I think we need to validate the input ourselves, not just see if Chronos vomits it back up or not :)

src/Controller/SchedulesByMonthController.php
tatitati: is validated in the routing.yml

BjornTwachtmann:
We also need to validate utcOffset parameter where we use/expect it. That will probably need some kind of helper or trait.

src/Controller/SchedulesByWeekController.php
tatitati: yeah, is the same validation than in routing. The main difference is that routing.yml our regex expression also allow 00 as a valid week number (from my point of view is wrong), so I validate it in here as well.

programmes-developers:

src/Controller/SchedulesByWeekController.php
```diff
@@ -176,4 +181,26 @@ private function twinService(Service $mainService, array $servicesInNetwork): ?S
...
+    private function isValidWeek($date): bool
+    {
+        // validate format
+        if (!preg_match('#\d{4}/\w\d{2}#', $date)) {
+            return false;
+        }
+
+        // validate content
+        list($year, $week) = explode('/', $date);
+        $dateTime = DateTime::createFromFormat('Y/m', $date);
```
votemike: This isn't how you create a date from a week.

src/Controller/SchedulesByYearController.php
```diff
@@ -29,4 +33,14 @@ private function serviceIsActiveDuringYear(Service $service, Date $startOfYear):
     {
         return (!$service->getStartDate() || $service->getStartDate() <= $startOfYear->endOfYear()) && (!$service->getEndDate() || $startOfYear < $service->getEndDate());
     }
+
+    private function isValidYear(string $date): bool
+    {
+        // validate format
+        if (!preg_match('#\d{4}#', $date)) {
```
votemike: This is also triggered if an error occurred. Is that OK?

tests/Controller/SchedulesByDayControllerTest.php
```diff
@@ -210,6 +210,41 @@ public function testNoScheduleNoResult()
...
+    }
+
+    public function invalidDatesForControllerValidationProvider(): array
+    {
+        // trigger HTTP NOT FOUND EXCEPTION
+        return [
+            'CASE 1: nonexistent month' => ['Specified date not found', '2012/13/20'],
+            'CASE 2: nonexistent month' => ['Specified date not found', '2012/00/20'],
+            'CASE 3: nonexistent day trigger HTTP NOT FOUND EXCEPTION' => ['Specified date not found', '2012/02/36'],
+            'CASE 4: nonexistent day trigger HTTP NOT FOUND EXCEPTION' => ['Specified date not found', '2009/02/29'],
```
votemike: 👍

src/Controller/SchedulesByWeekController.php
tatitati: thanks @michael. I couldn't spot it, tests were passing. I will fix it

src/Controller/SchedulesByDayController.php
```diff
@@ -233,4 +237,29 @@ private function getOtherNetworks(Service $service, NetworksService $networksSer
...
+
+        // validate format
+        if (!preg_match('#\d{4}/\d{2}/\d{2}#', $date)) {
+            return false;
+        }
+
+        // validate content
+        list($year, $month, $day) = explode('/', $date);
+
+        if (!in_array((int) $month, range(1, 12))) {
```
BjornTwachtmann: Is there a reason this doesn't say "$month >= 1 && $month <=12". This seems a bit convoluted.

src/Controller/SchedulesByDayController.php
```diff
@@ -233,4 +237,29 @@ private function getOtherNetworks(Service $service, NetworksService $networksSer
...
+            return false;
+        }
+
+        if (!in_array((int) $day, range(1, 31))) {
```
BjornTwachtmann: Again, as above

src/Controller/SchedulesByWeekController.php
```diff
@@ -176,4 +180,22 @@ private function twinService(Service $mainService, array $servicesInNetwork): ?S
...
+    private function isValidDate($date): bool
+    {
+        // validate format
+        if (!preg_match('#\d{4}/\w\d{2}#', $date)) {
+            return false;
+        }
+
+        // validate content
+        list($year, $week) = explode('/', $date);
+        $week = (int) preg_replace('/w/', '', $week);
```
votemike: The "w" will be the first character of `$week` won't it? Can't you just remove the first character instead of a preg?

src/Controller/SchedulesByWeekController.php
```diff
@@ -176,4 +180,22 @@ private function twinService(Service $mainService, array $servicesInNetwork): ?S
...
+
+        if ($week == 0 || $week > 53) {
```
votemike: What id the week is -1?

votemike:
There was a test for February 31st or something before. Where has that gone? Can we bring it back?

src/Controller/SchedulesByDayController.php
```diff
@@ -233,4 +237,29 @@ private function getOtherNetworks(Service $service, NetworksService $networksSer
...
     }
+
+    private function isValidDate(?string $date): bool
+    {
+        if (is_null($date)) {
+            return true;
+        }
+
+        // validate format
+        if (!preg_match('#\d{4}/\d{2}/\d{2}#', $date)) {
```
votemike: `\d{4}/\([0-9]|1[0-2])/\([0-9]|[1-2][0-9]|3[0-1])`?

src/Controller/SchedulesByMonthController.php
```diff
@@ -29,4 +33,21 @@ private function serviceIsActiveDuringMonth(Service $service, Date $firstOfMonth
     {
         return (!$service->getStartDate() || $service->getStartDate() <= $firstOfMonth->endOfMonth()) && (!$service->getEndDate() || $firstOfMonth < $service->getEndDate());
     }
+
+    private function isValidDate(string $date): bool
+    {
+        // validate format
+        if (!preg_match('#\d{4}/\d{2}#', $date)) {
```
votemike: `\d{4}/\([0-9]|1[0-2])`?

src/Controller/SchedulesByWeekController.php
```diff
@@ -176,4 +180,22 @@ private function twinService(Service $mainService, array $servicesInNetwork): ?S
 
         return reset($otherServices);
     }
+
+    private function isValidDate($date): bool
+    {
+        // validate format
+        if (!preg_match('#\d{4}/\w\d{2}#', $date)) {
```
BjornTwachtmann: This should be w, not \w

votemike: `\d{4}/w([0-4][0-9]|[5][0-3])`?

src/Controller/SchedulesByDayController.php
BjornTwachtmann: I'd rather have simpler regex and some numeric comparisons. Those kind of regexes are annoying to read and debug.

src/Controller/SchedulesByDayController.php
```diff
@@ -233,4 +237,29 @@ private function getOtherNetworks(Service $service, NetworksService $networksSer
...
+        }
+
+        // validate content
+        list($year, $month, $day) = explode('/', $date);
+
+        if ($month < 1 || $month > 12) {
+            return false;
+        }
+
+        if ($day < 1 || $day > 31) {
```
votemike: This allows February 31st. Which isn't a valid date.

tatitati: yeah, Bjorn and I decided to make it more generic that what I was doing

votemike: So we're happy to accept Feb 31st and just have it break?

tatitati: Actually is not broken. I ran a few tests and I saw that 
```$broadcastDay = new BroadcastDay($dateTimeToShow, $service->getNetwork()->getMedium());``` calculates dynamically the next correct date when the user introduces one that is not possible. 
So in this case if a user introduces ```2009/02/31``` we just display data from the day ```2009/03/03```.

@BjornTwachtmann any thought?, we want to keep this behaviour?

BjornTwachtmann: Fuck, seriously? I'd rather it just blew up with an invalidargumentexception than that. We can discuss this tomorrow.

PR #185, opened 2017-09-11
PROGRAMMES-5868 Adding ids of pages that are allowed to edit FB previ…

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5868


app/Resources/views/base_ds2013.html.twig
```diff
@@ -10,6 +10,9 @@
         <meta name="theme-color" content="{{ branding.getColours().highlight.bg|default('#2B2B2B') }}">
         <meta name="description" content="{{ meta_context.description() }}">
 
+        {#- Facebook pages that are allowed to edit FB preview titles and descriptions -#}
+        <meta property="fb:pages" content="194575130577797,7397061762,7833211321,129044383774217,401538510458,8251776107,273726292719943,107909022566650,21750735380,1477945425811579,944295152308991,154344434967,8244244903,512423982152360,156060587793370,215504865453262,260967092113,1392506827668140,193435954068976,7519460786,1143803202301544,1470145583204820,470911516262605,413132078795966,118883634811868,130593816777,228735667216,80758950658,260212261199,294662213128,1086451581439050,283348121682053,295830058648,239931389545417,304314573046,310719525611571,647687225371774,1159932557403143,286567251709437,1731770190373618,125309456546,742734325867560,168895963122035,292291897588734,658551547588605,156400551056385,185246968166196,1411916919051820,958681370814419,317278538359186,8585725981,1436581493296600,163571453661989,6025943146,193022337414607,1659215157653827">
```
votemike: Happy to hardcode these in the template? Or do we want to put them somewhere else? Maybe the MetaContext?

programmes-developers:

app/Resources/views/base_ds2013.html.twig
BPScott: I like the idea of putting this in the meta context, then we don't have to repeat ourselves (I like the idea of making ORB do there's a single point across the whole BBC even more but that's a different story)

src/ValueObject/MetaContext.php
```diff
@@ -108,6 +113,71 @@ public function showAdverts(): bool
         return $this->showAdverts;
     }
 
+    /**
+     * @return int[]
+     */
+    private function bbcFacebookPageIds(): array
+    {
+        return [
```
votemike: Do we want these indexed with the pages that they are IDs of?

BPScott: I don't think it's worth it, we'll never use those names anywhere

PR #186, opened 2017-09-12
added test for schedules

tatitati:
For weird dates is not to clear how is the behaviour of schedules (for instance, 31-feb, which is wrong). And there is no tests for that. So I added a couple.


tests/Controller/SchedulesByDayControllerTest.php
```diff
@@ -56,6 +56,32 @@ public function scheduleDateTestProvider(): array
...
+        $this->assertEquals(
+            $expectedDateDisplayed,
+            $crawler->filter("h3.broadcast__time[property='startDate']")->extract(['content'])[0]
+        );
+    }
+
+    public function specifiedDateProvider(): array
+    {
+        return [
+            'CASE 1: specify a correct date to display correct schedules date' => ['2017-05-22T03:00:00+01:00', '2017/05/22'],
```
votemike: Where does 3:00 come from?

tatitati: good question, I was just focus on the day/month/year, but the rest not sure

PR #187, opened 2017-09-12
PROGRAMMES-5884 validation of utcoffset

tatitati:
PROGRAMMES-5884 validation of utcoffset in schedules


programmes-developers:

PR #189, opened 2017-09-13
PROGRAMMES-5883 . fix for issue 2

tatitati:
Jira:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5883

This ticket fix the issue 2.


programmes-developers:

BjornTwachtmann:
Hmm. I think this needs to be more generic. If we're including query strings in what we send to comscore we should include all query strings

PR #191, opened 2017-09-13
PROGRAMMES-5734 Adding Coming Soon column to MAP

votemike:
For https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5734

Are people happy with the size of this? Or would you like me to see if I can split it in to two PRs so that you can see what's happening?

I have been using the following URLs to test:
- "Promo Takeover" (shows synopsis) http://sandbox.bbc.co.uk/programmes/p00bm305
- "Coming Soon" http://www.test.bbc.co.uk/programmes/b00cyj3f


programmes-developers:

src/DsAmen/Organism/Map/MapPresenter.php
```diff
@@ -91,8 +94,12 @@ public function __construct(
                 $this->rightGridClasses = '2/3@gel3b';
             }
 
-            $this->rightColumns[] = new OnDemandPresenter($programme, ['must_show_tx_column' => true]);
-            $this->rightColumns[] = new TxPresenter($programme, ['must_show_tx_column' => true]);
+            $this->rightColumns[] = new OnDemandPresenter($programme, ['must_show_tx_column' => $this->comingSoonTakeover]);
+            if ($hasComingSoon) {
+                $this->rightColumns[] = new ComingSoonPresenter($programme, $promotion, ['must_show_tx_column' => $this->comingSoonTakeover, 'show_mini_map' => $this->showMiniMap, 'show_synopsis' => $this->comingSoonTakeover]);
```
votemike: Depending on what the '1/1' and '1/2@gel1b' classes actually do, I think this could be simplified A LOT. Depending on when @BPScott is free to walk me through it, I'll either update this PR or do it as part of another PR.

src/Controller/FindByPid/TlecController.php
```diff
@@ -70,4 +78,34 @@ public function __invoke(
...
+        $comingSoon = $comingSoonBlock['content']['promotions'];
+        if (!array_key_exists('promotion_title', $comingSoon)) {
+            $comingSoon = reset($comingSoon);
+        }
+
+        $pid = new Pid($comingSoon['promoted_item_id']);
+        $synopses = new Synopses($comingSoon['short_synopsis']);
+        $image = $imagesService->findByPid($pid);
+        if (is_null($image)) {
+            throw new Exception('SOmething bad happened'); //@TODO what sort of error do we want here? Or maybe just not create the promotion?
```
votemike: What do we want to happen if the Image can't be loaded from the DB? Throw an Exception? Or just pass null through as the promotion so that it is treated as a non-promoted brand.

tests/DsAmen/Organism/Map/SubPresenter/ComingSoonPresenterTest.php
```diff
@@ -0,0 +1,56 @@
...
+class ComingSoonPresenterTest extends TestCase
+{
+    /**
+     * @dataProvider invalidOptionProvider
+     * @param mixed[] $options
+     * @param string $expectedExceptionMessage
+     */
+    public function testInvalidOptions(array $options, string $expectedExceptionMessage)
+    {
+        $programme = $this->createMock(Programme::class);
```
votemike: @tatitati Is $programme the correct name for this? Do you have any documentation about your new naming system?

src/DsAmen/Organism/Map/MapPresenter.php
```diff
@@ -75,13 +77,14 @@ public function __construct(
         }
 
         // Add columns to map
-        if ($this->comingSoonTakeover) {
-            $this->rightColumns[] = new OnDemandPresenter($programme, ['must_show_tx_column' => false]); // I hope once the other columns are done, we won't need to pass in must_show_tx_column
-            $this->rightColumns[] = new ComingSoonPresenter($programme);
-        } elseif ($this->isWorldNews()) {
```
votemike: I have refactored this line and added `!$this->comingSoonTakeover` as https://github.com/bbc/programmes/blob/344819df35c03883ce148658776317509253f23c/webapp/php/lib/BBC/Programmes/View/Map2016.php#L43 seems to suggest that the takeover takes precedent over choosing which columns are shown rather than is it is a World News programme.
Do people think it would be "more" correct to  keep this just as `if ($this->isWorldNews()) {`?

tests/DsAmen/Organism/Map/MapPresenterTest.php
```diff
@@ -43,8 +42,7 @@ public function testComingSoonTakeoverColumns()
     {
         $programmeContainer = $this->createMock(ProgrammeContainer::class);
         $programmeContainer->method('getAggregatedEpisodesCount')->willReturn(0);
-        $programmeContainer->expects($this->at(0))->method('getOption')->with('coming_soon')->willReturn(null);
-        $programmeContainer->expects($this->at(1))->method('getOption')->with('comingsoon_textonly')->willReturn('Coming Soon Text');
+        $programmeContainer->expects($this->at(0))->method('getOption')->with('comingsoon_textonly')->willReturn('Coming Soon Text');
```
tatitati: I think in this case we can remove the at(0) expectation as we only call ```getOption('comingsoon_textonly')```

tests/DsAmen/Organism/Map/SubPresenter/ComingSoonPresenterTest.php
tatitati: for me is fine. As personal preference I try no to keep dummy objects in any variable, in this way I make clear that dummy objects are not important at all in the test except to pass them as arguments and carry on with the test. But is just a sugar question.

src/DsAmen/Organism/Map/MapPresenter.php
BPScott: `1/1` applies the style `.1/1 {width:100%}`. `1/2@gel1b` applies the style `@media min-width:320px { 1/2@gel1b {width: 50%; } }.

`1/2@gel1b` is make this element 1/2 the width of its container when the viewport is at least the gel1b breakpoint (which is 320px).

tests/DsAmen/Organism/Map/MapPresenterTest.php
votemike: It doesn't seem to work if I remove it.... 🤔

tatitati: really?, 
can you try this?
``` php
$programmeContainer->expects($this->once())->method('getOption')->with('comingsoon_textonly')->willReturn('Coming Soon Text')
```

votemike: Still no

src/Controller/FindByPid/TlecController.php
BPScott: Treat it as though there as no promotion. A silent "fail" and recovery is better than falling over in a ball of flames

src/Controller/FindByPid/TlecController.php
```diff
@@ -60,7 +66,9 @@ public function __invoke(
             }
         }
 
-        $mapPresenter = $presenterFactory->mapPresenter($request, $programme, $upcomingEpisodesCount, $mostRecentBroadcast);
+        $promotion = $this->getPromotion($imagesService, $programme);
```
BPScott: getPromotion might be confusing as we've already got a list of promotions for this page. `getComingSoonPromo` perhaps?

PR #192, opened 2017-09-14
PROGRAMMES-5734 Refactoring MAP

votemike:
After doing the work for PROGRAMMES-5734 and speaking to Ben and Cathal, I understand the map a lot better. Not just what it does in V2, but also what it is _supposed_ to do. This is a small refactor to make the MAP more intelligible and to make https://github.com/bbc/programmes-frontend/pull/191/ simpler to review.


tatitati:
nice refactoring. Should we update some tests for this?. It would be nice a test where we can see in which cases are displayed or not these columns

votemike:
@tatitati Are the ones in MapPresenterTest not sufficient?

tatitati:
didn't check. I was just asking, jeje, sorry, seems that we have!, jeje

programmes-developers:

BjornTwachtmann:
I don't know the MAP logic well enough to eyeball this and say whether or not it's correct. This looks like a good structure though, and the explanation is clear and readable.

PR #193, opened 2017-09-14
PROGRAMMES-5874 Istats counter name fix differences with PAL in AWS

tatitati:
Jira: 
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5874

This ticket fix the problem of the ticket. 
Also cover rest of cases with counter name (which after comparing with PAL I saw they work properly).


programmes-developers:

tests/ValueObject/AnalyticsCounterNameTest.php
```diff
@@ -0,0 +1,138 @@
...
+     * @dataProvider urlDefaultProvider
+     */
+    public function testCounterNameIsBuildByDefaultInOtherCases(string $expectedValueOutput, string $relativePathProvided)
+    {
+        $analyticsCounterName = new AnalyticsCounterName(null, $relativePathProvided);
+
+        $this->assertEquals($expectedValueOutput, (string) $analyticsCounterName);
+    }
+
+    public function urlDefaultProvider(): array
```
votemike: Do we need a provider with one item in it?

tests/ValueObject/AnalyticsCounterNameTest.php
```diff
@@ -0,0 +1,138 @@
...
+            ]),
+
+        ]);
+
+        $analyticsCounterName = new AnalyticsCounterName($programmeContext, $relativePathProvided);
+
+        $this->assertEquals($expectedValueOutput, (string) $analyticsCounterName);
+    }
+
+    public function relativePathForProgrammeWithparentsProvider(): array
```
votemike: Do we need a provider with one item in it?

tatitati: yeah, I should remove this provider. I don't expect to add more scenarios for this case

programmes-developers:

PR #194, opened 2017-09-14
PROGRAMMES-5734 Making method name clearer

votemike:
https://github.com/bbc/programmes-frontend/pull/191#discussion_r138868033


programmes-developers:

PR #195, opened 2017-09-14
PROGRAMMES-5852 Fix param utcoffset repeated in browser

tatitati:
JIRA:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5885


felipeparaujo:
Should you be refactoring rewriteUrlWithUtcOffset instead of removing the attribute?

BjornTwachtmann:
@felipeparaujo that was my suggestion. Simply modify the search parameters instead of looking for a string 50 times. I can't see any downside to removing the attribute once the link is processed, I could be wrong though.

app/Resources/js/timezone-notification.js
```diff
@@ -50,6 +50,7 @@ define('timezone-notification', ['jquery-1.9', 'rv-bootstrap'], function ($) {
                     'Sat'
                 ]
             },
+            attrTargetAddUtcOffset: 'data-href-add-utcoffset',
```
votemike: This is already specified inside cssSelectors isn't it? Can't you access it from there?

tatitati: no because of the square brackets in that selector.

programmes-developers:

PR #196, opened 2017-09-14
PROGRAMMES-5881 put 1900 as limit year in schedules

tatitati:
JIRA:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5881


src/Controller/SchedulesByWeekController.php
```diff
@@ -195,7 +195,7 @@ private function isValidDate($date): bool
         list($year, $week) = explode('/', $date);
         $week = (int) str_replace('w', '', $week);
 
-        if ($week < 1 || $week > 53) {
+        if ($week < 1 || $week > 53 || $year < 1900) {
```
votemike: You haven't cast year to an int.

src/Controller/SchedulesByDayController.php
```diff
@@ -252,6 +252,11 @@ private function isValidDate(?string $date): bool
         }
 
         list($year, $month, $day) = explode('/', $date);
+
+        if ($year < 1900) {
```
votemike: Is 1900 a random year? Do we want it to be a constant somewhere?

src/Controller/SchedulesByWeekController.php
tatitati: yeah, is not necessary actually, at the moment of comparing php is treating it as an int (in tests I pass an string, and they work properly). Same with week

src/Controller/SchedulesByDayController.php
tatitati: I had a discussion with @BjornTwachtmann (actually was Jatin). We saw some weird behaviour in extreme cases. For instance if you select the year 0000, you see in the dates-bar the next stuff:
-0009 || -0008 || -0007 | ......|| 0000 || 0001 || 0002 || ........

votemike: So should it be a constant?

BjornTwachtmann: Yeah, I agree, make this a public constant in one of the Schedules controllers (don't care which one) and refer to it from all these locations for consistency.

src/Controller/SchedulesHomeController.php
```diff
@@ -9,6 +9,8 @@
 
 class SchedulesHomeController extends BaseController
 {
+    const LIMIT_YEAR = 1900;
```
votemike: Could this be better named? `LIMIT_YEAR` is a tiny bit vague. Not sure what though.... Maybe `MINIMUM_VALID_YEAR`?

programmes-developers:

tests/Controller/SchedulesByDayControllerTest.php
```diff
@@ -236,12 +236,13 @@ public function invalidFormatDatesProvider(): array
...
     public function invalidDatesForControllerValidationProvider(): array
     {
-        // trigger HTTP NOT FOUND EXCEPTION
+        // trigger HTTP NOT FOUND EXCEPTION (validation exception)
         return [
             'CASE 1: nonexistent month' => ['Invalid date supplied', '2012/13/20'],
             'CASE 2: nonexistent month' => ['Invalid date supplied', '2012/00/20'],
             'CASE 3: nonexistent day' => ['Invalid date supplied', '2012/02/36'],
             'CASE 4: nonexistent day' => ['Invalid date supplied', '2009/02/00'],
+            'CASE 5: invalid year, previous to 1900' => ['Invalid date supplied', '1800/02/00'],
```
votemike: This date is also invalid because of the day, how are you sure you are testing the error on the year?

PR #197, opened 2017-09-15
PROGRAMMES-5886 Forward URL params as part of Vanity Redirect

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5886


programmes-developers:

BjornTwachtmann:
So, I realise this has nothing to do with this PR (which is fine). But I'm going to take this oppurtunity to ask an annoying question anyway. Why in the name of Jebus are we bothering to do a database lookup for this service when all we want is a straight redirect based on the URL?

votemike:
@BjornTwachtmann you mean as part of the magic so that we get a whole `Service` passed in to the `__invoke()` method?  
@BPScott is there a way to bypass this?

BjornTwachtmann:
@votemike it's just dark magic based on type hints and variable names in the __invoke method. Change "Service $service" to "string $pid".

votemike:
@BjornTwachtmann updated

PR #205, opened 2017-09-21
PROGRAMMES-5893 flush cache

Tasiobg:
Flush the cache for the current request when the next query string is present `?__flush_cache=1`

Depend on this branding client branch: https://github.com/bbc/branding-client/tree/PROGRAMMES-5893-flush-cache

Composer will be update when the changes on branding get approved and merged


src/EventSubscriber/CacheFlushSubscriber.php
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public function __construct(ContainerInterface $container)
+    {
+        $this->container = $container;
+    }
+
+    public function setupCacheFlush(GetResponseEvent $event)
+    {
+        if ($event->getRequest()->get('__flush_cache') === '1') {
```
votemike: This makes me nervous. This isn't limited to admins nor environments. What if someone repeatedly hits these URLs. Either on purpose or not.

app/config/services.yml
```diff
@@ -80,6 +81,7 @@ services:
             - env: '%app.feed_env.orbit%'
               cacheKeyPrefix: 'frontend.%cosmos_component_release%.orbit'
               mustache: { cache: '%kernel.cache_dir%/mustache' }
+        public: true
```
BPScott: Symfony is moving towards very few public services, as for the most part they're not needed. In our app services are private by default. 

If it is not possible to have CacheFlushSubscriber implement ServiceSubscriberInterface, and instead we must pass in the main container it would be worth documenting why that is the case and why these services must be public

src/EventSubscriber/CacheFlushSubscriber.php
```diff
@@ -0,0 +1,44 @@
...
+
+class CacheFlushSubscriber implements EventSubscriberInterface
+{
+    /** @var ContainerInterface */
+    private $container;
+
+    public static function getSubscribedEvents()
+    {
+        return [
+            KernelEvents::REQUEST => [['setupCacheFlush']],
```
BPScott: REQUEST is very early in the event lifecycle - before we even know the controller or if the URL requested is even a valid route. Is it possible to make this happen later, so that we can avoid work when it is not needed?

Perhaps this could be done in the BaseController constructor instead of a dedicated event? That way we wouldn't need to make the services public

app/config/services.yml
BjornTwachtmann: Honestly, who cares? What is the downside to making these services "public"?

src/EventSubscriber/CacheFlushSubscriber.php
BjornTwachtmann: No, that's the wrong place to do this. It should happen before any data gets requested. The amount of work is trivial, altough we have enough subscribers now that I'm starting to wonder if we should merge some of them.

programmes-developers:

src/EventSubscriber/CacheFlushSubscriber.php
Tasiobg: This is mean to work on live too, we have discuss this issue before the implementation and if for any reason this become a load problem we could disable it.
If you have any question feel free to ask me in person and I'll give you more details about it.

votemike: Also, does it need to be 1? Surely it could just exist in the request? e.g. `?__flush_cache`

PR #206, opened 2017-09-25
PROGRAMMES-5735 - Collapsed broadcast object

felipeparaujo:
**TESTS WILL CAME IN A LATER PR**. I don't want to make this impossible to read, it's already annoying as it is.


src/DsAmen/Organism/Programme/CollapsedBroadcastPresenter.php
```diff
@@ -0,0 +1,99 @@
...
+            $this->collapsedBroadcast,
+            $this->router,
+            $this->helperFactory->getLiveBroadcastHelper(),
+            $options
+        );
+    }
+
+    protected function buildProgrammeCtaPresenter(): ?ProgrammeCtaPresenter
+    {
+        $ctaPresenter = null;
```
votemike: Does this need to be set as null? Could you not just return the values instantly?

src/DsAmen/Organism/Programme/CollapsedBroadcastPresenter.php
```diff
@@ -0,0 +1,99 @@
...
+            $this->subPresenterOptions('services_details_options')
+        );
+    }
+
+    public function getCollapsedBroadcastDatePresenter(): CollapsedBroadcastDatePresenter
+    {
+        return new CollapsedBroadcastDatePresenter(
+            $this->collapsedBroadcast,
+            $this->helperFactory->getLocalisedDaysAndMonthsHelper(),
+            $this->subPresenterOptions('date_options')
```
votemike: I can't see these being used inside the `CollapsedBroadcastDatePresenter`. Are they?

programmes-developers:

src/DsAmen/Organism/Programme/CollapsedBroadcastSubPresenters/CollapsedBroadcastDatePresenter.php
```diff
@@ -0,0 +1,51 @@
...
+    public function getTime(): DateTimeImmutable
+    {
+        return $this->collapsedBroadcast->getStartAt();
+    }
+
+    public function isInternational(): bool
+    {
+        return array_reduce(
+            $this->collapsedBroadcast->getServices(),
+            function (bool $isInternational, Service $service) {
```
votemike: Where is `$isInternational` first set? Do I not know how array_reduce works?

src/DsAmen/Organism/Programme/CollapsedBroadcastSubPresenters/CollapsedBroadcastProgrammeCtaPresenter.php
```diff
@@ -0,0 +1,83 @@
...
+    {
+        return $this->liveBroadcastHelper->isWatchableLive($this->collapsedBroadcast);
+    }
+
+    public function getTemplateVariableName(): string
+    {
+        return 'collapsed_broadcast_cta';
+    }
+
+    public function removeBackgroundIfStandalone(): string
```
votemike: The name of this method sounds like it should return a bool. Do you think it's worth renaming?

src/DsAmen/Organism/Programme/CollapsedBroadcastSubPresenters/CollapsedBroadcastServicesDetailsPresenter.php
```diff
@@ -0,0 +1,110 @@
...
+use App\Translate\TranslateProvider;
+use BBC\ProgrammesPagesService\Domain\ApplicationTime;
+use BBC\ProgrammesPagesService\Domain\Entity\CollapsedBroadcast;
+use BBC\ProgrammesPagesService\Domain\Entity\Service;
+use DateTimeImmutable;
+use RMP\Translate\Translate;
+
+class CollapsedBroadcastServicesDetailsPresenter extends Presenter
+{
+    /** @var  DateTimeImmutable */
```
votemike: TOO MANY SPACES!!!

src/DsAmen/Organism/Programme/CollapsedBroadcastSubPresenters/CollapsedBroadcastServicesDetailsPresenter.php
```diff
@@ -0,0 +1,110 @@
...
+    /** @var string */
+    public $serviceAvailabilityQualifier;
+
+    public function __construct(
+        CollapsedBroadcast $collapsedBroadcast,
+        TranslateProvider $translateProvider,
+        array $options = []
+    ) {
+        parent::__construct($options);
+        $this->now = ApplicationTime::getLocalTime();
```
votemike: Will this work for international services?

src/DsAmen/Organism/Programme/CollapsedBroadcastSubPresenters/CollapsedBroadcastServicesDetailsPresenter.php
```diff
@@ -0,0 +1,110 @@
...
+
+        return $this->translate->translate('broadcast_only_on', ['%1' => $serviceNames]);
+    }
+
+    private function getAllServices(): array
+    {
+        $allServices = [];
+
+        foreach ($this->collapsedBroadcast->getServices() as $service) {
+            foreach ($service->getNetwork()->getServices() as $networkService) {
```
votemike: Are these all pre-fetched?

src/DsAmen/Organism/Programme/CollapsedBroadcastSubPresenters/CollapsedBroadcastServicesDetailsPresenter.php
```diff
@@ -0,0 +1,110 @@
...
+            array_map(function (Service $service) {
+                return $service->getShortName();
+            }, $this->collapsedBroadcast->getServices())
+        );
+        $serviceNames = implode(', ', $serviceNames);
+
+        return $this->translate->translate('broadcast_only_on', ['%1' => $serviceNames]);
+    }
+
+    private function getAllServices(): array
```
votemike: An array of what? Services? Or strings? Maybe add a docblock?

src/DsAmen/Organism/Programme/CollapsedBroadcastSubPresenters/CollapsedBroadcastDatePresenter.php
felipeparaujo: `array_reduce` iterates over the array and reduces it to a single value. The first parameter of the callback function is the result of calling callback function itself on the previous element of the array. The starting value for `$isInternational` is the 3rd parameter of `array_reduce`, false in this case.

src/DsAmen/Organism/Programme/CollapsedBroadcastSubPresenters/CollapsedBroadcastProgrammeCtaPresenter.php
felipeparaujo: Do you have any suggestions? I'd be happy to rename this, only this is the best I could come up with

src/DsAmen/Organism/Programme/CollapsedBroadcastSubPresenters/CollapsedBroadcastServicesDetailsPresenter.php
felipeparaujo: FIXED

src/DsAmen/Organism/Programme/CollapsedBroadcastSubPresenters/CollapsedBroadcastProgrammeCtaPresenter.php
votemike: No good suggestions.... `backgroundClass`?

src/DsAmen/Organism/Programme/CollapsedBroadcastSubPresenters/CollapsedBroadcastServicesDetailsPresenter.php
felipeparaujo: You mean the network services? Yeah, I had to create a new query in programmes-pages-service for it.

src/DsAmen/Organism/Programme/CollapsedBroadcastPresenter.php
```diff
@@ -0,0 +1,78 @@
...
+    {
+        return new CollapsedBroadcastDetailsPresenter(
+            $this->collapsedBroadcast,
+            $this->translateProvider,
+            $this->helperFactory->getLocalisedDaysAndMonthsHelper(),
+            $this->options
+        );
+    }
+
+    public function getWatchFromStartCollapsedBroadcastCtaPresenter(): ?ProgrammeCtaPresenter
```
tatitati: Not sure, but I think the return type cannot be null. We return always a new CollapsedBroadcastProgrammeCta...

src/DsAmen/Organism/Map/MapPresenter.php
```diff
@@ -126,6 +130,8 @@ public function showMap(): bool
 
     private function constructThreeColumnMap(?Promotion $promotion, bool $hasComingSoon, bool $programmeHasEpisodes)
     {
+        $comingSoonTakeover = !$this->programme->getParent() && ($hasComingSoon && !$programmeHasEpisodes);
```
votemike: This was removed in a previous commit, https://github.com/bbc/programmes-frontend/commit/64dd9f91886fb3be9ad42cb8474e41fc507f3f34#diff-1071d9d469460846b22baa06bcd5958a have you added it on purpose? Or is this due to a rebase?

src/DsAmen/Organism/Map/SubPresenter/TxPresenter.php
```diff
@@ -14,14 +20,26 @@ class TxPresenter extends Presenter
     /** @var Programme */
     private $programme;
 
-    public function __construct(Programme $programme, array $options = [])
+    /** @var CollapsedBroadcast|null */
+    private $collapsedBroadcast;
+
+    protected $options = [
+        'context_programme' => null,
```
votemike: Where is this used?

src/DsAmen/Organism/Map/MapPresenter.php
felipeparaujo: I forgot to remove it from the changes =[. I'll undo it.

src/DsAmen/Organism/Map/SubPresenter/TxPresenter.php
felipeparaujo: Various things, e.g. when building the programme title, we go through the programme parents titles (Episode 1, Series 2, Doctor Who). If we're in the Doctor Who page, we don't want to show the redundant Doctor Who title, so we set it as the context.

src/DsAmen/Organism/Programme/CollapsedBroadcastSubPresenters/CollapsedBroadcastDetailsPresenter.php
```diff
@@ -0,0 +1,136 @@
...
+    /** @var DateTimeImmutable */
+    private $now;
+
+    /** @var CollapsedBroadcast */
+    private $collapsedBroadcast;
+
+    /** @var Translate */
+    private $translate;
+
+    /** @var string */
```
votemike: `/** @var string|null */`?

src/DsAmen/Organism/Programme/CollapsedBroadcastSubPresenters/CollapsedBroadcastDetailsPresenter.php
```diff
@@ -0,0 +1,136 @@
...
+        $this->now = ApplicationTime::getLocalTime();
+        $this->localisedDaysAndMonthsHelper = $localisedDaysAndMonthsHelper;
+        $this->collapsedBroadcast = $collapsedBroadcast;
+        $this->translate = $translateProvider->getTranslate();
+        $this->networkBreakdown = $broadcastNetworksHelper->getCollapsedBroadcastNetworkBreakdown($collapsedBroadcast);
+    }
+
+    public function getDateTimestamp(): string
+    {
+        return $this->collapsedBroadcast->getStartAt()->format(DATE_ATOM);
```
votemike: FYI, should also be able to do `return $this->collapsedBroadcast->getStartAt()->toAtomString();`

src/DsAmen/Organism/Programme/CollapsedBroadcastSubPresenters/CollapsedBroadcastDetailsPresenter.php
```diff
@@ -0,0 +1,136 @@
...
+                return $isInternational || $service->isInternational();
+            },
+            false
+        );
+    }
+
+    private function implodeNames(array $names): string
+    {
+        // Return the first five names followed by ' & $count more...' if there are more than 5 names
+        if (count($names) > 5) {
```
votemike: `count($names)` is done several times. You could pull it out in to a variable.

src/DsAmen/Organism/Programme/CollapsedBroadcastSubPresenters/CollapsedBroadcastDetailsPresenter.php
```diff
@@ -0,0 +1,136 @@
...
+        $names = [];
+        foreach ($services as $service) {
+            $names[] = $service->getShortName();
+        }
+
+        // Return all services with the relevant qualifier ('only' or 'except')
+        return $this->translate->translate($translation, ['%1' => $this->implodeNames($names)]);
+    }
+
+    public function getTime(): DateTimeImmutable
```
votemike: Does this not return a Chronos?

src/DsAmen/Organism/Programme/CollapsedBroadcastSubPresenters/CollapsedBroadcastDetailsPresenter.php
```diff
@@ -0,0 +1,136 @@
...
+    {
+        return $this->collapsedBroadcast->getStartAt();
+    }
+
+    public function isInternational(): bool
+    {
+        return array_reduce(
```
votemike: Would this not be more efficient as a foreach which returns as soon as `$service->isInternational()` is true?

src/DsAmen/Organism/Programme/CollapsedBroadcastSubPresenters/CollapsedBroadcastDetailsPresenter.php
tatitati: Actually not sure if is used this $serviceNames.....

src/DsShared/Helpers/LocalisedDaysAndMonthsHelper.php
```diff
@@ -62,4 +65,55 @@ public function localisedDaysAndMonths(): string
 
         return json_encode($localisedDaysAndMonths);
     }
+
+    public function getFormatedDay(Chronos $date)
```
votemike: Typehint the return type?

src/DsShared/ValueObject/CollapsedBroadcastNetworkAndServiceBreakdown.php
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    /**  @param Service[] $notBroadcastOnServices */
+    public function setNotBroadcastOnServices(array $notBroadcastOnServices)
+    {
+        $this->notBroadcastOnServices = $notBroadcastOnServices;
+    }
+
+    /** @param Network[] $networks */
+    public function setNetworks(array $networks)
```
votemike: From `addNetwork()` it seems this needs specific keys for the array. Should that be mentioned somewhere on this method or the property?

src/DsAmen/Organism/Programme/CollapsedBroadcastSubPresenters/CollapsedBroadcastDetailsPresenter.php
```diff
@@ -0,0 +1,136 @@
...
+            $this->collapsedBroadcast->getServices(),
+            function (bool $isInternational, Service $service) {
+                return $isInternational || $service->isInternational();
+            },
+            false
+        );
+    }
+
+    private function implodeNames(array $names): string
```
votemike: This method is slightly hard to read. Do you think it would be worth refactoring at all?

votemike: Maybe something like:
```
private function implodeNames(array $names): string
{
    $namesCount = count($names);
    if ($namesCount === 2) {
        return implode($names, ' & ');
    }

    if ($names <= 5) {
        return implode($names, ', ');
    }

    // Return the first five names followed by ' & $count more...' if there are more than 5 names
    $xMore = $this->translate->translate('x_more', ['%1' => $namesCount - 5]);
    return implode(array_slice($names, 0, 5), ', ') . ' & ' . $xMore . '...';
}
```
What do you think?

src/DsAmen/Organism/Programme/CollapsedBroadcastSubPresenters/CollapsedBroadcastDetailsPresenter.php
```diff
@@ -0,0 +1,136 @@
...
+    {
+        // Return the first five names followed by ' & $count more...' if there are more than 5 names
+        if (count($names) > 5) {
+            return implode(array_slice($names, 0, 5), ', ') .
```
votemike: So this will be "a, b, c, d, e". Is that right? It shouldn't be "a, b, c, d & e"?

felipeparaujo: That's not what v2 does, I'm going to keep in line with its logic

src/DsShared/ValueObject/CollapsedBroadcastNetworkAndServiceBreakdown.php
felipeparaujo: I added a comment after the class declaration. It's not really worth doing much more than that.

src/DsAmen/Organism/Programme/CollapsedBroadcastSubPresenters/collapsed_broadcast_details.html.twig
```diff
@@ -0,0 +1,29 @@
...
+            {{- collapsed_broadcast_details.getTime()|local_date('H:i') -}}
+            {%- if collapsed_broadcast_details.isInternational() -%}
+                <span class="timezonetimezone--note">{{- tr('UTC') -}}</span>
+            {%- endif -%}
+        </span>
+
+        {# Service details #}
+        <div class="broadcast-details__service">
+            <span class="gel-minion-bold" property="name">
+                {{- collapsed_broadcast_details.getNetworksNames() -}}
```
BjornTwachtmann: How will this look if we have multiple networks with service exceptions? I think it could break

PR #207, opened 2017-09-25
Only show two related links in promos on tlec page

BPScott:
PromotionPresenter option 'show_related_links' is replaced by
'related_links_count' which takes an integer.

See PROGRAMMES-5820

### DIFF
--- a/app/Resources/views/find_by_pid/tlec.html.twig
+++ b/app/Resources/views/find_by_pid/tlec.html.twig
@@ -34,6 +34,7 @@
                         {% set promo_widths = loop.first and has_double_width_first_promo ? '2/3@gel3b 1/2@gel5' : '1/2@gel3 1/3@gel3b 1/4@gel5' %}
                         <li class="grid__item {{ promo_widths }}">
                             {{- ds_amen('promotion', promotion, {
+                                'related_links_count': 2,
                                 'media_variant': 'media--column media--card',
                                 'cta_class': 'br-box-secondary',
                                 'branding_name': 'secondary',
--- a/src/DsAmen/Organism/Promotion/PromotionPresenter.php
+++ b/src/DsAmen/Organism/Promotion/PromotionPresenter.php
@@ -32,7 +32,7 @@ class PromotionPresenter extends Presenter
         // display options
         'show_image' => true,
         'show_synopsis' => true,
-        'show_related_links' => true,
+        'related_links_count' => 999,
 
         // classes & elements
         'h_tag' => 'h4',
@@ -122,11 +122,7 @@ class PromotionPresenter extends Presenter
      */
     public function getRelatedLinks(): array
     {
-        if (!$this->options['show_related_links']) {
-            return [];
-        }
-
-        return $this->promotion->getRelatedLinks();
+        return array_slice($this->promotion->getRelatedLinks(), 0, $this->options['related_links_count']);
     }
 
     /**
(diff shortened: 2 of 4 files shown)


programmes-developers:

src/DsAmen/Organism/Promotion/PromotionPresenter.php
```diff
@@ -122,11 +122,11 @@ public function getSynopsis(): string
      */
     public function getRelatedLinks(): array
     {
-        if (!$this->options['show_related_links']) {
+        if (!$this->options['related_links_count']) {
```
votemike: Do you think this would be clearer with an `== 0` now instead?

src/DsAmen/Organism/Promotion/PromotionPresenter.php
```diff
@@ -175,8 +175,8 @@ protected function validateOptions(array $options): void
             throw new InvalidOptionException('show_synopsis option must be a boolean');
         }
 
-        if (!is_bool($options['show_related_links'])) {
-            throw new InvalidOptionException('show_related_links option must be a boolean');
+        if (!is_int($options['related_links_count'])) {
```
votemike: Worth checking that it's positive?

tests/DsAmen/Organism/Promotion/PromotionPresenterTest.php
```diff
@@ -134,8 +134,8 @@ public function testDisabledOptions()
 
         $presenter = new PromotionPresenter($this->mockRouter, $promotion, [
             'show_synopsis' => false,
-            'show_related_links' => false,
             'show_image' => false,
+            'related_links_count' => 0,
```
votemike: And a new test to make sure it limits to the selected number if more links exist than you want to display?

app/Resources/views/find_by_pid/tlec.html.twig
```diff
@@ -34,6 +34,7 @@
                         {% set promo_widths = loop.first and has_double_width_first_promo ? '2/3@gel3b 1/2@gel5' : '1/2@gel3 1/3@gel3b 1/4@gel5' %}
                         <li class="grid__item {{ promo_widths }}">
                             {{- ds_amen('promotion', promotion, {
+                                'show_related_links': 2,
```
felipeparaujo: it feels a bit weird having this named show_related_links and it not being a bool. I don't a have a better name for this though, so if you don't want to change it, w.e.

tests/DsAmen/Organism/Promotion/PromotionPresenterTest.php
BPScott: There's an `$this->assertSame([], $presenter->getRelatedLinks());` assertion in this test, is that enough?

app/Resources/views/find_by_pid/tlec.html.twig
BPScott: Whoops, I made that change but missed adding it in this file. The property is now related_links_count.

src/DsAmen/Organism/Promotion/PromotionPresenter.php
BPScott: This whole check could be removed as  array_slice($bigArray, 0, 0) returns [] anyway.

votemike: But this would be less expensive, right? Is it worth keeping in?

tests/DsAmen/Organism/Promotion/PromotionPresenterTest.php
votemike: I mean if there are 4 items, but you only want to display 2.

src/DsAmen/Organism/Promotion/PromotionPresenter.php
BPScott: Added

BPScott: addded

tests/DsAmen/Organism/Promotion/PromotionPresenterTest.php
BPScott: added

PR #212, opened 2017-09-29
Fix prioritisation for cache flush subscriber so FindByPid also gets flushed

BjornTwachtmann:


programmes-developers:

PR #213, opened 2017-10-02
Programmes-5882-onnow

mattgreenham:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5882

Added SchedulesOnNowController


app/Resources/views/schedules/on_now_legacy_2013.html.twig
```diff
@@ -0,0 +1,23 @@
+<h2 class="text--shout centi no-margin">
+    {% if simulcastUrl %}
+        <a href="{{ simulcastUrl }}">{{ tr(isRadio ? "iplayer_listen_live" : "iplayer_watch_live") }}</a>
+    {% else %}
+        <span>{{ tr(isRadio ? "on_air" : "on_now") }}</span>
+    {% endif %}
+</h2>
+
+<h3 class="centi">{{ broadcast.getStartAt()|date('H:i') }}</h3>
```
votemike: Does this work for world service radio?

src/Controller/SchedulesOnNowController.php
```diff
@@ -0,0 +1,49 @@
...
+        CollapsedBroadcastsService $collapsedBroadcastsService,
+        NetworksService $networksService,
+        HelperFactory $helperFactory,
+        Request $request,
+        string $networkKey
+    ) {
+        $network = $networksService->findByUrlKeyWithDefaultService($networkKey);
+
+        if (!$network || !$network->getDefaultService()) {
+            throw $this->createNotFoundException('No network or service found.');
```
votemike: Do we want to mention the network key in the exception?

src/Controller/SchedulesOnNowController.php
```diff
@@ -0,0 +1,49 @@
...
+use App\DsShared\Helpers\HelperFactory;
+use BBC\ProgrammesPagesService\Data;
+use BBC\ProgrammesPagesService\Service\BroadcastsService;
+use BBC\ProgrammesPagesService\Service\CollapsedBroadcastsService;
+use BBC\ProgrammesPagesService\Service\NetworksService;
+use Symfony\Component\HttpFoundation\Request;
+
+class SchedulesOnNowController extends BaseController
+{
+
```
votemike: Why did the style linter not pick up on this blank line?

src/Controller/SchedulesOnNowController.php
```diff
@@ -0,0 +1,49 @@
...
+        if (!$network || !$network->getDefaultService()) {
+            throw $this->createNotFoundException('No network or service found.');
+        }
+
+        $broadcast = $broadcastsService->findOnNowByService($network->getDefaultService());
+        if (!$broadcast) {
+            throw $this->createNotFoundException('No broadcasts found.');
+        }
+
+        $designSystem = $request->query->get('partial');
```
votemike: Do we want this to be in the query string rather than as part of the main URL?

tests/DataFixtures/ORM/CollapsedBroadcastsFixture.php
```diff
@@ -0,0 +1,51 @@
...
+        $broadcast = $this->getReference('p5100002');
+
+        $collapsedBroadcast = new CollapsedBroadcast(
+            $broadcast->getProgrammeItem(),
+            (string) $broadcast->getId(),
+            (string) $broadcast->getService()->getId(),
+            '0',
+            new DateTime('2017-05-22 02:00:00'),
+            new DateTime('2017-05-22 02:45:00')
+            // NOTE: These are copy/pasted from BrodcastsFixture due to issues with UtcDateTimeType
```
votemike: BroadcastsFixture instead of BrodcastsFixture?

votemike:
Not a fan on the ticket itself. Have left a comment. https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5882?focusedCommentId=3594532&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-3594532

src/Controller/SchedulesOnNowController.php
```diff
@@ -0,0 +1,49 @@
+<?php
+declare(strict_types = 1);
+
+namespace App\Controller;
+
+use App\DsShared\Helpers\HelperFactory;
+use BBC\ProgrammesPagesService\Data;
```
votemike: What's this?

app/config/routing.yml
```diff
@@ -51,6 +51,10 @@ schedules_by_year:
     defaults: { _controller: App\Controller\SchedulesByYearController }
     requirements: { pid: '[0-9b-df-hj-np-tv-z]{8,15}', year: '\d{4}' }
 
+schedules_on_now:
+    path: /schedules/network/{networkKey}/on-now
+    defaults: { _controller: App\Controller\SchedulesOnNowController }
```
votemike: It feels like this could be part of the `SchedulesVanityRedirectController`

programmes-developers:

app/Resources/views/schedules/on_now_legacy_2013.html.twig
BjornTwachtmann: I believe what Mike is trying to say here is that date('H:i') should be local_date('H:i') in order to run our custom code to output this date in the correct Timezone

app/config/routing.yml
BjornTwachtmann: It would require duplicating the logic to do that, which I don't think is worth it. Leave it here.

src/Controller/SchedulesOnNowController.php
BjornTwachtmann: I'm not a fan of this either really, but Ben made the decision to do this this way, and it's the same for the other partial views with the "no_chrome" parameter so we might as well be consistent.

BjornTwachtmann:
I think another thing that's worth doing here is to instruct varnish to cache this page until the end of the currently on broadcast, I will explain this to you...

src/Controller/SchedulesOnNowController.php
votemike: I think that's a slightly different case. That renders the same page/HTML but only the HTML from the middle of the page.  
This is being used as both a vanity redirect and as a way to retrieve HTML partials.

app/Resources/views/schedules/on_now_legacy_2013.html.twig
votemike: Or possibly that the timezone needs to be set as UTC for international services.

src/Controller/SchedulesOnNowController.php
```diff
@@ -0,0 +1,69 @@
...
+        CollapsedBroadcastsService $collapsedBroadcastsService,
+        NetworksService $networksService,
+        HelperFactory $helperFactory,
+        Request $request,
+        string $networkKey
+    ) {
+        $network = $networksService->findByUrlKeyWithDefaultService($networkKey);
+
+        if (!$network || !$network->getDefaultService()) {
+            throw $this->createNotFoundException('No network or service found from network key.');
```
votemike: I was thinking more something like `'No network or service found from network key ' . $networkKey . '.'`

src/Controller/SchedulesOnNowController.php
```diff
@@ -0,0 +1,69 @@
...
+
+        $designSystem = $request->query->get('partial');
+        if (!in_array($designSystem, ['legacy_2013', 'legacy_amen'])) {
+            return $this->redirectToRoute('find_by_pid', ['pid' => (string) $broadcast->getProgrammeItem()->getPid()]);
+        }
+
+        $collapsedBroadcast = $collapsedBroadcastsService->findByBroadcast($broadcast);
+        $simulcastUrl = $helperFactory->getLiveBroadcastHelper()->simulcastUrl($collapsedBroadcast);
+
+        $secondsUntilBroadcastEndAt = $broadcast->getEndAt()->getTimestamp() - ApplicationTime::getLocalTime()->getTimestamp();
```
votemike: $should be able to do something like `$secondsUntilBroadcastEndAt = $broadcast->getEndAt()->diffInSeconds(ApplicationTime::getLocalTime())`. Or it might be the other way around.

src/Controller/SchedulesOnNowController.php
```diff
@@ -0,0 +1,69 @@
...
+
+        // Cache for 10 minutes, unless the broadcast finishes sooner
+        // Minimum cache life of 15 seconds
+        if ($secondsUntilBroadcastEndAt < 15) {
```
votemike: Maybe move the logic of this in to a private function. Along with calculating the `$secondsUntilBroadcastEndAt`?
So.... 
```
$cacheLifetime = $this->calculateCacheLifetime($broadcast->getEndAt());
$this->response()->setPublic()->setMaxAge($cacheLifetime);
```

app/Resources/views/schedules/on_now_legacy_2013.html.twig
```diff
@@ -0,0 +1,23 @@
...
+        <span>{{ tr(isRadio ? "on_air" : "on_now") }}</span>
+    {% endif %}
+</h2>
+
+<h3 class="centi">{{ broadcast.getStartAt()|local_date('H:i') }} {{ broadcast.getStartAt()|time_zone_note()}}</h3>
+
+<div class="programme block-link programme--grid">
+    <div class="programme__img one-whole programme__img--hasimage">
+        <div class="programme__img-box">
+            <div class="rsp-img lazy-img" data-img-blank="{{ asset('images/blank.png') }}" data-img-src-68="{{ broadcast.getProgrammeItem().getImage().getUrl(160, 90) }}"></div>
```
votemike: Is `data-img-src-68` a thing?

votemike: Is there not already a ds2013 molecule or something that displays a responsive image for a programme?

app/Resources/views/schedules/on_now_legacy_2013.html.twig
```diff
@@ -0,0 +1,23 @@
...
+        </div>
+    </div>
+    <div class="programme__body programme__body--flush">
+        <h4 class="programme__titles">
+            <a href="{{ path('find_by_pid', {'pid': broadcast.getProgrammeItem().getPid()}) }}" class="br-blocklink__link"><span class="programme__title">{{ broadcast.getProgrammeItem().getTleo().getTitle() }}</span></a>
+        </h4>
+    </div>
+    <a class="block-link__overlay-link" href="{{ path('find_by_pid', {'pid': broadcast.getProgrammeItem().getPid()}) }}" aria-hidden="true" tabindex="-1">View Programme information</a>
```
votemike: Should "View Programme information" be translatable instead of hardcoded?

app/Resources/views/schedules/on_now_legacy_amen.html.twig
```diff
@@ -0,0 +1,31 @@
...
+                </a>
+            </h4>
+        </div>
+    </div>
+    <div class="media__panel 1/1">
+        <div class="media__overlay-container ratiobox">
+            <img data-molecule="responsive-image"
+                 class="media__image"
+                 src="{{ broadcast.getProgrammeItem().getImage().getUrl(320, 180) }}"
+                 alt="" />
```
votemike: Should we add alt tags in? Maybe the programme title?

programmes-developers:

app/Resources/views/schedules/on_now_legacy_2013.html.twig
```diff
@@ -0,0 +1,27 @@
...
+    {% endif %}
+</h2>
+
+<h3 class="centi">{{ broadcast.getStartAt()|local_date('H:i') }} {{ broadcast.getStartAt()|time_zone_note() }}</h3>
+
+<div class="programme block-link programme--grid">
+    <div class="programme__img one-whole programme__img--hasimage">
+        <meta property="image" content="{{ programmeImage.getUrl(480,270) }}"/>
+        <div class="programme__img-box">
+            {{ ds_shared('image', programmeImage, 160, {770:'160px'}) }}
```
mattgreenham: @BjornTwachtmann @felipeparaujo is this correct use of ds_shared?

app/Resources/views/schedules/on_now_legacy_2013.html.twig
```diff
@@ -0,0 +1,27 @@
...
+    {% else %}
+        <span>{{ tr(isRadio ? "on_air" : "on_now") }}</span>
+    {% endif %}
+</h2>
+
+<h3 class="centi">{{ broadcast.getStartAt()|local_date('H:i') }} {{ broadcast.getStartAt()|time_zone_note() }}</h3>
+
+<div class="programme block-link programme--grid">
+    <div class="programme__img one-whole programme__img--hasimage">
+        <meta property="image" content="{{ broadcast.getProgrammeItem().getImage.getUrl(480,270) }}"/>
```
votemike: `getImage()` not `getImage`

tests/Controller/SchedulesOnNowControllerTest.php
```diff
@@ -0,0 +1,77 @@
...
+
+        $this->assertContains('Watch live', $crawler->filter('h2')->text());
+        $this->assertContains('03:00', $crawler->filter('h3')->text());
+        $this->assertContains('Early Episode', $crawler->filter('.programme__title')->text());
+        $this->assertEquals('/programmes/p3000002', $crawler->filter('.br-blocklink__link')->attr('href'));
+    }
+
+    public function testOnNowRedirectsWithNoPartial()
+    {
+        ApplicationTime::setTime((new Chronos('2017-05-21 14:01:00'))->getTimestamp());
```
tatitati: just wondering, why this dates?

BjornTwachtmann:
Looks fine to me. Might be worth looking at the URL structure again at some point, but I think it's fine for now. Give me a shout if you need help with the merge conflicts.

tests/Controller/SchedulesOnNowControllerTest.php
tatitati: Still not clear for me. Is redirecting because the date is in the past of what is set in the fixtures?

mattgreenham: When there is no partial in the URL i.e. a user goes to /schedules/network/bbcone/on-now it will redirect them to the programme page of the programme currently being broadcast. This time is just during a broadcast in the fixtures

PR #214, opened 2017-10-03
Programmes 5806 scenarios and fixtures

BjornTwachtmann:


programmes-developers:

src/Fixture/Doctrine/Entity/HttpFixture.php
```diff
@@ -0,0 +1,127 @@
...
+    {
+        return $this->envAgnosticUrl;
+    }
+
+    public function getResponseCode(): int
+    {
+        return $this->responseCode;
+    }
+
+    public function setResponseCode(int $responseCode)
```
Tasiobg: Not sure if is a good practice but you could set the return type to void
`public function setResponseCode(int $responseCode): void`
https://wiki.php.net/rfc/void_return_type

BjornTwachtmann: Probably a good idea, I'll set that...

PR #215, opened 2017-10-04
Added some comments.

tatitati:
Added some comments to explain how they look in cloudwatch AWS (I think is easy to understand what is being sent with this little comment and how Metrics work). 

In this way you don't need to go to AWS cloudwatch, search the metrics, and see how they look.


programmes-developers:

PR #216, opened 2017-10-04
PROGRAMMES-5896 Schedules network redirect

tatitati:
PROGRAMMES-5896 
Schedules network redirect


programmes-developers:

app/config/routing.yml
```diff
@@ -51,6 +51,12 @@ schedules_by_year:
     defaults: { _controller: App\Controller\SchedulesByYearController }
     requirements: { pid: '[0-9b-df-hj-np-tv-z]{8,15}', year: '\d{4}' }
 
+# Redirect: /schedules/network/<network_url_key>   --->   /schedules/<default service pid in network>
+schedules_by_network_url_key:
+    path: /schedules/network/{networkUrlKey}
+    defaults: { _controller: App\Controller\SchedulesByNetworkUrlKeyController }
+    requirements: { networkUrlKey: '[a-zA-Z]{2,35}' }
```
tatitati: The longest is actually 26, but I allow 35 (not sure if is good to put a very rigid limit on this)

felipeparaujo: Probably don't even need an upper limit there, but it's alright to keep it as 35 I think.

src/Controller/SchedulesByNetworkUrlKeyController.php
```diff
@@ -0,0 +1,136 @@
...
+
+        return $this->redirectToRoute('schedules_by_day', ['pid' => $defaultServicePid], 301);
+    }
+
+    /**
+     * @return string[]
+     */
+    public static function getNetworkDefaultServiceMap(): array
+    {
+        return array(
```
felipeparaujo: use `[]` instead of `array()`

src/Controller/SchedulesByNetworkUrlKeyController.php
```diff
@@ -0,0 +1,136 @@
+<?php
+namespace App\Controller;
+
+class SchedulesByNetworkUrlKeyController extends BaseController
+{
+    public function __invoke(string $networkUrlKey)
+    {
+        $defaultServicePid = $this->getDefaultServicePidInNetwork($networkUrlKey);
+
+        return $this->redirectToRoute('schedules_by_day', ['pid' => $defaultServicePid], 301);
```
votemike: What if `$defaultServicePid` is null?

src/Controller/SchedulesByNetworkUrlKeyController.php
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    /**
+     * @return string[]
+     */
+    public static function getNetworkDefaultServiceMap(): array
```
votemike: Is there not a way to automatically generate this from dbinfo?

tests/Controller/SchedulesByNetworkUrlKeyControllerTest.php
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * @covers App\Controller\SchedulesByNetworkUrlKeyController
+ */
+class SchedulesByNetworkUrlKeyControllerTest extends BaseWebTestCase
+{
+    private $client;
+
+    public function testCanRedirectFromNetworkUrlKeyToDefaultPidService()
+    {
+        $this->userRequestUrlUsingNetworkUrlKey('/schedules/network/bbcone');
```
votemike: Is it worth splitting these out in to two private methods? Not fussed. Just wondered how come you chose to do so.

src/Controller/SchedulesByNetworkUrlKeyController.php
votemike: Does the response need to be cached?

src/Controller/SchedulesByNetworkUrlKeyController.php
votemike: Does this need to be static?

tests/Controller/SchedulesByNetworkUrlKeyControllerTest.php
tatitati: just trying new approaches. But I think is very expressive these tests.
```php
$this->userRequestUrlUsingNetworkUrlKey('/schedules/network/bbcone');
$this->userSuffer301RedirectToDefaultServicePid('/schedules/p00fzl6p');
```

No needs to dig into the code at all. Is explaining the behaviour really nice I think

src/Controller/SchedulesByNetworkUrlKeyController.php
tatitati: no. We are changing the code by using database, jeje This will be changed

tatitati: yeah!, the big question. We are moving to that approach. 👍

src/Controller/SchedulesByNetworkUrlKeyController.php
```diff
@@ -0,0 +1,18 @@
...
+use BBC\ProgrammesPagesService\Service\NetworksService;
+
+class SchedulesByNetworkUrlKeyController extends BaseController
+{
+    public function __invoke(NetworksService $networkService, $networkUrlKey)
+    {
+        $network = $networkService->findByUrlKeyWithDefaultService($networkUrlKey);
+
+        if (is_null($network)) {
+            throw $this->createNotFoundException('Invalid network supplied');
```
felipeparaujo: This message should be `Network not found`

PR #221, opened 2017-10-05
PROGRAMMES-5737 on demand column

votemike:
PR for https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5737
That ticket has all the ACs etc... on it.


src/DsAmen/Organism/Map/SubPresenter/OnDemandPresenter.php
```diff
@@ -4,38 +4,132 @@
...
+            return $this->upcomingEpisode;
+        }
+
+        return $this->streamableEpisode;
+    }
+
+    /**
+     * @return mixed[]
+     */
+    public function getImageOptions(): array
```
votemike: I'm going to copy what Felipe has done once his PR is merged.

tests/DsAmen/Organism/Map/SubPresenter/OnDemandPresenterTest.php
```diff
@@ -0,0 +1,149 @@
...
+
+    public function invalidOptionProvider(): array
+    {
+        return [
+            'invalid-show_mini_map' => [['show_mini_map' => 'bar', 'show_synopsis' => false], 'show_mini_map option must be a boolean'],
+            'invalid-full_width' => [['show_mini_map' => true, 'full_width' => 'baz'], 'full_width option must be a boolean'],
+        ];
+    }
+
+    //@TODO make getTrailingLinkPresenter test
```
votemike: Or maybe not. Is that the right thing to do?

src/DsAmen/Organism/Map/SubPresenter/on_demand.html.twig
```diff
@@ -1,5 +1,43 @@
 <div class="grid__item map__item {{ on_demand.getClass() }}">
     <div class="map-card-wrap br-box-secondary" data-map-column="on_demand2016">
-        <h1>I Я On Demand for {{ on_demand.programme.title }}</h1>
+        {%- if not on_demand.getEpisode() -%}
+            <div class="island">
+                {%- if on_demand.getProgramme().isRadio() -%}
+                    <h2 class="map__heading gel-double-pica-bold">{{ tr('on_demand') }}</h2>
+                    {# @TODO Is this even needed? There's no translation for not_available_iplayer_radio #}
```
votemike: `not_available_iplayer_radio` is generated from https://github.com/bbc/programmes/blob/636e751443c3d6504282fd1d5c03eb202a235c61/webapp/php/lib/BBC/Programmes/View/Map/OnDemand.php#L45 but the string doesn't seem to exist. What should I do? Just use `not_available_iplayer` all the time?

src/DsAmen/Organism/Map/SubPresenter/OnDemandPresenter.php
```diff
@@ -4,38 +4,132 @@
...
+        return $this->programmeContainer;
+    }
+
+    public function getTrailingLinkPresenter(): TrailingLinkPresenter
+    {
+        return new TrailingLinkPresenter(
+            $this->programmeContainer,
+            $this->translateProvider,
+            $this->programmeContainer->isRadio() ? 'map_ondemand_all' : 'map_iplayer_all',
+            $this->programmeContainer->isRadio() ? 'all_available_episodes_radio' : 'all_available_episodes', // @TODO Is this even needed? There's no translation for all_available_episodes_radio
```
votemike: `all_available_episodes_radio` is generate from https://github.com/bbc/programmes/blob/daaec583d3f7e9de6af2a3c3d0c0525f9086e76a/webapp/php/sharedmodules/programmes/views/programmes/map/on_demand2016.spv#L59 but the string doesn't seem to exist. What should I do? Just use `all_available_episodes` all the time?

src/DsAmen/Organism/Map/SubPresenter/OnDemandPresenter.php
```diff
@@ -4,38 +4,130 @@
...
+        return $this->programmeContainer;
+    }
+
+    public function getTrailingLinkPresenter(): TrailingLinkPresenter
+    {
+        return new TrailingLinkPresenter(
+            $this->programmeContainer,
+            $this->translateProvider,
+            $this->programmeContainer->isRadio() ? 'map_ondemand_all' : 'map_iplayer_all',
+            $this->programmeContainer->isRadio() ? 'all_episodes_iplayer_radio' : 'all_available_episodes',
```
votemike: Do we want `all_available_episodes` or `all_episodes_iplayer`?

src/DsAmen/Organism/Map/SubPresenter/on_demand.html.twig
```diff
@@ -1,5 +1,42 @@
...
     <div class="map-card-wrap br-box-secondary" data-map-column="on_demand2016">
-        <h1>I Я On Demand for {{ on_demand.programme.title }}</h1>
+        {%- if not on_demand.getEpisode() -%}
+            <div class="island">
+                {%- if on_demand.getProgramme().isRadio() -%}
+                    <h2 class="map__heading gel-double-pica-bold">{{ tr('on_demand') }}</h2>
+                    <p>{{ tr('programme_availability_none_radio') }}</p>
+                {%- else -%}
+                    <h2 class="map__heading gel-double-pica-bold">{{ tr('available_on_iplayer_short') }}</h2>
+                    <p>{{ tr('not_available_iplayer') }}</p>
```
votemike: Do we want `not_available_iplayer` or `programme_availability_none`?

src/DsAmen/Organism/Map/SubPresenter/on_demand.html.twig
```diff
@@ -1,5 +1,59 @@
...
+
+            {%- if on_demand.getProgramme().getAvailableEpisodesCount() > 0 and not on_demand.showMiniMap() -%}
+                <div class="map__additional-info gel-pica br-keyline media__meta-row--separator">
+                    <p>
+                        <a
+                            data-linktrack="{{ on_demand.getAllLinkLocation() }}"
+                            {%- if on_demand.getProgramme().isRadio() -%}
+                                title="{{ tr('all_episodes_iplayer_radio', {'%1': on_demand.getProgramme().getTitle()}) }}"
+                            {%- else -%}
+                                title="{{ tr('all_available_episodes', {'%1': on_demand.getProgramme().getTitle()}) }}"
```
votemike: Do we want `all_available_episodes` or `all_episodes_iplayer`?

tests/DsAmen/Organism/Map/SubPresenter/OnDemandPresenterTest.php
```diff
@@ -0,0 +1,161 @@
...
+use BBC\ProgrammesPagesService\Domain\Entity\Series;
+use Cake\Chronos\Chronos;
+use PHPUnit\Framework\TestCase;
+
+class OnDemandPresenterTest extends TestCase
+{
+    /**
+     * @return bool[][]
+     */
+    public function trueFalseDataProvider(): array
```
tatitati: Just thinking, maybe is worth to change this name to ```isRadioProvider()``` ?
Is obvious that is providing a true/false value. But not too clear what is the point of it.

votemike: My thinking was that we could make this a generic provider in a base test class at some point when we need to test both true and false.

src/DsAmen/Organism/Map/SubPresenter/on_demand.html.twig
felipeparaujo: progamme_availability_none and programme_availability_none_radio seems to fit better

src/DsAmen/Organism/Map/SubPresenter/on_demand.html.twig
felipeparaujo: i would go with all_episodes_iplayer and all_episodes_iplayer_radio

tests/DsAmen/Organism/Map/SubPresenter/OnDemandPresenterTest.php
```diff
@@ -0,0 +1,174 @@
...
+    {
+        $episode = $this->createEpisode();
+        $programme = $this->createProgramme(true);
+        $collapsedBroadcast = $this->createMock(CollapsedBroadcast::class);
+        $collapsedBroadcast->method('getStartAt')->willReturn(new Chronos('-8 days'));
+        $odPresenter = new OnDemandPresenter($programme, null, $episode, $collapsedBroadcast);
+        $this->assertTrue($odPresenter->justMissed());
+    }
+
+    public function testNotJustMissed()
```
tatitati: I feel this test is trying to cover too much, seems that is trying to test different scenarios all in once.

votemike: I was wondering if I need the test at all. It seems a bit odd to check that something _isn't_ something.

tests/DsAmen/Organism/Map/SubPresenter/OnDemandPresenterTest.php
tatitati: yeah, that might be worth

tests/DsAmen/Organism/Map/SubPresenter/OnDemandPresenterTest.php
tatitati: XD , no idea. To be honest I didn't read the test at all. I just saw that was too big (visually) and nothing else. But yeah, maybe is a good point what you said 👍

src/Controller/FindByPid/TlecController.php
```diff
@@ -54,6 +54,12 @@ public function __invoke(
             $clips = $aggregationService->findDescendantClips($programme, 4);
         }
 
+        $upcomingEpisodes = null;
```
BjornTwachtmann: Might be worth wrapping these queries in an if ($programme->getAggregatedEpisodesCount() > 0), for ever so slight savings on brands with no episodes.

src/DsAmen/Organism/Map/SubPresenter/on_demand.html.twig
```diff
@@ -1,5 +1,74 @@
...
+            <h2 class="gel-double-pica-bold">{{ tr('on_demand') }}</h2>
+        {%- else -%}
+            <h2 class="gel-double-pica-bold">{{ tr('available_on_iplayer_short') }}</h2>
+        {%- endif -%}
+    </div>
+    {{ ds_amen('programme', episode, {
+        'context_programme': presenter.getProgramme(),
+        'media_variant': 'media media--column media--card map-card',
+        'branding_name': 'secondary',
+        'force_iplayer_linking': true,
```
BjornTwachtmann: That's not going to be correct for radio brands

src/DsAmen/Organism/Map/SubPresenter/on_demand.html.twig
```diff
@@ -1,5 +1,74 @@
+{% import _self as self %}
+
 <div class="grid__item map__item {{ on_demand.getClass() }}">
     <div class="map-card-wrap br-box-secondary" data-map-column="on_demand2016">
-        <h1>I Я On Demand for {{ on_demand.programme.title }}</h1>
+        {%- if not on_demand.getStreamableEpisode() and not on_demand.getUpcomingEpisode() -%}
+            {%- if on_demand.getProgramme().isRadio() -%}
+                {{ self.message('on_demand', 'programme_availability_none_radio') }}
+            {%- else -%}
+                {{ self.message('available_on_iplayer_short', 'progamme_availability_none') }}
```
BjornTwachtmann: progamme_availability_none doesn't exist, should be not_available_iplayer

BjornTwachtmann: I think

src/DsAmen/Organism/Map/SubPresenter/on_demand.html.twig
```diff
@@ -1,5 +1,74 @@
...
+        {%- endif -%}
+    </div>
+    {{ ds_amen('programme', episode, {
+        'context_programme': presenter.getProgramme(),
+        'media_variant': 'media media--column media--card map-card',
+        'branding_name': 'secondary',
+        'force_iplayer_linking': true,
+        'show_image': not presenter.showMiniMap(),
+        'title_options': {
+            'text_colour_on_title_link': true,
```
BjornTwachtmann: Why is this here? Looks like the wrong colouring to me compared with live, or am I missing something?

src/DsAmen/Organism/Map/SubPresenter/on_demand.html.twig
votemike: We hadn't decided to do the radio page by the time I coded this part. And I forgot to re-visit it.
What do we want the link location prefix to be for radio?

votemike: Ignore me. It turns out that this option affects multiple things. I was using it for something, but that's no longer needed.

votemike: Nope, actually we do need it. Maybe I could have some help with it?

tests/DsAmen/Organism/Map/SubPresenter/OnDemandPresenterTest.php
felipeparaujo: I don't really see the value of genericizing this sort of data provider. If you have a dataprovider for each test that needs it, you can name each test instance more meaningfully instead of just having 'true' or 'false'

tests/DsAmen/Organism/Map/SubPresenter/OnDemandPresenterTest.php
felipeparaujo: It's good to have a test for the edge cases. I think this could stay here, it's already written anyway

tests/DsAmen/Organism/Map/SubPresenter/OnDemandPresenterTest.php
votemike: I used to often find myself needing true/false providers as part of unit tests. You can see what the value is symbolising due to the name e.g. `$isRadio` https://github.com/bbc/programmes-frontend/pull/221/files#diff-66fa891a4ffb2ad9dc7fcc41ad40d9cfR53  
I can change it if people are really opposed to it.

src/DsAmen/Organism/Map/SubPresenter/on_demand.html.twig
```diff
@@ -1,5 +1,75 @@
...
+
+{% macro main_column(episode, presenter) %}
+    <div class="island">
+        {%- if presenter.getProgramme().isRadio() -%}
+            <h2 class="gel-double-pica-bold">{{ tr('on_demand') }}</h2>
+        {%- else -%}
+            <h2 class="gel-double-pica-bold">{{ tr('available_on_iplayer_short') }}</h2>
+        {%- endif -%}
+    </div>
+    {# @TODO sort link_location_prefix for radio #}
```
votemike: This should be done in a future ticket surrounding link tracking.

tests/DsAmen/Organism/Map/SubPresenter/OnDemandPresenterTest.php
BPScott: Having the provider in the same class as the test is fine, however I'm not too convinced that moving it into a base class is a valuable thing - the increased cognitive overhead of having to go look and find out what it does seems like much more effort that duplicating a trivial array. I reckon tests should be self contained, you shouldn't need to look for something so core to a test in a different file.

PR #223, opened 2017-10-06
Changed comment

tatitati:
I realized that this comment was wrong, so I changed it.


programmes-developers:

PR #224, opened 2017-10-06
Programmes 5894 excludelangs

mattgreenham:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5894


programmes-developers:

src/Controller/HomeController.php
```diff
@@ -14,11 +14,38 @@ public function __invoke(EntityManagerInterface $em, NetworksService $networksSe
         $programmeCount = $this->countProgrammesAndGroups($em);
 
         $serviceTypes = ['TV', 'National Radio', 'Regional Radio', 'Local Radio'];
-        $networks = $networksService->findPublishedNetworksByType(
+        $publishedNetworks = $networksService->findPublishedNetworksByType(
             $serviceTypes,
             NetworksService::NO_LIMIT
         );
 
+        $blacklistedNetworks = [
```
BPScott: Nitpick: Using a private class constant for this might be nice:

tatitati:
Seems ok to me Matt! 👍

src/Controller/HomeController.php
```diff
@@ -14,11 +14,38 @@ public function __invoke(EntityManagerInterface $em, NetworksService $networksSe
...
+            'bbc_persian_radio', // Persian
+            'bbc_russian_radio', // Russian
+            'bbc_sinhala_radio', // Sinhala
+            'bbc_tamil_radio', // Tamil
+            'bbc_tigrinya_radio', // Tigrinya
+            'bbc_urdu_radio', // Urdu
+            'bbc_uzbek_radio', // Uzbek
+        ];
+
+        $networks = array_filter($publishedNetworks, function ($n) use ($blacklistedNetworks) {
```
votemike: You could probably typehint $n (and maybe $blacklistedNetworks). Not fussed though.

src/Controller/HomeController.php
```diff
@@ -14,11 +14,38 @@ public function __invoke(EntityManagerInterface $em, NetworksService $networksSe
...
+            return !in_array($n->getNid(), $blacklistedNetworks);
```
BjornTwachtmann: This is going to be quite inefficient. You're iterating over a large array, then for each element of that array iterating over a smaller array and doing a string comparison for every member. 

I would be tempted to rewrite this to use isset() against an array key rather than in_array. That compiles down to a single hash table lookup and should be much more efficient without sacrificing readability significantly.

BPScott: Ah cunning. This sort of trick?

```
$blacklistedNetworks = [
  'bbc_blah' => true,
  'bbc_foo' => true,
];

//... Then in the array_filter
return isset($blacklistedNetworks[(string) $n->getNid()];
```

Dang PHP for not having a proper standard Set implementation (no the Data Structures extension doesn't count, it's bad)

BjornTwachtmann: Exactly :)

src/Controller/SchedulesHomeController.php
```diff
@@ -9,6 +9,43 @@
 
 class SchedulesHomeController extends BaseController
 {
+    // https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5792 and
```
felipeparaujo: I know you just copied this over from the previous stuff, but how about replacing these links by a quick explanation as to why these are here?

PR #226, opened 2017-10-06
Fix redirect url key with incomplete pattern to match route.

tatitati:
We saw that some url_keys were not matching to any route. So urls like these were failing (no redirecting to /schedules/< pid >):

/schedules/network/**radio250s**
/schedules/network/**radio7**
/schedules/network/**radio1**
/schedules/network/**radio3**

This PR has:
- Fix for redirect routes '/schedules/network/<url_key>' when url_key has digits.
- Tests to cover the fix


programmes-developers:

PR #227, opened 2017-10-06
Use consistent naming for route variables

BPScott:
We've got two routes next to each other that use slightly different
labels for the same thing. Might as well make them consistant.

/cc @mattgreenham @tatitati

### DIFF
--- a/app/config/routing.yml
+++ b/app/config/routing.yml
@@ -69,8 +69,9 @@ schedules_by_network_url_key:
     requirements: { networkUrlKey: '[a-zA-Z0-9]{2,35}' }
 
 schedules_on_now:
-    path: /schedules/network/{networkKey}/on-now
+    path: /schedules/network/{networkUrlKey}/on-now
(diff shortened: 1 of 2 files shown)


programmes-developers:

app/config/routing.yml
```diff
@@ -69,8 +69,10 @@ schedules_by_network_url_key:
     requirements: { networkUrlKey: '[a-zA-Z0-9]{2,35}' }
 
 schedules_on_now:
-    path: /schedules/network/{networkKey}/on-now
+    path: /schedules/network/{networkUrlKey}/on-now
     defaults: { _controller: App\Controller\SchedulesOnNowController }
+    requirements: { networkUrlKey: '[a-zA-Z0-9]{2,35}' }
+
```
mattgreenham: Maybe nitpicking but there's an extra newline here

BPScott: Oooh go on then :)

PR #230, opened 2017-10-09
Add missing param

votemike:
Added in missing minimap param


programmes-developers:

PR #231, opened 2017-10-09
PROGRAMMES-5735 - Tx column bugfixes

felipeparaujo:


programmes-developers:

src/DsAmen/Organism/Map/SubPresenter/TxPresenter.php
```diff
@@ -79,8 +79,14 @@ public function __construct(
...
     {
-        // Radio brand pages and repeats don't have badges
-        if ($this->contextProgramme->isRadio() || $this->collapsedBroadcast->isRepeat()) {
+        $parent = $this->collapsedBroadcast->getProgrammeItem()->getParent();
+
+        // Radio brand pages, repeats and programmes that don't belong to series don't have badges
+        if ($this->contextProgramme->isRadio() ||
+            $this->collapsedBroadcast->isRepeat() ||
+            !$parent ||
+            $parent->getType() !== 'series'
```
BPScott: use `instanceof` instead of getType()

tests/DsShared/Helpers/BroadcastNetworksHelperTest.php
```diff
@@ -50,6 +51,19 @@ public function testGetNetworksAndServicesDetailsWithOneNetworkShowingOnlyServic
...
+    {
+        $network1Services = $this->createServicesWithNetwork(['sid1', 'sid2', 'sid3', 'sid4'], 'nid1');
+        $network1Services[0]->method('getEndDate')->willReturn(new Chronos('2017-01-01'));
+
+        $broadcast = $this->createCollapsedBroadcast(array_slice($network1Services, 1, 2), '2017-01-02');
+        $breakdown = $this->helper->getNetworksAndServicesDetails($broadcast);
+
+        $this->assertCount(1, $breakdown);
+        $this->assertEquals('Network nid1', implode(array_keys($breakdown)));
+        $this->assertEquals('except Service sid4', array_values($breakdown)[0]);
```
tatitati: Just wondering, but cannot we use simply ```$breakdown[0]``` ?

src/Controller/FindByPid/TlecController.php
```diff
@@ -60,7 +60,10 @@ public function __invoke(
 
         $upcomingBroadcasts = [];
         if ($programme->getAggregatedEpisodesCount()) {
-            $upcomingBroadcasts = $collapsedBroadcastsService->findUpcomingByProgrammeWithFullServicesOfNetworksList($programme);
+            $upcomingBroadcasts = $collapsedBroadcastsService->findUpcomingByProgrammeWithFullServicesOfNetworksList(
+                $programme,
+                100
```
votemike: Why this number?

src/DsAmen/Organism/Map/SubPresenter/TxPresenter.php
```diff
@@ -79,8 +79,14 @@ public function __construct(
...
     public function getBadgeTranslationString(): string
     {
-        // Radio brand pages and repeats don't have badges
-        if ($this->contextProgramme->isRadio() || $this->collapsedBroadcast->isRepeat()) {
+        $parent = $this->collapsedBroadcast->getProgrammeItem()->getParent();
+
+        // Radio brand pages, repeats and programmes that don't belong to series don't have badges
+        if ($this->contextProgramme->isRadio() ||
+            $this->collapsedBroadcast->isRepeat() ||
+            !$parent ||
```
votemike: Is this a place where you can use something like `$this->collapsedBroadcast->isTleo()`?

src/DsShared/Helpers/BroadcastNetworksHelper.php
```diff
@@ -111,37 +111,47 @@ public function buildServicesNames(array $networkBreakdowns): array
...
         }
 
         // Build a list of the services from a network where the broadcast didn't happen
-        foreach ($networksAndServices as $nid => $networkAndServices) {
-            $networksAndServices[$nid]['not_on_services'] = array_udiff(
-                $networkAndServices['network']->getServices(),
-                $networkAndServices['on_services'],
-                function (Service $a, Service $b) {
-                    return strcmp((string) $a->getSid(), (string) $b->getSid());
+        foreach ($breakdowns as $nid => $breakdown) {
```
votemike: ❤️  array of array of arrays

src/DsShared/Helpers/BroadcastNetworksHelper.php
```diff
@@ -111,37 +111,47 @@ public function buildServicesNames(array $networkBreakdowns): array
...
+            $breakdowns[$nid]['not_on_services'] = [];
+            // Loop over all of the network services
+            foreach ($breakdown['network']->getServices() as $networkService) {
+                // Loop over all of the broadcast services
+                foreach ($breakdown['on_services'] as $broadcastService) {
```
votemike: Maybe move each of these foreaches in to a well named method? So I can ready what's happening.

src/DsShared/Helpers/BroadcastNetworksHelper.php
```diff
@@ -111,37 +111,47 @@ public function buildServicesNames(array $networkBreakdowns): array
...
+                    // If the network service is in the broadcast service list, or if the broadcast started after
+                    // the end date of the service, don't add it to the list of services where the broadcast
+                    // didn't happen
+                    if ((string) $networkService->getSid() === (string) $broadcastService->getSid() ||
+                        ($networkService->getEndDate() && $collapsedBroadcast->getStartAt()->gt($networkService->getEndDate()))
+                    ) {
+                        continue 2;
```
votemike: 🤢

src/Controller/FindByPid/TlecController.php
felipeparaujo: [Because v2 limits to 100 as well](https://github.com/bbc/programmes/blob/develop/webapp/php/sharedlib/BBC/Programmes/ProgrammeContainer.php#L50)

BPScott: Iirc V2 handles collapsing many broadcasts into a single collapsed broadcast in code and this number is big because it requests many broadcasts, as were getting collapsed broadcasts from the db we can probably drop this substantially

votemike: Should we instead do 2 queries? One for repeats, one for debuts?

src/DsAmen/Organism/Map/SubPresenter/TxPresenter.php
```diff
@@ -79,8 +80,12 @@ public function __construct(
 
     public function getBadgeTranslationString(): string
     {
-        // Radio brand pages and repeats don't have badges
-        if ($this->contextProgramme->isRadio() || $this->collapsedBroadcast->isRepeat()) {
+        // Radio brand pages, repeats and programmes that don't belong to series don't have badges
+        if ($this->contextProgramme->isRadio() ||
+            $this->collapsedBroadcast->isRepeat() ||
+            $this->collapsedBroadcast->getProgrammeItem()->isTleo() ||
+            !$this->collapsedBroadcast->getProgrammeItem()->getParent() instanceof Series
```
BPScott: Thinking about this more (sorry):

`!$this->collapsedBroadcast->getProgrammeItem()->getParent() instanceof Series` can be replaced with `$this->collapsedBroadcast->getProgrammeItem()->getParent()->isTleo()`, as what we care about if the parent is at the top (i.e something like eastenders where there is a container then episodes directly under it, rather than something like mongrels which has a brand, then a series, then an episode) then we don't want to show the 'new series' badge.

Though TBH I don't think we want this clause at all, because what should the "eastenders case" as above not deserve a "new" badge when it is new.

votemike: @BPScott is this also going to be the case in the On Demand column?

BPScott: Confirmed with Cathal that we do want this clause. Eastenders doesn't want a new badge on each episode.

@votemike If it's got the same badge logic, then I think so, yes

src/DsAmen/Organism/Programme/SubPresenters/ProgrammeTitlePresenter.php
```diff
@@ -20,7 +20,7 @@ class ProgrammeTitlePresenter extends Presenter
     /** @var TitleLogicHelper */
     private $titleHelper;
 
-    /** @var Programme */
+    /** @var Programme|ProgrammeItem */
```
votemike: If you have ProgrammeItem you don't need Programme

felipeparaujo: Tell that to phpstan

felipeparaujo: I was getting a warning complaining that `isAudio()` is not present in Programme in this line

https://github.com/bbc/programmes-frontend/blob/42dc8f1183a443f9e7ae3adf1639f8f779d77e70/src/DsAmen/Organism/Programme/SubPresenters/ProgrammeTitlePresenter.php#L98

That `$isEpisode` ensures that `isAudio` will exist, but PHPStan is not smart enough to figure that out. This was the simplest way to fix that.

felipeparaujo:
test this please

felipeparaujo:
test this please

PR #234, opened 2017-10-11
PROGRAMMES-5737 Fixing bug with upcoming episodes

votemike:
Should fix this: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5737?focusedCommentId=3605286&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-3605286

Do not merge before https://github.com/bbc/programmes-pages-service/pull/202 is merged and composer is updated in this PR.


programmes-developers:

src/DsAmen/Organism/Map/SubPresenter/OnDemandPresenter.php
```diff
@@ -106,7 +110,22 @@ public function justMissed(): bool
         }
 
         $sevenDaysAgo = ApplicationTime::getTime()->subDays(7);
-        return $this->lastOn->getStartAt()->lt($sevenDaysAgo) && $this->upcomingEpisode && !$this->streamableEpisode;
+        return $this->lastOn->getStartAt()->gt($sevenDaysAgo);
+    }
+
+    public function showImage(): bool
```
tatitati: Not sure, but because is returning a bool, might be a better name ```mustShowImage```, showImage seems to me that is displaying the image itself. Not only asking if should be displayed.

phpstan.neon
```diff
@@ -3,7 +3,8 @@ parameters:
         - '#Call to an undefined method DateTimeImmutable::\w*\(\).#'
         - '#Access to an undefined property Cake\\Chronos\\ChronosInterface::\$month.#' # https://github.com/cakephp/chronos/issues/151
         - '#Access to an undefined property Cake\\Chronos\\ChronosInterface::\$year.#'
-        - '#Method (\w*\\?)*::\w*\(\) should return Cake\\Chronos\\Chronos but returns DateTimeImmutable.#'
+        - '#Method [\w\\]*\w::\w*\(\) should return Cake\\Chronos\\Chronos but returns DateTimeImmutable.#'
```
votemike: @felipeparaujo I believe you added this line originally. It turns out that the original regex was a problem. See https://www.regular-expressions.info/catastrophic.html for more info.

votemike:
Test this please

votemike:
Test me please

src/DsAmen/Organism/Map/SubPresenter/OnDemandPresenter.php
```diff
@@ -97,16 +98,31 @@ public function getStreamableEpisode(): ?Episode
 
     public function justMissed(): bool
     {
-        if (!$this->lastOn) {
+        if (!$this->lastOn || $this->lastOn->getProgrammeItem()->getStreamableFrom()->isFuture()) {
             return false;
         }
 
-        if ($this->streamableEpisode && (string) $this->streamableEpisode->getPid() === (string) $this->lastOn->getProgrammeItem()->getPid()) {
+        if ($this->lastOnMatchesStreamable()) {
```
votemike: Do you think that the line https://github.com/bbc/programmes/blob/develop/webapp/php/sharedmodules/programmes/controllers/ProgrammesController.php#L1083 `$data->available_programmes[0]->pid == $data->last_on->programme_item->pid` in V2 is the equivalent of doing `$this->lastOn->getProgrammeItem()->isStreamable()` in V3?

src/DsAmen/Organism/Map/SubPresenter/OnDemandPresenter.php
```diff
@@ -74,7 +74,7 @@ public function getBadgeTranslationString(): string
             return '';
         }
 
-        if (!$this->streamableEpisode->getParent() instanceof Series) {
+        if (!$this->streamableEpisode->getParent() instanceof Series) { //@TODO did Ben say this should be something else?
```
votemike: @BPScott I seem to remember you said something about someone else's instanceof a Series.

src/DsAmen/Organism/Map/SubPresenter/OnDemandPresenter.php
BjornTwachtmann: I don't think that's the intention, the previous line is already checks for that via !$data->last_on->programme_item->is_available; . I do not know what the intention is but I can take a look if you want

votemike: I guess so. I have no idea what the intention of that check is.... If you could help, that'd be great.

tests/DsAmen/Organism/Map/SubPresenter/OnDemandPresenterTest.php
```diff
@@ -101,54 +109,125 @@ public function testNewSeriesBadge(bool $isRadio)
...
+        $this->assertFalse($odPresenter->shouldShowImage());
+
+        // MiniMap
+        $episodeImage = $this->createMock(Image::class);
+        $episodeImage->method('getPid')
+            ->willReturn(new Pid('v0t3m1k3'));
+        $episode = $this->createEpisode(1, '-1 day');
+        $episode->method('getImage')
+            ->willReturn($episodeImage);
+        $odPresenter = new OnDemandPresenter($programme, $episode, null, ['show_mini_map' => true]);
```
tatitati: I think this test looks like this:

```
Create scenario (programme)
assert
modify scenario (programme)
assert
modify scenario (programme)
assert
modify scenario (programme)
assert
```

Not sure, but seems that we can split this in different tests. The reason is that if the second assert fails, also will fails the next asserts because they are depending on the previous ones, but at the same time they are testing agains different 'programmes' states, so it can be a little confuse. We can finish with a test that complains saying that two or three behaviours are not working, but is only one failing.

votemike:
Test this please

BjornTwachtmann:
This looks fine, I can't speak for all of the logic, but merge it, the parts I understand look good.

PR #237, opened 2017-10-11
PROGRAMMES-5756 Programme Footer

mattgreenham:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5756
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5757


app/Resources/js/lazyload.js
```diff
@@ -1,74 +1,136 @@
...
+            data_outer_div: 'data-lazyload-outerdiv'
         },
-        setOptions: function (options) {
+        setOptions : function (options) {
             this.options = $.extend({}, this.options, options);
         },
-        init: function () {
-            var _this = this;
+        init : function () {
+            var _this = this, timer;
```
votemike: What's this witchcraft?

app/config/routing_3rd_party.yml
```diff
@@ -74,3 +80,8 @@ programme_broadcasts:
     defaults: { _controller: FrameworkBundle:Redirect:redirect, route: '' }
     requirements: { pid: '[0-9b-df-hj-np-tv-z]{8,15}' }
     schemes: [http]
+
+network:
+    path: /{networkUrlKey}
```
votemike: Are we having this route?

src/DsAmen/Organism/Footer/FooterPresenter.php
```diff
@@ -0,0 +1,77 @@
...
+use BBC\ProgrammesPagesService\Domain\Entity\Network;
+use BBC\ProgrammesPagesService\Domain\Entity\Programme;
+
+class FooterPresenter extends Presenter
+{
+    /** @var  Programme */
+    private $programme;
+
+    /** @var Network */
+    protected $network;
```
votemike: protected things above private ones please

src/DsAmen/Organism/Footer/FooterPresenter.php
```diff
@@ -0,0 +1,77 @@
...
+    {
+        parent::__construct($options);
+
+        $this->programme = $programme;
+        $this->network = $programme->getNetwork();
+    }
+
+    public function hasNetwork()
+    {
+        if ($this->network) {
```
votemike: Or just `return (bool) $this->network`?

src/DsAmen/Organism/Footer/FooterPresenter.php
```diff
@@ -0,0 +1,77 @@
...
+
+    public function __construct(Programme $programme, array $options = [])
+    {
+        parent::__construct($options);
+
+        $this->programme = $programme;
+        $this->network = $programme->getNetwork();
+    }
+
+    public function hasNetwork()
```
votemike: Return types for everything?

programmes-developers:

src/DsAmen/Organism/Footer/FooterPresenter.php
BPScott: This class isn't going to be inherited from, so this should be private anyway.

app/config/routing_3rd_party.yml
```diff
@@ -74,3 +80,8 @@ programme_broadcasts:
...
+    defaults: { _controller: FrameworkBundle:Redirect:redirect, route: '' }
+    requirements: { networkUrlKey: '[a-zA-Z]{2,35}' }
```
BPScott: urlkeys can have numbers in them too, check the format in the other routes

app/Resources/views/find_by_pid/tlec.html.twig
```diff
@@ -80,3 +80,5 @@
 
 {# TODO Related links block #}
 {% endblock %}
+
+{% block footer %}{{ ds_amen('footer', programme) }}{% endblock %}
```
BPScott: Having a separate block seems unneeded, we could stick the {{dsamen()}} call at the bottom of the body block

gulpfile.js
```diff
@@ -37,14 +37,15 @@ gulp.task('js', ['js:clean'], function () {
...
+            "jquery.appear": "../../../node_modules/jquery.appear/jquery.appear",
             "lazysizes": "../../../node_modules/lazysizes/lazysizes-umd",
             "eqjs": "../../../node_modules/eq.js/build/eq",
             "comscorews" : "../../../vendor/bbc-rmp/comscore/js-modules/comscorews",
             "rmpcomscore/base" : "../../../vendor/bbc-rmp/comscore/js-modules/base",
             "orb/cookies": "empty:"
         },
         // @TODO do not optimise on sandbox
-        //"optimize": "none",
+        "optimize": "none",
```
BPScott: Put this back

src/DsAmen/Organism/Footer/FooterPresenter.php
BPScott: This class isn't being inherited from, so should be private.

src/DsAmen/Organism/Footer/footer.html.twig
```diff
@@ -0,0 +1,96 @@
...
+                <div class="grid__item 1/3@gel4 1/4@gel3b 1/2@gel3">
+                    <div class="islet--horizontal footer__programmes footer__service-island">
+                        <h2 class="gel-double-pica-bold">{{ tr('similar_programmes') }}</h2>
+                        {%- if footer.getProgrammeGenreHierarchy() -%}
+                            <div class="islet--vertical">
+                                <h3 class="gel-brevier-bold text--shout">{{ tr('by_genre') }}:</h3>
+                                <ul class="list-raw list-unstyled text--shout gel-brevier-bold">
+                                    {%- for genreTree in footer.getProgrammeGenreHierarchy() -%}
+                                        <li>
+                                            {%- for genre in genreTree|reverse -%}
```
BPScott: If you make footer.getGenres() return a simple list, then this line can be `genre.getAncestry|reverse` which saves you having to loop through that array twice

app/Resources/js/lazyload.js
```diff
@@ -1,4 +1,4 @@
-define(['jquery-1.9'], function ($) {
+define(['jquery-1.9', 'jquery.appear'], function ($) {
```
BPScott: The original version of this didn't need appear, as it's functionality was pretty small, is there a way where we can not reintroduce that dependency?

src/DsAmen/Organism/Footer/FooterPresenter.php
```diff
@@ -0,0 +1,65 @@
...
+    {
+        return $this->network ? $this->network->getUrlKey() : "";
+    }
+
+    public function getNetworkName(): string
+    {
+        return $this->network ? $this->network->getName() : "";
+    }
+
+    public function getNavigationLinks(): array
```
votemike: Would you be able to have a look at this: https://gist.github.com/votemike/04d7f5c2859a6999a5aa3915ea40c982 Ask IRL if you don't understand

mattgreenham: Fixed

PR #238, opened 2017-10-11
Use @bp convention for per-breakpoint classes

BPScott:
Update island, text helpers and widths to follow the standard
convention of having per-breakpoint classes be denoted by a suffix.
i.e. "blah@bpw" not "bpw-blah".

Any in progress PRs that use these classes will need to be updated after this is merged.


programmes-developers:

BPScott:
For the record, easised way to check if this change affects you is to search your branch for the regex `bp[a-zA-Z0-9]?-` to find anything that uses the prefix style notation

PR #239, opened 2017-10-11
Run a composer update and yarn upgrade

BPScott:
Goin' LIVE soon, so lets update our dependencies


programmes-developers:

votemike:
On my machine, it looks like some packages still need upgrading.
`composer outdated`
squizlabs/php_codesniffer has a version 3.1.0 out but we seem to be on 2.9.1 for example.  
Similar thing with yarn where there are new major versions.

BPScott:
Updated based on yarn/composer outdated commands. Reponse is now:

```
 ❯❯❯ composer outdated                                                   
nette/php-generator                      v3.0.1             v3.1.0             🐘 Nette PHP Generator: generates n...
squizlabs/php_codesniffer                2.9.1              3.1.0              PHP_CodeSniffer tokenizes PHP, JavaSc...
symfony/translation                      v3.1.0             v3.3.10            Symfony Translation Component
❯❯❯ yarn outdated                                                       ✱ 
yarn outdated v0.23.2
Package   Current Wanted Latest    Package Type    URL
lazysizes 3.0.0   3.0.0  4.0.0-rc4 devDependencies https://github.com/aFarkas/lazysizes#readme
```

I'm not updating to an rc of a package, so that's the yarn deps covered.
Translation is version locked due to a bug in Symfony
PHPCS is locked because the the coding standards we use only supports phpcs2, not 3 (I note we're using a fork of the original symfony standards and the originial now supports 3, should we move back?)
php-generator is a downstream dep of phpstan which we can't control

votemike:
Cool. Raise a ticket to investigate/change the forked version?

PR #242, opened 2017-10-13
Programmes-5736 Promo priority presenter

BjornTwachtmann:
Also sets proper responsive image sizes for all TLEC page stuff. Which should probably have been in a different PR. But it isn't.


programmes-developers:

src/Controller/FindByPid/TlecController.php
```diff
@@ -80,12 +80,19 @@ public function __invoke(
...
+            $promotions[0] ?? null,
             $comingSoonPromo,
             $streamableEpisodes[0] ?? null,
             $upcomingRepeatsAndDebutsCounts['debuts'],
             $upcomingRepeatsAndDebutsCounts['repeats']
         );
 
+        // This is ugly but I don't know how else to do it. If promo priority is active the first promo moves
+        // into the MAP
+        if ($mapPresenter->isPromoPriority()) {
```
votemike: 🤢

src/DsAmen/Organism/Map/MapPresenter.php
```diff
@@ -155,18 +152,26 @@ public function getSocialPresenter(): SocialPresenter
         return new SocialPresenter($this->programme);
     }
 
+    public function isPromoPriority(): bool
+    {
+        if ($this->programme->getOption('brand_layout') === 'promo'
+            && $this->firstPromo
+            && is_null($this->programme->getParent())
```
votemike: Is this equivalent to `$this->programme->getParent()->isTleo`? or Tlec? Might be clearer.

src/DsAmen/Organism/Map/MapPresenter.php
```diff
@@ -216,18 +221,42 @@ private function constructTwoColumnMap()
...
+                [
+                    'is_three_column' => $this->countTotalColumns() === 3,
+                    'show_mini_map' => $this->showMiniMap(),
+                ]
+            );
+        } else {
+            $this->leftColumn = new ProgrammeInfoPresenter(
+                $this->programme,
+                [
+                    'is_three_column' => $this->countTotalColumns() === 3,
```
votemike: If the options are the same in both presenters, you could construct them before the `if`.

src/DsAmen/Organism/Map/SubPresenter/ComingSoonPresenter.php
```diff
@@ -3,13 +3,16 @@
...
 
+use App\DsAmen\Organism\Map\SubPresenter\Traits\RightColumnImageSizeTrait;
 use App\DsAmen\Presenter;
 use App\Exception\InvalidOptionException;
 use BBC\ProgrammesPagesService\Domain\Entity\Programme;
 use BBC\ProgrammesPagesService\Domain\Entity\Promotion;
 
 class ComingSoonPresenter extends Presenter
 {
+    use RightColumnImageSizeTrait;
```
votemike: 😍

src/DsAmen/Organism/Map/SubPresenter/promo_priority.html.twig
```diff
@@ -0,0 +1,12 @@
+<div class="br-box-secondary" data-map-column="promo">
+    {{ ds_amen('promotion', promo_priority.getPromotion(), {
+        'branding_name' : 'secondary',
+        'show_synopsis' : true,
+        'related_links_count' : 2,
```
votemike: What does 2 this relate to?

src/Controller/FindByPid/TlecController.php
BPScott: How about dragging this function into the Controller, doing this check before constructing the map, and only passing a promo into the map if  it's in priority mode?

Then in the map work out if it should show a promo based on the presence or absence of that promo.


```php
$mapPromo = null;
if ($mapPresenter->isPromoPriority()) {
$mapPromo = array_shift($promotions);
}

$mapPresenter = $presenterFactory->mapPresenter(
//...
$mapPromo,
//...
);

BjornTwachtmann: Yeah. It's not nice. If I pull isPromoPriority into the controller I also have to pull in the showMiniMap logic in its entireity. Which puts a lot of presentation logic in the controller that doesn't really belong there. Other option would be to pull all that logic out into a helper or trait. 

Personally I think this is less likely to result in b0rkage than either of those two, but I don't mind pulling the logic out into another class if you two would prefer that.

src/DsAmen/Organism/Map/MapPresenter.php
BjornTwachtmann: $this->programme->isTleo(), but yeah, that would be clearer :)

src/DsAmen/Organism/Map/SubPresenter/promo_priority.html.twig
BjornTwachtmann: It's just a hardcoded number of related links that are allowed to be shown in a promo priority column. Can add a comment if that's not clear.

votemike: Just wondered.
Do you have an example URL with 2 links?

BjornTwachtmann: @votemike http://www.bbc.co.uk/programmes/b00cyj3f

PR #243, opened 2017-10-13
DO NOT REVIEW PROGRAMMES-5737 bugfixes

votemike:
Fixing these two problems....

https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5737?focusedCommentId=3610913&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-3610913

Do not merge before https://github.com/bbc/programmes-pages-service/pull/204
I also renamed `lastOnNotAvailableYet()`

### DIFF
--- a/src/Controller/FindByPid/TlecController.php
+++ b/src/Controller/FindByPid/TlecController.php
@@ -57,7 +57,7 @@ class TlecController extends BaseController
         if ($programme->getAggregatedEpisodesCount() > 0) {
             $streamableEpisodes = $aggregationService->findStreamableDescendantEpisodes($programme, 1);
             if (empty($streamableEpisodes)) {
-                $hasUpcomingEpisodes = null !== $aggregationService->findUpcomingStreamableDescendantEpisodes($programme, 1); //@TODO maybe change to count
+                $hasUpcomingEpisodes = $aggregationService->countUpcomingStreamableDescendantEpisodes($programme) > 0;
             }
         }
 
--- a/src/Controller/FindByPid/TlecController.php
+++ b/src/Controller/FindByPid/TlecController.php
@@ -52,10 +52,13 @@ class TlecController extends BaseController
             $clips = $aggregationService->findDescendantClips($programme, 4);
         }
 
-        $upcomingEpisodes = null;
+        $hasUpcomingEpisodes = false;
         $streamableEpisodes = null;
         if ($programme->getAggregatedEpisodesCount() > 0) {
             $streamableEpisodes = $aggregationService->findStreamableDescendantEpisodes($programme, 1);
+            if (empty($streamableEpisodes)) {
+                $hasUpcomingEpisodes = null !== $aggregationService->findUpcomingStreamableDescendantEpisodes($programme, 1); //@TODO maybe change to count
+            }
         }
 
         if ($programme->getOption('show_gallery_cards')) {
@@ -82,6 +85,7 @@ class TlecController extends BaseController
             $lastOn,
             $comingSoonPromo,
             $streamableEpisodes[0] ?? null,
+            $hasUpcomingEpisodes,
             $upcomingRepeatsAndDebutsCounts['debuts'],
             $upcomingRepeatsAndDebutsCounts['repeats']
         );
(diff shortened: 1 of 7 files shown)


programmes-developers:

src/Controller/FindByPid/TlecController.php
```diff
@@ -52,10 +52,13 @@ public function __invoke(
             $clips = $aggregationService->findDescendantClips($programme, 4);
         }
 
-        $upcomingEpisodes = null;
+        $hasUpcomingEpisodes = false;
         $streamableEpisodes = null;
         if ($programme->getAggregatedEpisodesCount() > 0) {
             $streamableEpisodes = $aggregationService->findStreamableDescendantEpisodes($programme, 1);
+            if (empty($streamableEpisodes)) {
+                $hasUpcomingEpisodes = null !== $aggregationService->findUpcomingStreamableDescendantEpisodes($programme, 1); //@TODO maybe change to count
```
votemike: Should I change this to be a count method? Or does a count method already exist?

src/DsAmen/Organism/Map/SubPresenter/OnDemandPresenter.php
```diff
@@ -66,7 +80,7 @@ public function getClass(): string
         return $this->class;
     }
 
-    public function getUpcomingEpisode(): ?Episode
+    public function getPendingEpisode(): ?Episode
```
votemike: Happy with this new name?

src/Controller/FindByPid/TlecController.php
felipeparaujo: I don't think there's a count for this already. A count would be faster, although probably not by much. You should evaluate it.

src/DsAmen/Organism/Map/SubPresenter/OnDemandPresenter.php
felipeparaujo: If you leave a little comment explaining what a pending episode is just above the method, yeah.

votemike: I have added comments here: https://github.com/bbc/programmes-frontend/pull/243/files#diff-d04f71ad0d6c828a552d1ed9031238fbR34 and here: https://github.com/bbc/programmes-frontend/pull/243/files#diff-d04f71ad0d6c828a552d1ed9031238fbR142 . Is that good enough? Or do you want it here too?

src/Controller/FindByPid/TlecController.php
```diff
@@ -52,10 +52,13 @@ public function __invoke(
             $clips = $aggregationService->findDescendantClips($programme, 4);
         }
 
-        $upcomingEpisodes = null;
+        $hasUpcomingEpisodes = false;
         $streamableEpisodes = null;
         if ($programme->getAggregatedEpisodesCount() > 0) {
             $streamableEpisodes = $aggregationService->findStreamableDescendantEpisodes($programme, 1);
+            if (empty($streamableEpisodes)) {
+                $hasUpcomingEpisodes = $aggregationService->countUpcomingStreamableDescendantEpisodes($programme) > 0;
```
BjornTwachtmann: I've looked at the logic for this in V2 and I wonder if this query is necessary. https://github.com/bbc/programmes/blob/develop/webapp/php/lib/BBC/Programmes/View/Map/OnDemand.php#L64 . Could we just look at $upcomingBroadcast->isStreamable() (below)?

BjornTwachtmann: It's probably not a 100% match. But it's probably close enough that nobody will ever notice

BjornTwachtmann: Or rather $upcomingBroadcast->getProgrammeItem()->getStreamableFrom() && !$upcomingBroadcast->getProgrammeItem()->isStreamable()

votemike: Should we OK this with Cathal, Claudio or the testers?

votemike: Thinking of something like this? https://github.com/bbc/programmes-frontend/compare/PROGRAMMES-5737-bugfixes-2?expand=1

PR #244, opened 2017-10-16
PROGRAMMES-5735 - Title Truncation

felipeparaujo:


programmes-developers:

tests/DsAmen/Organism/Programme/SubPresenters/ProgrammeTitlePresenterTest.php
```diff
@@ -116,6 +116,36 @@ public function getMainAndSubTitlesProvider(): array
         ];
     }
 
+    public function testLongerThan60CharacterLongMainAndSubTitles(): void
```
votemike: Want to add a test with multi-byte characters?

tests/DsAmen/Organism/Programme/SubPresenters/ProgrammeTitlePresenterTest.php
```diff
@@ -116,6 +116,36 @@ public function getMainAndSubTitlesProvider(): array
...
+            $this->router,
+            $this->mockTitleLogicHelper,
+            [
+                'context_programme' => $this->mockContext,
+                'title_format' => 'item::ancestry',
+            ]
+        );
+
+        $this->assertSame('Very Very Very Very Very Very Very Very Very Very Long Clip…', $titlePresenter->getMainTitle());
+        $this->assertSame('Very Very Very Very Long Series Name, Very Very Very Very…', $titlePresenter->getSubTitle());
```
tatitati: I understand that the first string is 60 characters length because of the option ``` 'max_title_length' => 60,```
but the second one is 58 chars length, why?

tests/DsAmen/Organism/Programme/SubPresenters/ProgrammeTitlePresenterTest.php
```diff
@@ -116,6 +116,36 @@ public function getMainAndSubTitlesProvider(): array
...
+            $this->mockClip,
+            $this->router,
+            $this->mockTitleLogicHelper,
+            [
+                'context_programme' => $this->mockContext,
+                'title_format' => 'item::ancestry',
+            ]
+        );
+
+        $this->assertSame('Very Very Very Very Very Very Very Very Very Very Long Clip…', $titlePresenter->getMainTitle());
```
tatitati: Maybe would be useful to add an assert such like:
```$this->assertEquals(60, strlen($titlePresenter->getMainTitle()))```

**Comment by BjornTwachtmann** [comment]
Logic looks good, or at least the same as in Amen and nobdy's complained about that.

File: tests/DsAmen/Organism/Programme/SubPresenters/ProgrammeTitlePresenterTest.php (hunk 3 above)
Comment (felipeparaujo) [reply tests/DsAmen/Organism/Programme/SubPresenters/ProgrammeTitlePresenterTest.php]: Not really... The way the truncation works is: when there are more than 60 characters, truncate the title on the last word so that there are 60 or less characters in the title. So, we could have less that 60 characters in the title

Comment (tatitati) [reply tests/DsAmen/Organism/Programme/SubPresenters/ProgrammeTitlePresenterTest.php]: ou, I see. Cool!

File: tests/DsAmen/Organism/Programme/SubPresenters/ProgrammeTitlePresenterTest.php (hunk 1 above)
Comment (felipeparaujo) [reply tests/DsAmen/Organism/Programme/SubPresenters/ProgrammeTitlePresenterTest.php]: done

File: src/DsAmen/Organism/Programme/SubPresenters/ProgrammeTitlePresenter.php [hunk 4]
```diff
@@ -119,6 +122,21 @@ protected function validateOptions(array $options): void
...
+            throw new InvalidOptionException('max_title_length option must be an integer. HINT: use -1 for unlimited title length');
+        }
+    }
+
+    private function truncate(string $string, string $suffix = '…'): string
+    {
+        $maxLength = $this->getOption('max_title_length');
+
+        if ($maxLength > 0 && mb_strlen($string) > $maxLength) {
+            return mb_substr($string, 0, mb_strrpos($string, ' ', $maxLength - mb_strlen($string))) . $suffix;
```
votemike: Tiny possible improvement. This is the 2nd time you're calling `mb_strlen($string)`. Maybe extract it to a variable?

PR #246, opened 2017-10-16
PROGRAMMES-5758 footer watchlive

mattgreenham:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5758


programmes-developers:

app/Resources/js/lazyload.js
```diff
@@ -13,16 +29,18 @@ define(['jquery-1.9'], function ($) {
...
+            data_delay_lazyload : 'data-lazyload-delay'
         },
         setOptions: function (options) {
             this.options = $.extend({}, this.options, options);
         },
         init: function () {
             var _this = this;
 
-            this.loadContents();
+            _this.loadContents();
```
votemike: This shouldn't need changing, should it?

app/Resources/js/lazyload.js
```diff
@@ -34,18 +52,71 @@ define(['jquery-1.9'], function ($) {
...
+            var window_top = $window.scrollTop();
+            var offset = $element.offset();
+            var left = offset.left;
+            var top = offset.top;
+
+            // load just before in viewport
+            var lazy_top_offset = 500;
+            var lazy_left_offset = 500;
+
+            if (top + $element.height() >= window_top &&
```
votemike: Can the logic in this `if` be extracted to a method to make it more readable?

app/Resources/js/lazyload.js
```diff
@@ -34,18 +52,71 @@ define(['jquery-1.9'], function ($) {
...
+                top - lazy_top_offset <= window_top + $window.height() &&
+                left + $element.width() >= window_left &&
+                left - lazy_left_offset <= window_left + $window.width()) {
+                return true;
+            } else {
+                return false;
+            }
+        },
+        addOnScrollHandler : function (context) {
+            var _this = this;
```
votemike: Why do we keep doing this instead of `.bind(this)`?

app/Resources/js/lazyload.js
```diff
@@ -34,18 +52,71 @@ define(['jquery-1.9'], function ($) {
...
+
+            var check_in_view = debounce(function () {
```
votemike: Why is this function assigned to a variable? Why is it not just a named method?

app/Resources/js/lazyload.js
mattgreenham: I don't think so, that's what this entire method does?

votemike: Extract each line to a variable instead.  
Then `return a && b && c && d;`

app/Resources/js/lazyload.js
```diff
@@ -4,6 +4,22 @@ define(['jquery-1.9'], function ($) {
         this.setOptions(options);
     };
 
+    var toAppear = [];
```
votemike: What is this an array of? I can't figure it out.

app/Resources/js/lazyload.js
```diff
@@ -4,6 +4,22 @@ define(['jquery-1.9'], function ($) {
...
+
+    function debounce(func, wait) {
+        var timeout;
+        return function() {
+            var context = this, args = arguments;
```
votemike: Does `this` need to be assigned to `context`? Can't you just pass `this` straight in to `func.apply()`?

app/Resources/js/lazyload.js
```diff
@@ -34,18 +52,68 @@ define(['jquery-1.9'], function ($) {
...
+                    if (!this.isDelayed(content) || this.isInViewport(content)) {
+                        this.loadAjaxContent(content);
+                    }
+                    else {
+                        toAppear.push(content);
+                    }
                 }
             }
+
+            this.addOnScrollOrResizeHandler(context);
```
votemike: Could you explain why this is here. And what `context` is?

app/Resources/js/lazyload.js
```diff
@@ -34,18 +52,68 @@ define(['jquery-1.9'], function ($) {
...
+            var right_of_lhs = (element_left + $element.width() >= window_left);
+            var left_of_rhs = (element_left - lazy_left_offset <= window_left + $window.width());
+
+            return below_window_top && above_window_bottom && right_of_lhs && left_of_rhs;
+        },
+        checkInView : function (context) {
+            return debounce(function () {
+                if (toAppear.length > 0) {
+                    var content;
+                    for (var i = 0; (content = toAppear[i]) != null; i++) {
```
votemike: Is this not better-off as a `.each()`?

app/Resources/js/lazyload.js
mattgreenham: Lazy elements which haven't yet been loaded i.e. not yet in view. On page load we find all the 'lazy-module' elements on the page so on scroll we only need to check if elements in this array are in view.

app/Resources/js/lazyload.js
mattgreenham: No. Yes. Changed.

app/Resources/js/dsamen-bootstrap.js
```diff
@@ -10,6 +10,9 @@ define(['eqjs', 'respimg', 'lazysizes'], function(EQ) {
             window.addEventListener('lazyload-complete',function(e){
                 EQ.query(document.querySelectorAll(selector));
             },false);
+
+            var lazyload = new Lazyload();
```
tatitati: I suspect that is not necessary to create a new lazyload object. You are passing already an object in there. Just try (maybe I'm wrong)

app/Resources/js/lazyload.js
votemike: So.... a bunch or jQuery objects? A bunch of HTML tags? What?

mattgreenham: jQuery objects

app/Resources/js/dsamen-bootstrap.js
mattgreenham: Doesn't work without it

tatitati: Interesting. Thanks Matt. Sorry for my "experimental feedback" jeje

app/Resources/js/lazyload.js
votemike: Add a comment to this, I still get confused as to what it is.

app/Resources/js/lazyload.js
```diff
@@ -34,18 +51,66 @@ define(['jquery-1.9'], function ($) {
...
+
+            return below_window_top && above_window_bottom && right_of_lhs && left_of_rhs;
+        },
+        checkInView : function () {
+            return debounce(function () {
+                var _this = this;
+                toAppear.forEach(function (element, i) {
+                    if (_this.shouldBeRenderedNow(element)) {
+                        _this.loadAjaxContent(element);
+                        toAppear.splice(i, 1);
```
votemike: `i` looks to similar to `1`. Rename it. Maybe to `key`?

app/Resources/js/lazyload.js
```diff
@@ -34,18 +51,66 @@ define(['jquery-1.9'], function ($) {
 
             return (lazy_contents.length) ? lazy_contents : false;
         },
+        isOverThreshold : function (lazyloaded_object) {
+            var threshold = lazyloaded_object.getAttribute(this.options.data_threshold);
```
votemike: Can you rename `threshold` to `element_threshold`?

votemike: Or maybe even `minimum_render_threshold` or something similar.

tatitati:
COOOOOOOOOOOOOOOOL

PR #247, opened 2017-10-17
PROGRAMMES-5735 - Fixing broadcast details spacing issue

felipeparaujo:


programmes-developers:

src/DsAmen/Organism/Programme/_collapsed-broadcast.scss
```diff
@@ -3,8 +3,8 @@
  */
 
 .broadcast-details__service {
-    span[property='name'] {
-        margin-right: $gel-unit-small / 2;
+    span ~ span {
```
votemike: What is this? Can you not use class names instead?

felipeparaujo: This means that every span preceded by a span gets that property.

felipeparaujo: I could use classes, but it would just be the same

votemike: span ~ span is gross and confusing. What if it is `<span><span><span></span></span></span>`?

felipeparaujo: None of them gets the styling. This makes it easier to implement stuff like "add spacing to all children except for the first one". If I remove this, the template code will look uglier.

PR #250, opened 2017-10-17
OPTIMIZE-IMPORTS Remove unused imports

votemike:
Removing unused imports


tests/TwigEnvironmentProvider.php
```diff
@@ -2,8 +2,8 @@
 declare(strict_types = 1);
 namespace Tests\App;
 
-use App\Ds2013\PresenterFactory as Ds2013PresenterFactory;
 use App\Ds2013\PresenterFactory;
```
votemike: Should this even be here?

BjornTwachtmann: Please don't change this. It's there for a reason https://github.com/bbc/programmes-frontend/blob/eb31694fa34e8ac8f279caad378a5cd7bf43df68/tests/TwigEnvironmentProvider.php#L120 . That reason is not my personal amusement.

votemike: Grim.
Is it OK to re-order it? Intellij also re-order when it cleans up unused imports.

votemike: Wait, your line doesn't even use this line. It uses the one above/below it.

programmes-developers:

tests/TwigEnvironmentProvider.php
BjornTwachtmann: use App\Ds2013\PresenterFactory as Ds2013PresenterFactory;
...
  self::$ds2013PresenterFactory = new Ds2013PresenterFactory($translateProvider, $router, $helperFactory); 

Unless I'm being more dense than usual that import is needed. If you want to refactor it to use the fully qualified class name or something do so, I'm completely unconcerned either way.

votemike: My comment is on the `use App\Ds2013\PresenterFactory;` line. I can't see a `PresenterFactory;` being used in this file.

BjornTwachtmann: Ah, okay, I was somewhat confused by the diff showing the used import being removed. The other one probably isn't needed.

votemike: The diff should be clearer now...

votemike:
Test this please

PR #253, opened 2017-10-17
PROGRAMMES-5924 - Moving away from b-g-p to avoid box-sizing issues

felipeparaujo:
To make our box-sizing situation more sane (and stop breaking Orbit styling badly) we moved to this strategy to set `box-sizing: border-box` everywhere: 
https://github.com/bbc/programmes-frontend/blob/6667e4bb6b978a485e55d725141c9b8fbda39005/src/DsAmen/Core/Generic/_main.scss#L5-L13

This means we don't override every single item's box-sizing to a fixed value, it now inherits from its parent.

Unfortunately, this also means we have to stop using `.b-g-p` (and possibly `b-g-m` in the future, although that's not used anywhere for now) because it assumes `box-sizing: content-box` when setting its breakpoints and `max-width`s. To fix this I introduced `.p-g-p`, which updates `b-g-p`s `max-width`s to assume `box-sizing: border-box`

I think this is a good move, but it comes with downsides: If `b-g-p` (which comes from Orbit) starts supporting more breakpoints, we'll have to our `.p-g-p` class. Also, we have more CSS to maintain.


programmes-developers:

votemike:
Where is `b-g-p` defined? Should you remove that too?

felipeparaujo:
`b-g-p` comes from Orbit, we can't change it.

votemike:
Don't really know enough about our CSS to approve this. But from what I can tell, it seems sound.

PR #254, opened 2017-10-18
PROGRAMMES-5737 When the previous broadcast's episode is no longer

votemike:
Change can be seen by switching between develop and this branch when looking at the Dr. Who brand page on live and locally.


programmes-developers:

src/DsAmen/Organism/Map/SubPresenter/OnDemandPresenter.php
```diff
@@ -91,7 +97,7 @@ public function getBadgeTranslationString(): string
         }
         // If the parent is the TLEO (e.g. Eastenders) we don't want to show a new badge for each episode
         // Otherwise (e.g. Mongrels) the parent will be a series, which we do want badges for each episode.
-        if (!$this->streamableEpisode->getParent() || $this->streamableEpisode->getParent()->isTleo()) {
+        if ($this->streamableEpisode === null || !$this->streamableEpisode->getParent() || $this->streamableEpisode->getParent()->isTleo()) {
```
felipeparaujo: We usually use `is_null` but it's all the same to me

votemike: My IDE complains about is_null for some reason... Dunno if there's something in the .editorconfig.

PR #255, opened 2017-10-18
Sourcemaps in SANDBOX 

tatitati:
After running YARN we have this nice structure in our assets folder where we can see that sourcemaps are received from different files in a separated folder:

[PROGRAMMES-5845](https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5845)


programmes-developers:

tatitati:
@BPScott you are also familiar with asset build. Please let me know your thoughts about these changes.

gulpfile.js
```diff
@@ -10,6 +10,7 @@ const del = require('del');
 const requirejsOptimize = require('gulp-requirejs-optimize');
 const autoprefixer = require('gulp-autoprefixer');
 const override = require('gulp-rev-css-url');
+const yargs = require('yargs');
```
tatitati: Added yargs library to pass all kind of arguments to our scripts. Very useful to make them more dynamic and easy to use.

votemike:
Do we want the js stuff in different files?

votemike:
Sourcemaps do not work properly with revisioned files.

BPScott:
As per IRL conversation, I reckon we should have the files and their maps in the same folder rather than a separate sourcemaps folder - use `sourcemaps.write('.')`

package.json
```diff
@@ -20,7 +21,7 @@
     },
     "scripts": {
         "gulp": "gulp default",
-        "gulp-ci": "gulp distribution",
+        "gulp-ci": "gulp distribution --SOURCEMAPS false",
```
votemike: I thought we were doing Source Maps for all environments?

tatitati: Yeah. But our rev pluging doesnt support our sourcemaps plugin. So we cannot revision and add sourcemaps at the same time. :(
I discussed this with @BPScott 

https://github.com/gulp-sourcemaps/gulp-sourcemaps/wiki/Plugins-with-gulp-sourcemaps-support

votemike: O. That's rubbish. Is there a different sourcemap plugin you can use? I'm sure I've used both together before.

tatitati: let me try something I just realized....maybe works

tatitati: Well, I can try, but to be honest, I don't see the problem having sourcemaps only in sandbox. For me is more logic to have debug info in my sandbox that in production.

tatitati: @BPScott ?, @BjornTwachtmann ?

BjornTwachtmann: Seems okay to disable this on int/test/live to me, for now at least. We can revisit if it becomes annoying.

programmes-developers:

gulpfile.js
```diff
@@ -19,6 +21,14 @@ const imageMatch = '/images/*';
 
 var throwError = true;
 
+var optimizeWithSourcemaps = (function() {
```
votemike: Do you need this function? Could you not just do `.pipe(gulpif(yargs.argv.SOURCEMAPS !== 'false', sourcemaps.init()))`

tatitati: Yeah, but that logic becomes repetitive because we need that check in multiple places. So I prefered to extract it to only one place. I less error prone.

package.json
BPScott: Gulp-rev does not modify the contents of files by design. The issue is that when you run gulp-rev the sourcemap file is revisioned (so the filename becomes `blah-abc123.css.map`) but the reference to the sourcemap in the JS or CSS file remains as `blah.css`. There needs to be an extra step that rewrites the CSS/JS file so that it correctly points to the revisioned sourcemap file.

gulpfile.js
BPScott: Is there much point having this as an immediately invoked function?

`var optimizeWithSourcemaps = yargs.argv.SOURCEMAPS != 'false';` seems a lot simpler.

gulpfile.js
```diff
@@ -32,28 +43,10 @@ gulp.task('js', ['js:clean'], function () {
...
-        "map": {
-            "*": {
-                "jquery": "jquery-1.9"
-            }
-        }
-    };
-
     return gulp.src(modulesToOptimize)
-        .pipe(requirejsOptimize(config))
+        .pipe(gulpif(isSandbox, sourcemaps.init()))
```
tatitati: Seems trivial this check but it isn't. 
The reason to keep this check is to avoid sourcemaps when revisioning files. This cannot happens at the same time because sourcemaps doesn't support revisioning files.

PR #256, opened 2017-10-18
PROGRAMMES-5924 - Only DSAmen pages should use p-g-p

felipeparaujo:


BjornTwachtmann:
Looks fine, make sure you've checked all the 2013 routes.

PR #258, opened 2017-10-19
Enforce PHPSCS rule that avoids else/elseif after an early return

BPScott:
When writing code with early returns inside an if statement there is no
point carrying on in an else or an elseif - either write logic or have
a second if statement.

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -25,8 +25,6 @@
         <!-- We don't care about enforcing using sprintf in error messages -->
         <exclude name="Symfony.Errors.ExceptionMessage.Invalid"/>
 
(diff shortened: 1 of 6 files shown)


programmes-developers:

BjornTwachtmann:
I've got to say I really don't like code sniffers having opinions about this sort of thing. Indentaion and spacing is one thing, structure is anouther. I don't see much wrong with the code you've changed. If the consensus is that this is fine I'm not going to throw my toys out of the pram, but I'd rather not have this.

BPScott:
Usually I'm with you (see yoda conditionals being in the bin), but I'm ok with this one as pointing out "you don't need an else clause if you're returning early" is a structure thing I try and point out in reviews if I spot it as I think it's a good format to use.

PR #259, opened 2017-10-19
Fix countername for world service online

BjornTwachtmann:


programmes-developers:

PR #260, opened 2017-10-19
Use a null cache adapter for the status controller

Tasiobg:
The status controller should not connect to redis and setting the TTL to CacheInterface::NONE is not avoiding the connection, this PR updates the dependency injection to use a null cache adapter on the status controller.

One of the side effects of the previous solution is to display a lot of cache miss on the CloudWatch Dashboad because the health check runs at least once every second even when we set that to run every 5 minutes


programmes-developers:

src/Controller/StatusController.php
```diff
@@ -16,22 +19,34 @@
...
+    }
+
     public function __invoke(
         Request $request,
         Connection $dbalConnection,
-        ProgrammesService $programmesService,
-        BroadcastsService $broadcastsService,
-        VersionsService $versionsService,
-        SegmentEventsService $segmentEventsService
+        EntityManagerInterface $entityManager
```
BjornTwachtmann: I don't think this is used any more, is it?

BjornTwachtmann:
Other than that one tiny thing, fine, merge it.

src/Controller/StatusController.php
Tasiobg: nop, that was for the old aproach

PR #261, opened 2017-10-20
PROGRAMMES-5733 Adding priority text

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5733

Example on programmes/b00cyj3f?__flushcache=1


app/Resources/views/find_by_pid/tlec.html.twig
```diff
@@ -21,11 +21,25 @@
 
     <div class="p-g-p">
         <div class="leading-gap cf">
-            {# TODO Priority text block #}
+            {#{%- if programme.getOption('prioritytext_block') or isbbcdotcom or hasrecipes -%} TODO #}
+            {%- if programme.getOption('prioritytext_block') -%}
```
votemike: Do we want to get the prioritytext_block like this? Or do we want to retrieve it in the controller and pass it to the view?

programmes-developers:

app/Resources/views/find_by_pid/tlec.html.twig
Tasiobg: Looks good to me. 
My personal preference is to keep templates highly decoupled by sending an array of parameters instead of calling functions from inside a template but this is just personal preference I think no one else shared this point when I mentioned that before

tatitati:
Ei michael, the code looks ok to me. But still (even with the links) I don't know where appear or how it looks that priority text in the url you put in the comment. Might you explain a little more the comment to make easier (for everybody) to find that text in the site?

votemike:
@tatitati Is it not clear from the ticket? I've linked to the ticket on the description. Otherwise I'll just be copy and pasting stuff from the ticket.

PR #262, opened 2017-10-20
Fix IE display bugs in the overlay complicated watch live/watch from …

BjornTwachtmann:
…start mess


programmes-developers:

PR #263, opened 2017-10-20
PROGRAMMES-5952 Adding vote priority

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5952

Can be viewed at programmes/b006m8dq?__flush_cache=1


app/Resources/views/find_by_pid/tlec.html.twig
```diff
@@ -13,7 +13,15 @@
 
     {# TODO BBC Live Promo #}
 
-    {# TODO Priority Vote block #}
+    {% if programme.getOption('brand_layout') == 'vote' and programme.getOption('ivote_block') %}
+        <div class="p-g-p">
+            <div class="grid leading-gap">
+                <div class="grid__item">
+                    {{ dump(programme.getOption('brand_layout'), programme.getOption('ivote_block')) }}
```
Tasiobg: Is this `dump()` right? isn't this only for debugging?

votemike: From what I can tell, the ticket it only to get the placeholder ready. Do you also think that's what the ticket is asking for?
I didn't know what else to put inside the div if we're not yet implementing the vote thing.

src/DsAmen/Organism/Map/MapPresenter.php
```diff
@@ -288,9 +288,13 @@ private function showMiniMap(): bool
     {
         if ($this->request->query->has('__2016minimap')) {
             return (bool) $this->request->query->get('__2016minimap');
-        } else {
-            // if ($this->is2016BrandPage() && $this->isVotePriority() ) return true;
-            return filter_var($this->programme->getOption('brand_2016_layout_use_minimap'), FILTER_VALIDATE_BOOLEAN);
         }
+
+        if ($this->programme->getOption('brand_layout') === 'vote') {
```
BjornTwachtmann: If the TLEC template is checking for the presence of the programme-options ivote bit, that should probably be the same here too. People will probably delete the vote and forget to change the layout back.

app/Resources/views/find_by_pid/tlec.html.twig
BjornTwachtmann: it's fine for now I think

Tasiobg: Sorry I didn't realised it was just a placeholder, in that case looks good

BjornTwachtmann:
There's some extra markup and stuff for iVote in V2, but as we're not implementing iVote I don't think it's worth replicating. This is good as a place to put the testing code for the new module I think

src/DsAmen/Organism/Map/MapPresenter.php
votemike: So check for both?

BjornTwachtmann: yep

BjornTwachtmann: I could probably have expressed that better :)

programmes-developers:

src/DsAmen/Organism/Map/MapPresenter.php
```diff
@@ -288,9 +293,13 @@ private function showMiniMap(): bool
...
-        } else {
-            // if ($this->is2016BrandPage() && $this->isVotePriority() ) return true;
-            return filter_var($this->programme->getOption('brand_2016_layout_use_minimap'), FILTER_VALIDATE_BOOLEAN);
         }
+
+        if ($this->isVotePriority) {
+            return true;
+        }
+
+        // if ($this->is2016BrandPage()) return true;
```
votemike: Do we need to worry about this any more? Or is everything now the 2016 brand page?

votemike:
Test me please

PR #264, opened 2017-10-20
PROGRAMMES-5766 - Group object

felipeparaujo:
This PR actually encompasses more than the Group Object. I moved some code around to take advantage of the overlapping logic in these objects. The objects now live in DsAmen/Organism/Object. The object presenters and its subpresenters extend the base classes (in Base) and use the Shared blocks (which are concrete implementations of the abstract classes in Base, without any changes). Easiest way to review this will be to check it out locally and observe the changes in intellij or similar.


programmes-developers:

src/DsAmen/Organism/Object/Base/ObjectPresenter.php
```diff
@@ -1,31 +1,31 @@
...
 use App\DsAmen\Presenter;
 use App\DsShared\Helpers\HelperFactory;
 use App\Exception\InvalidOptionException;
+use BBC\ProgrammesPagesService\Domain\Entity\CoreEntity;
 use BBC\ProgrammesPagesService\Domain\Entity\Programme;
-use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeItem;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 
-class ProgrammePresenter extends Presenter
+abstract class ObjectPresenter extends Presenter
```
votemike: What is an Object? Is it a CoreEntity?

src/DsAmen/Organism/Object/Group/GroupPresenter.php
```diff
@@ -0,0 +1,63 @@
...
+            $this->programme,
+            $options
+        );
+    }
+
+    public function getImagePresenter(array $options = []): ImagePresenter
+    {
+        $options = array_merge($this->subPresenterOptions('image_options'), $options);
+
+        $options['cta_options'] = array_key_exists('cta_options', $options) ?
```
votemike: Is this set of options used?

src/DsAmen/Organism/Object/Base/SubPresenters/BodyPresenter.php
```diff
@@ -1,30 +1,30 @@
...
 use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeItem;
 use Cake\Chronos\Date;
 
-class ProgrammeBodyPresenter extends Presenter
+abstract class BodyPresenter extends Presenter
 {
-    /** @var Programme */
-    private $programme;
+    /** @var CoreEntity */
+    protected $programme;
```
votemike: `protected $programme;` => `protected $coreEntity;`?

src/DsAmen/Organism/Object/Shared/SubPresenters/SharedImagePresenter.php
```diff
@@ -0,0 +1,16 @@
...
+
+namespace App\DsAmen\Organism\Object\Shared\SubPresenters;
+
+use App\DsAmen\Organism\Object\Base\SubPresenters\CtaPresenter;
+use App\DsAmen\Organism\Object\Base\SubPresenters\ImagePresenter;
+use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeItem;
+
+class SharedImagePresenter extends ImagePresenter
+{
+    public function getCtaPresenter(): ?CtaPresenter
```
votemike: Are there no cta_options that might be passed in to this?

felipeparaujo: not for this one because this is just a getter, we don't instantiate the ctaPresenter in the image presenter

src/DsAmen/Organism/Object/Base/ObjectPresenter.php
felipeparaujo: In this case I meant for object to mean 'something that contains a core entity that gets displayed in a page', such as Programme Object, Collapsed Broadcast Object or Group Object

src/DsAmen/Organism/Object/Group/GroupPresenter.php
felipeparaujo: No, but this is what I do in the collapsed broadcast object to allow it to set options for the cta in the image presenter. I thought it would be a nice to have, in case we ever need it.

src/DsAmen/Organism/CoreEntity/Base/SubPresenter/BaseTitlePresenter.php
```diff
@@ -1,49 +1,47 @@
 <?php
-declare(strict_types = 1);
+declare(strict_types=1);
```
votemike: Why removing the spaces everywhere? I thought we always put spaces in?

src/DsAmen/Organism/CoreEntity/Programme/ProgrammePresenter.php
```diff
@@ -0,0 +1,76 @@
...
+use App\DsAmen\Organism\CoreEntity\Shared\SubPresenter\SharedTitlePresenter;
+use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeItem;
+
+class ProgrammePresenter extends BaseCoreEntityPresenter
+{
+    public function getBodyPresenter(array $options = []): BaseBodyPresenter
+    {
+        $options = array_merge($this->subPresenterOptions('body_options'), $options);
+
+        $options['cta_options'] = array_key_exists('cta_options', $options) ?
```
votemike: This is duplicated many places. Is it worth refactoring?

src/DsAmen/Organism/CoreEntity/Programme/ProgrammePresenter.php
```diff
@@ -0,0 +1,76 @@
...
+use App\DsAmen\Organism\CoreEntity\Programme\SubPresenter\CtaPresenter;
+use App\DsAmen\Organism\CoreEntity\Shared\SubPresenter\SharedImagePresenter;
+use App\DsAmen\Organism\CoreEntity\Shared\SubPresenter\SharedTitlePresenter;
+use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeItem;
+
+class ProgrammePresenter extends BaseCoreEntityPresenter
+{
+    public function getBodyPresenter(array $options = []): BaseBodyPresenter
+    {
+        $options = array_merge($this->subPresenterOptions('body_options'), $options);
```
votemike: This (or similar) is duplicated many places. Is it worth refactoring?

votemike:
Looks pretty good to me.

src/DsAmen/Organism/CoreEntity/Base/SubPresenter/BaseTitlePresenter.php
felipeparaujo: auto formatting. We're not that consistent with this actually. I could make it use spaces everywhere

src/DsAmen/Organism/CoreEntity/Programme/ProgrammePresenter.php
felipeparaujo: refactoring this would mean having a method with a single function call. it makes sense for the cta one, up above, but for this one I don't see much value

PR #265, opened 2017-10-20
PROGRAMMES-5955 fix missing translations from SchedulesOnNowController.php

Tasiobg:
Translations were not working beucase is required to render a page with "renderWithChrome" otherwhise the language is not set.


src/Controller/BaseController.php
```diff
@@ -98,6 +98,9 @@ protected function setContext($context)
         } elseif ($context instanceof Service) {
             $this->setBrandingId($context->getNetwork()->getOption('branding_id'));
             $this->fallbackBrandingId = 'br-00002';
+        } elseif ($context instanceof Network) {
```
BPScott: I might be being blind but if the context is a network won't we go into the first if statement, so this elseif doesn't do anything?

PR #268, opened 2017-10-23
Remove the 'sponsor' advert slot

BPScott:
AdOps have confirmed that the sponsor block shall not be sold again, so
there is no need to have a sponsor placeholder.


tatitati:
seeeeeeeeems ok

PR #269, opened 2017-10-23
PROGRAMMES-5755-Recommendations

mattgreenham:
This is the first stage in implementing https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5755
This contains the new RecEngService which ultimately outputs an array of programme objects


programmes-developers:

src/RecEng/RecEngService.php
```diff
@@ -0,0 +1,138 @@
...
+        int $limit
+    ): array {
+        $programmeEpisode = $this->getProgrammeEpisode($programme, $latestEpisode, $upcomingEpisode, $lastOnEpisode);
+
+        if (!$programmeEpisode) {
+            return [];
+        }
+
+        $programmePid = $programmeEpisode->getPid();
+        $recEngKey = $programmeEpisode->isTv() ? $this->videoKey : $this->audioKey;
```
felipeparaujo: @BPScott this always gets me confused. Should he be using `isTv()` or `isVideo()` in this case? I think this should be `isVideo()`

src/RecEng/RecEngService.php
```diff
@@ -0,0 +1,138 @@
...
+
+        $requestUrl = $this->recEngBase . '?key=' . $recEngKey . '&id=' . $programmePid;
```
felipeparaujo: Use the router for this. Register receng's endpoint as a 3rd party route and use it

src/RecEng/RecEngService.php
```diff
@@ -0,0 +1,138 @@
...
+
+        if ($response->getStatusCode() === 200) {
+            $body = $response->getBody()->getContents();
+            $result = json_decode($body, true);
+            $pids = $this->parseResult($result, $limit);
+
+            return $this->programmesService->findByPids($pids);
+        }
+
+        return [];
```
tatitati: this two return []  can be confuse at some point. They both return the same, but one of them is cause of a guzzle error.

src/DsAmen/Organism/Footer/FooterPresenter.php
```diff
@@ -17,11 +17,14 @@ class FooterPresenter extends Presenter
     /** @var Network */
     private $network;
 
-    public function __construct(Programme $programme, array $options = [])
+    private $recommendations;
```
votemike: Array of what?

src/RecEng/RecEngService.php
```diff
@@ -0,0 +1,138 @@
...
+    {
+        $this->client = $client;
+        $this->audioKey = $audioKey;
+        $this->videoKey = $videoKey;
+        $this->programmesService = $programmesService;
+    }
+
+    public function getRecommendations(
+        Programme $programme,
+        ?Programme $latestEpisode,
```
votemike: Should these 3 not all be `Episode`s?

src/RecEng/RecEngService.php
```diff
@@ -0,0 +1,138 @@
...
+            }
+        }
+
+        return null;
+    }
+
+    /**
+     * Returns an array of at most limit Pid objects from the given result response array
+     *
+     * @param array $results
```
votemike: An array of what?

src/RecEng/RecEngService.php
```diff
@@ -0,0 +1,138 @@
...
+        foreach ($result as $recItem) {
+
+            if (preg_match('/^urn:bbc:pips:[0-9b-df-hj-np-tv-z]{8,}$/', $recItem['ref'])) {
+                $pid = new Pid(substr($recItem['ref'], 13));
+                $pids[] = $pid;
+                $numResults++;
+            }
+
+            if ($numResults >= $limit) {
+                break;
```
votemike: `break`s are gross. `return $pids;` here.

src/RecEng/RecEngService.php
```diff
@@ -0,0 +1,138 @@
...
+
+        $requestUrl = $this->recEngBase . '?key=' . $recEngKey . '&id=' . $programmePid;
+
+        try {
+            $response = $this->client->request('GET', $requestUrl);
+        } catch (GuzzleException $e) {
+            return [];
+        }
+
+        if ($response->getStatusCode() === 200) {
```
votemike: Tiny thing, maybe switch this check.

src/RecEng/RecEngService.php
```diff
@@ -0,0 +1,138 @@
...
+
+            return $this->programmesService->findByPids($pids);
+        }
+
+        return [];
+    }
+
+    // recEng requires an episode pid, so this determines which to use based on the type of programme passed into it
+    // Takes nullable args of latest, upcoming and last on episodes as this is called in TLEC controller and these are already fetched
+    private function getProgrammeEpisode(Programme $programme, ?Programme $latestEpisode, ?Programme $upcomingEpisode, ?Programme $lastOnEpisode): ?Programme
```
votemike: Does this not return `?Episode`?

src/RecEng/RecEngService.php
```diff
@@ -0,0 +1,156 @@
...
+     * Will return an empty array if not results were found or RecEng cannot be reached
+     *
+     * @param Programme $programme
+     * @param Episode|null $latestEpisode
+     * @param Episode|null $upcomingEpisode
+     * @param Episode|null $lastOnEpisode
+     * @param int $limit
+     * @return Programme[]
+     */
+    public function getRecommendations(
```
votemike: Is the TLEC controller the only place that the RecEngService will be called? I feel that  the RecEngService should be generic and not need to accept a Programme and 3 Episodes. You should probably figure out which Programme you want in the controller and then pass that in to the RecEngService. That's if I'm understanding all of this correctly.

src/RecEng/RecEngService.php
```diff
@@ -0,0 +1,156 @@
...
+            UrlGeneratorInterface::ABSOLUTE_URL
+        );
+
+        try {
+            $response = $this->client->request('GET', $requestUrl);
+        } catch (GuzzleException $e) {
+            return [];
+        }
+
+        if (!$response->getStatusCode() == 200) {
```
votemike: Tiny thing, could also do `if ($response->getStatusCode() !== 200) {`

src/RecEng/RecEngService.php
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    /**
+     * Returns an array of at most limit Pid objects from the given result response array
+     *
+     * @param string[] $results
+     * @param int $limit
+     * @return Pid[]
+     */
+    private function parseResult(array $results, $limit = 10): array
```
votemike: I can't get my sandbox working at home in order to check the results returned from the RecEng, but this method doesn't feel quite right. Please wait before merging this.

src/RecEng/RecEngService.php
```diff
@@ -0,0 +1,156 @@
...
+use BBC\ProgrammesPagesService\Domain\Entity\Episode;
+use BBC\ProgrammesPagesService\Domain\Entity\Programme;
+use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeContainer;
+use BBC\ProgrammesPagesService\Domain\ValueObject\Pid;
+use BBC\ProgrammesPagesService\Service\ProgrammesService;
+use GuzzleHttp\Client;
+use GuzzleHttp\Exception\GuzzleException;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+
+class RecEngService
```
votemike: Is the THE MAIN RecEng file? If it is, could you put a tiny DocBlock saying the it is the Recommendation Engine. A la https://github.com/bbc/programmes-frontend/blob/develop/src/DsAmen/Organism/Map/MapPresenter.php#L24 and https://github.com/bbc/programmes-frontend/blob/develop/src/Controller/FindByPid/TlecController.php#L20

src/RecEng/RecEngService.php
mattgreenham: Sure

src/RecEng/RecEngService.php
```diff
@@ -0,0 +1,156 @@
...
+        } catch (GuzzleException $e) {
+            return [];
+        }
+
+        if (!$response->getStatusCode() == 200) {
+            return [];
+        }
+
+        $body = $response->getBody()->getContents();
+        $result = json_decode($body, true);
```
votemike: Any reason you're doing the `json_decode` outside of `parseResult()`?

votemike: Any reason you are passing `true` as the second param? I don't think it's needed.

src/RecEng/RecEngService.php
```diff
@@ -0,0 +1,156 @@
...
+    /**
+     * Returns an array of at most limit Pid objects from the given result response array
+     *
+     * @param string[] $results
+     * @param int $limit
+     * @return Pid[]
+     */
+    private function parseResult(array $results, $limit = 10): array
+    {
+        $result = $results['recommendations'];
```
votemike: `$recommendations = $results->recommendations;`?

src/RecEng/RecEngService.php
```diff
@@ -0,0 +1,156 @@
...
+        $pids = [];
+        $numResults = 0;
+
+        if ($limit <= 0) {
+            return [];
+        }
+
+        foreach ($result as $recItem) {
+            if (preg_match('/^urn:bbc:pips:[0-9b-df-hj-np-tv-z]{8,}$/', $recItem['ref'])) {
```
votemike: Why the preg match? Do recommendations also contain non-pids?

votemike:
```
$recommendation = str_replace('urn:bbc:pips:', '', $recItem->ref);
if ($this->isALegitPid($recommendation)) {
    $pids[] = new Pid($recommendation);
    $numResults++;
}

.... blah
```

src/RecEng/RecEngService.php
mattgreenham: So it returns an associative array

votemike: What's wrong with objects?

tests/RecEng/RecEngServiceTest.php
```diff
@@ -0,0 +1,169 @@
...
+
+        $mockEpisode = $this->createMock(Episode::class);
+        $mockRecProgrammeOne = $this->createMock(Programme::class);
+        $mockRecProgrammeTwo = $this->createMock(Programme::class);
+
+        $mockRecProgrammeOne->method('getPid')->willReturn(new Pid('p04vjd23'));
+        $mockRecProgrammeTwo->method('getPid')->willReturn(new Pid('p0576pm5'));
+
+        $programmesServiceResult = [$mockRecProgrammeOne, $mockRecProgrammeTwo];
+        $this->mockProgrammesService->method('findByPids')->with([new Pid('p04vjd23'), new Pid('p0576pm5')])->willReturn($programmesServiceResult);
```
BjornTwachtmann: expects($this->once())->method('findByPids')

BjornTwachtmann: That'll assert it's actually called that way

BjornTwachtmann: Probably a couple of other places in these tests that might be handy to add

src/RecEng/RecEngService.php
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    /**
+     * Returns an array of at most limit Pid objects from the given response
+     *
+     * @param string $response
+     * @param int $limit
+     * @return Pid[]
+     */
+    private function parseResult(string $response, $limit = 10): array
```
BjornTwachtmann: Probably sensible to set the default for limit on the public method, and make it 2, as that's how many we usually want

src/DsAmen/Organism/Footer/FooterPresenter.php
```diff
@@ -85,4 +89,12 @@ public function getFormats(): array
 
         return $formats;
     }
+//
```
BjornTwachtmann: Probably don't need to comment this if it's not called. I realise I told you to. The lesson here is don't listen to me.

programmes-developers:

tests/RecEng/RecEngServiceTest.php
```diff
@@ -0,0 +1,169 @@
...
+        return [
+            'video' => [true, 'imASecretVideoKey'],
+            'audio' => [false, 'imASecretAudioKey'],
+        ];
+    }
+
+    public function testResultLimit()
+    {
+        $response = new Response(200, [], "{\"recommendations\":[{\"ref\":\"urn:bbc:pips:p04vjd23\"},{\"ref\":\"urn:bbc:pips:p0576pm5\"}]}");
+        $this->mockHandler->append($response);
```
tatitati: morning @mattgreenham! I think would be great if we can extract this two lines (use in most of tests) to abstract this behaviour a little. Something like this I think would increase the readability:

```php
private function thirdPartyResplyWith($responseCode, $headers, $jsonResponse)
{

        $this->mockHandler->append(
               new Response($responseCode, $headers, $jsonResponse)
        );
}

File: src/RecEng/RecEngService.php [hunk 21]
```diff
@@ -0,0 +1,186 @@
...
+        int $limit
+    ): array {
+        $programmeEpisode = $this->getProgrammeEpisode($programme, $latestEpisode, $upcomingEpisode, $lastOnEpisode);
+
+        if (!$programmeEpisode) {
+            return [];
+        }
+
+        $programmePid = $programmeEpisode->getPid();
+        $recEngKey = $programmeEpisode->isVideo() ? $this->videoKey : $this->audioKey;
```
tatitati: I think would be useful to have a value object that keeps both keys: video and audio, instead of two different attributes (one per key). Is weird to see two key attributes. What about a unique $keyRecomStore?

In this way we can even avoid this check and make something like:
recEngKey = $this->keyRecomStore->getKeyForProgramme($programmeEpisode)

THe reason for this is that this protect the keys (a value object is immutable), and also both keys are a related  concept (so keeping both attributes together make sense). The cool thing about it is that this value object can have also its own policy to choose what key to use for a programme.

votemike: Sounds like a reasonable idea.

src/RecEng/RecEngService.php
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    /**
+     * Returns an array of at most limit Pid objects from the given response
+     *
+     * @param string $response
+     * @param int $limit
+     * @return Pid[]
+     */
+    private function parseResult(string $response, $limit): array
```
votemike: Typehint for limit?

tests/RecEng/RecEngServiceTest.php
```diff
@@ -0,0 +1,170 @@
...
+    {
+        return [
+            'video' => [true, 'imASecretVideoKey'],
+            'audio' => [false, 'imASecretAudioKey'],
+        ];
+    }
+
+    public function testResultLimit()
+    {
+        $this->queueMockResponse(200, [], "{\"recommendations\":[{\"ref\":\"urn:bbc:pips:p04vjd23\"},{\"ref\":\"urn:bbc:pips:p0576pm5\"}]}");
```
votemike: Would this be clearer if you made an array/object and json_encoded it?

src/RecEng/RecEngService.php
```diff
@@ -0,0 +1,186 @@
...
+
+            if ($numResults >= $limit) {
+                return $pids;
+            }
+        }
+
+        return $pids;
+    }
+
+    private function isValidPid(string $testPid): bool
```
tatitati: I think I would add a new static function to Pid value object:
```php
    public static function isValidPid($pid)
    {
           return preg_match('/^[0-9b-df-hj-np-tv-z]{8,}$/', $pid);
    }
```

Surely is useful.

src/RecEng/RecEngService.php
mattgreenham: After conversation with Bjorn, we've decided that we should leave as is, as this is the only place these keys are used and the only place these checks take place

src/RecEng/RecEngService.php
mattgreenham: Have updated and use existing validation in Pid value object :)

src/RecEng/RecEngService.php
```diff
@@ -0,0 +1,208 @@
...
+        $recEngRespose = $response->getBody()->getContents();
+
+        try {
+            $pids = $this->parseResult($recEngRespose, $limit);
+        } catch (ParseException $e) {
+            $this->logger->warning($e->getMessage() . "Url was: " . $requestUrl);
+            return [];
+        }
+
+        $result = ($pids === []) ? [] : $this->programmesService->findByPids($pids);
```
votemike: Slightly weird. But o well..

PR #270, opened 2017-10-23
Tidy

votemike:
A bit of tidying.  
Added a new partial template for a simple column message.  
Added an abstract base class for the right hand column presenters.


src/DsAmen/Organism/Map/SubPresenter/RightColumnPresenter.php
```diff
@@ -0,0 +1,40 @@
+<?php
+declare(strict_types = 1);
+
+namespace App\DsAmen\Organism\Map\SubPresenter;
+
+use App\DsAmen\Presenter;
+use App\Exception\InvalidOptionException;
+
+//@TODO do an abstract text template
+abstract class RightColumnPresenter extends Presenter
```
votemike: Should I add a constructor that accepts a ProgrammeContainer and sets it? Could get rid of a bit of logic from the child classes.

programmes-developers:

src/DsAmen/Organism/Map/SubPresenter/message.html.twig
```diff
@@ -0,0 +1,4 @@
+<div class="map__prose island">
```
BjornTwachtmann: Any idea what the execution penalty is on including this in a view rather than inline. I'm kind of conscious of the fact that our presentation layer is already far too deeply nested.

tests/DsAmen/Organism/Map/MapPresenterTest.php
```diff
@@ -97,21 +97,6 @@ public function testDefaultColumns()
         $this->assertColumns($presenter, [OnDemandPresenter::class]);
     }
 
-    public function testShowPromoPriority()
```
BjornTwachtmann: I get why you've removed this, but it's kind of useful to actually have this tested...

votemike: Can we test controllers? I could move it to a controller test.

src/DsAmen/Organism/Map/SubPresenter/message.html.twig
votemike: Not sure. Do we have a way to test it easily?

BjornTwachtmann: There's no prior art, no.

tests/DsAmen/Organism/Map/MapPresenterTest.php
BjornTwachtmann: You *can*. A full controller test is messy and slow though. 
You could mock the controller, mark the method public using reflection and test it that way maybe?

src/DsAmen/Organism/Map/SubPresenter/message.html.twig
votemike: Rendering the twig for the TLEC page seems to take 75.49ms. Of that, the OD column takes 2.82ms/4% and the TX column 1.01ms/1% when each of those columns use the message partial. Is that acceptable?

tests/DsAmen/Organism/Map/MapPresenterTest.php
votemike: It's not in a method at the moment. Do you think I should move it to one?

src/Controller/FindByPid/TlecController.php
```diff
@@ -52,22 +50,18 @@ public function __invoke(
...
             $galleries = $aggregationService->findDescendantGalleries($programme, 4);
         }
 
-        $upcomingBroadcast = null;
-        if ($programme->getAggregatedEpisodesCount()) {
-            $upcomingBroadcast = $collapsedBroadcastsService
-                ->findNextDebutOrRepeatOnByProgrammeWithFullServicesOfNetworksList($programme);
-        }
-
         $upcomingRepeatsAndDebutsCounts = $collapsedBroadcastsService->countUpcomingRepeatsAndDebutsByProgramme($programme);
```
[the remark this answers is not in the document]
votemike: I was going to, but `Broadcast`s (and `CollapsedBroadcast`s) are different things to `Episode`s so I'm not sure if that is the correct thing to do. Is it?

tests/DsAmen/Organism/Map/MapPresenterTest.php
votemike: Done

src/Controller/FindByPid/TlecController.php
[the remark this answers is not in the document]
BjornTwachtmann: Something can't have broadcasts if it doesn't have episodes, so moving it into the if block should be valid

PR #272, opened 2017-10-24
PROGRAMMES-5953 Change on-now controller to always return a 200

Tasiobg:
Change on-now controller to always return a 200 status code even when no broadcasts are found and set the cache-control to 300 seconds


src/Controller/SchedulesOnNowController.php
```diff
@@ -33,7 +33,10 @@ public function __invoke(
 
         $broadcast = $broadcastsService->findOnNowByService($network->getDefaultService());
         if (!$broadcast) {
-            throw $this->createNotFoundException('No broadcasts found.');
+            $response = $this->response();
+            $response->setStatusCode(200);
+            $response->setPublic()->setMaxAge(300);
+            return $response;
```
votemike: So, what does this return?

src/Controller/SchedulesOnNowController.php
```diff
@@ -33,7 +33,10 @@ public function __invoke(
 
         $broadcast = $broadcastsService->findOnNowByService($network->getDefaultService());
         if (!$broadcast) {
-            throw $this->createNotFoundException('No broadcasts found.');
+            $response = $this->response();
+            $response->setStatusCode(200);
+            $response->setPublic()->setMaxAge(300);
```
BPScott: The status is 200 by default, and setPublic() is already called in the constructor.All the setters on $response are a fluent interface so I believe this should suffice:

```
return $this->response()->setMaxAge(300);
```

Tasiobg: thanks!

programmes-developers:

src/Controller/SchedulesOnNowController.php
Tasiobg: this return a 200 response with no content

votemike: Should it be a 204 instead?

Tasiobg: Good point.
204 is usefull for APIs responses, in our case is an AJAX response and I'm not completly sure if using 204 is an improvement or not, seems a bit more clear to understand that the response was empty on purpose.
Anyone else wants to share their views?

votemike: Well, this is loaded by Javascript. How can you tell if it failed or just returned blank content?

Tasiobg: After having a chat with bjorn the 204 looks technically more correct but he is not too sure how the varnish layer will handle the 204 codes so the advice is to leave it as 200

PR #273, opened 2017-10-25
PROGRAMMES-5979 return 404 if week is invalid

Tasiobg:


programmes-developers:

tatitati:
We should cover this case with a test, please add in here the next line:

[SchedulesWeekByControllerTest](https://github.com/bbc/programmes-frontend/blob/develop/tests/Controller/SchedulesByWeekControllerTest.php#L33)
```php
'CASE 5: valid week for a year with ' => ['Invalid date', '2016/w53'],
```

programmes-developers:

PR #276, opened 2017-10-30
PROGRAMMES-5930 Adding Last On column

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5930

It looks like on LIVE the times in this column get localised. Do we want to do that as part of this ticket? Or a different ticket?


felipeparaujo:
Are any of the other columns localising as well? I don't think the Tx column is. If not, we should probably do all of them at once

votemike:
@felipeparaujo Yeah, the TX column also localises e.g. http://www.bbc.co.uk/programmes/n27vnphm. Let's do it in a different ticket....

votemike:
Localisation ticket: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6000

PR #278, opened 2017-10-31
Programmes-5755-RecEng

mattgreenham:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5755


programmes-developers:

src/RecEng/RecEngService.php
```diff
@@ -113,6 +123,7 @@ public function getRecommendations(
 
         $recEngRespose = $response->getBody()->getContents();
 
+
```
felipeparaujo: WHITESPCAE!!!!!!!11111!!!!

felipeparaujo:
looks alright, but as you renamed `getDuration()` to `showDuration()` you'll have to change that in the other amen objects that use the CTA (which is all of them)

mattgreenham:
From conversation IRL, `getDuration()` hasn't been removed so no need to change

src/RecEng/RecEngService.php
```diff
@@ -40,45 +42,54 @@ class RecEngService
...
      */
     public function getRecommendations(
         Programme $programme,
-        ?Episode $latestEpisode,
-        ?Episode $upcomingEpisode,
-        ?Episode $lastOnEpisode,
         int $limit = 2
     ): array {
-        $programmeEpisode = $this->getProgrammeEpisode($programme, $latestEpisode, $upcomingEpisode, $lastOnEpisode);
+        $programmeEpisode = $this->getProgrammeEpisode($programme);
```
mattgreenham: Is it worth also caching this result or is the PPS caching sufficient?

app/Resources/js/lazyload.js
```diff
@@ -126,6 +126,8 @@ define(['jquery-1.9'], function ($) {
                 success: function (data) {
                     var loaded_content = $($.parseHTML(data, document, true)).addClass(_this.options.lazy_css_state.loading);
                     content.replaceWith(loaded_content);
+                    loaded_content.addClass(_this.options.lazy_css_state.complete);
+                    loaded_content.removeClass(_this.options.lazy_css_state.loading);
```
votemike: Should these be in the order, or should they be swapped?

app/Resources/views/recommendations/ds2013_recommends_footer.html.twig
```diff
@@ -0,0 +1 @@
+<h2 class="footer__heading gel-double-pica-bold">{{- tr('you_might_also_like') -}}</h2>
```
votemike: Do 2013 pages not show recommendations?

src/Controller/RecommendationsController.php
```diff
@@ -0,0 +1,40 @@
...
+        $recommendations = $recEngService->getRecommendations($programme, 2);
+
+        if ($designSystem === 'ds_amen') {
+            return $this->renderWithChrome('recommendations/dsamen_recommends_footer.html.twig', [
+                'recommendations' => $recommendations,
+            ]);
+        }
+
+        if ($designSystem === 'ds_2013') {
+            return $this->renderWithChrome('ds2013_recommends_footer.html.twig', [
```
votemike: Does this not need to be prefixed with "recommendations/"?

tests/Controller/FindByPid/Tlec/TlecControllerClipsListTest.php
```diff
@@ -20,8 +20,8 @@ public function setUp()
     public function testBrandWithNoClipsDoesntShowGrid()
     {
         $crawler = $this->client->request('GET', '/programmes/prstdbrnd1');
-        $grid = $crawler->filter(".grid");
-        $this->assertEquals(2, $grid->count());
+        $grid = $crawler->filter(".leading-gap")->filter(".grid");
+        $this->assertEquals(0, $grid->count());
```
votemike: I believe there is an `assertCount` in PHPUnit that _might_ work here.

src/RecEng/RecEngService.php
```diff
@@ -40,45 +42,54 @@ class RecEngService
...
+    ) {
         $this->client = $client;
         $this->audioKey = $audioKey;
         $this->videoKey = $videoKey;
         $this->programmesService = $programmesService;
         $this->router = $router;
         $this->logger = $logger;
         $this->cache = $cache;
+        $this->collapsedBroadcastsService = $collapsedBroadcastsService;
+        $this->aggregationService = $aggregationService;
```
tatitati: I really think that this service is doing too much. Usually we have a service doing/fetching one thing. But this service is using more services. So this service has the responsibilities of: recEngService+CollapsedBroadcastService+aggregationsService.

I'm wondering if we can in this service just return the list of PIDs that we fetched from the HTTP request. 
And in the controller take these PIDs and invoke the rest of services. 

This simplify the design (more specialized) and is gonna be much easier to test. So far we have three different data sources that we have to stub in our mocks: from the HTTP requests, from the collapsedBroadcastRepository, and from AggregationRepository.

app/Resources/views/recommendations/dsamen_recommends_footer.html.twig
```diff
@@ -0,0 +1,44 @@
...
+                    'cta_options': {
+                        'cta_class': 'br-box-secondary',
+                        'show_duration': false,
+                    }
+                }) -}}
+            </li>
+        {%- endfor -%}
+    </ul>
+    <script>
+        require(['jquery-1.9','istats-1'], function($, istats) {
```
BjornTwachtmann: I'd probably omit this iStats code for now, and examine this when we come to refactor our iStats stuff. This is ugly as sin and I'd rather find a way to do this with data attrbitues (we might even already have one somewhere). 

Leave a note here saying todo instead?

programmes-developers:

src/DsAmen/_dsAmen-components.scss
```diff
@@ -50,6 +50,7 @@ $ie: false !default;
 //@import 'Molecule/Cover/cover';
 @import 'Molecule/icon';
 //@import 'Molecule/Letterbox/letterbox';
+@import 'Molecule/lazyload';
```
BjornTwachtmann: won't work I think

PR #279, opened 2017-10-31
PROGRAMMES-6003 fix for next week for edges cases

tatitati:
[PROGRAMMES-6003](https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6003)


programmes-developers:

src/Ds2013/Page/Schedules/ByWeekPage/SchedulesByWeekPagePresenter.php
```diff
@@ -78,6 +78,21 @@ public function getStartOfWeek(): Chronos
         return $this->startOfWeek;
     }
 
+    public function nextWeek(): Chronos
```
votemike: This looks like it is a bug in Chronos.... We should submit a patch to them instead of fixing it in our app.

src/Ds2013/Page/Schedules/ByWeekPage/schedules_by_week_page.html.twig
```diff
@@ -12,7 +12,7 @@
                     {{ ds2013_presenter(schedules_by_week_page.getDateListItem(offset)) }}
                 {% endfor %}
                 <th class="date-list__page box-link week-guide__table__nav br-box-highlight trail">
-                    <a id="next-week" class="br-page-bg-onbg--hover br-page-linkhover-ontext--hover text--no-ul text--shout" rel="next" data-href-add-utcoffset="true" href="{{ path('schedules_by_week', {pid: schedules_by_week_page.getService().getPid(), date: schedules_by_week_page.getStartOfWeek().addWeek()|local_date('Y/\\wW')}) }}">
```
BPScott: Spitballing, Could we fix this by changing this to `schedules_by_week_page.getEndOfWeek().addWeek()`?

tatitati: Yes, done. The problem is that there is another format to represent Years. ```Y``` and ```o```


| Format | Description | Example | 
| --- | --- | --- | 
| o | ISO-8601 week-numbering year. This has the same value as             Y, except that if the ISO week number             (W) belongs to the previous or next year, that year             is used instead. (added in PHP 5.1.0) | Examples: 1999 or 2003 | 
| Y | A full numeric representation of a year, 4 digits | Examples: 1999 or 2003 |

src/Controller/SchedulesVanityRedirectController.php
```diff
@@ -34,7 +34,7 @@ public function __invoke(Request $request, string $pid, string $vanity, UrlGener
                 $time = $time->subWeek();
             }
 
-            $params['date'] = $time->format('Y/\\wW');
+            $params['date'] = $time->format('o/\\wW');
```
votemike: To avoid this happening again with any future development, should we split this string out in to a constant somewhere?

tests/Controller/SchedulesVanityRedirectControllerTest.php
```diff
@@ -54,8 +55,45 @@ public function vanityTestProvider(): array
...
     }
 
+    /**
+     * @dataProvider edgeDatesProvider
+     */
+    public function testEdgeCasesForNextWeek($vanityWord, $currentWeek, $expectedWeek)
+    {
+        $date = $this->createDateFromWeekNumber($currentWeek);
+
+        ApplicationTime::setTime((Chronos::instance($date))->getTimestamp());
```
votemike: Why are you making a `DateTime` and then creating a `Chornos` from that? Why don't you just create the `Chronos` in the first place?

tests/Controller/SchedulesVanityRedirectControllerTest.php
```diff
@@ -54,8 +55,45 @@ public function vanityTestProvider(): array
...
+    }
+
     protected function tearDown()
     {
         ApplicationTime::blank();
     }
+
+    private function createDateFromWeekNumber($currentWeek)
+    {
+        $piecesDateUrl = explode("/", $currentWeek);
```
votemike: Tiny thing, maybe use `list()` here?

tests/Ds2013/Page/Schedules/ByWeekPage/SchedulesByWeekPagePresenterTest.php
```diff
@@ -0,0 +1,152 @@
...
+        // convert string url to date
+        $piecesDateUrl = explode("/", $stringDateUrl);
+        $year = $piecesDateUrl[1];
+        $weekNumberOfYear = str_replace('w','',$piecesDateUrl[2]);
+        $date = new DateTime();
+        $date->setISODate($year, $weekNumberOfYear, 1); //year , week num of year , day of week
+
+        // set current date in presenter
+        $this->weekPresenter = new SchedulesByWeekPagePresenter(
+            $this->buildService(), Chronos::instance($date), [], '', []
```
votemike: Again, why are you passing a DateTime into a Chronos?

tests/Ds2013/Page/Schedules/ByWeekPage/SchedulesByWeekPagePresenterTest.php
```diff
@@ -0,0 +1,152 @@
...
+use BBC\ProgrammesPagesService\Domain\Entity\Service;
+
+class SchedulesByWeekPagePresenterTest extends TestCase
+{
+    private $weekPresenter;
+
+    public function testNormalBehavioursForADayInMiddleOfTheYear()
+    {
+        $this->userVisitWeek('/2012/w33');
+        $this->assertEquals('/2012/w34', $this->userPassToNextWeek());
```
votemike: Does this test pass?

tatitati: Yes. Actually this test is not testing anything special. Is just a normal case, a week in the middle of a month with no possibilities of changing the year.

tests/Ds2013/Page/Schedules/ByWeekPage/SchedulesByWeekPagePresenterTest.php
tatitati: my goal was create a date based on the year and week number instead of the classic "day-month-year". In this way I can write a test-case more similar to the real world. 
I didn't find how to do this directly with chronos, but because is not production code I thought that I shouldn't be too worried about being nice when writing a helper in tests. The way I build the date is completely secondary.

src/Controller/SchedulesVanityRedirectController.php
tatitati: Not happy with that idea. I found that date-expression in just a couple of places, so is not being used too much. Also I consider that for the future we should have tests covering this cases. In the presenter I couldn't find any tests, so I wrote the first ones. We should be aware of this.

tests/Ds2013/Page/Schedules/ByWeekPage/SchedulesByWeekPagePresenterTest.php
```diff
@@ -0,0 +1,165 @@
+<?php
```
BPScott: Is this a useful test? We've not changed anything in this PR that affects the Presenter (the fix was in the template). A test in the SchedulByWeekControllerTest might be handy though

programmes-developers:

PR #281, opened 2017-11-02
fix tests

tatitati:
We have some tests complaining like:
```
There were 6 warnings:

1) Tests\App\RecEng\RecEngServiceTest::testParsesRecEngResponseCorrectly
Trying to configure method "findByPids" which cannot be configured because it does not exist
....
```

**The problem** is that none release was created in PPS when the new functionality 'findByPids' was merged into master. So programmes-frontend didn't update to the version of PPS that include this function.

**The fix:** PPS version updated from v3.1.1 to v3.2.0


programmes-developers:

PR #283, opened 2017-11-07
Timezone js fixes

BjornTwachtmann:
## Why this PR:
So far the only context we use to set UTC offset in our schedules links is the current user time.
This means that if the user is in Jun(BST, UTC+1), then this user will get all the links with UTC+1 (even for the schedule of December, which is out of BST).

## The fix:
We want to use "link-date-context". And for this we need to process the date of every ```/schedules/<pid>/yyyy/mm/dd``` link.

## Exactly what links are gonna be processed?
    In schedules by day:
     - links in days-nav bar
     - links In the bottom section "Please choose the appropriate schedule"

    In schedules by week:
     - links in week-days nav bar

For more details: [PROGRAMMES-5964](https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5964)


app/Resources/js/timezone-notification.js
```diff
@@ -187,12 +176,76 @@ define('timezone-notification', ['jquery-1.9', 'rv-bootstrap'], function ($) {
...
+         * Convert string in format yyyy/mm/dd to Date
+         *
+         * @param {string}
+         * @returns {Date|null}
+         */
+        convertStringToDate: function(stringWithDate) {
+            var dateRegex = new RegExp('[0-9]{4}\/[0-9]{2}\/[0-9]{2}');
+            if (dateRegex.test(stringWithDate)) {
+                var pieces = stringWithDate.split('/');
+                var year = parseInt(pieces[0]);
```
BjornTwachtmann: should be parseInt(thing, 10); JS is weird.

app/Resources/js/timezone-notification.js
```diff
@@ -178,6 +164,9 @@ define('timezone-notification', ['jquery-1.9', 'rv-bootstrap'], function ($) {
             // Trigger lazyload.js to load in any newly created lazyloads
             $('body').trigger('lazyload-refresh');
         },
+        /**
+         * rewrite url of links containing "yyyy/mm/dd" with the new utcoffset param
+         */
         rewriteScheduleLinks: function (content) {
             if (!this.gmtOffset) {
```
BjornTwachtmann: This needs to go now really.

app/Resources/js/timezone-notification.js
```diff
@@ -187,12 +176,76 @@ define('timezone-notification', ['jquery-1.9', 'rv-bootstrap'], function ($) {
...
+                    var extractedDate = _this.extractDateFromLink(href);
+                    if (extractedDate != null && extractedDate instanceof Date) {
+                        offset =  _this.getGmtOffset(extractedDate);
+                    }
+
                     // append ?utcoffset=xx:yy to any relevant URLs to prevent a double load
-                    $this.attr('href', _this.rewriteUrlWithUtcOffset(href));
+                    // we only want to rewrite links that has different utcoffset. The utcoffset that will
+                    // be added will be the given by the link date.
+                    if (offset != '') {
```
BjornTwachtmann: if (offset)
should be fine here

app/Resources/js/timezone-notification.js
```diff
@@ -187,12 +176,76 @@ define('timezone-notification', ['jquery-1.9', 'rv-bootstrap'], function ($) {
             scheduleLinks.each(function () {
                 var $this = $(this);
                 var href = $this.attr('href');
+                // Here look for links of the format /schedules/pid/yyyy/mm/dd and get local time from yyyy mm dd
+                // and put into offset
                 if (href) {
+                    var offset = null;
+                    var extractedDate = _this.extractDateFromLink(href);
+                    if (extractedDate != null && extractedDate instanceof Date) {
```
BjornTwachtmann: can't we just say extractedDate instanceof Date ?

src/Ds2013/Page/Schedules/ByDayPage/schedules_by_day_page.html.twig
```diff
@@ -63,7 +63,12 @@
 {% endif %}
 
 {% if schedules_by_day_page.hasBroadcasts() %}
-    <ol class="list-unstyled g-c-l">
+    {% set routeDate = '' %}
+    {% if schedules_by_day_page.getRouteDate() != null %}
+        {% set routeDate = schedules_by_day_page.getRouteDate()|date('Y/m/d') %}
```
BjornTwachtmann: This is wrong. You're passing a string into a date formatting function. Move all of this into the presenter.

programmes-developers:

app/Resources/js/timezone-notification.js
```diff
@@ -178,21 +190,82 @@ define('timezone-notification', ['jquery-1.9', 'rv-bootstrap'], function ($) {
...
                 }
             });
+        },
+        /**
+         * Check if url link has a date yyyy/mm/dd, and convert it to Date
+         * @param {string} href - Expected format: http...../schedules/<pid>/yyyy/mm/dd
+         * @returns {(Date|null)}
+         */
+        extractDateFromSchedulesLink: function (href) {
+            var urlHrefRegex = new RegExp('.*\/schedules\/[0-9b-df-hj-np-tv-z]{8,15}\/([0-9]{4}\/[0-9]{2}\/[0-9]{2})$');
```
BjornTwachtmann: The .* at the beginning of this regex isn't needed

src/Ds2013/Page/Schedules/ByDayPage/SchedulesByDayPagePresenter.php
```diff
@@ -189,6 +189,19 @@ public function hasSiblingServiceLink(): bool
...
+     * Expected output in format data-page-time="2014/08/23"
+     * @return string
+     */
+    public function dataPageTimeAttr(): string
+    {
+        if (is_null($this->getRouteDate())) {
+            return '';
+        }
+
+        return 'data-page-time=' . $this->getRouteDate();
```
BjornTwachtmann: You're missing double quotes around $this->getRouteDate. This should read
return 'data-page-time="' . htmlspecialchars($this->getRouteDate()) . '"';

BjornTwachtmann:
Other than the 2 comments this looks good now. Make sure you get the testers to regression test this on multiple browsers/devices though as this could break some shit.

src/Ds2013/Page/Schedules/ByDayPage/SchedulesByDayPagePresenter.php
tatitati: yeah, is not casual. If I add the quotes then in the template I have this:
```html
<ol data-page-time=""2017/05/22"">
```

src/Ds2013/Page/Schedules/ByDayPage/SchedulesByDayPagePresenter.php
```diff
@@ -189,6 +189,19 @@ public function hasSiblingServiceLink(): bool
...
+     * Expected output in format data-page-time="2014/08/23"
+     * @return string
+     */
+    public function dataPageTimeAttr(): string
+    {
+        if (is_null($this->getRouteDate())) {
+            return '';
+        }
+
+        return 'data-page-time="' . $this->getRouteDate() . '"';
```
BjornTwachtmann: This still needs to be wrapped in htmlspecialchars...

tests/Controller/SchedulesByDayControllerTest.php
```diff
@@ -298,6 +298,34 @@ public function invalidsUtcOffsetsProvider(): array
...
+        $this->assertEquals(1, $crawler->filter('[data-page-time]')->count());
+        $this->assertEquals('2017/05/22', $crawler->filter('[data-page-time]')->attr('data-page-time'));
+    }
+
+    public function testDataPageTimeIsNotAddedInHtmlWhenNoRouteDateIsPassedInUrl()
+    {
+        $this->loadFixtures(["BroadcastsFixture"]);
+
+        $timeNow = '2017/05/22';
+        if (!is_null($timeNow)) {
```
votemike: Why would `$timeNow` be null?

tatitati: oups!, because I copied/pasted too much (trying to write quick tests), my fault. Good spot

PR #287, opened 2017-11-09
Electron models

BjornTwachtmann:
And circuit breakers.


programmes-developers:

app/config/services.yml
```diff
@@ -192,7 +196,13 @@ services:
         arguments:
             - '@null_cache_service_factory'
 
-    App\RecEng\RecEngService:
+    # API services
+    App\ExternalApi\Electron\ElectronService:
+        class: '%app.electron.service.class%'
+        arguments:
+            $env: 'live'
```
votemike: Does this also need to be 'test' in the test config or anything like that?

src/ExternalApi/CircuitBreaker/CircuitBreaker.php
```diff
@@ -0,0 +1,111 @@
...
+        if ($this->apcu->fetch($this->makeApcKey('is_open'))) {
+            $this->logger->warning('CircuitBreaker logFailure called when breaker open. Something is wrong in calling code');
+            return;
+        }
+        $failureCountKey = $this->makeApcKey('failure_count');
+
+        // This is potentially leaky
+        // multiple requests running at the same time could end up setting $countCache to 1
+        // But for the purposes of this exercise, 99% accurate is good enough
+        if (!$this->apcu->exists($failureCountKey)) {
```
votemike: Maybe flip this check so that the else case isn't a double negative?

src/ExternalApi/CircuitBreaker/CircuitBreaker.php
```diff
@@ -0,0 +1,111 @@
...
+        $maxFailures = $this->maxFailsPerMinute * (self::MONITORING_INTERVAL / 60);
+        $failureCount = (int) $this->apcu->fetch($failureCountKey);
+        if ($failureCount > $maxFailures) {
+            $this->logger->error("Closing circuit breaker for $this->apiName due to failureCount ($failureCount) > maxFailureCount ($maxFailures");
+            $this->metricsManager->addApiCircuitBreakerOpenMetric($this->apiName);
+            $this->apcu->store($this->makeApcKey('is_open'), 'true', $this->secondsToOpenWhenFailed);
+        }
+    }
+
+    public function isOpen(): bool
```
votemike: Maybe a docblock to explain the "Open" means that we are not going to do that call?

src/ExternalApi/Electron/ElectronService.php
```diff
@@ -0,0 +1,111 @@
...
+use BBC\ProgrammesPagesService\Domain\ValueObject\Pid;
+use GuzzleHttp\Client;
+use GuzzleHttp\Exception\ClientException;
+use GuzzleHttp\Exception\GuzzleException;
+use GuzzleHttp\Psr7\Response;
+use Psr\Log\LoggerInterface;
+
+class ElectronService
+{
+    private const ELECTRON_WEBSERVICE_BASEURL = 'https://api.live.bbc.co.uk/electron';
```
votemike: Should these be params in config files for the different environments?

src/ExternalApi/Electron/ElectronService.php
```diff
@@ -0,0 +1,111 @@
...
+    private $xmlParser;
+
+    /** @var SupportingContentMapper */
+    private $supportingContentMapper;
+
+    /** @var LoggerInterface */
+    private $logger;
+
+    public function __construct(
+        Client $guzzleClient,
```
votemike: Should this be a `GuzzleHttp\ClientInterface` so that we can mock it etc..?

src/ExternalApi/Electron/ElectronService.php
```diff
@@ -0,0 +1,111 @@
...
+        SupportingContentMapper $supportingContentMapper,
+        LoggerInterface $logger,
+        string $env = 'live'
+    ) {
+        $this->guzzleClient = $guzzleClient;
+        $this->cache = $cache;
+        $this->xmlParser = $xmlParser;
+        $this->supportingContentMapper = $supportingContentMapper;
+        $this->logger = $logger;
+        $this->env = $env;
```
votemike: Is feels weird to pass an environment in to a class. Should you not be passing in the config value based on the environment instead?

src/ExternalApi/Electron/ElectronService.php
```diff
@@ -0,0 +1,111 @@
...
+            $baseUrl = self::ELECTRON_WEBSERVICE_BASEURL_TEST;
+        } elseif ($this->env === 'int') {
+            $baseUrl = self::ELECTRON_WEBSERVICE_BASEURL_INT;
+        }
+        $url = $baseUrl . '/atom/v1/programmessupportingcontent';
+        $url .= '/' . urlencode($lastLetterOfPid) . '/' . urlencode($pid);
+        return $url;
+    }
+
+    private function parseResponseBody(string $responseBody)
```
votemike: Typehint for what this returns? It looks like an array of something, but I'm not sure what.

src/ExternalApi/XmlParser/XmlParser.php
```diff
@@ -0,0 +1,58 @@
...
+
+namespace App\ExternalApi\XmlParser;
+
+use App\ExternalApi\Exception\ParseException;
+use SimpleXMLElement;
+use LibXMLError;
+
+class XmlParser
+{
+    public function parse(string $body, array $config = []): SimpleXMLElement
```
votemike: We have an XmlParser in our Github repo, does that not do what needs to be done instead of writing a new one?

app/config/services.yml
BjornTwachtmann: Hmm, good point, it's currently test in the forge test env. Will talk to Christain about whether that's useful.

src/ExternalApi/XmlParser/XmlParser.php
BjornTwachtmann: The XmlParser in that repo is designed for ease of development. It is however not efficient, it creates a new PHP object for every XML element. 

It's going to be vastly more efficient to use the built in SimpleXml stuff, which this is a thin wrapper/error handler around. Unfortunately.

src/ExternalApi/Electron/ElectronService.php
BjornTwachtmann: I was kind of wondering this. 
We could move our endpoints into cosmos config. The advantage to that would be ease of overriding/changing them. The disadvantage is it's more annoying to setup and maintain.

Could also just live in the config.yml. Or here. 

Any thoughts @BPScott

app/config/services.yml
BjornTwachtmann: Christian says it's useful to have the test content on test. So I'll change that once we decide where these URLs are coming from...

src/ExternalApi/Electron/ElectronService.php
BPScott: We'd need to override in the INT/TEST environments anyway, having that be `test` or a url feels like much of a muchness, so we might as well go for the url as it'll means less string fiddling in the class

src/ExternalApi/CircuitBreaker/CircuitBreakerFactory.php
```diff
@@ -0,0 +1,62 @@
...
+        $this->logger = $logger;
+        $this->apcu = $apcu;
+    }
+
+    public function getBreakerFor(string $apiName): ?CircuitBreaker
+    {
+        if (!ApiTypeEnum::isValid($apiName)) {
+            throw new InvalidArgumentException("$apiName is not a valid API name");
+        }
+        if (!isset(self::BREAKER_PARAMETERS[$apiName])) {
```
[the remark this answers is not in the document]
BjornTwachtmann: Not all valid ApiTypeEnum values get a circuit breaker. Branding for instance. But all invalid ApiTypeEnum values throw an exception. There's a difference between the behaviours. 

Basically this will tell you in development if you've made a mess this way.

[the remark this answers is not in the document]
BPScott: Gotcha, thanks

app/config/config_prod_int.yml
```diff
@@ -5,3 +5,8 @@ parameters:
     app.feed_env.orbit: 'test'
     app.feed_env.branding: 'live'
 
+services:
+    App\ExternalApi\Electron\Service\ElectronService:
+        class: '%app.electron.service.class%'
+        arguments:
+            $baseUrl: 'https://api.test.bbc.co.uk/electron'
```
BPScott: A parameter rather than redefining the service perhaps? 

```
parameters:
    app.electron.base_url: 'https://api.test.bbc.co.uk/electron'
```

src/ExternalApi/CircuitBreaker/CircuitBreaker.php
```diff
@@ -82,7 +82,10 @@ public function logFailure(): void
         $maxFailures = $this->maxFailsPerMinute * (self::MONITORING_INTERVAL / 60);
         $failureCount = (int) $this->apcu->fetch($failureCountKey);
         if ($failureCount > $maxFailures) {
-            $this->logger->error("Closing circuit breaker for $this->apiName due to failureCount ($failureCount) > maxFailureCount ($maxFailures");
+            $this->logger->error(
+                "Closing circuit breaker for $this->apiName for $this->secondsToOpenWhenFailed " .
+                "due to failureCount ($failureCount) > maxFailureCount ($maxFailures)."
```
BPScott: Should we stick "seconds" in there otherwise it will be unitless fun: "Closing circuit breaker for thing for 30 due to failureCount"?

src/ExternalApi/Electron/Service/ElectronService.php
```diff
@@ -0,0 +1,109 @@
...
+     * @param string $responseBody
+     * @return SupportingContentItem[]
+     */
+    private function parseResponseBody(string $responseBody): array
+    {
+        $simpleXml = $this->xmlParser->parse($responseBody);
+        if (!isset($simpleXml->content->pages)) {
+            throw new ParseException("Electron XML response does not contain pages element");
+        }
+        return $this->supportingContentMapper->mapItems($simpleXml->content->pages);
```
BPScott: The convention that we have elsewhere is that a mapper maps a single element to a single domain object and that it is the responsibility of the the thing that calls the mapper to do the looping through to generate an array of stuff.

I reckon mapItems should be in this file rather than the mapper.

PR #289, opened 2017-11-13
Start for the new section "styleguide"

tatitati:
### Added new routes:
 - "..../styleguide/amen"
 - ".../styleguide/amen/promotion"

### Added new builders for:
 - Promotion

And additional builders to assemble easily a Promotion such as:
 - related links
 - episode
 - image

### Created new structure folder:
Templates
```
app/Resources
└── styleguide
    └── amen
        ├── base.html.twig
        ├── intro.html.twig
        └── promotion.html.twig
```
Controllers
```
src/Controller
├── Styleguide
│   └── Amen
│       ├── IntroController.php
│       └── PromotionController.php
```
Builders
```
src/Builders
├── BuilderInterface.php
├── EpisodeBuilder.php
├── ImageBuilder.php
├── PromotionBuilder.php
└── RelatedLinkBuilder.php
```
Routes
```
app/config
└── styleguide.yml
```
Tests
```
tests/Controller/Styleguide/Amen
└── Styleguide
    └── Amen
        ├── IntroControllerTest.php
        └── PromotionControllerTest.php
```
---------
no Tests added yet. Firstly I want to make sure that the people is happy with this approach. Test will be quick to add then.


app/Resources/views/styleguide/amen/base.html.twig
```diff
@@ -0,0 +1,12 @@
+{% extends'base_ds_amen.html.twig' %}
+
+{% block title %}BBC - Programmes Styleguide{{ block('section_name') is defined ? ' - ' ~ block('section_name') }}{% endblock %}
+
+{% block body %}
+    <div class="b-g-p">
+        <h1>Styleguide/AMEN</h1>
+        <h1>{% block section_name %}{% endblock %}</h1>
```
votemike: 2 H1s?

app/config/routing.yml
```diff
@@ -75,3 +75,6 @@ cloud_labs:
 # Include any 3rd party routes. This should be the last thing in the file
 _3rd_party:
     resource: routing_3rd_party.yml
+
+styleguie:
```
votemike: styleguide

src/DTO/RenderDescriptorDto.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+namespace App\DTO;
+
+/**
+ * This class contains info about an item that is bein displayed.
+ */
+class RenderDescriptorDto
```
votemike: What's DTO?

app/Resources/views/styleguide/amen/base.html.twig
tatitati: yeah, im not concern at this point about how it looks. The UI will be improved in later PR's. For now I just want an skeleton

src/Controller/Styleguide/PromotionController.php
```diff
@@ -0,0 +1,118 @@
...
+     * Use case: The user see the same promotions, same fields, but the layout is different.
+     * List of options to customize render of promotion:
+     * @see \App\DsAmen\Organism\Promotion\PromotionPresenter
+     * @return RenderDescriptorDto[]
+     */
+    private function buildPromotionsWithDifferentHtmlAttributes()
+    {
+        $defaultPromotion = PromotionBuilder::default()->build();
+
+        return [
```
votemike: Not sure what PHPs `yield` is actually used for, but would it be applicable here?

src/DTO/RenderDescriptorDto.php
tatitati: is Data transfer object. Data transfer objects are objects with no behaviours, the classic "bag of properties". I thought that a render descriptor have semantic meaning and would fit much better than a simple array that I pass to the template. A descriptor make more sense. These TDO doesn need tests (no behaviours), and are used to pass data between processes/layers/systems. So passing this from the controller to the layout also helped me to decide to use them.

src/Controller/Styleguide/PromotionController.php
tatitati: I thought about it. No, yield is for sequencial extraction. In this case I want to pass a pack of data in once.

programmes-developers:

votemike:
Do you mention the defaults anywhere?

src/Builders/RelatedLinkBuilder.php
```diff
@@ -0,0 +1,90 @@
...
+
+    /** @var string */
+    private $type;
+
+    /** @var bool */
+    private $isExternal;
+
+    private function __construct()
+    {
+        $this->title = 'my title';
```
votemike: `my external link`?

tatitati: yeah, can be handy. Also because we make a distinction between both maybe we can create different function:
```php
relatedLink:internal()->build()
relatedLink:external()->build()
```

votemike: OK, maybe tweak them after constructing the default one but before you build?

app/config/styleguide.yml
```diff
@@ -0,0 +1,9 @@
+styleguide_intro:
+    path: /programmes/styleguide/{engine}
+    defaults: { _controller: App\Controller\Styleguide\Amen\IntroController, engine: amen}
```
BPScott: As we're doing dedicated controllers for Amen and 2013, the engine variable is no longer used, so we can remove it

src/Controller/Styleguide/Amen/PromotionController.php
```diff
@@ -0,0 +1,118 @@
+<?php
+declare(strict_types = 1);
+namespace App\Controller\Styleguide\Amen;
+
+use Tests\App\DataFixtures\Builders\EpisodeBuilder;
```
BPScott: The Tests namespace is only mapped in development, not in production. If you want to use your builders you'll need to create them in the `src` folder. Otherwise this will throw errors when deployed

src/Controller/Styleguide/Amen/PromotionController.php
```diff
@@ -0,0 +1,118 @@
...
+     * List of options to customize render of promotion:
+     * @see \App\DsAmen\Organism\Promotion\PromotionPresenter
+     * @return array
+     */
+    private function buildPromotionsWithDifferentHtmlAttributes()
+    {
+        $defaultPromotion = PromotionBuilder::default()->build();
+
+        return [
+            $this->toArray(
```
BPScott: Having this as a regular array where the key is the description, and the value is an array of arguments that get passed into a presenter feels a lot simpler than this indirection

src/Controller/Styleguide/Amen/PromotionController.php
tatitati: wow, thanks @BPScott, this was really a good spot

PR #290, opened 2017-11-14
Supporting content

BjornTwachtmann:
HTML/CSS/Presenters for supporting content. See PROGRAMMES-5752.

### DIFF
--- a/app/Resources/views/find_by_pid/tlec.html.twig
+++ b/app/Resources/views/find_by_pid/tlec.html.twig
@@ -137,7 +137,13 @@
                     {%- endfor -%}
                 {%- endif -%}
             </ul>
-            {# TODO Supporting Content blocks #}
+            <ul class="grid">
+                {%- for supportingContentItem in supportingContentItems -%}
+                    <li class="1/1 1/2@gel5 grid__item">
+                        {{- ds_amen('supportingContent', supportingContentItem) -}}
+                    </li>
+                {%- endfor -%}
+            </ul>
 
             {# TODO Regular Vote block #}
         </div>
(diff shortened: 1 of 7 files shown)


app/Resources/views/find_by_pid/tlec.html.twig
```diff
@@ -137,7 +137,15 @@
                     {%- endfor -%}
                 {%- endif -%}
             </ul>
-            {# TODO Supporting Content blocks #}
+            {%- if supportingContentItems -%}
+                <ul class="grid">
+                    {%- for supportingContentItem in supportingContentItems -%}
+                        <li class="1/1 1/2@gel4 grid__item">
```
BPScott: Don't need `1/1` as `grid_item` is width:100% by default

src/DsAmen/Organism/SupportingContent/_supporting-content.scss
```diff
@@ -0,0 +1,26 @@
+.feature {
```
BPScott: The CSS block should should match the filename, it's confusing when the names don't line up

src/DsAmen/Organism/SupportingContent/supporting_content.html.twig
```diff
@@ -0,0 +1,21 @@
+{%- set grid_classes = "feature feature--noimage" -%}
+{%- if supporting_content.showImage() -%}
+    {%- set grid_classes = "feature grid" -%}
+{%- endif -%}
+<div class="br-box-secondary br-secondary-link">
+    <div class="{{ grid_classes }}">
+        {%- if supporting_content.showImage() -%}
+            <div class="grid__item feature__item 1/1 1/2@gel2b br-box-subtle">
```
BPScott: If we use flexbox for positioning the text and  image (configuring the text container to take up as much space as is available), then we could get rid of having to use the grid for this layout, which would save on having to do ths conditional logic around the grid__item classes.

BjornTwachtmann: You got an example? That media object seems to be where sanity goes to die.

PR #291, opened 2017-11-14
ImagesFixture now using builders

tatitati:
ImagesFixture now using builders


programmes-developers:

tests/DataFixtures/PagesService/ImagesFixtures.php
```diff
@@ -3,92 +3,78 @@
...
-            'BBC One Logo',
-            'standard',
-            'jpg'
-        );
+        return ImageBuilder::default()
+           ->withPid('p02lrnkz')
+           ->withTitle('BBC One Logo')
+           ->withShortSynopses('BBC One Logo')
+           ->withLongestSynopsis('BBC One Logo')
+           ->build();
```
BjornTwachtmann: Your default type in ImageBuilder is 'this is the type'. So the objects you're creating aren't the same. I would probably change that to 'standard' in the ImageBuilder class.

tatitati: my bad!, it was in my mind, but I forgot it, sorry

PR #292, opened 2017-11-14
episodes builder applied

tatitati:
episodes builder applied


src/Builders/EpisodeBuilder.php
```diff
@@ -235,79 +235,109 @@ public function withOptions(Options $options)
         return $this;
     }
 
-    public function withParent(Programme $parent)
+    /**
+     * @param Programme|null $parent
```
BPScott: These docblocks don't provide any new information as the info is already there in the function definition. They're noise we have to scroll through, so get rid of them

tests/DataFixtures/PagesService/EpisodesFixtures.php
```diff
@@ -3,88 +3,86 @@
...
-            new DateTimeImmutable('2017-07-01 20:00:00') //Streamable until
-        );
+        return EpisodeBuilder::default()
+            ->withDbAncestryIds([1])
+            ->withPid('p0000001')
+            ->withTitle('An Episode of Eastenders')
+            ->withSearchTitle('An Episode of Eastenders')
+            ->withSynopses(new Synopses('Short Synopsis', 'Medium Synopsis', 'Long Synopsis'))
+            ->withImage(ImagesFixtures::eastenders())
+            ->withPromotionsCount(0)
```
BPScott: Are things like the zero counts already the default in the builders? If so we should remove them to reduce noise

programmes-developers:

src/Builders/EpisodeBuilder.php
tatitati: true

tests/DataFixtures/PagesService/EpisodesFixtures.php
tatitati: well Ben, I really like your comment because yesterday I was thinking about how much is worth to this modification, it was in my mind to drop some comments about this today. 

From my point of view there is not even point of doing this with a Builder, I know that we discussed to do it. A builder is to create objects based on very specific tweaks (an small amount, they are designed to gain time, not to specify many attributes). In this case we are using them to create completely different objects for our database test, so we have to specify in our builder every parameter, and in this case we don't take advantage of builders at all. I would like to listen to @BjornTwachtmann about this.

Cause of this  I deleted the part I changed in EpisodesFixture.php to keep it as it was (a simple constructor). I just left in this PR some simple modifications learned during the process. Let me know any point of view, I will be happy to proceed in any way.

src/Builders/EpisodeBuilder.php
```diff
@@ -109,7 +109,7 @@ private function __construct()
         $this->pid = new Pid('d00744wz');
         $this->title = 'my episode title';
         $this->searchTitle = 'my search episode title';
-        $this->synopses = new Synopses('My short synopsis', 'my medium synopsis', 'my long synopsis');
+        $this->synopses = new Synopses('My short synopsis', 'my a very medium no too much synopsis', 'my extremely boring and endless text for my long synopsis');
```
tatitati: this change is not necessary for this PR, but I thought that was so tiny....I just thought that the length of the synopses might be a little improved to make them more obvious. Simple

src/Builders/EpisodeBuilder.php
```diff
@@ -235,79 +235,73 @@ public function withOptions(Options $options)
...
-     */
-    public function withFormats(array $formats)
+    public function withFormats(?array $formats)
     {
         $this->formats = $formats;
         return $this;
     }
 
-    public function withFirstBroadcastDate(DateTimeImmutable $firstBroadcastDate)
+    public function withFirstBroadcastDate(?DateTimeImmutable $firstBroadcastDate)
```
votemike: Why is this being changed to accept null?

tatitati: firstBoradcastDate is an optional parameter in the Episode constructor, so it can be null:
https://github.com/bbc/programmes-pages-service/blob/master/src/Domain/Entity/Episode.php#L44

But in our builder by default we create an object with a value different to null
```php
$this->firstBroadcastDate = new DateTimeImmutable('2012-06-05');
```

Because in some cases we will want to build it with firstBroadcastDate=null our EpisodeBuilder need to set up this value as well.

BjornTwachtmann:
This looks good mate, as long as the tests pass. One suggestion for the EpisodeBuilder. I think you're probably defaulting a bit too much stuff to being set. Some of the properties you're setting as default are obscure and unlikely to come up too often. I'd probably go with a more minimal set, but up to you

src/Builders/EpisodeBuilder.php
votemike: In the builder, the docblock doesn't appear to say it's null.
Don't you think that by default, objects should start with as little information as possible? So if any of these can be null, they should default to null and then set them if you want to. Then there's no need to allow null to be passed in to these methods.

tatitati: 1. Yeah, true, in the dockblock I forgot to add that info. I will fix it now, thanks for the spot.
2. For me is a trade off. If you create very basic objects by default, then you will spend more time setting properties for complex objects (exactly what we want to avoid). In the other hand, if you build complex objects by default, then you will need set null parameters in case you wish to build simple objects. 

Why then I choose to create complex objects?. The reason is that usually is more complex to add complex properties (which maybe imply to use another builder, set new attributes, etc) than set them to null. So I thought that would be faster to set as much properties as possible (mainly those that are complex)

votemike: I disagree. I think the default object should be as simple as possible. *If* the object needs extra options, then you should add them after the default is made. Otherwise you might be testing stuff that is set in the default object, not explicitly as part of the test.

tatitati: yeah, I thought about it as well.....let me check something

votemike: Why can you still set these as null? If the are set up with null in the first place, there's no need to change them to null....

PR #293, opened 2017-11-15
Cache prefix change

votemike:
Change so that blogs can be blogs-frontend.


programmes-developers:

PR #294, opened 2017-11-16
PROGRAMMES-5748 - Recipes Models

felipeparaujo:


programmes-developers:

src/ExternalApi/Recipe/Service/RecipeService.php
```diff
@@ -0,0 +1,78 @@
...
+use GuzzleHttp\ClientInterface;
+use GuzzleHttp\Exception\ClientException;
+use GuzzleHttp\Exception\GuzzleException;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+
+class RecipeService
+{
+    /** @var ClientInterface */
+    private $guzzleClient;
```
votemike: Tiny thing, should this just be named `client`? Otherwise, if we switch Guzzle out for something else, then name GuzzleClient won't make sense.

src/ExternalApi/Recipe/Service/RecipeService.php
```diff
@@ -0,0 +1,78 @@
...
+    public function fetchRecipesByProgramme(Programme $programme, ?int $limit = null, int $page = 1): array
+    {
+        $cacheKey = $this->cache->keyHelper(__CLASS__, __FUNCTION__, (string) $programme->getPid(), $limit, $page);
+        $cacheItem = $this->cache->getItem($cacheKey);
+
+        if ($cacheItem->isHit()) {
+            return $cacheItem->get();
+        }
+
+        $url = 'https://api.live.bbc.co.uk/food/recipes/by/programme/' . (string) $programme->getPid();
```
votemike: Should this be a config parameter like the Electron URL?

src/ExternalApi/Recipe/Service/RecipeService.php
```diff
@@ -0,0 +1,78 @@
...
+    private $mapper;
+
+    /** @var LoggerInterface */
+    private $logger;
+
+    /** @var UrlGeneratorInterface */
+    private $router;
+
+    public function __construct(
+        ClientInterface $guzzleClient,
```
tatitati: I think is a little intrusive to have this third dependency leak in our service. I would consider to create an adapter only responsible for making the request. That makes sense and also simplify tests as we won't need to mock guzzle (which is hard and breaks the principle of "don't mock what you don't own".

src/ExternalApi/Recipe/Domain/Recipe.php
```diff
@@ -0,0 +1,71 @@
...
+    private $hasImage;
+
+    /** @var Chef|null */
+    private $chef;
+
+    public function __construct(
+        string $id,
+        string $title,
+        string $description,
+        bool $hasImage = null,
```
BPScott: Default to false?

src/ExternalApi/Recipe/Mapper/RecipeMapper.php
```diff
@@ -0,0 +1,55 @@
...
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+
+class RecipeMapper
+{
+    /** @var UrlGeneratorInterface */
+    private $router;
+
+    public function __construct(UrlGeneratorInterface $router)
+    {
+        $this->router = $router;
```
BPScott: Doesn't seem to be used. Remove?

src/ExternalApi/Recipe/Service/RecipeService.php
```diff
@@ -0,0 +1,78 @@
...
+        CacheInterface $cache,
+        RecipeMapper $mapper,
+        LoggerInterface $logger,
+        UrlGeneratorInterface $router
```
BPScott: Doesn't seem to be used. Remove?

src/ExternalApi/Recipe/Service/RecipeService.php
```diff
@@ -0,0 +1,78 @@
...
+            return $cacheItem->get();
+        }
+
+        $url = 'https://api.live.bbc.co.uk/food/recipes/by/programme/' . (string) $programme->getPid();
+        $url .= '?page=' . $page . '&pageSize=' . $limit . '&sortBy=lastModified&sortSense=desc';
+
+        try {
+            $response = $this->guzzleClient->request('GET', $url);
+        } catch (GuzzleException $e) {
+            if ($e instanceof ClientException && $e->getResponse() && $e->getResponse()->getStatusCode() === 404) {
```
BPScott: It looks like Guzzle has a `http_errors` option you can set which disables throwing exceptions for 400/500 statuses. Should use that to avoid jumping through try/catch hoops? What do you think @BjornTwachtmann ?

src/ExternalApi/Recipe/Mapper/RecipeMapper.php
```diff
@@ -0,0 +1,55 @@
...
+            $recipe->description ?? '',
+            $recipe->hasImage === 'true',
+            $chef
+        );
+    }
+
+    /**
+     * @return Recipe[]
+     */
+    public function mapItems(stdClass $items, string $pid): array
```
BPScott: A mapper should only know how to map a single entity. It's the responsibility of whatever calls the mapper to loop through data and call the mapper for each item. Otherwise the Mapper has responsibility for things greater than "how to I create a single item"

This method should be moved into the RecipeService.

src/ExternalApi/Recipe/Service/RecipeService.php
BPScott: Same goes for the LoggerInterface too

src/ExternalApi/Recipe/Service/RecipeService.php
BPScott: The "Service" layer is the layer of adapters that abstracts our domain models and business logic from how we retrieve that data.

If we ever need to change how we request recipe data, we update the RecipeService (and thus we don't need to update our controllers). When doing things like controller tests we mock the calls to RecipeService

src/Controller/FindByPid/TlecController.php
```diff
@@ -78,6 +80,7 @@ public function __invoke(
         $isPromoPriority = $this->isPromoPriority($programme, $showMiniMap, !empty($promotions));
 
         $supportingContentItems = $electronService->fetchSupportingContentItemsForProgramme($programme);
+        $recipes = $recipeService->fetchRecipesByProgramme($programme);
```
BjornTwachtmann: I'm not sure (for now) that we should be passing this directly into the TLEC page. I think maybe it should be a lazy load. Dunno, thoughts?

felipeparaujo: I'll address that in the frontend work ticket. This is just to make sure it runs. I'll remove that for now.

src/ExternalApi/Recipe/Domain/Chef.php
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    public function getName(): string
+    {
+        return $this->name;
+    }
+
+    public function getImageUrl(string $width): ?string
+    {
+        if ($this->hasImage) {
```
BjornTwachtmann: Is there a way we can make this work with the existing responsive image stuff? Possibly by extending the pages service Image object and overriding the URL method? 

It's not catastrophic if not, but it would be nice to be able to re-use the existing responsive image code.

src/ExternalApi/Recipe/Service/RecipeService.php
tatitati: Yeah I agree Ben

src/ExternalApi/Recipe/Domain/Chef.php
BPScott: This might be fiddly as the valid widths in their ichef recipes might not match up to the valid widths in our ichef recipes, but I believe we can solve that by configuring the srcsets option when we call the ImagePresenter though.

src/ExternalApi/Recipe/Service/RecipeService.php
BjornTwachtmann: This is how everything is currently handling this. It's also useful because the CircuitBreaker can throw a GuzzleException when things go wrong and it's treated as an HTTP error by the calling class without doing any more work. 

I think it's fine this way.

src/ExternalApi/ApiType/UriToApiTypeMapper.php
```diff
@@ -14,6 +14,7 @@ class UriToApiTypeMapper
         ApiTypeEnum::API_BRANDING => '%^branding\.(int\.|test\.|stage\.)?files\.bbci\.co\.uk%i',
         ApiTypeEnum::API_RECOMMENDATIONS => '%^open\.live\.bbc\.co\.uk/recommend/items%i',
         ApiTypeEnum::API_ELECTRON => '%^api\.(int\.|test\.|stage\.|live\.)bbc\.co\.uk/electron%i',
+        ApiTypeEnum::API_RECIPE => '%^api\.(int\.|test\.|stage\.|live\.)bbc\.co\.uk/food/recipes/by/programme%i',
```
BPScott: Should the `by/programme` bit got into the Service instead?

src/ExternalApi/Recipes/Domain/ChefImage.php
```diff
@@ -0,0 +1,31 @@
...
+ * getUrl.
+ */
+class ChefImage extends Image
+{
+    /** @var string */
+    private $id;
+
+    public function __construct(string $id)
+    {
+        parent::__construct(new Pid('p0000001'), '', '', '', '', 'jpg');
```
BPScott: Is the id you get back the image Pid?

src/ExternalApi/Recipes/Domain/RecipeImage.php
```diff
@@ -0,0 +1,31 @@
...
+ * getUrl.
+ */
+class RecipeImage extends Image
+{
+    /** @var string */
+    private $id;
+
+    public function __construct(string $id)
+    {
+        parent::__construct(new Pid('p0000001'), '', '', '', '', 'jpg');
```
BPScott: Is the id you get back the image Pid?

app/config/services.yml
```diff
@@ -6,7 +6,9 @@ parameters:
     app.metric_backend.class: App\Metrics\Backend\CloudWatchMetricBackend
     app.receng.service.class: App\ExternalApi\RecEng\Service\RecEngService
     app.electron.service.class: App\ExternalApi\Electron\Service\ElectronService
+    app.recipe.service.class: App\ExternalApi\Recipes\Service\RecipesService
     app.electron.baseurl: 'https://api.live.bbc.co.uk/electron'
+    app.recipe.baseurl: 'https://api.live.bbc.co.uk/food/recipes/by/programme'
```
BPScott: Should by/programme be in the base URL? It more like "one of the many recipe apis" rather than as a base

src/ExternalApi/Recipes/Domain/ChefImage.php
felipeparaujo: No. The id food's ichef comes from the recipe's api, it's not a pid.

src/ExternalApi/Recipes/Domain/RecipeImage.php
felipeparaujo: Same as above

PR #295, opened 2017-11-20
PROGRAMMES-5743 - Link location stats implementation

Tasiobg:
This PR enables istats tracking and adds or updates the brand pages to meet the ticket requirements.

Notes:
istats-tracking.js is same we have in V2 with a small update, I have added "hardcodedItems" function to track item where we can't add the "data-linktrack" attribute


programmes-developers:

programmes-developers:

src/DsAmen/Organism/CoreEntity/CollapsedBroadcast/SubPresenter/CtaPresenter.php
```diff
@@ -21,7 +21,7 @@ class CtaPresenter extends BaseCtaPresenter
 
     protected $additionalOptions = [
         'link_to_start' => false,
-        'link_location_prefix' => 'programmeobject_',
+        'link_location_suffix' => '',
```
BjornTwachtmann: This shouldn't need to be here if it's being set in the base presenter. This should only ever be instantiated from there.

Tasiobg: This is the suffix which is only used on this class, should it be moved to the base class? not sure if any other class will need it

BjornTwachtmann: Shit. Misread that. Ignore me.

PR #297, opened 2017-11-21
PROGRAMMES-6052 invalidate $broadcast when stale

Tasiobg:


src/Controller/SchedulesOnNowController.php
```diff
@@ -42,6 +42,11 @@ public function __invoke(
         }
 
         $collapsedBroadcast = $collapsedBroadcastsService->findByBroadcast($broadcast);
+        if (!$collapsedBroadcast) {
+            // if this happen, $broadcast is stale on redis and should be expired
+            $broadcastsService->flushOnNowByService($network->getDefaultService());
+            return $this->response()->setMaxAge(300);
```
BjornTwachtmann: This should not be anywhere near 5 minutes. 10 seconds maybe.

programmes-developers:

PR #298, opened 2017-11-23
Cta horror

BjornTwachtmann:
This fixes a bug where Amen programme objects are showing as watchable live and/or streamable when they shouldn't be. 

It also refactors CTA presetners into something specific to a type of CTA rather than a type of programme. 
It also simplifies (hopefully) some of the CSS for standalone CTAs and the 11 classes that got added to do that in V2
It also fixes an IOS9 bug that makes icons space wrongly.
It also does not make the tea.


programmes-developers:

PR #299, opened 2017-11-23
Fix Footer Recommendations

mattgreenham:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5755
We now don't display the title if there are no recommendations, fixed title format & icon colours.


programmes-developers:

PR #302, opened 2017-11-24
Fix error page when ORB is dead so it doesn't require the ORB. 

BjornTwachtmann:
Exceptions thrown in an ExceptionHandler are bad mkay.


src/Controller/ExceptionController.php
```diff
@@ -36,20 +38,37 @@ public function __invoke(Request $request, FlattenException $exception, DebugLog
...
-            ], [
-                'searchScope' => $branding->getOrbitSearchScope(),
-                'skipLinkTarget' => 'programmes-content',
-            ]);
+            try {
+                $orb = $this->orbitClient->getContent([
+                    'variant' => $branding->getOrbitVariant(),
+                    'language' => 'en_GB',
+                ], [
+                    'searchScope' => $branding->getOrbitSearchScope(),
```
BPScott: If there was a Branding exception, this will error

BjornTwachtmann: Shit, good point

BPScott: https://github.com/bbc/branding-client/blob/master/src/Branding.php has a few defaults -> variant of 'default' and search scope of null

PR #306, opened 2017-11-27
refactoring in builders

tatitati:
For styleguide I have to create builders for clip, episode, and more programmes, but they use common constructors (programmeItem), so this is to avoid too much duplication in builders.


BjornTwachtmann:
AvailableClipsCount only exists on Episode. 
Also, don't you need a ClipBuilder?

PR #310, opened 2017-11-29
PROGRAMMES-6061

Tasiobg:
- for the on-now partial, if there is no default service, return 200 instead of 400
- changes on pages-services for findByUrlKeyWithDefaultService can now return networks with no default service, fix SchedulesByNetworkUrlKeyController for this case

Requires https://github.com/bbc/programmes-pages-service/pull/218


programmes-developers:

PR #311, opened 2017-11-29
Revert "Styleguide - replicate structure urls and folder "

tatitati:
This reverts commit bdafadfbfc8f56e63779e9bc470f8e067172be8d.

For any reason the previous commit was breaking a test in jenkins but passing in sandbox. There was nothing in logs to debug it either. 
Anyways, that previous commits is not necessary anymore because the Ds folder structure and the URLs for styleguide are gonna change with the new DS folder structure.  This quick fix solved the problem. 

The annoying thing is that we don't know why this previous commit broke the test only in jenkins. Frustrating


programmes-developers:

PR #312, opened 2017-11-29
move css to css folder

tatitati:
Moved Atom and Core folders to Css folder.

From:
```
Ds2013
    ├── Atom
    └── Core
```

to:
```
Ds2013
└── css
    ├── Atom
    └── Core
```


programmes-developers:

BPScott:
Rather than making lots of big changes, how about writing up a proposal as an ADR (along with the reasons why things should be done) then the team can discuss the wider ideology instead of talking through lots of little pull requests where we can't see the system as a whole. That'll also save you a lot of work if we decide to improve / iterate on the initial suggestions

votemike:
Are you not moving the DsAmen stuff too?  
Are we keeping the name "Atom"?

tatitati:
ei @BPScott we already discussed it in the office. We saw that one of the main difficulties for the team members is having a clear difference between "small" and "big", for instance, between molecules and organism. So we moved away from that idea of "complexity size", and started to think of contexts. Then we came up with this new folder structure. We think that is more intuitive and self-explanatory. 

```
/src
├── Ds2013
    └── css
    |    ├── Atom
    |    └── Core
    | 
    ├── AssemblyContext

           ├── Application
                    - calendar
                    - datelist
                    - Navigation
                    - Duration
                           - DurationPresenter.php
                           - Duration.html.twig
                           - Duration.scss                           

           ├── Domain
                    - Broadcast                    
                    - Programe
                    - CollapsedBroadcast
                            - CollapsedBroadcastPresenter.php
                            - CollapsedBroadcast.html.twig
                            - CollapsedBroadcast.scss

           ├── Section
                    - footer                        
                    - header
                    - map

           ├── Page
                    ─ Schedules
                        ─ by day
                        ─ by week
                        ─ by month
├── DsAMen
    .....
```

But you are right. I thought that make a big PR for all this would impossible to analyze, so I thought that the other option would be making small PR's, but you are right, is gonna be many PR's.  Any suggestion? @votemike, @BPScott

votemike:
Still not keen on the wording of "AssemblyContext", didn't Bjorn suggest "Presenters"? And "Application" seems a bit vague. How about "Components" or "Widgets"?

tatitati:
@votemike yeah, @BjornTwachtmann suggested Presenters, but the true is that there is no only presenters in these folders. There are also Twig, scss, and presenters/subpresenters. So I want to keep the idea of "assembly different things: scss with twigs and presenters", and on different levels:
first we build utilities and domain stuff, these are injected later in sections, and later these sections build pages. So for me is a kind of assembly chain. 

Maybe "assembly context" is not a good one, but maybe "context", or "layer" ? @votemike 
=====
About "Application":
Anyway I agree, what do you think about Utilities instead of Application?, you said that yesterday and for me is also a really good option. @votemike ?

BjornTwachtmann:
@BPScott aren't you supposed to be on holiday mate :P

src/Ds2013/_ds2013-components.scss
```diff
@@ -99,13 +99,13 @@
 // @import 'modules/atoz/atoz-grid';
 // @import 'modules/atoz/atoz-loader';
 // @import 'modules/atoz/navigation';
-@import 'Core/Modules/Atoz/search';
+@import 'Css/PageSpecfic/Atoz/search';
```
votemike: Why are these in CSS and not in the Page Presenters?

BjornTwachtmann: Because they're not tied to a presenter. We could move them there and have them standalone also. I don't really care, this just seemed more obvious to me.

votemike: Ahh, I assumed we had page presenters for these.

PR #325, opened 2017-12-04
PROGRAMMES-6020 bbc_site using value from isite when set.

tatitati:
when working with Istats we were setting the value for the "bbc_site" variable based on the core_entity's network or masterbrand. So the value selected in Isite was not having any real effect in this value. Now we have isite values ingested in our core_entity's "options" column, so we can fix this easily.

PR dependency  in pages-service (merged):
https://github.com/bbc/programmes-pages-service/pull/220


programmes-developers:

src/ValueObject/IstatsAnalyticsLabels.php
```diff
@@ -144,7 +144,10 @@ private function setBbcSiteLabel($context): void
 
         if (($context instanceof CoreEntity || $context instanceof Service) && $context->getNetwork()) {
             $nid = (string) $context->getNetwork()->getNid();
-            if ($nid === 'bbc_music') {
+            if (!$context instanceof Service && !empty($context->getOption('bbc_site'))) {
```
BjornTwachtmann: I'm not sure this belongs in this block actually...

PR #326, opened 2017-12-04
Changes to use on-demand sort for on-demand column on TLEC page

BjornTwachtmann:


src/Controller/FindByPid/TlecController.php
```diff
@@ -52,14 +52,15 @@ public function __invoke(
...
         }
 
         $upcomingBroadcast = null;
-        $streamableEpisodes = null;
+        $onDemandEpisode = null;
         $upcomingRepeatsAndDebutsCounts = ['debuts' => 0, 'repeats' => 0];
         if ($programme->getAggregatedEpisodesCount() > 0) {
-            $streamableEpisodes = $aggregationService->findStreamableDescendantEpisodes($programme, 1);
+            $onDemandEpisodes = $aggregationService->findStreamableOnDemandEpisodes($programme, 1);
+            $onDemandEpisode = $onDemandEpisodes[0] ?? null;
```
felipeparaujo: why is this not being done in pps?

BjornTwachtmann: well, PPS is returning an array, as you might want multiple of these

felipeparaujo: ah nevermind, that was happening before as well, you just moved it up there

programmes-developers:

PR #327, opened 2017-12-05
Programmes 5956 social bar tracking

Tasiobg:
- add tracking to social bar by created a function to enable tracking custom link_location for external link
- convert quotes in istats-tracking.js to single quotes, looks more consistent


programmes-developers:

PR #329, opened 2017-12-06
PROGRAMMES-5958 add tracking to links in 'supporting content'

Tasiobg:
Add tracking to links in 'supporting content' in brand page. This is already supporter by our implementation of tracking, only adding the data property is required


programmes-developers:

PR #330, opened 2017-12-06
200 response gap broadcasts

tatitati:
Response 200 when there is no broadcasts in past dates and include in this cases a 
```html
<meta name="robots" content="noindex">
```


programmes-developers:

app/Resources/views/base_ds2013.html.twig
```diff
@@ -4,6 +4,7 @@
 <html class="{{ branding.getOrbitThemeClasses()|raw }} {% block html_classes %}{% endblock %}" lang="{{ tr('language_code') }}">
     <head>
         <meta charset="UTF-8">
+        {% if meta_no_index is defined and meta_no_index == true %}<meta name="robots" content="noindex">{% endif %}
```
BjornTwachtmann: We already have a meta_context object. Would that be better in there?

src/Controller/SchedulesByDayController.php
```diff
@@ -93,9 +93,14 @@ public function __invoke(
...
 
         // If there are broadcasts, 200
-        // If there are no broadcasts, but the page is less than 35 days away, still 200
+        // If there are no broadcasts, but the page is LESS than 35 days away, still 200
+        // If there are no broadcasts, but the page is in the past, still 200, but we don't want to index the page
+        // If there are no broadcasts, and the page is MORE than 35 days in the future then 404
         // This means the date nav won't link to 404 pages.
-        if (!$broadcasts && !$broadcastDay->start()->isWithinNext('+35 days')) {
+        if (!$broadcasts && $dateTimeToShow->isPast()) {
+            $this->response()->setStatusCode(200);
```
BjornTwachtmann: I don't think you need to set 200 explicitly, that's the default

BjornTwachtmann:
We probably also need something like this for the week controller. Can you also please check that we're not linking to anything past the service start/end dates.

tests/BaseWebTestCase.php
```diff
@@ -8,13 +8,14 @@
 {
     const FIXTURES_PATH = 'Tests\App\DataFixtures\ORM\\';
 
-    public function assertResponseStatusCode($client, $expectedCode)
+    public function assertResponseStatusCode($client, $expectedCode, string $text = '')
```
tatitati: I thought that assert messages like ``` 'Failed asserting that the response status code "%s" matches expected "%s"``` provide still little info when a test is failing and response codes get more complex. Now we can add an explanation about what/why we expect something

tests/Controller/SchedulesByWeekControllerTest.php
```diff
@@ -129,6 +129,41 @@ public function testDataPageTimeIsSetProperlyInHtmlResponseForWeek()
         $this->assertEquals('2017/05/15', $crawler->filter('[data-page-time]')->attr('data-page-time'));
     }
 
+    public function testReturns404ForWeeksWithInactiveService()
```
tatitati: This case (404 response code) was not being covered by any test.

src/Controller/BaseController.php
```diff
@@ -144,7 +144,7 @@ protected function renderWithChrome(string $view, array $parameters = [])
 
             $parameters = array_merge([
                 'orb' => $orb,
-                'meta_context' => new MetaContext($this->context, $this->getCanonicalUrl()),
+                'meta_context' => new MetaContext($this->context, $this->getCanonicalUrl(), $parameters['meta_no_index'] ?? false),
```
votemike: Is `$parameters` the place we want to pass this through?

tests/BaseWebTestCase.php
votemike: `$text` isn't very descriptive

src/Controller/BaseController.php
tatitati: I must say that $parameters is a very generic name, but so far we use that array to pass parameters to the view (orbit, meta, analytics, comscore, ...). Maybe we can rename it one day to something more descriptive. Is used in this way for now:
```return $this->render($view, $parameters, $this->response);```

File: tests/BaseWebTestCase.php (hunk 3 above)
Comment (tatitati) [reply tests/BaseWebTestCase.php]: assertCaption?
assertDescription?

File: src/Controller/BaseController.php (hunk 5 above)
Comment (votemike) [reply src/Controller/BaseController.php]: I was always under the impression it was just for things referenced in the views. How about a second array of params for non-view things? `$additionalParams`?

File: tests/BaseWebTestCase.php (hunk 3 above)
Comment (votemike) [reply tests/BaseWebTestCase.php]: `$description`? `$additionalText`?

File: src/Controller/BaseController.php (hunk 5 above)
Comment (BjornTwachtmann) [reply src/Controller/BaseController.php]: Yeah, parameters is for things being passed into the view. Which this is not. 

I would suggest that you look at how the Canonical URL stuff is being handled
https://github.com/bbc/programmes-frontend/blob/376132d730727d9fdf314da87c5dbcb93347262b/src/Controller/BaseController.php#L71
there's a property in the baseController that sub-controllers can override if they so choose. Might as well be consistent.

File: src/Controller/BaseController.php [hunk 6]
```diff
@@ -44,6 +44,9 @@
 
     protected $canonicalUrl;
 
+    /** @var bool */
+    protected $metaNoIndex;
```
votemike: Are controllers supposed to have state/properties?

src/Controller/BaseController.php
```diff
@@ -245,4 +248,13 @@ private function logger(): LoggerInterface
     {
         return $this->container->get('logger');
     }
+
+    private function getMetaNoIndex(): bool
+    {
+        if (!isset($this->metaNoIndex)) {
```
votemike: When would this not be set?

src/Controller/BaseController.php
tatitati: yeah, the true is that we have some states for the controller already:
```
    private $context;
    private $istatsExtraLabels = [];
    private $istatsProgsPageType;
    protected $canonicalUrl;
    protected $metaNoIndex;
```
but yeah, this is breaking a generic rule. 🤔

tatitati: I agree, I'm thinking how to improve this. Thanks @votemike.

BjornTwachtmann: This is a PHP app, it's not like the state is going to persist between requests. I can't really see any issue with it.

src/Controller/BaseController.php
tatitati: Well, we want to avoid to index pages in schedules pages with no broadcasts. So far we only manage this "noindex" for schedules with info about broadcasts (i.e.: by day and by month). For "by month schedules" it would be very expensive to know what broadcasts we have as you wrote in the ticket.

So this left a few controllers that don't set this variable such as: schedules by month, brand page, ....

The bad thing about this is that we are setting a property that the rest of controllers don't need. The good thing is that in the future might be need it, but this sounds to me a little like "YANGI". Is not perfect ( i know)

src/Controller/BaseController.php
tatitati: Thanks bjorn, that is true

src/Controller/BaseController.php
votemike: Also, why do you need a getter for it? Just access the property directly.

BjornTwachtmann: It seems sensible to have something like this. Enforce a type and a default (though you could just set a default I guess).

PR #332, opened 2017-12-13
PROGRAMMES-6027 Related Links

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6027


src/Controller/FindByPid/TlecController.php
```diff
@@ -56,6 +58,11 @@ public function __invoke(
             $clips = $aggregationService->findStreamableDescendantClips($programme, 4);
         }
 
+        $relatedLinks = null;
+        if ($programme->getRelatedLinksCount() > 0) {
+            $relatedLinks = $relatedLinksService->findByRelatedToProgramme($programme); //@TODO it looks like in V2 there is no limit, should I change to be no limit?
```
votemike: I believe the default limit is 300, but should I make it unlimited?

programmes-developers:

tatitati:
seems ok

app/Resources/views/find_by_pid/tlec.html.twig
```diff
@@ -170,7 +170,22 @@
         </div>
     </div>
 
-    {# TODO Related links block #}
+    {% if relatedLinks %}
+        <div class="p-g-p">
+            <div class="grid">
```
BPScott: If all the links are always full width then you don't need the grid/grid__item divs

src/Controller/FindByPid/TlecController.php
BPScott: I reckon keep it limited to 300. There's potential for weirdness if we let unbounded SQL queries run and return loads of results. Looking at the DB the most links a core entity has is 50 anyway.

src/Controller/FindByPid/TlecController.php
```diff
@@ -56,6 +58,11 @@ public function __invoke(
             $clips = $aggregationService->findStreamableDescendantClips($programme, 4);
         }
 
+        $relatedLinks = null;
```
BPScott: Should the default be `[]` so the variable is always an array?

src/Twig/DesignSystemPresenterExtension.php
```diff
@@ -103,10 +103,11 @@ public function dsAmenPresenter(
         Twig_Environment $twigEnv,
         DsAmenPresenter $presenter
     ) {
-        return $twigEnv->render(
+        // Intellij adds a newline when saving templates. Trim it.
```
BPScott: A final newline in text files is [the POSIX standard](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline) and something we enforce in our .editorconfig.

This feels a little surpising, why do we need to do this rather than have
`{{- ds_amen('relatedLink', link) -}}` (note the hyphens) when we render a presenter that needs its surrounding whitespace supressed?

src/DsAmen/Presenters/Domain/RelatedLink/_related-link.scss
```diff
@@ -0,0 +1,30 @@
+.related-links {
+    margin-left: $gel-unit-small;
+    margin-right: $gel-unit-small;
+    margin-bottom: $gel-unit-small;
+
+    > .related-links__item {
```
BPScott: You're already hinting that this class lives within `.related-links` through the BEM naming convention, so you don't need to create this additional level of specificity (this compiles to the selector `.related-links > .related-links__item` when `.related-links` is fine).

src/Twig/DesignSystemPresenterExtension.php
votemike: (3 of us) Tried all of that. Didn't seem to work. Feel free to have a go. Would love to be proved wrong.

src/Controller/FindByPid/TlecController.php
votemike: I'm used to the returned things being `IsiteResult`s...

app/Resources/views/find_by_pid/tlec.html.twig
votemike: I need the grid__item for the bottom margin. And I need the grid if I have the grid__item.  
I think I can remove the 1/1 class though.

votemike: I _could_ add some margin bottom to .related-links I suppose. Is that preferable?

votemike:
Test this please

app/Resources/views/find_by_pid/tlec.html.twig
BPScott: Yeah that sounds preferable

.gitignore
```diff
@@ -10,3 +10,4 @@
 /composer.phar
 /app/Resources/gelicons/
 /script/fixture.ini
+yarn-error.log
```
Tasiobg: Should this have a leading slash as well?

votemike: I don't think so. Why would it?

votemike:
@Tasiobg have what more consistent?

.gitignore
Tasiobg: all the other files seems to have it, not sure what is the difference but if there is no leading slash and I add this file to the root folder phpstorm doesn't seems to recognice the file as ignored.
Ignore this comment if you think is fine, I haven't investigate it enough

Tasiobg:
@votemike we alredy use hyphens in other places where we include templates, would be nice if this could be use everywhere but nothing too important, your solution looks good and Ben hasn't replay so I'm assuming the other option doesn't work in this case

.gitignore
BPScott: The leading / makes it treat pattern as a glob relative to the route directory so `foo.txt` would match the file at `/foo.txt` and `/stuff/foo.txt`, while `/foo.txt` would match the file at `/foo.txt` but NOT `/stuff/foo.txt`.

https://git-scm.com/docs/gitignore#_pattern_format.

src/Twig/DesignSystemPresenterExtension.php
BPScott: OK, I think I've worked out what's going on here. Twig's whitespace control stuff (the hyphens in {{-blah -}} / {%-
 blah -%}) compacts the whitespace around that particular block, but does not touch anything within said block. I reckon this is the correct thing to because `{{- my_function() -}}` shouldn't manipulate the string output of my_function().

I'm torn on this.

On the one hand having our presenters all without leading/trailing whitespace sounds clean and consistent, but it introduces a large amount of trim() calls that are totally unneeded. It's /probably/ not a performance concern but why add code that does nothing?

On the other hand those trims are not needed. In this, and the vast majority of other cases presenters are rendering block or flexbox elements which discard trailing whitespace so there are is no rendering issue when displaying the page. In the rare cases where a template is spitting out whitespace that results in a visual change that we don't want we can manually add a `|trim` to the calling location (e.g. `{{ my_function()|trim }}`).

Is there some visual issue that is caused by not trimming these items?

.gitignore
votemike: So are we happy leaving this as-is?

app/Resources/views/find_by_pid/tlec.html.twig
```diff
@@ -170,7 +170,18 @@
...
     </div>
 
-    {# TODO Related links block #}
+    {% if relatedLinks %}
+        <div class="p-g-p">
+            <div class="br-box-secondary">
+                <h2 class="gel-double-pica-bold island">{{ tr('related_links') }}</h2>
+                <div class="related-links br-keyline media__meta-row--separator">
+                    {%- for link in relatedLinks -%}
+                        {{ ds_amen('relatedLink', link)|trim|raw }}
```
votemike: If we only use the trim filter, then the HTML is printed out rather than rendered correctly, so we have to use the raw tag too...... Is it safe enough to print this raw?

Tasiobg: I think this is what we want, if we want, by default twig escapes the HTML https://twig.symfony.com/doc/2.x/filters/escape.html

BPScott: @votemike What is the visual difference when we use trim vs when we don't? I couldn't see any change when I added/removed that bit of whitespace?

votemike: On Spring Watch, there are 8 links. *With* the trim, the links are in 2 rows of 4. *Without* the trim, it's 3 rows. 2 rows of 3 and 1 row of 2.

PR #333, opened 2017-12-13
PROGRAMMES-5915 Adding synopsis for clips and promos

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5915

I _think_ this does everything the ticket asks for.


programmes-developers:

src/DsAmen/Presenters/Domain/CoreEntity/Base/SubPresenter/BaseTitlePresenter.php
```diff
@@ -80,6 +80,10 @@ public function getMainTitle(): string
 
     public function getSubTitle(): string
     {
+        if ($this->coreEntity->isRadio()) {
```
BPScott: Why are we putting the synopsis in the the title are? That seems like the wrong place to me.

Programme presenters should also have a `show_synopsis` option in the same way the promotion ones do

src/DsAmen/Presenters/Domain/CoreEntity/Base/SubPresenter/BaseTitlePresenter.php
```diff
@@ -80,6 +81,10 @@ public function getMainTitle(): string
 
     public function getSubTitle(): string
     {
+        if (is_a($this->coreEntity, Clip::class) && $this->coreEntity->isRadio()) {
```
votemike: Do we want to do this? Or should I create a ClipTitlePresenter?

felipeparaujo: Having a presenter for such a specific case feels like over-engineering. This is fine in my opinion

src/DsAmen/Presenters/Domain/CoreEntity/Programme/ProgrammePresenter.php
```diff
@@ -12,12 +12,16 @@
...
 use App\DsAmen\Presenters\Domain\CoreEntity\Shared\SubPresenter\TitlePresenter;
 use App\DsAmen\Presenters\Domain\CoreEntity\Shared\SubPresenter\StreamableCtaPresenter;
+use BBC\ProgrammesPagesService\Domain\Entity\Clip;
 
 class ProgrammePresenter extends BaseCoreEntityPresenter
 {
     public function getBodyPresenter(array $options = []): BaseBodyPresenter
     {
         $options = array_merge($this->subPresenterOptions('body_options'), $options);
+        if (is_a($this->coreEntity, Clip::class) && $this->coreEntity->isRadio()) {
```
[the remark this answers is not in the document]
felipeparaujo: Same as above ^

votemike:
Test this please

src/DsAmen/Presenters/Domain/CoreEntity/Programme/ProgrammePresenter.php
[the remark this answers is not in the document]
BPScott: This looks like you're making it impossible to disable showing the synopsis for radio clips. That seems like a bad thing.

I really don't think this belongs here. PROGRAMMES-5915 looks like it's about changing just this configuration on the TLEC page. I'm think is the exception rather than the rule, thus we should be setting the [show_synopsis option in the TLEC controller](https://github.com/bbc/programmes-frontend/blob/89a1404879f774282d66dbb031a0c2bd8db3fd6b/app/Resources/views/find_by_pid/tlec.html.twig#L100) rather than changing the default value within the presenter.

src/DsAmen/Presenters/Domain/CoreEntity/Base/SubPresenter/BaseTitlePresenter.php
BPScott: I don't think we want to do this at all.

This is the exception rather than the rule - it's something just for the TLEC page, not something we want to apply globally. Other places where we have radio clips we show their subtitle, e.g. http://www.bbc.co.uk/programmes/b006qsq5/topics/Nonlinear_narrative_films and http://www.bbc.co.uk/programmes/b01mrh21/clips. Turning this off globally would be a bad idea and would mess up our other page layouts.


The Subtitle in this case is the context of the programme. Programme items within a context don't show the  titling of that context, e.g. if you were on the Archers homepage, you would'nt want to see "... The Archers" after every programme title because you know you're in the context of the archers.

Consider a Clip with the hierarchy "Brand -> Series -> Episode -> Clip". 

* When there is no [`context_programme`](https://github.com/bbc/programmes-frontend/blob/89a1404879f774282d66dbb031a0c2bd8db3fd6b/app/Resources/views/find_by_pid/tlec.html.twig#L101) defined then the Title is "Clip" and the Subtitle is "Episode, Series, Brand".
* When the `context_programme` is set to be the Brand then the Title is "Clip" and the Subtitle is "Episode, Series".

Thus we can hide the subtitles by manipulating the context_programme on the TLEC page and claiming it is the clip itself rather than doing any modification within the Programme presenter.  Change https://github.com/bbc/programmes-frontend/blob/89a1404879f774282d66dbb031a0c2bd8db3fd6b/app/Resources/views/find_by_pid/tlec.html.twig#L101 to be `'context_programme': programme.isRadio() ? clip : programme,` and you'll hide the subtitle on the TLEC page without it affecting global defaults.

BPScott: Also I guess Radio would want the same hiding of the context on Galleries too. You can do the same thing when rendering those too.

BPScott:
Nice, that all seems a lot simpler. :)I asked in a comment on PROGRAMMES-5915 if the "don't show the context subtitle" idea should apply to galleries as well as clips but didn't get a response. Can you poke Cathal about that on Monday?

votemike:
@BPScott he's off until the 3rd.

votemike:
If it needs changing a new Ticket/PR can be made...

PR #334, opened 2017-12-13
PROGRAMMES-5965 Changing all available link for radio episodes

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5965


programmes-developers:

app/config/routing.yml
```diff
@@ -22,6 +22,10 @@ find_by_pid:
     defaults: { _controller: '!find_by_pid' }
     requirements: { pid: '[0-9b-df-hj-np-tv-z]{8,15}' }
 
+episodes_player:
```
votemike: This is the first new PID route... do we have a preference on the naming convention?

felipeparaujo: Not for the cloud stuff, but I think episodes_player is fine

BPScott:
Y'all missed the bit where this breaks things when you deploy it to INT because the MetricManager assumes that every route has a valid controller defined (which this one does not).

Also this route is served by the PAL app so it should live in routing_3rd_party as per the comment at the top of routing.yml, and you missed the scheme definition (which saves a https->http redirect).

I've got you covered https://github.com/bbc/programmes-frontend/commit/f08778b65c3612e9de7182bf2a01d94b623e8d37

votemike:
@BPScott Did it make it 504? That might be what our 504s were. Cheers for the fix. Is there any way we could have a test that makes sure this won't happen again?

BPScott:
Yep. the EC2 boxes were throwing 500s, which meant www.test couldn't proxy through to them which I'm guessing was what was throwing the 504 (gateway timeout to save you a google) as it couldn't connect to the upstream host (i.e. the EC2 boxes)

PR #335, opened 2017-12-14
Fix autowiring deprecation messages

BPScott:
Explicitly add a "GuzzleHttp\ClientInterface" service which is aliased
to the guzzle client we create. This means we are now doing explict
autowiring based on service names rather than implicit (based on types,
which is deprecated)

/cc @felipeparaujo


programmes-developers:

PR #337, opened 2017-12-14
Adapt the new styleguide to the new DS folder structure

tatitati:
Just adapting the new styleguide to the new DS folder structure


programmes-developers:

PR #338, opened 2017-12-14
PROGRAMMES-6000 Localising MAP

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6000


programmes-developers:

src/Controller/FindByPid/TlecController.php
```diff
@@ -40,8 +42,14 @@ public function __invoke(
         ProgrammesAggregationService $aggregationService,
         ImagesService $imagesService,
         RecEngService $recEngService,
-        ElectronService $electronService
+        ElectronService $electronService,
+        HelperFactory $helperFactory
     ) {
+        if ($programme->getMasterBrand() && $programme->getMasterBrand()->getNetwork()->isInternational()) {
```
tatitati: can we have something like $programme->isInternational() ?
I see that this controller knowing too much about all of them, is coupled to Programme, Masterbrand and Network.

app/Resources/views/find_by_pid/tlec.html.twig
```diff
@@ -185,3 +185,21 @@
         </div>
     </div>
 {% endmacro %}
+
+{% block javascripts %}
+    {% if programme.getNetwork() and programme.getNetwork().isInternational() %}
```
BPScott: Add a reference to render `{{ parent() }}` otherwise any global contents of this block will be lost https://twig.symfony.com/doc/2.x/functions/parent.html

PR #339, opened 2017-12-15
Add first pass at ADA integration

BPScott:
We can get two types of data back from ADA - Classes or ProgrammeItems.
For the work on the Radio brand page we only care about Classes. There
is a AdaClassService because eventually there will be an
AdaProgrammeItemService (or something named similar).

Tests for the mapper and domain model are in there, but have not yet
done a test for the service yet.

--- 

I still need to add a test for the service, but here's what I've got so far for review. Please mention anything that is unclear and I'll try and add inline docs.


BPScott:
We've been repeating the process of creating stub services for each of our external APIs. I wonder would it be preferable to instead stub out the guzzle client?

programmes-developers:

app/config/services.yml
```diff
@@ -7,8 +7,10 @@ parameters:
     app.receng.service.class: App\ExternalApi\RecEng\Service\RecEngService
     app.electron.service.class: App\ExternalApi\Electron\Service\ElectronService
     app.recipes.service.class: App\ExternalApi\Recipes\Service\RecipesService
```
BPScott: Is it worth alphabetising these service class / baseurls?

src/ExternalApi/Ada/Mapper/AdaClassMapper.php
```diff
@@ -0,0 +1,36 @@
...
+
+namespace App\ExternalApi\Ada\Mapper;
+
+use App\ExternalApi\Ada\Domain\AdaClass;
+use BBC\ProgrammesPagesService\Domain\Entity\Image;
+use BBC\ProgrammesPagesService\Domain\ValueObject\Pid;
+
+class AdaClassMapper
+{
+    public function mapItem(array $adaClass): AdaClass
```
BPScott: In pages-service the format for Mappers is to have the public function be `getDomainModel()` but in ExternalAPI we're using `mapItem`. Is it worth doing a bit of renaming so the naming is consistent? `getDomainModel()` sounds nicer to me

src/ExternalApi/Ada/Service/AdaClassService.php
```diff
@@ -0,0 +1,174 @@
...
+        // If $countWithinTleo is false, then the programme_item_count returned
+        // shall be the number of items across the entire BBC
+        $contextPid = ($countWithinTleo ? (string) $programme->getTleo()->getPid() : null);
+
+        $stringPid = (string) $programme->getPid();
+
+        $cacheKey = $this->cache->keyHelper(__CLASS__, __FUNCTION__, $stringPid, $countWithinTleo);
+        $cacheItem = $this->cache->getItem($cacheKey);
+
+        if ($cacheItem->isHit()) {
```
BPScott: There ends up being a lot of common boilerplate when making an API call, is there some sensible way that we can abstract that away slightly rather than all the copy+paste?

We're up to 4 ExternalApis now, is that time to spot the similar patterns?

src/ExternalApi/Ada/Service/AdaClassService.php
```diff
@@ -0,0 +1,174 @@
...
+
+            if ($e instanceof ClientException && $e->getResponse() && $e->getResponse()->getStatusCode() === 404) {
+                // 404s get cached for a shorter time
+                $this->cache->setItem($cacheItem, [], CacheInterface::NORMAL);
+            }
+
+            return [];
+        }
+
+        $response = json_decode($response->getBody()->getContents(), true);
```
BPScott: I'm using json_decode($blah, true) here to get back the contents of the response as an array rather than an object. pages-service tends to act upon arrays and I generally find associative arrays nicer to manipulate than a stdClass. Would it be worth hunting down the otherjson apis and have them use arrays too?

src/ExternalApi/ApiType/ApiTypeEnum.php
```diff
@@ -10,13 +10,15 @@ class ApiTypeEnum
...
     public const API_RECIPE = 'RECIPE';
+    public const API_ADA = 'ADA';
 
     private const API_TYPES = [
         self::API_BRANDING => true,
         self::API_ORBIT => true,
         self::API_RECOMMENDATIONS => true,
         self::API_ELECTRON => true,
         self::API_RECIPE => true,
+        self::API_ADA => true,
```
BPScott: Is it worth making this alphabetical so that it's slightly easier to scan if you're looking for something?

src/ExternalApi/ApiType/UriToApiTypeMapper.php
```diff
@@ -15,6 +15,7 @@ class UriToApiTypeMapper
         ApiTypeEnum::API_RECOMMENDATIONS => '%^open\.live\.bbc\.co\.uk/recommend/items%i',
         ApiTypeEnum::API_ELECTRON => '%^api\.(int\.|test\.|stage\.|live\.)bbc\.co\.uk/electron%i',
         ApiTypeEnum::API_RECIPE => '%^api\.(int\.|test\.|stage\.|live\.)bbc\.co\.uk/food/recipes%i',
+        ApiTypeEnum::API_ADA => '%^ada\.(int\.|test\.|stage\.)?api\.bbci\.co\.uk%i',
```
BPScott: Is it worth making this alphabetical so that it's slightly easier to scan if you're looking for something?

app/config/services.yml
votemike: Probably

app/config/services.yml
```diff
@@ -7,8 +7,10 @@ parameters:
...
+    app.ada.class_service.class: App\ExternalApi\Ada\Service\AdaClassService
     app.electron.baseurl: 'https://api.live.bbc.co.uk/electron'
     app.recipes.baseurl: 'https://api.live.bbc.co.uk/food/recipes'
+    app.ada.baseurl: 'https://ada.api.bbci.co.uk'
```
votemike: I don't see a test version of this config anywhere. Should we be using the LIVE URL in our sandbox?

src/ExternalApi/Ada/Domain/AdaClass.php
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    public function getId(): string
+    {
+        return $this->id;
+    }
+
+    public function getTitle(): string
+    {
+        return urldecode($this->title);
```
votemike: Should we not store the URL as decoded?

src/ExternalApi/Ada/Mapper/AdaClassMapper.php
votemike: An array of what? Mixed array?

src/ExternalApi/Ada/Service/AdaClassService.php
votemike: I find arrays more than 1 level deep harder to read than objects more than 1 level deep. I don't know how other people feel....

src/ExternalApi/Ada/Service/AdaClassService.php
```diff
@@ -0,0 +1,174 @@
...
+        int $page = 1
+    ) {
+        return $this->baseUrl . '/classes?page=' . $page . '&page_size=' . $limit .
+            ($programmePid ? '&programme=' . $programmePid : '') .
+            ($contextPid ? '&count_context=' . $contextPid : '') .
+            ($countItemType ? '&count_item_type=' . $countItemType : '') .
+            ($type ? '&type=' . $type : '') .
+            ($threshold ? '&threshold=' . $threshold : '') .
+            ($order ? '&order=' . $order : '') .
+            ($orderDirection ? '&direction=' . $orderDirection : '');
```
votemike: Will this ever be blank?

src/ExternalApi/ApiType/UriToApiTypeMapper.php
votemike: What's this, never noticed these before...

app/config/services.yml
BPScott: These are all read-only apis, so calling LIVE where possible has no downside and means we get to work with more complete data.

In v2 we call Ada LIVE in all environments.

app/config/services.yml
BPScott: Have done the alphabetising on develop and rebased this branch onto latest develop

src/ExternalApi/ApiType/UriToApiTypeMapper.php
BPScott: It's some tooling for the Guzzle middleware that generates metric counts for each api, this lets it match URL patterns and says "url X was hit so increment the metric Y"

src/ExternalApi/Ada/Service/AdaClassService.php
BPScott: FAO @felipeparaujo / @Tasiobg / @tatitati: Do any of you have preferences on if we deal things decoded as objects or as arrays?

src/ExternalApi/Ada/Domain/AdaClass.php
BPScott: Yep, Looking at the old code I've got no idea why we were doing this, I can't find anything in the existing titles that might result in this (topics with apostrophes "Alumni of St John's College Cambridge" like or things with diacritics like "École Normale Supérieure alumni".

src/ExternalApi/Ada/Mapper/AdaClassMapper.php
BPScott: Yeah, the array is a single item, with the keys being named after fields. Pending peoples thoughts on the json_decode bits above I might end up changing this accept an stdClass.

src/ExternalApi/ApiType/UriToApiTypeMapper.php
votemike: Why are environments mentioned here? Why aren't they set via config?

src/ExternalApi/Ada/Service/AdaClassService.php
BPScott: Potentially if the user passes in an empty string, however that is stunningly unlikely as it is a private method. I reckon keep it like this to be consistent with the other parameters

src/ExternalApi/Ada/Service/AdaClassService.php
felipeparaujo: I like arrays better

tatitati: I prefer arrays as well. The reason is that using stdClass or using arrays in this case is not making any difference apart of the visual. StdClass gives the impression of being a well defined object, but is not, so like with arrays you still have to figure out what keys you have to work with this data. This said I think that also the syntax ```$item->{'another_key'}``` is not too natural. So for me arrays.

src/ExternalApi/Ada/Service/AdaClassService.php
Tasiobg: An investiagation could be worth now that we don't have many ExternalApis

src/ExternalApi/Ada/Service/AdaClassService.php
Tasiobg: If $item is an array I understand that it only contains data so seem easy and safer to work with but I'm not too sure if could be any adventage of using object instead, any option works fine as far as is consistent with other code.

src/ExternalApi/Ada/Service/AdaClassStubService.php
```diff
@@ -0,0 +1,25 @@
...
+
+use App\ExternalApi\Ada\Domain\AdaClass;
+use BBC\ProgrammesPagesService\Domain\Entity\Programme;
+
+/**
+ * Class ElectronStubService
+ *
+ * Stub class for unit tests
+ */
+class AdaClassStubService extends AdaClassService
```
votemike: Remind me again why we use stub services?

src/ExternalApi/Ada/Service/AdaClassService.php
```diff
@@ -0,0 +1,174 @@
...
+            if ($e instanceof ClientException && $e->getResponse() && $e->getResponse()->getStatusCode() === 404) {
+                // 404s get cached for a shorter time
+                $this->cache->setItem($cacheItem, [], CacheInterface::NORMAL);
+            }
+
+            return [];
+        }
+
+        $response = json_decode($response->getBody()->getContents(), true);
+        $result = array_slice($this->deduplicateClasses($this->mapItems($response)), 0, $limit);
```
votemike: Should you write a test for this method that includes making sure the deduplication works as expected?

src/ExternalApi/Ada/Service/AdaClassStubService.php
BPScott: So we avoid make network requests during unit tests which would be slooooooooow. I think there might be a way to stub out the Guzzle Client, or create mocks within the service container rather than having explicit stubs for every service to avoid this proliferation of services but I've not thought about it much.

src/ExternalApi/Ada/Service/AdaClassService.php
BPScott: I've not wrote any tests for AdaClassService yet, but this is on the list :)

src/ExternalApi/Ada/Service/AdaClassStubService.php
votemike: Yeah, I think stubbing the Guzzle Client calls would be best. Like in the `RecipesControllerTest`.

programmes-developers:

src/ExternalApi/Ada/Service/AdaClassService.php
BjornTwachtmann: I'm utterly unconcerned as to whether a specific piece of code uses arrays or stdClasses. I've tended to use stdClass purely because it's the default. If anyone wants to change that, go right ahead. 

One thing I'd mention here, this code is going to fail in a non-obvious way if what it's pasing isn't valid JSON. Probably check that the JSON items you expect are there and throw a parseException if not would be good.

src/ExternalApi/Ada/Service/AdaClassService.php
BjornTwachtmann: This could probably be moved into a base class or a helper or something. We do pretty much exactly this in about 5 different places now, yeah.

src/ExternalApi/Ada/Service/AdaClassService.php
BPScott: @BjornTwachtmann How deep do we want to fall down that validity rabbit hole? Checking for the existence of the `items` key is trivial but what if a key within a single item (such as the "title" of a class in Ada's case) is missing? Currently we'd end up throwing a TypeError (same as if the `items` key is missing in the current code, or if the JSON was not valid - and thus` json_decode()` returned null).

Sticking a bunch of `isset()` checks in each mapper feels like creating a bunch of junky boilerplate, and writing a JsonSchema / XSD for each API and validating it on each request is massive overkill.

That said, the Orbit/Branding clients check for valid JSON and a top level element only so a sanity check of that first level seems like a decent compromise between checking nothing and everything. That does mean we'd end up in a state where we handle invalid json in two ways:

* Blatant invalidness (i.e. missing top level key) is powered through and we get an empty result set back
* Subtler invalidness (i.e. missing a key within a single item) throws big exceptions.

BPScott:
I've tossed a take-two in here, that adds a base that can be used for all services, along with a base testcase for services and my service tests. @bbc/programmes can you take a look and see what you reckon?

I think this is good to go, and afterwards I'll take a look into rejigging the other ExternalAPI services to use this new base service class.

src/ExternalApi/AbstractHttpService.php
```diff
@@ -0,0 +1,87 @@
...
+        $this->client = $client;
+        $this->cache = $cache;
+        $this->logger = $logger;
+        $this->baseUrl = $baseUrl;
+    }
+
+    protected function makeCachedRequest(
+        string $url,
+        string $cacheKey,
+        callable $parseResponseBody,
```
votemike: Why is this a callable? Why not just have an abstract method in this class that people *have* to implement to parse the response body?

src/ExternalApi/Ada/Service/AdaClassService.php
```diff
@@ -0,0 +1,149 @@
...
+     * Classes that are returned from the API may be of type "tag" or "category"
+     * however in our listings we do not differentiate between these two. Thus
+     * if ADA returns a list of classes that contain a tag and a category with
+     * the same id we should remove one of them to avoid it looking like there
+     * are duplicate items in the list.
+     *
+     * @param AdaClass[] $classes
+     * @return AdaClass[]
+     */
+    private function deduplicateClasses(array $classes)
```
votemike: PHP 7 return typehint?

src/ExternalApi/Ada/Service/AdaClassService.php
```diff
@@ -0,0 +1,149 @@
...
+        ?string $programmePid = null,
+        ?string $contextPid = null,
+        ?string $countItemType = null,
+        ?string $type = null,
+        ?int $threshold = null,
+        ?string $order = null,
+        ?string $orderDirection = 'descending',
+        int $limit = 10,
+        int $page = 1
+    ) {
```
votemike: PHP7 return typehint?

BjornTwachtmann:
@BPScott , I've also been working on a common base class, as a prep for doing async requests. It looks remarkably similar to yours really...

I'll have a think about best ways to integrate the two approaches.

BPScott:
@BjornTwachtmann who wants to go first? Shall I get this merged in and you can rebase atop or should I wait for your base and go work on something else on the meantime?

BjornTwachtmann:
@BPScott , if this is ready to merge, merge it. I'm going to have to rework all the other services anyway, so one more is not much of an issue. I'll have to make some changes to deal with the additional fun imposed by parallellelllellism anyway

src/ExternalApi/AbstractHttpService.php
BPScott: I did the abstract method thing first but then I was wondering what would happen if you needed a different way of parsing within the same class, e.g. if you've got two endpoints that return the same types of entities but with a slightly different envelope. An abstract method would mean needing a new class.

My thinking was one class per entity type (potentially via many fees endpoints) rather than one class per feed (and thus potentially many classes per entity). That said I break that idea in PagesService when getting programmes, having multiple services by how entities are conceptually related (e.g. here's the a-z related service calls) to avoid having a single huuuuge service.

BPScott:
@BjornTwachtmann this is good to go (pending addressing Mike's comments). I'll update to reflect those and get this merged in later today

BjornTwachtmann:
Okay, don't worry too much about the exact details of the base class. I will have to change it substantially for the parallell stuff to avoid code duplication anyway.

BPScott:
I'll hold fire on updating the other services to use the base class if you were planning on doing it as part of a wider refactor

BjornTwachtmann:
Yeah, please dont change the other services, it's just wasted effort. I'm toying with the idea of a separate HttpApiClient class anyway to avoid some really long parameter lists I'm seeing doing sync/async requests in the same base class.

src/ExternalApi/AbstractHttpService.php
votemike: Pass a class that can do the parsing?

src/ExternalApi/Ada/Mapper/AdaClassMapper.php
BjornTwachtmann: My PR changes all json_decode stuff to return Array. So this should be consistent now

PR #340, opened 2017-12-15
Symfony34

BPScott:
Update to Symfony 3.4.1 and CsaGuzzleBundle 3.0.1  …
The ability to modify the container before using it in tests was
undeprecated so we don't need to tag the tests that used that
pattern as legacy anymore.

Liip\FunctionalTestBundle currently throws deprecations, so overwrite
the method that caused them with a version that doesn't check deprecated
features.

Updating the framework files (e.g. AppKernel, the frontcontrollers etc)
to use their latest v3.4 versions will be done in a later commit, as
part of moving the folder structure around to follow Flex conventions.

We can't update to Symfony 4.0 yet as Liip\FunctionalTestBundle doesn't
support it yet.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -8,16 +8,16 @@
     "packages": [
         {
             "name": "aws/aws-sdk-php",
-            "version": "3.36.25",
+            "version": "3.47.1",
             "source": {
                 "type": "git",
                 "url": "https://github.com/aws/aws-sdk-php.git",
-                "reference": "c7a93ab8a4aff2915f4dd1545c99f9f59e13bafd"
+                "reference": "a08d11a8fda59f1e6a21f1676ccd66c97e43aa11"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/aws/aws-sdk-php/zipball/c7a93ab8a4aff2915f4dd1545c99f9f59e13bafd",
-                "reference": "c7a93ab8a4aff2915f4dd1545c99f9f59e13bafd",
+                "url": "https://api.github.com/repos/aws/aws-sdk-php/zipball/a08d11a8fda59f1e6a21f1676ccd66c97e43aa11",
+                "reference": "a08d11a8fda59f1e6a21f1676ccd66c97e43aa11",
                 "shasum": ""
             },
             "require": {
@@ -39,7 +39,7 @@
                 "ext-dom": "*",
                 "ext-openssl": "*",
                 "nette/neon": "^2.3",
-                "phpunit/phpunit": "^4.8.35|^5.4.0",
+                "phpunit/phpunit": "^4.8.35|^5.4.3",
                 "psr/cache": "^1.0"
             },
             "suggest": {
@@ -84,7 +84,7 @@
                 "s3",
                 "sdk"
             ],
-            "time": "2017-10-11T19:12:01+00:00"
+            "time": "2017-12-14T21:45:01+00:00"
         },
         {
             "name": "bbc-rmp/cloudwatch-monitoringhandler",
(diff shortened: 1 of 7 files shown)


programmes-developers:

BPScott:
In order to ease that path to 4.0, we need to update PHPStan to 0.9.x. I tried doing that and it kicked up some more errors that look like bugs which I'm not entirely sure how to solve as some look like coding gotchas rather than "oh we just need to shuffle some of the code around"

BPScott:
Oh good, just after I made this 3.4.2 landed, which contains the translations fix. @Tasiobg If I were to update this to use 3.4.2 and remove the reference to symfony/translate is there some way for me to test this works as expected with the double pipe thing? Is there already a unit test or will I have to check something manually?

votemike:
Would it be good to make tickets for each stage of upgrading to 4.0 and reference them in your comment about what is going to be done as part of future work?

tests/BaseWebTestCase.php
```diff
@@ -66,4 +66,29 @@ protected function loadFixtures(array $fixtureNames, $omName = null, $registryNa
...
         parent::loadFixtures($classNames, $omName, $registryName, $purgeMode);
     }
+
+    /**
+     * Taken from Symfony/Bundle/FrameworkBundle/Test/KernelTestCase (this
+     * class's grandparent), as the method in
+     * Liip\FunctionalTestBundle\Test\WebTestCase (this class's parent) is
+     * currently using the KERNEL_DIR param that was deprecated in Symfony 3.4.
+     *
+     * Delete me once Liip\FunctionalTestBundle has been updated to support
```
votemike: Is there a ticket for this?

composer.lock
```diff
@@ -4,20 +4,20 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "fc9fbe9c034f0e2b5532ce21465c1d7d",
+    "content-hash": "024ae0ee335ff69310ef88ab7dff6c54",
     "packages": [
         {
             "name": "aws/aws-sdk-php",
-            "version": "3.36.25",
+            "version": "3.47.1",
```
votemike: Did you update ALL the composer packages?

Tasiobg:
@BPScott there is no unitest for that but is easy to test, I have sent you an email with a screenshot and instructions

votemike:
@BPScott Have upgraded PHPStan for you https://github.com/bbc/programmes-frontend/pull/341 Why do you think it would be useful in order to upgrade to Symfony 4?
Should we also bump the PHPStan level up to the maximum?

composer.lock
BPScott: Yeah, check the first commit. Before updating symfony I ran `composer update` to get the latest everything (symfony is locked to a single version in composer.json).

I figured that would minimise the chance of anything breaking because dependant packages weren't updated for 3.4. I didn't want to spend a bunch of time diagnosing breakages that could be fixed by having the latest dependencies.

BPScott:
@votemike PHPStan 0.8 depended on symfony/console 3.x but didn't support 4.x. PHPStan 0.9 updated its composer.json to claim it could be installed with 4.x. If we kept on PHPStan 0.8 then it would conflict and limit our max symfony version to 3.4.

This was why we had to point blogs at PHPStan dev-master for a bit https://github.com/phpstan/phpstan/issues/573.

I'm all for bumping PHPStan to max. And now that they have https://github.com/phpstan/phpstan-phpunit it might be worth installing that and getting PHPStan to cover our tests as that extension means we don't have to manually do `@var Foo|PHPUnit_Framework_MockObject_MockObject` for every single dang mock. (Previously we stopped running PHPStan on our tests because we decided that having to do all that typehinting was too damn annoying

votemike:
@BPScott https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6097

PR #342, opened 2017-12-15
Styleguide programmeobject

tatitati:
# PLEASE DONT REVIEW. WORK IN PROGRESS

Builders for:
- programme items:
      Episode (refactoring)
      Clip
- programme containers:
      Brand
      Series

### DIFF
--- a/src/Builders/AbstractCoreEntityBuilder.php
+++ b/src/Builders/AbstractCoreEntityBuilder.php
@@ -1,0 +1,135 @@
+<?php
+
+namespace App\Builders;
+
+use BBC\ProgrammesPagesService\Domain\Entity\Image;
+use BBC\ProgrammesPagesService\Domain\Entity\MasterBrand;
+use BBC\ProgrammesPagesService\Domain\Entity\Options;
+use BBC\ProgrammesPagesService\Domain\Entity\Programme;
+use BBC\ProgrammesPagesService\Domain\ValueObject\Pid;
+use BBC\ProgrammesPagesService\Domain\ValueObject\Synopses;
+
+abstract class AbstractCoreEntityBuilder implements BuilderInterface
+{
+    /** @var Programme|null */
+    protected $parent;
+
+    /** @var Pid */
+    protected $pid;
+
+    /** @var int[] */
+    protected $dbAncestryIds;
+
+    /** @var string */
+    protected $title;
+
+    /** @var string */
+    protected $searchTitle;
+
+    /** @var Synopses  */
+    protected $synopses;
+
+    /** @var Image */
+    protected $image;
+
+    /** @var int */
+    protected $promotionsCount;
+
+    /** @var int */
+    protected $relatedLinksCount;
+
+    /** @var int */
+    protected $contributionsCount;
+
+    /** @var Options */
+    protected $options;
+
+    /** @var MasterBrand|null */
+    protected $masterBrand;
+
+    protected function __construct()
+    {
+        $this->dbAncestryIds = [1212];
+        $this->pid = new Pid('d00744wz');
+        $this->title = 'my programme title';
+        $this->searchTitle = 'my programme title';
+        $this->synopses = new Synopses('My short synopsis', 'my a very medium no too much synopsis', 'my extremely boring and endless text for my long synopsis');
+        $this->image = ImageBuilder::default()->build();
+        $this->promotionsCount = 10;
+        $this->relatedLinksCount = 5;
+        $this->contributionsCount = 2;
+        $this->options = new Options();
+    }
+
+    public function withDbAncestryIds(array $dbAncestries)
+    {
+        $this->dbAncestryIds = $dbAncestries;
+        return $this;
+    }
+
+    public function withPid(string $pid)
+    {
+        $this->pid = new Pid($pid);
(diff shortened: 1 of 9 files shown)


programmes-developers:

src/Builders/AbstractCoreEntityBuilder.php
```diff
@@ -0,0 +1,135 @@
...
+namespace App\Builders;
+
+use BBC\ProgrammesPagesService\Domain\Entity\Image;
+use BBC\ProgrammesPagesService\Domain\Entity\MasterBrand;
+use BBC\ProgrammesPagesService\Domain\Entity\Options;
+use BBC\ProgrammesPagesService\Domain\Entity\Programme;
+use BBC\ProgrammesPagesService\Domain\ValueObject\Pid;
+use BBC\ProgrammesPagesService\Domain\ValueObject\Synopses;
+
+Abstract class AbstractCoreEntityBuilder
```
votemike: Lower case?

src/Builders/BrandBuilder.php
```diff
@@ -0,0 +1,58 @@
+<?php
+
+namespace App\Builders;
+
+use BBC\ProgrammesPagesService\Domain\Entity\Brand;
+
+class BrandBuilder extends AbstractProgrammeContainerBuilder implements BuilderInterface
+{
+    protected function __construct()
```
votemike: Is this useful?

src/Builders/AbstractCoreEntityBuilder.php
votemike: Should this include a `build()` method?

src/Builders/BrandBuilder.php
tatitati: I want to avoid wrong use of the builders. So I'm trying to avoid someone doing something like:
```$clip = new ClipBuilder()```

Instead I want something more friendly like:
```$clip = ClipBuilder::bigClip()->build()```
or 
```$clip = ClipBuilder::default()->build()```

In this way tests become more expressive

File: src/Builders/AbstractCoreEntityBuilder.php (hunk 1 above)
Comment (tatitati) [reply src/Builders/AbstractCoreEntityBuilder.php]: No. AbstractBuilders are "programmeItem", 'ProgrammeContainer'. But we don't want to build those objects. Instead we want to build Clips, episodes, brands, series, etc. Abstract classes are used in here just to reduce configuration of default values in childs. Otherwise these classes would have lot of default values and maintain our tests would be hard.

File: src/Builders/BrandBuilder.php (hunk 2 above)
Comment (votemike) [reply src/Builders/BrandBuilder.php]: Could you add a comment in the code making that clear?

File: src/Builders/AbstractCoreEntityBuilder.php (hunk 1 above)
Comment (votemike) [reply src/Builders/AbstractCoreEntityBuilder.php]: So you don't want `abstract public function build()`?

Comment (tatitati) [reply src/Builders/AbstractCoreEntityBuilder.php]: So far I'm using an interface BuildInterface.

Comment (votemike) [reply src/Builders/AbstractCoreEntityBuilder.php]: I can't see `BuildInterface` anywhere in this PR...

File: src/Builders/BrandBuilder.php [hunk 3]
```diff
@@ -0,0 +1,58 @@
+<?php
+
+namespace App\Builders;
+
+use BBC\ProgrammesPagesService\Domain\Entity\Brand;
+
+class BrandBuilder extends AbstractProgrammeContainerBuilder
+{
+    protected function __construct()
```
votemike: Why is this needed?

src/Builders/ClipBuilder.php
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    public static function default()
+    {
+        return new self();
+    }
+
+    public function build()
+    {
+        return new CLip(
```
votemike: `Clip`?

src/Builders/SeriesBuilder.php
```diff
@@ -0,0 +1,61 @@
...
+
+        $this->title = 'My series title';
+        $this->searchTitle = 'My search title';
+    }
+
+    /**
+     * Create a builder with default state.
+     * This default state reduce the amount of steps necessary when trying to get a final entity in an specified state
+     */
+    public static function default()
```
votemike: Could this be moved to the parent?

tatitati: I just tried it. The parent is abstract, the app crash saying that is not possible to instantiate an abstract class.

votemike: Does it work with `new static()` instead of `new self()`?

src/Builders/ClipBuilder.php
```diff
@@ -0,0 +1,54 @@
...
+{
+    protected function __construct()
+    {
+        parent::__construct();
+
+        $this->title = 'my clip title';
+        $this->searchTitle = 'my search clip title';
+    }
+
+    public static function default()
```
votemike: Do you need this method if the parent has it?

src/Builders/SeriesBuilder.php
votemike: Do you need this method if the parent has it?

PR #347, opened 2017-12-19
Added Faker to our builders for unit tests

tatitati:
Updated with Faker:
-ImageBuilder
-RelatedLinkBuilder
-PromotionBuilder
-EpisodeBuilder

Note: 
Episode builder will be modified accordingly in the next PR, with the rest of new builders related all with programmes.


src/Builders/AbstractBuilder.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace App\Builders;
+
+abstract class AbstractBuilder
+{
+    protected function __construct(){
```
tatitati: I force the people to use the static methods to create our test objects.
For example:
```RelatedLinkBuilder::externalLink()->build()```

File: src/Builders/ImageBuilder.php [hunk 2]
```diff
@@ -55,7 +46,7 @@ public function withShortSynopses(string $shortSynopses)
 
     public function withLongestSynopsis(string $longestSynopses)
     {
-        $this->shortSynopsis = $longestSynopses;
+        $this->longestSynopsis = $longestSynopses;
```
tatitati: this is actually a fix. It was wrong

programmes-developers:

src/Builders/AbstractBuilder.php
votemike: Would that be useful as a comment in the code?

src/Builders/ImageBuilder.php
```diff
@@ -73,7 +64,17 @@ public function withExtension(string $extension)
 
     public static function default()
     {
-        return new self();
+        $faker = Faker\Factory::create();
```
tatitati: I will think about this third party dependency leak. But I can do it in a future PR

src/Builders/ImageBuilder.php
```diff
@@ -73,7 +64,17 @@ public function withExtension(string $extension)
...
+
+        $self = new self();
+        $self->withPid('b00755wz')
```
votemike: You should be able to generate PIDs too if you write a provider. https://github.com/fzaninotto/Faker#faker-internals-understanding-providers

src/Builders/ImageBuilder.php
```diff
@@ -73,7 +64,17 @@ public function withExtension(string $extension)
...
+            ->withTitle($faker->text)
+            ->withShortSynopses($faker->text(20))
```
votemike: Is text appropriate here? Would `sentence` be better?

src/Builders/PromotionBuilder.php
```diff
@@ -101,7 +85,22 @@ public function withRelatedLinks(array $relatedLinks)
...
     public static function default()
     {
-        return new self();
+        $faker = Faker\Factory::create();
+
+        $self = new self();
+        $self->withPid('b00744wz')
+            ->withPromotedEntity(ImageBuilder::default()->build())
+            ->withSynopses(new Synopses('short synopsys', $faker->text(100), $faker->text(250)))
+            ->withTitle('This is the title')
```
votemike: Is there a reason this is hardcoded still?

src/Builders/RelatedLinkBuilder.php
```diff
@@ -69,10 +60,41 @@ public function withIsExternal(bool $isExternal)
...
     }
 
+    public static function internalLink()
+    {
+        $faker = Faker\Factory::create();
+
+        $self = new self();
+        $self->withTitle('This is an internal link')
+            // by default we create an external link
+             ->withUri('https://www.bbc.co.uk/something')
```
votemike: This doesn't look like an external link....

src/Builders/AbstractBuilder.php
tatitati: yeah. I will do it

src/Builders/ImageBuilder.php
tatitati: yeah, I was thinking to do it in a next PR. Firstly I would like to have faker working and later just go through all the builders doing that in the pid, yeah, but I will

src/Builders/ImageBuilder.php
tatitati: love it. Yeah, I will change it

src/Builders/ImageBuilder.php
votemike: All of these changes seem to be taking a lot of PRs.... Is it not just easier to do it in one?

BPScott:
Yay 3rd part fooling rather than writing our own stuff. 

Taking a 3 minute look at Faker it seems like it will create random data each time it is called. Will this be confusing if for instance we're looking at a styleguide page that renders objects built with Faker, as then all the titles would change every page refresh, thus making it hard to see what affect our code changes had.

Looking at their readme they mention https://github.com/thephpleague/factory-muffin as a related library that looks like it might be useful. It seems like instead of doing

```php
$myImage = ImageBuilder::default()
  ->withPid('b0000001')
  ->withExtension('png')
  ->build();
```

We could do:

```php
$fm = new FactoryMuffin();
$fm->loadFactories('some_folder_that_contains_the_definitions_of_default_values_for_each_class.php');
$myImage = $fm->->create(Image::class, [
  'pid' => 'b0000001',
  'extension' => 'png'
]);
```

I think factory muffin would mean we wouldn't need lots of builder classes where we have to define all the with* functions (and update them whenever somebody adds a new one function to the underlying model), we'd only need to specify the defaults of each particular class.


@tatitati / @votemike: What do you think?

BPScott:
I think for that random thing, it can be solved by seeding the generator in the Styleguide pages https://github.com/fzaninotto/Faker#seeding-the-generator. Then we'd always get the same data.

tatitati:
@BPScott  yeah, the seed indeed is gonna be useful.
About the muffin might work, i will give it a try to see if we can avoid to have builders. I will let you know.

votemike:
Muffin looks good. Would be great to not have to write our own builders.
As far as the random data goes, I thought it would be useful to have random data in the style guide. I'm not 100% clear on what/who/when the styleguide is for, but if it is used during development I don't think it's a good idea to only look at one set of data. Looking at how things like with 5 or 30 characters, 1/2 or 200 lines would help us see edge cases that we need to fix.

BPScott:
There was talk of Faker being used to power our styleguide pages. In that case it should be in "require" rather than "require-dev", otherwise all of our styleguide pages won't be renderable when we deploy to INT/TEST/LIVE

PR #348, opened 2017-12-20
Flex folder structure

BPScott:
Here is the first step towards moving to a Flex-based folder structure.

Everything in the app folder has been moved into new locations as per https://github.com/symfony/symfony/blob/master/UPGRADE-4.0.md. Check the individual commits for the particular moves.

* Templates now live in /templates
* Translations now live in /translations
* Sass, js and gelicons now live in /assets
* Config now lives in /config (but is still laid out in a Symfony 3.x style, reworking this is next on the list)
* The /tmp folder is dead, replaced with /var, as is the standard
* The console now lives in /bin (eventually we should move the rest of the stuff in /scripts into there too but that shall be done later)
* The App Kernel now lives in src, but still looks like a 3.x kernel, updating the Kernel to a Symfony 4 style goes hand-in-hand with reworking the config folder).

I've deployed this branch to TEST (well I did a build and there's an auto trigger there which punted it to TEST) and it seems to work as expected. It's release number 586

Please review this, and I'll get it merged into develop tomorrow and then move onto looking at restructuring the Kernel, entry points (console and the web frontcontroller) and config to follow a V4 style.


programmes-developers:

.sass-lint.yml
```diff
@@ -1,7 +1,7 @@
 options:
     formatter: stylish
 files:
-    include: ['app/Resources/sass/*.scss', 'src/Ds*/**/*.scss']
+    include: ['assets/sass/*.scss', 'src/Ds*/**/*.scss']
```
votemike: Is this right? Isn't it `resources/sass/*.scss`? That's what it is in /blogs.

votemike:
When I remove the vendor directory and try to composer install I get the error:
```
[RuntimeException]
  An error occurred when executing the "'cache:clear --no-warmup'" command:
  Could not open input file: src/console
```

votemike:
Looks pretty good so far. Just the few concerns I've mentioned.

felipeparaujo:
Shouldn't gelicons get moved to assets as well? Or are you planning on that on a later PR?

felipeparaujo:
Having the same problems as Mike when running cache:clear. Otherwise looks good.

BPScott:
Pushed a fix for the `Could not open input file: src/console`. It's been rebased into the commit that would have broke this, so we can have a stream of working commits, as I'm going to merge these commits directly rather than squashing them to preserve history.

 There's some fun with various folder structures in the SensioDistributionBundle but I can trick it into always using the new style. That bundle will be removed and the hack can go away once we start using Flex (as we use Flex's post-install hooks rather than SensioDistributionBundle's)

.sass-lint.yml
BPScott: https://github.com/symfony/symfony/blob/master/UPGRADE-4.0.md notes that assets should go in /assets. http://symfony.com/doc/current/best_practices/web-assets.html also claims it is best practice to "Store your assets in the assets/ directory at the root of your project."

I think there was a little bit of moving bits around during the beta stage and I picked up an older concept with blogs. It sounds like we should update the blogs codebase at some point.

BPScott:
Felipe has confirmed the `Could not open input file: src/console` thing works for him now, so have merged to minimise potential for drift.

PR #350, opened 2017-12-21
STYLEGUIDE: Programme-object

tatitati:
Include:
- New Builders for Clip, Brand, Series (episode was done already)
- twig template to display different programmes types/options
- Added tests for all builders: relatedLinks, Images, promotions, Brand, Episode, .....
- New end point to display Programmes in stylguide, so we have now:
        /programmes/styleguide/amen/domain/promotion
        /programmes/styleguide/amen/domain/programme


programmes-developers:

src/Builders/BrandBuilder.php
```diff
@@ -0,0 +1,49 @@
...
+
+class BrandBuilder extends AbstractBuilder
+{
+    protected function __construct()
+    {
+        $faker = Faker\Factory::create();
+
+        $this->classTarget = Brand::class;
+        // configure order of params to use Episode constructor. You are free to choose the key names, but no the order.
+        $this->blueprintConstructorTarget = [
```
tatitati: This array of params doesn't make me happy. But is this or create a comples hierarchy for builders, which makes harder to maintain them. Also using third party libraries (I was trying) you finished configuring also something like this. The good thing is that is very obvious what you are creating and upate them.

tatitati:
@BPScott after tying a couple of this party libraries and have more ideas I decided to go on my own, so I created very simple builders. I came with the idea of Muffin. So For example we changed:
```php 
return ImageBuilder::default()
           ->withPid('p01vg679')
           ->withLongestSynopsis('Image for Eastenders Brand')
           ->build();
```
To:
``` php
return ImageBuilder::anyWith([
           'pid' => new Pid('p01vg679'),
           'longestSynopsis' => 'Image for Eastenders Brand'
]);
```

As you can see is less verbose, you don't need to call build() and basically all the With* functions were replaced for only two: any() and anyWith(). Is less intuitive because now we have lost autocomplete for With* functions, but instead they are much simpler to maintain and easy to understand in a glance.

src/Controller/Styleguide/Amen/Domain/ProgrammeController.php
```diff
@@ -0,0 +1,163 @@
...
+
+class ProgrammeController extends BaseController
+{
+    public function __invoke()
+    {
+        return $this->renderWithChrome('styleguide/amen/domain/programme.html.twig', [
+            'itemsOfDifferentTypes' => $this->buildProgrammesItemsOfDifferentTypes(),
+            'itemsWithDifferentDisplayOptions' => $this->buildItemsWithDifferentDisplays(),
+            'itemsWithDifferentHtmlStructure' => $this->buildItemsWithDifferentHtmlStructure(),
+
```
votemike: WHY BLANK LINE!!!???

src/Controller/Styleguide/Amen/Domain/ProgrammeController.php
```diff
@@ -0,0 +1,163 @@
+<?php
+declare(strict_types = 1);
+namespace App\Controller\Styleguide\Amen\Domain;
+
+use App\Builders\ClipBuilder;
+use App\Builders\EpisodeBuilder;
+use App\Builders\BrandBuilder;
```
votemike: Can you run "Optimize Import" over your code please?

src/Controller/Styleguide/Amen/Domain/ProgrammeController.php
```diff
@@ -0,0 +1,163 @@
...
+    /**
+     * Use case: The user see two different programmes
+     * @return array
+     */
+    private function buildProgrammesItemsOfDifferentTypes()
+    {
+        return [
+            // programme items
+            'Programme item is Episode' => [
+                'item' => EpisodeBuilder::anyWith(['isStreamable' => true]),
```
votemike: 👍

src/Controller/Styleguide/Amen/Domain/ProgrammeController.php
```diff
@@ -0,0 +1,163 @@
...
+use App\Builders\BrandBuilder;
+use App\Builders\SeriesBuilder;
+use App\Controller\BaseController;
+use BBC\ProgrammesPagesService\Domain\ValueObject\PartialDate;
+
+class ProgrammeController extends BaseController
+{
+    public function __invoke()
+    {
+        return $this->renderWithChrome('styleguide/amen/domain/programme.html.twig', [
```
votemike: Instead of Building a load of Episodes/Clips from scratch each time, why not just create a single one with the builder. Then change the desired properties for each different display in the StyleGuide, that way it is clearer what is changing between each item in the styleguide. I believe @BPScott raised a similar issue in a different PR.

tatitati: Yeah. I did that initially, but I thought that people was trying to have objects in more direct way. Anyway, I restored that feature. Now we can delay the creation of the final object such as:

```php
$builder = BrandBuilder::any()
            ->with(['isStreamable' => false])
            ->with(['position' => 99999]);
$builder->with(['relatedLinksCount' => '100']);
$brand = $builder->build();
```
This allow to create only one initial object, and change it little by little as you said

src/Controller/Styleguide/Amen/Domain/ProgrammeController.php
```diff
@@ -0,0 +1,162 @@
...
+
+    /**
+     * Use case: The user see the same programme, but each one displays different fields. However they are the same.
+     * They have different displays only.
+     *
+     * @return array
+     */
+    private function buildItemsWithDifferentDisplays()
+    {
+        $clipBuilderDefault = ClipBuilder::any()->with(['isStreamable' => true, 'duration' => 1000]);
```
tatitati: @votemike now we use only an initial builder with the "state". This avoid to start each $clip from scratch

src/Builders/BrandBuilder.php
```diff
@@ -0,0 +1,49 @@
...
+use BBC\ProgrammesPagesService\Domain\Entity\Options;
+use BBC\ProgrammesPagesService\Domain\ValueObject\Pid;
+use BBC\ProgrammesPagesService\Domain\ValueObject\Synopses;
+use Faker;
+
+class BrandBuilder extends AbstractBuilder
+{
+    protected function __construct()
+    {
+        $faker = Faker\Factory::create();
```
votemike: Can you `use` this?

src/Builders/ClipBuilder.php
```diff
@@ -0,0 +1,50 @@
...
+
+        $this->classTarget = Clip::class;
+        // configure order of params to use Episode constructor. You are free to choose the key names, but no the order.
+        $this->blueprintConstructorTarget = [
+            'dbAncestryIds' => [$faker->randomNumber()],
+            'pid' => new Pid($faker->regexify('[0-9b-df-hj-np-tv-z]{8,15}')),
+            'title' => $faker->sentence(3),
+            'searchTitle' => $faker->sentence(4),
+            'synopses' => new Synopses($faker->sentence(5), $faker->sentence(15), $faker->sentence(30)),
+            'image' => ImageBuilder::any()->build(),
```
votemike: Is this a good idea? What if you override "image"? You've then built this Image for now reason...

tatitati: Well. Domain\Clip need an image to be built. Is mandatory. If you really care about an specific image then you override it. But this also make sense for me.

votemike: Yeah, I understand that, it just seems like wasted processing if it's just going to get overwritten straight away.

tatitati: yeah....and is true. But maybe is not too heavy

votemike: I suppose

PR #353, opened 2018-01-03
Move to a Flex style config layout

BPScott:
Use Flex as a tool for config management, update dependencies to pull in only the components we use, update frontcontrollers and console based on the new flex recipes (yay one index.php). Update config directory to use the flex style

As environment variables have caused us (Bjorn) pain in the past
continue to use a generated parameters.yaml for such config instead of
using environment variables. Thus there is no .env.dist (and a generated
.env), but parameters.yaml.dist and a generated parameters.yaml.

Comment out the mention of dotenv in index.php and bin/console to make
it clear that we know it should be there but we're choosing not to use
it

Of particular note is removing the "dynamically setting the cache provider through a parameter thing" as in Symfony 3.4 the redis connection is lazy so running the console doesn't trigger a connection failure (as it doesn't try to connect).

I've deployed this to TEST and it seems to work as expected, but if somebody else is at a loose end having them kick the tyres would be nice too. I need to confirm the fixtures bits are working.

To check this locally:
* Update your nginx config https://github.com/bbc/programmes-cloud-sandbox/pull/17
* Move config/parameters.yml to config/parameters.yaml
* Delete the var/cache folder to remove any old compiled containers


programmes-developers:

composer.json
```diff
@@ -31,59 +31,78 @@
...
         "cakephp/chronos": "^1.1",
         "csa/guzzle-bundle": "^3.0.1",
-        "doctrine/doctrine-bundle": "^1.6.8",
-        "doctrine/orm": "^2.5",
-        "incenteev/composer-parameter-handler": "^2.0",
-        "sensio/distribution-bundle": "^5.0.19",
+        "doctrine/doctrine-bundle": "^1.8",
+        "doctrine/orm": "2.5.*",
+        "incenteev/composer-parameter-handler": "^2.1",
+        "phpstan/phpstan-phpunit": "^0.9.3",
```
votemike: PHPStan shouldn't be in require surely. It should be in require-dev.

composer.json
```diff
@@ -31,59 +31,78 @@
...
         "escapestudios/symfony2-coding-standard": "^3.0.0",
         "fzaninotto/faker": "^1.7",
         "liip/functional-test-bundle": "^1.7",
         "phpstan/phpstan": "^0.9.1",
-        "phpstan/phpstan-phpunit": "^0.9.2",
-        "phpunit/phpunit": "^6.1",
         "squizlabs/php_codesniffer": "^3.0",
-        "symfony/phpunit-bridge": "^3.4.2"
+        "symfony/css-selector": "^3.4",
+        "symfony/dotenv": "^3.4",
```
votemike: Do we need this if we're not using dotenv?

composer.json
```diff
@@ -31,59 +31,78 @@
...
+        "symfony/phpunit-bridge": "^3.4",
+        "symfony/thanks": "^1.0",
```
votemike: Does this thank from a personal repository? You have to run `composer thanks` for this to actually *do* anything, right? It doesn't star everything when you composer install?
Would this be better in the suggest section?

web/index.php
```diff
@@ -0,0 +1,64 @@
...
+// The check is to ensure we don't use .env in production
+// if (!isset($_SERVER['APP_ENV'])) {
+//     if (!class_exists(Dotenv::class)) {
+//         throw new \RuntimeException('APP_ENV environment variable is not defined. You need to define environment variables for configuration or add "symfony/dotenv" as a Composer dependency to load variables from a .env file.');
+//     }
+//     (new Dotenv())->load(__DIR__.'/../.env');
+// }
+
+
+$env = $_SERVER['APP_ENV'] ?? 'dev';
```
BjornTwachtmann: Is it better to default to prod here. The idea of my dev environment having this wrong scares me a lot less than the idea of debugging stuff being exposed in production.

config/fixture_db.yaml
```diff
@@ -44,6 +44,9 @@ stof_doctrine_extensions:
             timestampable: true
 
 services:
+    App\Fixture\:
```
BjornTwachtmann: Why have you added this? Was something breaking? Most of the stuff there shouldn't want/need to be a service.

src/Kernel.php
```diff
@@ -18,53 +25,53 @@ public function __construct($environment, $debug)
...
+        $contents = require $this->getProjectDir() . '/config/bundles.php';
+        foreach ($contents as $class => $envs) {
+            if (isset($envs['all']) || isset($envs[$this->environment])) {
+                yield new $class();
+            }
+        }
     }
 
-    public function getLogDir()
+    protected function configureContainer(ContainerBuilder $container, LoaderInterface $loader)
```
BjornTwachtmann: This looks like the magic that's loading all our parameters/per-env config. Have you checked all this is cached etc. ? I have no idea how this works.

BjornTwachtmann:
That is a lot of changes. It broadly looks fine to me, it's going to need a lot of testing though...

web/index.php
votemike: The environment defaulting to dev shouldn't automatically enable debugging...

composer.json
BPScott: Goof when manually resolving conflicts while rebasing onto latest `develop`. Moved to require-dev

composer.json
BPScott: Good point. Removed entirely (and tweaked web/index.php and bin/console to reflect that.

composer.json
BPScott: Yeah. This was a copy+paste job from https://github.com/symfony/skeleton which is used to bootstrap Flex apps. Symfony have now changed how they want to do this, so I've removed it from our composer.json

config/fixture_db.yaml
BPScott: Wiring ScenarioManager was complaining that App\Fixture\UrlMatching\UrlReverseEvangelist wasn't defined, as I've turned on the switches for stricter autowiring where you have to explicitly define every class rather than it trying to discover dependant classes.

I put this in to avoid a game of service dependency whack-a-mole, but it turns out only that one class was missing, so have added that instead of registering all.

src/Kernel.php
BPScott: Yep this is still all cached. The same stuff as before is happening under the hood, but with slightly different convenience functions at the top level. This Kernel is taken from the Kernel you get when you create a new Flex app - https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/3.3/src/Kernel.php

I've compared the dumped output of the containers from before/after this and it's all the equivalent aside from a few changes we don't care about such as enabling some optimisations and no longer automatically registering the some of the symfony translation stuff (e.g twig extensions) as use our own implementation

BPScott:
Replying to @BjornTwachtmann here for visibility as GH has decided to collapse the original comment:

> Is it better to default to prod (when setting $env ins web/index.php)? The idea of my dev environment having this wrong scares me a lot less than the idea of debugging stuff being exposed in production.

We could default to prod, if we set the APP_ENV=dev param in the server config in the sandbox, or bring back a .env file that only contains the APP_ENV parameter which seems annoying for such a small thing. 

The line after this explicitly disables debug mode in our prod and prod_* environments so the only way debug mode can be triggered is using the dev environment, which we would immediately notice on INT/TEST/LIVE as everything would explode - notably requesting Orbit would fail due to incorrectly configured cert/PKI locations.  Thus I think the chance of accidentally enabling debug / the wrong env is very slim and not worth the effort of doing a workaround

PR #354, opened 2018-01-03
PROGRAMMES-6040 Adding preloading

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6040


programmes-developers:

config/config.yml
```diff
@@ -25,6 +25,8 @@ framework:
     http_method_override: false
     assets:
         base_path: 'assets'
+    web_link:
```
BPScott: When you rebase this onto the Flex stuff you'll need to `composer require symfony/web-link`, after which the framework config shall automatically detect its presence  and enable it so you won't need to add this config line

BPScott:
I've been doing some reading of https://www.w3.org/TR/preload/, https://dexecure.com/blog/http2-push-vs-http-preload/ and https://jakearchibald.com/2017/h2-push-tougher-than-i-thought/.

HTTP/2 push sounds scronky and more hassle than it's worth, this includes the pushes that may be triggered by a preload Link header, so whenever we do this stuff I think it's worth adding the `nopush` directive as per [the spec](https://www.w3.org/TR/preload/#server-push-http-2). I'm fairly sure this is a non-issue anyway as the connection for www.bbc.co.uk isn't considered authoritative for any of our *.files.bbci.co.uk static asset hosts (check the certificate for www.bbc.co.uk and note how those wildcards aren't mentioned). Thus push would never be triggered for any of our static assets but it's probably worth being explicit.


---

```html
<link href="{{ preload(asset('css/ds2013.css'), { as: 'style' }) }}" />
<link rel="stylesheet" href="{{ asset('css/ds2013.css') }}" />
```

This seems wrong to me. That first link isn't valid HTML as it is missing a `rel` tag and your use of the preload() function adds the `link` header, thus making having a `<link rel="preload">` element in the html superflous anyway - as the browser parsers the headers before parsing the html. The aim of preload seems to be helping to load resources before they would otherwise be discovered by any kind of [lookahead preparser](https://andydavies.me/blog/2013/10/22/how-the-browser-pre-loader-makes-pages-load-faster/). But you've got nothing between these two links - the browser preloader would have discovered this asset at the same time you're hinting so it wouldn't have sped anything up at all. It would be useful if you had a load of slow, blocking javascript between the `<link rel="preload ...>` and the `<link rel="stylesheet" ...>` but we don't.


The correct format for this should be:

```html
<link rel="stylesheet" href="{{ preload(asset('css/ds2013.css'), { as: 'style' }) }}" />
```

This will render the html `<link rel="stylesheet" href="big-old-path-prefix/css/ds2013-hash.css">` and also add a `link:<big-old-path-prefix/css/ds2013-hash.css>; rel="preload"; as="style"` HTTP header.

templates/schedules/by_day.html.twig
```diff
@@ -10,6 +10,16 @@
     </div>
 {%- endblock -%}
 
+{% block inline_head %}
+    {{ parent() }}
+    {% if service.isInternational() %}
+        <link href="{{ preload(asset('js/timezone-notification.js'), { as: 'script' }) }}" />
+    {% endif %}
+    {%- if other_networks -%}
+        <link rel="preload" href="{{ asset('images/logos/services_channel_sprite.png') }}" as="image">
```
BPScott: This is right at the bottom of the page and doesn't block any kind of rendering or layout. I don't think it counts as a critical resource worthy of preloading.

BPScott:
As for the rv-bootstrap stuff, I think this might be a race condition due to bad r.js config...

On the client side `require()` loads modules. These tend to map onto files, e.g. rv-bootstrap.js contains the module 'rv-bootstrap'. Thanks to the r.js optimizer (which runs as part of gulp) it may be the case that a single file contains many modules - as dependencies are inlined. Indeed if you take a look into rv-bootstrap.js you'll see `define('rv-bootstrap'` along with `define('lazyload'` and all the other modules rv-bootstrap depends upon (i.e. [all the things in the first line of the source file](https://github.com/bbc/programmes-frontend/blob/develop/assets/js/rv-bootstrap.js#L1)).

In the case of `timezone-notification` it depends upon `rv-bootstrap`, and because of our current r.js config that means all of `rv-bootstrap` and its dependencies get inlined into `timezone-notification.js`. 


On the client side that file gets loaded, if  `timezone-notification.js` is really fast at being loaded it might not need to make a request for `rv-bootstrap.js` as part of [this call](https://github.com/bbc/programmes-frontend/blob/PROGRAMMES-6040-http2-preload-push/templates/base_ds2013.html.twig#L97) because the rv-bootstrap module was already loaded - as it was part of `timezone-notification.js`.

This is pretty dumb and excluding the contents of rv-bootstrap from timezone-notification.js would make it a bit more sensible - which is what I've done in #355.

votemike:
@BPScott in what was does HTTP/2 push sound scronky? Are you saying that you don't want us to use it?
Looking at the network tab for https://www.bbc.co.uk/schedules and limiting to `domain:*.bbc.co.uk` it looks like there aren't many resource on the domain we're authoritative for, which does sort of limit the usefulness of _push_ and the `link:` header. Are there any resources in that subset that you think we could load in this way, or shall I just ditch the whole push thing?
With regards to the HTML, I was just following the examples from https://symfony.com/blog/new-in-symfony-3-3-asset-preloading-with-http-2-push which seem to show the preload and the HTML that actually uses the asset as separate tags.
I realise that having the HTML tags right next to each other is not useful if you are only relying on the HTML, but if we keep using `link:`, then the resource will be requested well before the actual tag if found.

From your comments so far, I get the impression you don't think any of the resources would benefit from preloading. Or am I getting that wrong? Which ones do you think should be preloaded? And do you think this should just be done via the HTML tag rather than Symfony's preload function?

BjornTwachtmann:
@votemike This is about preload rather than push right? We can't really use push as our assets are on a physically separate IP than our HTML and thus won't share an HTTPS connection. Explicitly indicating this to the browser seems sensible. 
Preload still seems valuble, if not that well supported right now.

votemike:
Preload only seems useful for requesting assets that will be requested from JS or CSS files.... Were there some assets you had in mind?

BjornTwachtmann:
I dunno, I haven't really been though everything that gets loaded. You don't think requesting preload of the CSS/JS will speed things up?

votemike:
Not as far as I can tell.

PR #355, opened 2018-01-04
Allow gulp r.js config to exclude modules

BPScott:
The timezone-notification module depends upon rv-bootstrap but should
not have the contents of rv-bootstrap inlined into it. This reduces the
size of the timezone-notification JS from 7.4kb to 1.9kb. rv-bootstrap
should be loaded as a separate file from the client side.

Look at https://programmes.files.bbci.co.uk/programmes-frontend/js/timezone-notification-7f8974d903.js and note that it contains all the contents of rv-bootstrap module and it's dependants. Duplication as the user has already probably already loaded rv-bootstrap.js so they don't need all those module definitions all over again.

This PR excludes the contents of rv-bootstrap from timezone-notification to reduce the bundle size.


programmes-developers:

PR #356, opened 2018-01-04
PROGRAMMES-6063 Improving Brand Page Querying

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6063

Only changed this one line as I don't think we need to cache the queries in the way suggested on the ticket.


src/Controller/FindByPid/TlecController.php
```diff
@@ -86,7 +86,7 @@ public function __invoke(
             $galleries = $aggregationService->findDescendantGalleries($programme, 4);
         }
 
-        $lastOn = $collapsedBroadcastsService->findPastByProgrammeWithFullServicesOfNetworksList($programme, 1);
+        $lastOn = $collapsedBroadcastsService->findPastByProgramme($programme, 1);
```
votemike: @felipeparaujo You originally wrote this line, is there a reason you chose `findPastByProgrammeWithFullServicesOfNetworksList()` over `findPastByProgramme()`?

programmes-developers:

src/Controller/FindByPid/TlecController.php
felipeparaujo: The list of services for the network is necessary to generate the network and services details for a collapsed broadcast, [here](https://github.com/bbc/programmes-frontend/blob/develop/src/DsShared/Helpers/BroadcastNetworksHelper.php) . Looking at the code, it doesn't look like that's happening, although I can't remember why I changed it. Your change is probably safe.

PR #357, opened 2018-01-04
PROGRAMMES-6009 Fix wrong message for case

tatitati:
From [map_on_demand_column.feature](https://github.com/bbc/programmes-frontend-automation-aws/blob/master/cucumber/features/map_on_demand_column.feature#L165) we can see that we need only three messages:

|                          | Msg when No episodes                         | Msg when There will be episodes soon                      |
|--------------------------|---------------------------------------------|-----------------------------------------------------------|
| user in TV brand page    | "No available"                              | "This programme will be available shortly after broadcast" |
| user in Radio brand page | "There are currently no available episodes" | -                                                         |

Instead we have this:

|                          | Msg when No episodes                         | Msg when There will be episodes soon                      |
|--------------------------|---------------------------------------------|-----------------------------------------------------------|
| user in TV brand page    | < same >                              | < same > |
| user in Radio brand page | "This programme is not currently available on BBC iPlayer Radio" | -                                                         |

Logic of template is here:
https://github.com/bbc/programmes-frontend/blob/develop/src/DsAmen/Presenters/Section/Map/SubPresenter/on_demand.html.twig#L5-L17

URLs  For testing the message after fixing it:
http://sandbox.bbc.co.uk/programmes/b00ty69k   -> MSG: "not available"
http://sandbox.bbc.co.uk/programmes/b01d76m6 -> MSG: "There are currently no available episodes"


programmes-developers:

translations/programmes/en.po
```diff
@@ -1528,7 +1528,7 @@ msgstr "This programme is not currently available on BBC iPlayer"
 #. Example URL: http://www.bbc.co.uk/programmes/b03kpkyz .
 #. Priority for: Programmes
 msgid "programme_availability_none_radio"
-msgstr "This programme is not currently available on BBC iPlayer Radio"
+msgstr "There are currently no available episodes"
```
votemike: This string already exists twice in this file. Do we really want to make it a third?

votemike: What about other languages?

votemike: You'll also need to updated the `English:` bit in programmes.pot

tatitati: Good question, @BjornTwachtmann, should I translate the rest of texts on my own?, is there someone else translating to different languages?. I just did it in the english version, not sure if I should carry on with the rest of languages

tatitati: True, actually we have all this:
```
msgid "available_count"
msgid_plural "available_count %count%"
msgstr[0] "There are currently no available episodes"
....

msgid "available_count_short"
msgid_plural "available_count_short %count%"
msgstr[0] "There are currently no available episodes"
....
```

@BjornTwachtmann so far I consider that we can have repeated translations for different contexts and they can change to different things without to be worried about other contexts. This said, is it fine to add another translation that we already have?

tatitati:
Also I saw that we have for that text different translations, for french, arabic,  and other lenguages. 
Might be useful If I add a translation to the spanish? (I wish I knew more lenguages, but..jeje)

src/DsAmen/Presenters/Section/Map/SubPresenter/OnDemandPresenter.php
```diff
@@ -110,6 +117,12 @@ public function getTitleTranslationString(): string
         return $this->programmeContainer->isRadio() ? 'available_now' : 'available_on_iplayer_short';
     }
 
+    public function getOnDemandNotAvailableTranslationString(): string
+    {
+        $tr = $this->translateProvider->getTranslate();
+        return $this->programmeContainer->isRadio() ? $tr->translate('available_count', [], 0) : $tr->translate('not_available_iplayer');
```
tatitati: there are different options for "available_count":
```
msgid "available_count"
msgid_plural "available_count %count%"
msgstr[0] "There are currently no available episodes"
msgstr[1] "There is currently %count% available episode"
msgstr[2] "There are currently %count% available episodes"
```

In order to force the 0 option, I decided to inject the translateProvider into the onDemandPresenter.

translations/programmes/en.po
```diff
@@ -1527,6 +1527,7 @@ msgstr "This programme is not currently available on BBC iPlayer"
 #. English: "This programme is not currently available on BBC iPlayer Radio".
 #. Example URL: http://www.bbc.co.uk/programmes/b03kpkyz .
 #. Priority for: Programmes
+#. The only use of this key was replaced for the available_count key
```
tatitati: Should I delete this translation?

src/DsAmen/Presenters/Section/Map/SubPresenter/OnDemandPresenter.php
BjornTwachtmann: That's fine. But there's also a getBadgeTranslationString function here. It's not necessarily clear to the caller that one returns a translated string and the other returns a string to be translated.

You might as well just call tr from the other function in here too if you're passing in the provider.

BjornTwachtmann: or just rename this function if that's easier. Either way.

translations/programmes/en.po
BjornTwachtmann: mmm... leave it for now. People tend to change their minds and it might be useful to have.

PR #358, opened 2018-01-04
Master patch fix apcu

BjornTwachtmann:


programmes-developers:

votemike:
Do we have a test covering this?

src/Metrics/Cache/MetricCacheApcu.php
```diff
@@ -207,17 +207,12 @@ private function tryToLockOnKey(string $key, int $timeout) : bool // 0 timeout =
...
-                $key,
-                function ($key) use (&$lock) {
-                    $lock = true;
-                    return 1;
-                },
-                $timeout
-            );
+            /**
+             * apcu_entry is broken (horribly) in APCu >= 5.1.8 . That actually supports locking.
+             * This doesn't really. There's a slight risk of more than one request getting hold of this lock.
```
votemike: What happens if two requests get a hold of the lock?

src/Metrics/Cache/MetricCacheApcu.php
```diff
@@ -207,17 +207,12 @@ private function tryToLockOnKey(string $key, int $timeout) : bool // 0 timeout =
...
+             */
+            $lock = apcu_add($key, 1, $timeout);
```
votemike: From the PHP docs:
> Note: Unlike many other mechanisms in PHP, variables stored using apcu_add() will persist between requests (until the value is removed from the cache).
Are you aware of this?

tatitati: yeah, is confusing, I also read this in ttl:
> Time To Live; [....] If no ttl is supplied (or if the ttl is 0), the value will persist until it is removed from the cache manually, or otherwise fails to exist in the cache (clear, restart, etc.).

PR #359, opened 2018-01-05
Update recipes baseurl on int/test and make errors from external APIs…

BjornTwachtmann:
… more visible in logs


programmes-developers:

PR #362, opened 2018-01-08
Add findByPid routes to metrics, as they are not dicoverable from rou…

BjornTwachtmann:
…ting YAML


programmes-developers:

PR #364, opened 2018-01-09
PROGRAMMES-6086 Improving accessibility

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6086

This PR disabled tabbing to SVGs and the CTA links. This means that when tabbing through the schedules page, you should only tab across the programme links rather than the iPlayer links (except the live "watch from start" link).
I added the time to the programme title when it is read out, but this requires a fair amount of re-jigging, do you think it's more detrimental than not having the time read out.
For some reason on the TLEC page, when tabbing to the iPlayer link, the PID was read out instead of the invisible span, so I moved it to be an aria-label, which I think is more semantically correct anyway.
 
Do people think these are the correct way to go about making these items more accessible?


programmes-developers:

bassani2016:
Thinking about it abit further if someone was blind then they would never be able to watch the episode live and only from the beginning of the episode. Also if the svg's of programmes that have already been broadcasted are disabled then a blind person won't be able to play their programme directly from the schedules page and instead will have to go to the episode page to then go to the iplayer page.

Bare in mind that tabbing to links is a common way for someone who's visually impaired to navigate but it's not the only way; depending on the reader tool (NVDA /JAWS for exp) it may allow the user to navigate element by element on the page so have we checked to see how the reader reacts when navigating to the SVG using element by element?

Just to clarify, are we saying we're not including the time when navigating to the programme title? If so then how is a blind person going to know what time their programmes on when it's selected?

BjornTwachtmann:
How horrific would it be to add the time in here? It is kind of an important element of the schedule. I don't want to make a mess of the codebase though. 

What's here looks fine apart from that. At some point when we're under less time pressure it would be good to look at a way to get the iplayer links working too. Especially with the way radio playout is going.

votemike:
@bassani2016 We spoke with Cathal yesterday and he was OK with disabling the tabbing to CTA buttons. The people can always go to iPlayer via the programme page. The SVGs are only tabbed to in IE (which is why you needed to tab twice to get off of them).

Do a significant number of our users navigate our schedule pages SVG by SVG?

Yes, no time will be narrated at present. In the same way it doesn't on LIVE at the moment.

@BjornTwachtmann It would require overriding the Programme Title presenter so that it accepts a Broadcast. Or modifying the current presenter to optionally accept a broadcast.

I think a redesign to avoid nesting links would be a good start to get the iPlayer links working.... nit sure if that is asking too much.

votemike:
I have now added in the time to the links to programme pages as an aria-label.

src/Ds2013/Presenters/Domain/Programme/BroadcastSubPresenters/BroadcastProgrammeTitlePresenter.php
```diff
@@ -0,0 +1,48 @@
...
+        Programme $programme,
+        array $options = []
+    ) {
+        parent::__construct($router, $titleHelper, $programme, $options);
+        $this->broadcast = $broadcast;
+    }
+
+    public function getAriaTitle(): string
+    {
+        $title = $this->broadcast->getStartAt()->setTimezone(ApplicationTime::getLocalTimeZone())->format('H:i') . ': ';
```
BjornTwachtmann: Looking at the output, e.g.
"06:00: Breakfast, 10/01/2018"
that colon might not be the best separator here

PR #365, opened 2018-01-09
Fix for a bug where services of networks weren't being fetched for world news

felipeparaujo:


programmes-developers:

src/Controller/FindByPid/TlecController.php
```diff
@@ -86,7 +86,12 @@ public function __invoke(
             $galleries = $aggregationService->findDescendantGalleries($programme, 4);
         }
 
-        $lastOn = $collapsedBroadcastsService->findPastByProgramme($programme, 1);
+        if ($programme->getNetwork() && $programme->getNetwork()->isWorldNews()) {
```
votemike: Are you going to use the new `isWorldNews()` method in the MapPresenter?

votemike:
Maybe a test that fails without this change?

src/Controller/FindByPid/TlecController.php
```diff
@@ -181,6 +171,21 @@ private function getComingSoonPromotion(ImagesService $imagesService, ProgrammeC
         );
     }
 
+    private function getLastOn(ProgrammeContainer $programme, CollapsedBroadcastsService $collapsedBroadcastsService)
```
votemike: return Typehint?

PR #367, opened 2018-01-10
Http Api Common Code Refactor

BjornTwachtmann:
* Move common code related to caching/guzzle exceptions into a separate class
* Allow that class to make async requests (but don't do so yet)
* Make all uses of json_decode return an array for consistency while I'm in here anyway.


programmes-developers:

src/ExternalApi/RecEng/Service/RecEngService.php
```diff
@@ -194,10 +163,13 @@ private function parseResult(string $response, int $limit): array
             }
 
             if ($numResults >= $limit) {
-                return $pids;
+                break;
```
votemike: Does this `break` suggest this method should be refactored?

phpstan.neon
```diff
@@ -14,3 +14,4 @@ parameters:
         - '#Call to an undefined method BBC\\ProgrammesPagesService\\Domain\\Entity\\CoreEntity::getDuration()#'
         - '#Call to an undefined method Tests\\App\\Translate\\TranslatableTraitTest::tr().#'
         - '#Call to an undefined method Tests\\App\\Translate\\TranslatableTraitTest::localDateIntl().#'
+        - '#Call to an undefined method GuzzleHttp\\Exception\\GuzzleException::getMessage\(\)#'
```
votemike: If this is not ignored, where is it triggered?

src/ExternalApi/Client/HttpApiClient.php
```diff
@@ -0,0 +1,154 @@
...
+
+    public function __construct(
+        ClientInterface $client,
+        CacheInterface $cache,
+        LoggerInterface $logger,
+        string $cacheKey,
+        string $requestUrl,
+        callable $parseResponse,
+        array $parseResponseArguments,
+        $nullResult,
```
votemike: What's this?

src/ExternalApi/Client/HttpApiClient.php
```diff
@@ -0,0 +1,154 @@
...
+            return $this->handleGuzzleException($e);
+        }
+        return $this->handleResponse($response);
+    }
+
+    public function makeCachedPromise(): PromiseInterface
+    {
+        $this->cacheItem = $this->cache->getItem($this->cacheKey);
+        if ($this->cacheItem->isHit()) {
+            return new FulfilledPromise($this->cacheItem->get());
```
votemike: So the item is cached when the promise is resolved? So API calls will continue being sent out until that point?

src/ExternalApi/Client/HttpApiClient.php
felipeparaujo: What to return in case of a non-200 response from the api

votemike: I think there should be a docblock or something explaining that. Or do we not want to enforce it to something for consistency?

src/ExternalApi/Client/HttpApiClientFactory.php
```diff
@@ -0,0 +1,58 @@
...
+            $requestUrl,
+            $parseResponse,
+            $parseResponseArguments,
+            $nullResult,
+            $standardTTL,
+            $notFoundTTL
+        );
+    }
+
+    public function keyHelper(string $className, string $functionName, ...$uniqueValues): string
```
votemike: Would you expect a key helper to be in a factory?

src/ExternalApi/Client/HttpApiClient.php
BPScott: We can't enforce because the result may differ between services. In most cases we return arrays of domain items, so nullResult is `[]`, however in some cases (hi recipes) we return a result object `new RecipesApiResult([], 0);`

src/ExternalApi/Client/HttpApiClientFactory.php
votemike: Would the keyHelper method be better as some sort of global helper function rather than specifically on the Cache?

BPScott: Not really, but the alternative is making the whole Cache object public. Having a smaller api area by not exposing that is preferable IMO.

src/ExternalApi/Client/HttpApiClient.php
votemike: Docblocks can also document as well as enforcing type.

BPScott: Agreed that documentation of that would be useful :)

src/ExternalApi/Client/HttpApiClient.php
BjornTwachtmann: No, the item is already retrieved from the cache and is wrapped in a sort of fake promise in order that cache hits don't return something different from cache misses. No extra API calls will be triggered on a cache hit.

src/ExternalApi/Client/HttpApiClientFactory.php
BjornTwachtmann: Yeah, what ben said really. This is a bit counter-intuitive, but the alternatives all look worse :)

src/ExternalApi/RecEng/Service/RecEngService.php
BjornTwachtmann: This could probably be pulled out into a separate method, yeah, I'll do that...

src/ExternalApi/Client/HttpApiClient.php
BjornTwachtmann: Will add some documentation on what the parameters are, yeah, not all obvious.

phpstan.neon
BjornTwachtmann: I'm catching GuzzleException (which is an interface implemented by all Guzzle exceptions) and PHPStan is complaining that it doesn't have the getMessage method. However, getMessage is on the base PHP Exception class, so no exception will ever not have that method. Hence this.

PR #371, opened 2018-01-11
Making tester's life easier by adding an attribute to recipes placeholder

felipeparaujo:


programmes-developers:

votemike:
I'm not convinced that we should be making tester's lives easier....

PR #372, opened 2018-01-12
Add Ada top topics to Brand page

BPScott:
Use [In Our Time](http://www.bbc.co.uk/programmes/b006qykl) as an example. 

Eventually we'll need to update to make the tracking link location and the visibility of count configurable when we do the episode page, but that can happen when we need those options.

### DIFF
--- a/src/DsAmen/Presenters/Section/RelatedTopics/_related-topics.scss
+++ b/src/DsAmen/Presenters/Section/RelatedTopics/_related-topics.scss
@@ -1,0 +1,41 @@
+.related-topics {
+    // Bottom padding is less to account for the spacing that occurs due to the
+    // margin-bottom on the individual li items
+    padding: 8px 8px 0;
+    margin-bottom: 8px;
+
+    line-height: 1;
+    // Remove whitespace caused by `inline-block`.
+    letter-spacing: -0.31em;
+    word-spacing: -0.43em;
+
+    @include mq($from: gel3) {
+        padding: 16px 16px 8px;
+        margin-bottom: 16px;
+    }
+}
+
+.related-topics li {
+    display: inline-block;
+    list-style: none;
+    margin-right: 8px;
+    margin-bottom: 8px;
+
+    letter-spacing: normal;
+    word-spacing: normal;
+}
+
+.related-topics li:last-child {
+    margin-right: 0;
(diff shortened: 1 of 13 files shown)


programmes-developers:

src/DsAmen/Presenters/Section/RelatedTopics/_related-topics.scss
```diff
@@ -0,0 +1,41 @@
...
+    letter-spacing: -0.31em;
+    word-spacing: -0.43em;
+
+    @include mq($from: gel3) {
+        padding: 16px 16px 8px;
+        margin-bottom: 16px;
+    }
+}
+
+.related-topics li {
```
votemike: Any reason these aren't all nested inside the `.related-topics` block above?

src/DsAmen/Presenters/Section/RelatedTopics/related_topics.html.twig
```diff
@@ -0,0 +1,9 @@
+<div class="br-box-secondary">
+    <h2 class="gel-double-pica-bold island">{{ tr('top_topics') }}</h2>
+    <ol class="related-topics br-keyline media__meta-row--separator">
+        {%- for adaClass in related_topics.getAdaClasses() -%}
+            {%- set url = path(adaClass.getProgrammeItemCountContext() ? 'programme_topics_topic' : 'topics_topic', {pid: adaClass.getProgrammeItemCountContext(), topic: adaClass.getId()}) -%}
```
votemike: Should this not be a method in the presenter? Amongst other things, so that it can be tested.

src/DsAmen/Presenters/Section/RelatedTopics/_related-topics.scss
BPScott: I strongly dislike nesting and instead prefer to keep to a style similar to how you'd write plain CSS where possible. I feel it is much more readable and makes it more obvious what the outputted CSS looks like, especially if you're not too familiar with how Sass works.

BjornTwachtmann:
This looks good to me. I have very little idea how the ADA pages work, so my opinion is probably not worth an awful lot, but I'm happy for this to be merged.

PR #373, opened 2018-01-12
No truncation by default in Amen Programme objects

BPScott:
The DsAmen behaviour now matches the Ds2013 programme object.

I've renamed the option to 'truncation_length' to match the name used by the Ds2013 programme object.

Programmes/Broadcasts in the MAP should still be truncated to 60 chars explicitly.


programmes-developers:

votemike:
Could we have the ticket that a PR is for in the Pull Request so that we can make sure that the PR is actually fulfilling the ticket when we review?

BPScott:
@votemike Check the commit messages. Second one states "Fixes PROGRAMMES-6059"

votemike:
@BPScott So we should go in to the commits tab and click on each one to open up the extended commit message to see which ticket a PR is fixing?
Would it not be easier to mention it in the PR description (or title) to avoid searching for it?

PR #374, opened 2018-01-15
Drop support for rendering Franchise pages

BPScott:
Franchises are being phased out and will not be supported in v3.
Explictly throw an error when finding a page that is a franchise.


programmes-developers:

src/ArgumentResolver/ContextEntityByPidValueResolver.php
```diff
@@ -86,6 +86,10 @@ public function resolve(Request $request, ArgumentMetadata $argument): Generator
                 (new ReflectionClass($type))->getShortName()
             );
 
+            if ($entity instanceof Franchise) {
+                throw new NotFoundHttpException(sprintf('The item with PID "%s" was a franchise, which v3 does not support', $pid));
```
BjornTwachtmann: Is this message going to be visible to the user? If so I would suggest writing something they'd be more likely to understand.

BPScott: These messages are only visible in debug mode. e.g. contrast https://www.bbc.co.uk/schedules/p0000000 and https://sandbox.bbc.co.uk/schedules/p0000000

PR #375, opened 2018-01-15
Add skeleton for /programmes/:pid/episodes/player

BPScott:
Move the route into the main routes.yaml, adds a controller and view and
builds the context titling.

Fixes PROGRAMMES-6140 and PROGRAMMES-6137

---

There's a ProgrammeEpisodes namespace there because there's about 6 or so routes that hang off /programmes/:pid/episodes so grouping them would be handy.

I should probably start some controller tests for this but it feels like slightly early days given its still missing the subnav and pagination

### DIFF
--- a/src/Controller/BaseController.php
+++ b/src/Controller/BaseController.php
@@ -84,6 +84,25 @@ abstract class BaseController extends AbstractController
         return $this->canonicalUrl;
     }
 
+    protected function getPage(): int
+    {
+        $pageString = $this->request()->query->get(
+            'page',
+            '1'
+        );
+
+        if (ctype_digit($pageString)) {
+            $page = (int) $pageString;
+            // Have a controlled upper-bound to stop people putting in clearly
+            // absurdly large page sizes
+            if ($page >= 1 && $page <= 9999) {
+                return $page;
+            }
+        }
+
+        throw $this->createNotFoundException('Page parameter must be a number between 1 and 9999');
+    }
+
     protected function setBrandingId(string $brandingId)
     {
         $this->brandingId = $brandingId;
--- a/tests/Twig/HtmlUtilitiesExtensionTest.php
+++ b/tests/Twig/HtmlUtilitiesExtensionTest.php
@@ -4,14 +4,12 @@ namespace Tests\App\Twig;
 
 use App\Twig\HtmlUtilitiesExtension;
 use PHPUnit\Framework\TestCase;
-use PHPUnit_Framework_MockObject_MockObject;
 use Symfony\Component\Asset\Packages;
 
 class HtmlUtilitiesExtensionTest extends TestCase
 {
     public function testAssetJs()
     {
-        /** @var Packages|PHPUnit_Framework_MockObject_MockObject $mockPackages */
         $mockPackages = $this->createMock(Packages::class);
         $mockPackages->expects($this->once())->method('getUrl')
             ->with('some/example/path.js')
--- a/config/routes.yaml
+++ b/config/routes.yaml
@@ -22,10 +22,19 @@ find_by_pid:
     defaults: { _controller: '!find_by_pid' }
     requirements: { pid: '[0-9b-df-hj-np-tv-z]{8,15}' }
 
+# Programme information pages
+
+programme_episodes_player:
+    path: /programmes/{pid}/episodes/player
+    defaults: { _controller: App\Controller\ProgrammeEpisodes\PlayerController }
+    requirements: { pid: '[0-9b-df-hj-np-tv-z]{8,15}' }
+
 partial_programme_recipes:
     path: /programmes/{pid}/recipes.ameninc
     defaults: { _controller: App\Controller\Partial\RecipesController }
     requirements: { pid: '[0-9b-df-hj-np-tv-z]{8,15}' }
+
+# Schedules
 
 schedules_home:
     path: /schedules
--- a/config/routes/3rd_party.yaml
+++ b/config/routes/3rd_party.yaml
@@ -78,12 +78,6 @@ programme_episodes_guide:
     requirements: { pid: '[0-9b-df-hj-np-tv-z]{8,15}' }
     schemes: [http]
 
-programme_episodes_player:
-    path: /programmes/{pid}/episodes/player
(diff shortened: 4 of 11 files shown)


programmes-developers:

src/Controller/BaseController.php
```diff
@@ -84,6 +84,25 @@ protected function getCanonicalUrl(): string
...
     }
 
+    protected function getPage(): int
+    {
+        $pageString = $this->request()->query->get(
+            'page',
+            '1'
+        );
+
+        if (ctype_digit($pageString)) {
```
votemike: Is this better than `is_int` in some way?

src/Twig/HtmlUtilitiesExtension.php
```diff
@@ -75,4 +83,34 @@ public function truncate(string $string, ?int $length, string $suffix = '…')
...
+            );
+        };
+
+        $trailingLinks = implode(', ', array_map($getEntityLink, $ancestry));
+        $links = $getEntityLink($tleo) . ($trailingLinks ? ' ' . $trailingLinks : '');
+
+        return '<span class="context">' . $links . '</span>';
+
+
+        // return  '<span class="context">%s ' '</span>'
```
votemike: What's this bit?

tests/Twig/HtmlUtilitiesExtensionTest.php
```diff
@@ -33,4 +40,50 @@ public function testBuildCssClasses()
...
+            'one item' => [$greatGrandparent, '<span class="context"><a class="context__item" href="/programmes/b0000001">great-grandparent</a></span>'],
+            'two items' => [$grandparent, '<span class="context"><a class="context__item" href="/programmes/b0000001">great-grandparent</a> <a class="context__item" href="/programmes/b0000002">grandparent</a></span>'],
+            'three items' => [$parent, '<span class="context"><a class="context__item" href="/programmes/b0000001">great-grandparent</a> <a class="context__item" href="/programmes/b0000002">grandparent</a>, <a class="context__item" href="/programmes/b0000003">parent</a></span>'],
+            'four items' => [$child, '<span class="context"><a class="context__item" href="/programmes/b0000001">great-grandparent</a> <a class="context__item" href="/programmes/b0000002">grandparent</a>, <a class="context__item" href="/programmes/b0000003">parent</a>, <a class="context__item" href="/programmes/b0000004">child</a></span>'],
+        ];
+    }
+
+    private function buildMockCoreEntity(string $pid, string $title, array $parents)
+    {
+        $programme = $this->createMock(CoreEntity::class);
```
votemike: Should this use the builders?

src/Controller/BaseController.php
Tasiobg: ctype_digit check for numbers inside a string while is_int will fail if the variable is a string

tests/Twig/HtmlUtilitiesExtensionTest.php
Tasiobg: I think builders make sense for complex objects or for reusability, not sure if for this case a builder would be best but I'm intersted in Ben other people point of view.

src/Twig/HtmlUtilitiesExtension.php
BjornTwachtmann: It generates the context links around the main H1 (look at the "series 10" bit at http://www.bbc.co.uk/programmes/b08n5s4l/episodes/player)

I think this function is trying to do a bit too much. It's also missing some of the RDFa stuff from V2. Would this be better as a small presenter or something? Not quite sure how to structure it, but this seems a little confusing maybe?

BjornTwachtmann:
In general this looks very solid. Not entirely sure about the context link bit though.

src/Controller/BaseController.php
BPScott: Yep ^^ query params are always strings initially. I don't want to cast it to an int immediatly because things like "+10e2", "2.4", "9fff1" are considered validly castable (parsing 1000, 2 and 9 respectively) when I only want plain digits.

tests/Twig/HtmlUtilitiesExtensionTest.php
BPScott: I need the pid, title and ancestry to be fixed so I can assert on them, so I need to explicitly state those, thus this function would change from:

```php
$programme = $this->createMock(CoreEntity::class);
$programme->method('getPid')->willReturn(new Pid($pid));
$programme->method('getTitle')->willReturn($title);
$programme->method('getAncestry')->willReturn(array_merge([$programme], $parents));

return $programme;
```

to

```php
return $builder = SeriesBuilder::any()->with([
    'pid' => new Pid($pid),
    'title' => $title,
    'parent' => $parent,
])->build();
```

which seems no more or less complicated than using boring old php unit mocking (slight pro: you can mock just the parent rather than the whole ancestry, slight con: indirection if you're not sure what the deal with builders are)

BPScott:
Other notes: v2 allowed the user to customise the number of items on the page with a `limit` parameter however it was never exposed by the application so I bet nobody every discovered that. It was easter egg functionality which I don't think is worth keeping and having to maintain and test.

BPScott:
Rejigged the entity context helper and have moved it into a dsShared utility presenter. It looks pretty reasonable and has a lot less annoying boilerplate than i feared.

I've done a bit of looking at RDFa stuff, and I don't think providing the RDFa tags for the context of an aggregation page provides any actual value. This is because there doesn't seem to be a way to delimit between "here are the episodes in the list" and "here is the thing that everything in this list belongs to". In most cases that belonging to bit is already covered by 'part of' references within the episode object itself anyway.

src/Controller/BaseController.php
votemike: My thoughts are.... If someone is putting +10e2 into the address bar, they can have whichever page my application feels like serving to them after casting to int. But up to you. Just wondered why this wasn't just being cast to an int.

tests/Twig/HtmlUtilitiesExtensionTest.php
votemike: I was under the impression we should always use builders from now on....
Do we have a rough rule on when to use builders and when not to?

PR #376, opened 2018-01-17
PROGRAMMES-6130 Add Radio Favourites Button

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6130

Will write tests soon as long as no one completely disagrees with the structure of the PR.


programmes-developers:

config/services.yaml
```diff
@@ -22,6 +22,8 @@ parameters:
     app.ada.class_service.class: App\ExternalApi\Ada\Service\AdaClassService
     app.electron.baseurl: 'https://api.live.bbc.co.uk/electron'
     app.electron.service.class: App\ExternalApi\Electron\Service\ElectronService
+    app.favourites.url: 'https://www.bbc.co.uk/modules/plugin/musicfavourite/init'
```
votemike: Do we want to point to LIVE by default?

src/Controller/FindByPid/TlecController.php
```diff
@@ -134,6 +136,11 @@ public function __invoke(
 
         $this->setIstatsLabelsForTlec($onDemandEpisode, $upcomingBroadcast, $lastOn);
 
+        $favouritesButton = null;
+        if ($programme->isRadio()) {
```
votemike: I'm only checking if the programme is a radio programme here but V2 checks all this: https://github.com/bbc/programmes/blob/develop/webapp/php/sharedmodules/programmes/views/helpers/HasFavouritesButton.php#L15
Do we really need all that? Or is this sufficient?

src/DsAmen/Presenters/Section/Map/SubPresenter/programme_info.html.twig
```diff
@@ -10,5 +10,6 @@
                 {{ ds_amen('synopsis', programme_info.getProgramme().getSynopses(), 250) }}
             </div>
         {% endif %}
+        <div class="favourites-button buttons--favourites-small"></div>
```
votemike: Acceptable class name?

src/DsAmen/Presenters/Section/Map/_map.scss
```diff
@@ -79,6 +79,17 @@ $map-gap: 2px;
     margin-right: 11px; // due to weird paddings etc
 }
 
+.buttons--favourites-small {
```
votemike: OK for these 2 new styles to live here?

config/services_prod_int.yaml
```diff
@@ -2,4 +2,5 @@ parameters:
     app.feed_env.orbit: 'test'
     app.feed_env.branding: 'live'
     app.electron.baseurl: 'https://api.test.bbc.co.uk/electron'
+    app.favourites.url: 'https://www.test.bbc.co.uk/modules/plugin/musicfavourite/init'
```
BjornTwachtmann: Is there an int version? Does it matter what env the HTML comes from (i.e. will this work on int with the HTML from test?)

src/Controller/FindByPid/TlecController.php
BjornTwachtmann: This will do I think

src/DsAmen/Presenters/Section/Map/_map.scss
BjornTwachtmann: The favourites button is used elsewhere, might be an idea to put it somewhere more generic. Maybe a file under Css\Components

src/ExternalApi/FavouritesButton/FavouritesButton.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace App\ExternalApi\FavouritesButton;
```
BjornTwachtmann: If you look at the other ExternalApi bits they have a folder structure for Domain , Service etc. Would be good to keep that consistent

config/services.yaml
felipeparaujo: If it works, I think it's fine to point to live

src/ExternalApi/FavouritesButton/FavouritesButtonClient.php
```diff
@@ -0,0 +1,47 @@
...
+declare(strict_types = 1);
+
+namespace App\ExternalApi\FavouritesButton;
+
+use App\ExternalApi\Client\HttpApiClientFactory;
+use App\ExternalApi\Exception\ParseException;
+use Closure;
+use GuzzleHttp\Psr7\Response;
+
+class FavouritesButtonClient
```
BjornTwachtmann: The other external API fetchy things are called SomethingService

templates/find_by_pid/tlec.html.twig
```diff
@@ -11,6 +11,8 @@
     {% if programme.getNetwork() and programme.getNetwork().isInternational() %}
         {% do preload(asset('js/timezone-notification.js'), { as: 'script', nopush: true }) %}
     {% endif %}
+    {{ favouritesButton.getHead()|raw }}
+    <script>{{ favouritesButton.getScript()|raw }}</script>
```
BjornTwachtmann: Should this go in the head or the foot? Foot seems less likely to block things, but I don't know what they put in the JS.

config/services_prod_int.yaml
votemike: There is an int version. But orbit and electron are pointing at test and branding is pointing at live in this file.... so not sure what the right thing to do is....
The different envs result in different URLs in the retrieved HTML.

src/ExternalApi/FavouritesButton/FavouritesButtonClient.php
votemike: I was basing this off of Orbit, which calls it Client. I can change it to service if you like.

templates/find_by_pid/tlec.html.twig
votemike: Could probably go in either.

src/Controller/FindByPid/TlecController.php
felipeparaujo: ` $service->product_group->has_favourites` this worries me a bit though... How is this field getting populated?

config/services_prod_int.yaml
BjornTwachtmann: yeah, I'm not sure test orbit on int is really ideal either. Change this one to int for now and we'll see how it goes.

src/Controller/FindByPid/TlecController.php
BjornTwachtmann: as far as I can see it's hardcoded for radio stuff here https://github.com/bbc/programmes/blob/develop/webapp/php/sharedlib/BBC/Programmes/ProductGroup.php#L33

config/services.yaml
BPScott: name it `app.favourites.baseurl` to be consistent with the naming of other ExternalApi params?

BjornTwachtmann:
LGTM if you want to merge

PR #377, opened 2018-01-17
Control whitespace of related links within the presenter

BPScott:
Remove the whitespace at the source instead of having to trim it
whenever the presenter is called

FAO @votemike / @Tasiobg This is a nicer fix that the stuff we talked about in https://github.com/bbc/programmes-frontend/pull/332#discussion_r157725529


programmes-developers:

votemike:
I don't understand how this works. There will be a 5th line, why will that not be rendered?

PR #378, opened 2018-01-18
Ensure Redirect handling is optimised

BPScott:
- Ensure 3xx redirect exceptions do not get logged by creating a custom
  activation strategy for how the FingersCrossed log handler is fired.
  This may not be needed in the future as if
  https://github.com/symfony/symfony/pull/25533 is merged we would be
  able to do this with Symfony config rather than writing custom code.
- Ensure redirect exceptions don't have any body content (and thus
  don't make any calls to Orbit or Branding)

You will still see exception notices in dev for redirects because the
errors are only swallowed if you are using the FingersCrossed log
handler, which is only the case in production.

Fixes PROGRAMMES-5835

### DIFF
--- a/config/packages/prod/monolog.yaml
+++ b/config/packages/prod/monolog.yaml
@@ -3,10 +3,8 @@ monolog:
         main:
             type: fingers_crossed
             channels: ['!app_access']
-            action_level: error
+            activation_strategy: '@App\Exception\Handler\HttpCodeActivationStrategy'
             handler: nested
-            excluded_404s:
-                - ^/
         app_access:
             type: stream
             channels: ['app_access']
@@ -19,3 +17,10 @@ monolog:
         console:
             type: console
             process_psr_3_messages: false
+
+
+services:
+    App\Exception\Handler\HttpCodeActivationStrategy:
+        arguments:
+            excludedCodes: [301, 302, 404]
+            actionLevel: error
--- a/src/Controller/ExceptionController.php
+++ b/src/Controller/ExceptionController.php
@@ -47,6 +47,12 @@ class ExceptionController extends BaseExceptionController
         $code = $exception->getStatusCode();
         $orb = $branding = null; //No need for Orb or Branding when developing locally
 
+        // If this is a Redirect exception don't return any body or show the
+        // debug page, just redirect everything
+        if ($code >= 300 && $code < 400) {
+            return new Response('', $code, $exception->getHeaders());
+        }
+
         if (!$showException) {
             try {
                 $branding = $this->brandingClient->getContent('br-00001');
@@ -76,6 +82,7 @@ class ExceptionController extends BaseExceptionController
         if (!$showException && $code >= 400 && $code <= 499) {
             $headers['Cache-Control'] = 'public, max-age=60';
         }
+
         // The 200 status here is a misnomer, it is a default and shall be
         // overridden later.
         return new Response($this->twig->render(
--- a/src/Exception/Handler/HttpCodeActivationStrategy.php
+++ b/src/Exception/Handler/HttpCodeActivationStrategy.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace App\Exception\Handler;
+
+use Monolog\Handler\FingersCrossed\ErrorLevelActivationStrategy;
+use Symfony\Component\HttpKernel\Exception\HttpException;
+use Symfony\Component\HttpFoundation\RequestStack;
+
+/**
+ * Activation strategy that ignores certain HTTP codes.
+ */
+class HttpCodeActivationStrategy extends ErrorLevelActivationStrategy
+{
+    private $excludedCodes;
+    private $requestStack;
+
+    public function __construct(RequestStack $requestStack, array $excludedCodes, $actionLevel)
+    {
+        parent::__construct($actionLevel);
+
+        $this->requestStack = $requestStack;
+        $this->excludedCodes = $excludedCodes;
+    }
+
+    public function isHandlerActivated(array $record)
+    {
+        $isActivated = parent::isHandlerActivated($record);
+
+        if ($isActivated && isset($record['context']['exception']) && $record['context']['exception'] instanceof HttpException) {
+            return !in_array($record['context']['exception']->getStatusCode(), $this->excludedCodes);
+        }
+
+        return $isActivated;
+    }
+}
--- a/src/Exception/ProgrammeOptionsRedirectHttpException.php
+++ b/src/Exception/ProgrammeOptionsRedirectHttpException.php
@@ -6,7 +6,10 @@ use Symfony\Component\HttpKernel\Exception\HttpException;
 
 /**
  * Symfony provides several exceptions that result in the Framework emiting a
- * response, but it doesn't provide one for 301/302 redirects.
+ * response, but it doesn't provide one for 301/302 redirects - which is
+ * reasonable as they're not exceptions.
+ * However for us it is useful to eject out of code to trigger a redirect in a
+ * few places, notably when triggering programme options redirects.
  *
  * This Exception takes a location and status code, and shall send a redirect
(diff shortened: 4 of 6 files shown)


programmes-developers:

BjornTwachtmann:
Jesus, this is a getting a bit too magic for me. I don't like using exceptions to signal status codes like this. 

Let's have a chat when you're online, see if we can think of another approach.

Tasiobg:
Code looks good if we want to redirect based on exceptions but personally I found it a bit hacky.
What people thing about other ways of doing the redirection? I was thinking of replacing `throw new ProgrammeOptionsRedirectHttpException` for something like this:
```
$event->setResponse(new Response('', 303, ['location' => 'cbbc/shows/arthur']));
```
Or something more traditional
```
header('Cache-Control: public, max-age=60') ;
header('Location: http://sandbox.bbc.co.uk/cbbc/shows/arthur') ;
exit;
```

Those examples are working fine on my sandbox but are just quick test to see if the idea works.

BPScott:
I'm also not a full-on fan of Exception handling for redirect wrangling, its not really designed for redirects but the big headache is we currently have to deal with Programme Options redirects at two locations in the code (one of which doesn't support returning a Response to cancel out of the event loop), and  aborting the request lifecycle using  `HttpException` felt somewhat familar as we do that in other cases, e.g. `throw $this->createNotFoundException()` (which creates a `NotFoundHttpException`). I came up with three potential solutions to this, and I think this PR results in the least additional complexity / spagetti.

We currently have to check for the options redirect in two places:

* In `EventSubscriber\FindByPidRouterSubscriber` which is used when we hit the main find by pid route - `/programmes/:pid`. This resolves the entity that relates to `:pid` by making one to three DB calls (look for a CoreEntity, if nothing found look for a version, if nothing found look for a segment), and sets the controller action of the find_by_pid route to the desired controller, and stores the retrieved entity so that it can be used by the appropriate controller without requiring an additional DB query later. If the found CoreEntity contains a redirect in its options then we want to trigger the redirect. If no entity is found we throw a `NotFoundHttpException` to trigger a 404. You can read more about this in ADR-1 in the docs/arch folder.
* In `ArgumentResolver\ContextEntityByPidValueResolver` which is used when we hit any other route that contains a `:pid` parameter and contains an argument that is a CoreEntity or subclass thereof, or a Service. This looks at what the typehint of the controller argument is, and makes a DB call to the appropirate service, including limiting to the appropriate type (e.g. if you typehint ProgrammeItem, then it will do the lookup limiting it to episodes and clips). If the found entity contains a redirect in its options then we want to trigger the redirect. If no entity is found we throw a `NotFoundHttpException` to trigger a 404. Note that because of how we set the retrieved entity in `FindByPidRouterSubscriber` this ArgumentResolver does not trigger for `/programmes/:pid` and thus we avoid a duplicate query. You can read more about this in ADR-2 in the docs/arch folder.

Currently these two systems work well in isolation with no coupling between them.

`FindByPidRouterSubscriber` triggers at a point in the event lifecycle where we can do `$event->setResponse(new Response('', $redirectStatus, ['location' => $redirectLocation]));` which will result Symfony returning that response, rather than carrying on with the rest of the "what controller am i, lets call that etc" lifecycle and everything is fine and dandy.

`ContextEntityByPidValueResolver` offers no such escape hatch. Throwing exceptions is the only way to signal something other than "return an entity that will be passed into the controller".

Setting some explict headers and `die()`ing is always an option but that feels like giving up with an obscene hand gesture to the framework. That means we never do things like call our stats reporting code (as kernel.terminate events never fire if we `die()` immediatly). I think this is even more hacky than exception fiddling.


At about this point you're probably starting to ponder "Well can we rewrite these two things so that all the DB lookup stuff happens in an EventSubscriber where we can bail by doing `$event->setResponse()`?"

The answer is yes, kinda but I think it ends up being more complex than what we have now.

We could do something like the following:

* Change `FindByPidRouterSubscriber` so that it triggers on all routes with `:pid` as a parameter. It will:
    * Look up the Pid in the CoreEntity, then Version then Segment (like it does now)
    * If a CoreEntity with options is found, trigger a redirect with `$event->setResponse()`
    * If no entity is found throw a `NotFoundHttpException` (like it does now)
    * Store found entity within the request's attributes
* Change `ContextEntityByPidValueResolver` so that it will:
    * Check for the existence of an entity stored on the request's attributes
    * If one is found, and the type does not match against the type defined of the controller argument throw a `NotFoundHttpException` - because we want to handle the "/programmes/:pid/episodes only accepts ProgrammeContainers" case
    * Yield the entity

This helps to isolate the redirect logic to a single place where we can handle it without custom exceptions (we'd still want to throw NotFoundHttpExceptions), but it does mean the EventSubscriber and ArgumentResolver's functionality are intertwined which I find unclear. There is also an extra bit of complexity required as this naive solution results in checking entity types we know we don't care about (e.g. /programmes/:pid/episodes will try and lookup a version pid if no CoreEntity is found). This can be circumvented by adding hints to routes regarding what type of entities they should look up (CoreEntity would be the default, the find_by_pid and the schedule routes would require a hint), but I think this is kludgy and non-obvious compared to "specify a typehint in the controller and it just works" which is in place currently. It rapidly builds to the point where this is more weird and complex than the Exception-based approach.

BPScott:
If https://github.com/symfony/symfony/pull/25533 gets merged then all of this PR except the extra lines in ExceptionController can be replaced with two lines in the monolog config once we upgrade to Symfony 4.1.

Tasiobg:
@BPScott Thanks for the detailed answer :ok_hand:

I agree die() isn't a graceful way to end a request and I'm not masively convinced but I couldn't think of any negative impact because at the point when we need to do the redirect I belive there is nothing else the frameworks needs to do that could have a bad impact in the current frontend (I can be wrong, I haven't researched too much).

The way ArgumentResolvers and EventSubscribers works makes the code a bit hard to understand because of the magic behind and adding at the top of this a mechanims to do redirections within exceptions makes things even more magic and feels like at some point this can be complex for us to maintain but I understand why you did things this way, so I'm not agains it, just a bit worried if the amount of magic keep growing.

I think you can go ahead with this change unless someone else could suggest any improvements

BPScott:
> I couldn't think of any negative impact [of calling die()] because at the point when we need to do the redirect I belive there is nothing else the frameworks needs to do that could have a bad impact in the current frontend

The stats handling that gets called in kernel.terminate event would never get called, thus we'd loose stats about page requests - the number of pages would be slightly underreported :)

> Magic

I appreciate that the use of ArgumentResolvers and EventSubscribers reduce the explictness of the code compared to a good old procedural list of steps :) They are non-obvious extension points if you aren't highly familar with Symfony however I spent a fair amount of time investigating alternatives and these approaches were the best I could come up with. The alternative approaches require tradeoffs that I do not think are worth it. (Alternatives to FindByPidRouterSubscriber require either using Subrequests - which have proven to be confusing and complex in the past in v2 - or a loss of granularity in stats reporting as we stop being able to tell the difference between the various FindByPid page types. Alternatives to ContextEntityByPidValueResolver require boilerplate logic in every page that has a pid context - which is about 2/3rds of all our routes).

I agree that too many of these layers make our code harder to understand and that we should be prudent by avoiding adding them unless required.  However I have strived to keep as close to the ideas that Symfony already uses rather than inventing our own. I'm not convinced by calling these various approaches 'magic' - they're using the built in tools that Symfony provides.

In this case Symfony already regularly uses Exceptions for returning a 404 (or other 4xx status) from any arbitary point, and we use that system ofent. Using that but to handle redirects is conceptually exactly the same except the status code changes, it doesn't feel like a radical departure to handling things that aren't a 404.

---

If leveraging the HttpException stuff for redirects still feels weird, then we could create our own separate exception that doesn't use the ExceptionController stuff, as per https://www.trisoft.ro/blog/56-symfony-redirecting-from-outside-the-controller. This was the third approach I considered

programmes-developers:

PR #381, opened 2018-01-18
Make HTTP requests Parallel

BjornTwachtmann:
This does a few things:
* Modify all ExternalApi services to have an async method that returns a promise. Said promise will do all necessary postprocessing on any API calls when wait() is called and return correct domain objects or null results.
* Modify BaseController to optionally pre-request branding when setting context. Request branding in parallel together with all other API requests (see changes in branding client)
* Modify TLEC controller to parallelise all HTTP requests
* Modify other controllers in line with new behaviours
* Make "renderWithCrome" render with chrome. Make a new "renderWithoutChrome" to render without chrome


programmes-developers:

src/Controller/FindByPid/TlecController.php
```diff
@@ -136,25 +135,32 @@ public function __invoke(
...
+        ];
+
+        $parameterPromises = [
+            'recommendations' => $recommendationsPromise,
+            'relatedTopics' => $relatedTopicsPromise,
+            'supportingContentItems' => $electronService->fetchSupportingContentItemsForProgrammeAsync($programme),
+            'favouritesButton' => $favouritesButtonPromise,
+        ];
+
+        $parameters = array_merge($parameters, $this->resolvePromises($parameterPromises));
```
BjornTwachtmann: What do people think of this way of resolving promises? Does it make sense to you?

There's no point creating the promises before making various SQL queries by the way. HTTP requests are only started once wait is called on our promises.

src/Controller/SchedulesByDayController.php
```diff
@@ -101,6 +101,9 @@ public function __invoke(
 
         $this->setIstatsExtraLabels($this->getIstatsExtraLabels($date, $broadcastDay->start()->isYesterday()));
 
+        if ($this->request()->query->has('no_chrome')) {
+            return $this->renderWithoutChrome('schedules/by_day.html.twig', $viewData);
```
BPScott: I initially thought if we aren't rendering chrome then do we need branding, but then i realised we still need it to work out what language to display.

src/Controller/FindByPid/TlecController.php
BPScott: Looks like it makes sense to me. The special case in resolvePromises around the branding item feels a bit odd but I'm not familiar enough with promises to instantly come up with a better way deal with that off the top of my head.

At some point recommendations is going to be pushed into the BaseController as it will be required for all pages with a programme footer. I guess that request would need to do a similar thing as branding but that can happen later as long as we've got an understood path for how to handle that.

votemike: Should promises be resolved in each controller? Or should they be passed as an extra parameter into `renderWithChrome` and be resolved from the `BaseController`?

votemike:
I agree with Ben about the Async thing. I don't see the need for 2 versions of each method. If someone wants the value straight away, they can just call `->wait()` instantly.

src/Controller/FindByPid/TlecController.php
BjornTwachtmann: @BPScott , there are other ways to deal with resolving that branding promise, but they start falling into dark magic and relying on implementation details of Guzzle's Async stuff. To me, this is clear to the reader (make all HTTP requests at once in a single method), if not exactly pretty. 

@votemike the reason these promises aren't resolved in renderWithChrome is so that the results of API calls can be used in the controller if necessary. Felipe's working on something that needs that right now as it happens.

BjornTwachtmann:
@votemike , Agree the BaseController is getting a bit big. It might be worth seeing if bits of this functionality can be put into separate helper classes with a single job or something. But I think that's one for a another PR.

src/Controller/FindByPid/TlecController.php
votemike: Well, if something was needed in the controller, you could always resolve the current promises and pass them in as parameters?
Just a thought.

PR #382, opened 2018-01-19
PROGRAMMES-5747 - Lx Promo

felipeparaujo:


programmes-developers:

src/ExternalApi/Morph/Service/LxPromoService.php
```diff
@@ -0,0 +1,88 @@
...
+use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeContainer;
+use BBC\ProgrammesPagesService\Domain\Exception\DataNotFetchedException;
+use GuzzleHttp\Promise\FulfilledPromise;
+use GuzzleHttp\Promise\PromiseInterface;
+use Psr\Log\LoggerInterface;
+use RMP\Translate\TranslateFactory;
+
+class LxPromoService
+{
+
```
votemike: WHY IS THERE A BLANK LINE!? 😡

composer.json
```diff
@@ -14,6 +14,8 @@
         { "type": "vcs", "url": "git@github.com:bbc/gel-iconography-assets.git" },
         { "type": "vcs", "url": "git@github.com:bbc/DoctrineExtensions.git" },
         { "type": "vcs", "url": "git@github.com:bbc/programmes-pages-service.git" },
+        { "type": "vcs", "url": "git@github.com:bbc/programmes-morph-library.git" },
+        { "type": "vcs", "url": "git@github.com:bbc/programmes-caching-library.git" },
```
votemike: Why is this needed? Is this a dependency of programmes-morph-library?

src/ExternalApi/Morph/Service/LxPromoService.php
```diff
@@ -0,0 +1,87 @@
...
+    private $logger;
+
+    /** @var TranslateFactory */
+    private $translateFactory;
+
+    public function __construct(
+        LoggerInterface $logger,
+        TranslateFactory $translateFactory,
+        MorphClient $client,
+        string $env
```
votemike: Do you think this env param could do with a bit more of a description?

src/ExternalApi/Morph/Service/LxPromoService.php
```diff
@@ -0,0 +1,87 @@
...
+    ) {
+        $this->env = $env;
+        $this->client = $client;
+        $this->logger = $logger;
+        $this->translateFactory = $translateFactory;
```
votemike: Where's this used?

src/Controller/FindByPid/TlecController.php
```diff
@@ -106,6 +109,35 @@ public function __invoke(
...
+
+        $promises = [
+            'recommendations' => $recommendationsPromise,
+            'relatedTopics' => $relatedTopicsPromise,
+            'supportingContentItems' => $electronService->fetchSupportingContentItemsForProgramme($programme),
+            'favouritesButton' => $favouritesButtonPromise,
+            'lxPromo' => $lxPromoPromise,
+        ];
+
+        $parameterPromises = $this->resolvePromises($promises);
```
BjornTwachtmann: I don't think calling this variable promises when it now contains data is that clear.

composer.json
felipeparaujo: it's used both by morph-library and frontend now. We need to set the repo here because the package is not available from packagist. So everything that uses caching-libray, or that uses something that uses caching-library, need to define the repo.

src/ExternalApi/Morph/Service/LxPromoService.php
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public function fetchByProgrammeContainer(ProgrammeContainer $programme): ?PromiseInterface
```
BjornTwachtmann: Can this return null? It should always return a promise really.

composer.json
votemike: Where is it specifically required in this composer file?

felipeparaujo: you're right, it's not necessary. i've removed it.

PR #383, opened 2018-01-19
PROGRAMMES-6156 - Fix for network and service with same name

felipeparaujo:


programmes-developers:

src/DsShared/Helpers/BroadcastNetworksHelper.php
```diff
@@ -124,14 +133,27 @@ public function getCollapsedBroadcastNetworkBreakdown(CollapsedBroadcast $collap
                     ];
                 }
 
-                $breakdowns[$nid]['on_services'][(string) $service->getSid()] = $service;
+                // If the service has the same name as the network, ignore it
+                if (strtoupper($service->getShortName()) !== strtoupper($service->getNetwork()->getName())) {
```
votemike: You should be able to use `strcasecmp()`, dunno if that's cleaner.

src/DsShared/Helpers/BroadcastNetworksHelper.php
```diff
@@ -70,8 +71,10 @@ public function buildServicesNames(array $networkBreakdowns): array
             $broadcastOnServices = $networkBreakdown['on_services'];
             $notBroadcastOnServices = $networkBreakdown['not_on_services'];
 
-            if (empty($notBroadcastOnServices)) {
+            if (empty($notBroadcastOnServices) || empty($broadcastOnServices)) {
```
votemike: When would `$broadcastOnServices` be empty?

felipeparaujo: When the only service that got the broadcast has the same name as the network

votemike: Our system is SO grim and convoluted 😿

PR #384, opened 2018-01-22
Cleaning data-map-column

votemike:
Adding data-map-column to last-on column.


programmes-developers:

PR #385, opened 2018-01-22
Add more reasonable HTTP timeouts

BjornTwachtmann:


programmes-developers:

config/packages/prod/csa_guzzle.yaml
```diff
@@ -8,5 +8,5 @@ csa_guzzle:
                 cert: '/etc/pki/tls/certs/client.crt'
                 ssl_key: '/etc/pki/tls/private/client.key'
                 verify: false
-                timeout: 60
+                timeout: 20
```
votemike: Is there a way we can alert someone if they try changing this to something higher than the page timeout?

BjornTwachtmann: The app doesn't really know what the page timeout is. That's implemented higher up the chain. Although, we should maybe think about having our own timeout in the app.

PR #387, opened 2018-01-24
First pass at ds2013 footer

BPScott:
Has everything but recommendations.
The Presenter and its tests are copied verbatim from the amen version, just the template content changes.

I've got some issues with recommendations that I've outlined in PROGRAMMES-6143. If people could have a think about that and so we can have a chat about it some point that'd be appreciated.


programmes-developers:

PR #389, opened 2018-01-25
Remove deprecations by removing overrides for render() and redirect()

BPScott:
These methods are marked @final in Symfony 3.4 so will trigger
deprecations if you try and override them in the BaseController. So
instead create alternative functions that shall pick up on the caching
headers.

Instead of render() use renderWithChrome() or renderWithoutChrome()
to ensure caching headers are set.

Instead of redirect() / redirectToRoute(), use cachedRedirect() and
cachedRedirectToRoute() to ensure caching headers are set.

### DIFF
--- a/src/Controller/BaseController.php
+++ b/src/Controller/BaseController.php
@@ -191,33 +191,35 @@ abstract class BaseController extends AbstractController
     }
 
     /**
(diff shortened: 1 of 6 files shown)


programmes-developers:

PR #390, opened 2018-01-25
Fix for exception thrown when Duration of Radio Episode is null

felipeparaujo:
https://programmes-frontend.int.api.bbc.co.uk/programmes/p01n4ldk 

happens in this brandpage


programmes-developers:

src/DsAmen/Presenters/Section/Map/SubPresenter/TxPresenter.php
```diff
@@ -109,7 +109,10 @@ public function getTitleTranslationString(): string
             return 'on_now';
         }
 
-        if ($this->programmeContainer->getNetwork()->isInternational() || $this->upcomingBroadcast) {
+        if (!$this->programmeContainer->getNetwork() ||
```
felipeparaujo: is this what we want to happen? Have a look at http://bbc.co.uk/programmes/p04ts87g. It shows both next on and last on, but it's a 2013 brand page. We might have to talk to UX

PR #392, opened 2018-01-29
Recommendations in ds2013

BPScott:
Lazy load recommendations into the ds2013 footer.

Simplify RecEng service signature to only accept Episodes. All the logic for picking what episode should be passed into it should be done in the RecommendationsController. (i.e. if you passed in a container, use the latest available ondemand episode, else the next on broadcast, else try the last on broadcast). Currently that logic is duplicated in the TlecController but I think we should seek to make amen pages use a lazy loaded partial for recommendations at some point in the future.

Read the commit messages for more info.

Currently the ordering of recommendations is wrong on some pages (e.g. b006v5y2), this is because calling findByPids doesn't guarantee returning results in the same order as the order of pids you specified. https://github.com/bbc/programmes-pages-service/pull/227 fixes that

### DIFF
--- a/src/Controller/FindByPid/TlecController.php
+++ b/src/Controller/FindByPid/TlecController.php
@@ -109,13 +109,14 @@ class TlecController extends BaseController
             $relatedTopicsPromise = $adaClassService->findRelatedClassesByContainer($programme, $usePerContainerValues);
         }
 
-        $recommendationsPromise = $recEngService->getRecommendations(
-            $programme,
-            $onDemandEpisode,
-            $upcomingBroadcast ? $upcomingBroadcast->getProgrammeItem() : null,
-            $lastOn ? $lastOn->getProgrammeItem() : null,
-            2
-        );
+        $recommendationEpisode = $onDemandEpisode ?? ($upcomingBroadcast ? $upcomingBroadcast->getProgrammeItem() : null) ?? ($lastOn ? $lastOn->getProgrammeItem() : null);
+        $recommendationsPromise = new FulfilledPromise([]);
+        if ($recommendationEpisode) {
+            $recommendationsPromise = $recEngService->getRecommendations(
+                $recommendationEpisode,
+                2
+            );
+        }
 
         $favouritesButtonPromise = new FulfilledPromise(null);
         if ($programme->isRadio()) {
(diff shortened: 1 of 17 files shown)


programmes-developers:

src/Controller/FindByPid/TlecController.php
```diff
@@ -109,13 +109,14 @@ public function __invoke(
...
         }
 
-        $recommendationsPromise = $recEngService->getRecommendations(
-            $programme,
-            $onDemandEpisode,
-            $upcomingBroadcast ? $upcomingBroadcast->getProgrammeItem() : null,
-            $lastOn ? $lastOn->getProgrammeItem() : null,
-            2
-        );
+        $recommendationEpisode = $onDemandEpisode ?? ($upcomingBroadcast ? $upcomingBroadcast->getProgrammeItem() : null) ?? ($lastOn ? $lastOn->getProgrammeItem() : null);
```
votemike: 😻

BPScott: I am expecting Bjorn to scowl at me for this line

BjornTwachtmann: Would have been better if you shelled out to a Perl script for the ulitmate incomprehinsible one-liner. Not a bad effort though :D

BjornTwachtmann:
This looks fine to me. I haven't actually tested it. 
I still don't think the partial should be making queries anything like that complex though. Find the last episode by on_demand_sort_date for a ProgrammeContainer and be done with it. Can't imagine anyone will notice or care.

BPScott:
I agree that simplifying the queries to just have the latest episode by on_demand_sort_date for a ProgrammeContainer would be preferable, however Cathal was reticent to change this when I raised it in PROGRAMMES-6143's comments. We've not had a video chat for a while, lets do one of them tomorrow to discuss this.


@votemike: check that ticket ^^ for a bunch of research and my suggestions on what to do with the recommendations stuff. I believe the TL;DR for a lazy-loading strategy is:

Try to avoid it when loading content within the main page is preferable (e.g. clips/galleries lists on the brand page, which used to be lazy loaded because of slow APIs but now the requests for data is fast). However in the case of the recommendations footer, the lowish chance that a use scrolls down that far, the number of queries required and need to inject them on every page results in complexity that makes lazy-loading viable for this low-impact piece of content.

PR #395, opened 2018-01-30
Fix api cert issues

BjornTwachtmann:
This fixes a couple of issues
* INT/TEST certs not valid for live ADA
* Dev cert breaks requests to non-int bbc envs for favourites button
* Favourites button needs metrics

It allows consumers of the HttpApiClient class to specify guzzle options on instantiation.


programmes-developers:

votemike:
Approved apart from the failing tests

BjornTwachtmann:
I just ran a proper build on Jenkins and the tests pass. Think it's just the PR job that's misbehaving again.

PR #396, opened 2018-01-31
PROGRAMMES-6130 Fave button on promo priority

votemike:
The radio favourite button was missing on promo-priority column.
In order to render the button, I needed to pass the ProgrammeContainer in to the PromoPriorityPresenter, which made it more similar to other columns. So I also did a refactor.

Can be tested with The Today Programme: http://sandbox.bbc.co.uk/programmes/b006qj9z?__2016minimap=false


src/DsAmen/Presenters/Section/Map/SubPresenter/programme_info.html.twig
```diff
@@ -1,16 +1,16 @@
 <div class="br-box-secondary" data-map-column="programme-info">
     {% if not programme_info.showMiniMap %}
-        {{ ds_shared('image', programme_info.getProgramme().getImage(), programme_info.getDefaultImageSize(), programme_info.getImageSizes(), {'is_lazy_loaded': false}) }}
+        {{ ds_shared('image', programme_info.getProgrammeContainer().getImage(), programme_info.getDefaultImageSize(), programme_info.getImageSizes(), {'is_lazy_loaded': false}) }}
     {% endif %}
-    <div class="island" vocab="http://schema.org/" typeof="{{ rdfa_schema_programme_typeof(programme_info.programme) }}" resource="{{ path('find_by_pid', {'pid': programme_info.programme.pid}) }}">
-        <h1 class="invisible" property="name">{{ programme_info.programme.title }}</h1>
```
votemike: The refactor brought to light all these property accessors that were using Twig magic rather than calling the get methods. Yay for refactoring.

src/DsAmen/Css/Components/_buttons.scss
```diff
@@ -8,3 +8,12 @@
 .favourites-button {
     margin-top: 18px;
 }
+
+.favourites-button-island {
```
votemike: I imagine @BPScott will hate all of this. But wasn't sure of a nicer way to do it with existing classes.

programmes-developers:

src/DsAmen/Css/Components/_buttons.scss
BPScott: I'd be tempted to avoid using @extend as much as possible. It has the potential to get [quite clunky quite quickly](https://www.sitepoint.com/avoid-sass-extend/). The island CSS is small and straightfoward, so I reckon just set the paddings that you need here rather than try to reuse/extend other utility classes.

At that point do you need a new wrapper element or can it be a modifier of .favouirites-button - `.favourites-button--blah`? It is is a modifier then we can totally sidestep the question of how do we state a hierarchy (i.e. it's not obvious that a .favourite-button must going inside a .favourites-button-island)

votemike: So.... for the programme info column, I need to push the button down 18 px from the text.
But in the promo_priority column, the text is within the promotion object. So there is no surrounding island class. So I need to make my own island class. But that has a load of top margin (maybe padding), so I don't need to push the button down as far....
Not familiar enough with our CSS enough to know which is the best way to go.

votemike: I have made a change. Want to see what you think?

PR #398, opened 2018-01-31
PROGRAMMES-6154 Speed up test on jenkins

tatitati:
WITH_TEST_COVERAGE is the new boolean parameter setup in jenkins. This allows to the user to choose whether running tests with coverage or not


programmes-developers:

script/ci-docker
```diff
@@ -33,7 +33,11 @@ php --version
 yarn gulp-ci
 
 # Run PHPUnit
-bin/phpunit || RESULT=1
+if [ $WITH_TEST_COVERAGE == true ]; then
+    bin/phpunit || RESULT=1
+else
+    bin/phpunit --no-coverage || RESULT=1
+fi
```
tatitati: Should I add some kind of ```echo 'running with code coverage'``` and ```echo 'running without code coverage'``` to make clearer the output and have some kind of "reminder" about how was ran a build?

Do we want to add also something like ```time bin/phpunit ....``` to check better how long they take

script/ci-docker
```diff
@@ -33,7 +33,11 @@ php --version
 yarn gulp-ci
 
 # Run PHPUnit
-bin/phpunit || RESULT=1
+if [ $WITH_TEST_COVERAGE == true ]; then
```
BjornTwachtmann: This will test for the string true, rather than a boolean value. Is that what you want?

script/ci-docker
BjornTwachtmann: Yeah, sounds useful

script/ci-docker
```diff
@@ -33,7 +33,13 @@ php --version
 yarn gulp-ci
 
 # Run PHPUnit
-bin/phpunit || RESULT=1
+if $WITH_TEST_COVERAGE ; then
```
BjornTwachtmann: I think this should probably read 
```if [[ ! -z "$WITH_TEST_COVERAGE" ]]```
That tests for any non-empty string

tatitati: I tested it. I guess that you were thinking that the value is set only when the checkbox is checked. To make sure of this I tested your idea, but is matching always the same condition ($WITH_TEST_COVERAGE has always a value, when checked or unchecked).

BjornTwachtmann: Fair enough. If it works.

PR #399, opened 2018-01-31
Don't use last/next broadcast episode when looking for recommendations

BPScott:
Simplify and speed up recommendations by only looking for the streamable
episodes. As far as I can tell RecEng only returns results when you give
it a streamable episode.

For example:
Doctor who has no available episodes and no recommendations - I'm sure this had recommendations a few days ago when the Christmas special was still available.
Strictly has no available episodes and no recommendations
Weakest Link has one available episode and has recommendations. When passing the available episode into RecEng there are episodes, if you pass any of its sibling episodes into RecEng you get no results.

### DIFF
--- a/src/Controller/FindByPid/TlecController.php
+++ b/src/Controller/FindByPid/TlecController.php
@@ -109,11 +109,10 @@ class TlecController extends BaseController
             $relatedTopicsPromise = $adaClassService->findRelatedClassesByContainer($programme, $usePerContainerValues);
         }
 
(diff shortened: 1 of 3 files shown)


programmes-developers:

BPScott:
I've asked Cathal to see if he can confirm this with RecEng product people. If that is confirmed then I shall add a comment into the $recEngService->getRecommendations() service stating that you must pass in an available episode to get results.

PR #402, opened 2018-02-01
Moving Schedules controllers into their own package

felipeparaujo:
It was annoying me that schedules controllers didn't have their own package, so I created one.


programmes-developers:

PR #403, opened 2018-02-01
Episode Guide WIP

BPScott:
* Add controller and make data calls
* Basic page header

Still need to add rendering of episodes

Conveniently we've already got all the data APIs for this page (PROGRAMMES-6163) thanks to this page being powered by the episode guide feed from Clifton.


programmes-developers:

BPScott:
(I'm on holiday for the next week so merge this in whenever)

src/Controller/ProgrammeEpisodes/GuideController.php
```diff
@@ -1,16 +1,47 @@
...
+        );
+
+        // If you visit an out-of-bounds page then throw a 404. Page one should
+        // always be a 200 so search engines don't drop their reference to the
+        // page if a programme has no episodes
+        if (!$episodes && $page !== 1) {
+            throw $this->createNotFoundException('Page does not exist');
+        }
+
+        $totalEpisodes = $programmesService->countEpisodeGuideChildren($programme);
```
votemike: If it's page 1 and there are no episodes, is it worth doing this query?

config/routes.yaml
```diff
@@ -21,9 +21,9 @@ programme_episodes:
     requirements: { pid: '[0-9b-df-hj-np-tv-z]{8,15}' }
 
 programme_episodes_guide:
-    path: /programmes/{pid}/episodes/guide
-    defaults: { _controller: App\Controller\ProgrammeEpisodes\GuideController }
-    requirements: { pid: '[0-9b-df-hj-np-tv-z]{8,15}' }
+    path: /programmes/{pid}/episodes/guide{extension}
+    defaults: { _controller: App\Controller\ProgrammeEpisodes\GuideController, extension: '' }
+    requirements: { pid: '[0-9b-df-hj-np-tv-z]{8,15}', extension: '$|\.inc2013' }
```
votemike: What's the extension all about?

felipeparaujo: http://www.bbc.co.uk/programmes/b006q2x0/episodes/guide take a look at this guide. If you click on any of the seasons, it loads a partial, which is the .inc2013 route

votemike: So in that case, the PID would be a PID for a season?

votemike:
Do we still need to lazy load this stuff on AWS? Would it be better to just load it on page load?

PR #408, opened 2018-02-06
PROGRAMMES-6165 - Broadcast Event Object

felipeparaujo:
This also includes quite a bit of the guide page, we just need the series collapsible object here


programmes-developers:

src/DsShared/Helpers/BroadcastNetworksHelper.php
```diff
@@ -105,6 +121,17 @@ public function buildServicesNames(array $networkBreakdowns): array
         return $serviceNames;
     }
 
+    private function buildUrlKeys(array $networkBreakdowns): array
```
Tasiobg: I'm assuming this will be used in a future, if so please ignore comment

felipeparaujo: nope, i just forgot to push the commit that removes it lol

src/Controller/ProgrammeEpisodes/GuideController.php
```diff
@@ -42,6 +49,19 @@ public function __invoke(
             }
         }
 
+        $upcomingBroadcastCount = $collapsedBroadcastService->countUpcomingByProgramme($programme, CacheInterface::MEDIUM);
+        $totalAvailableEpisodes = $programmeAggregationService->countStreamableOnDemandEpisodes($programme);
```
votemike: Are all streamable episodes not on-demand? And vice versa?

BPScott: Things you can watch/listen to online == Available on demand == Streamable.

Editorial wording tends to talk about things being "available on demand", while the field in our DB that says if something is available on iplayer is "isStreamable", which is where the two names for the same concept came from.

IMO `countStreamableOnDemandEpisodes` should just be `countStreamableEpisodes` as it's saying the same thing twice but I didn't want to grump about naming things after a PR was merged :)

votemike: I ❤️ grumping

PR #409, opened 2018-02-06
Guide episode new

tatitati:
[PROGRAMMES-6164](https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6164)

---

- **Examples to test:**
http://sandbox.bbc.co.uk/programmes/b006q2x0/episodes/guide
http://sandbox.bbc.co.uk/programmes/b08n5s4l/episodes/guide.2013inc
http://sandbox.bbc.co.uk/programmes/b00rhg2r/episodes/guide
http://sandbox.bbc.co.uk/programmes/b006m8wd/episodes/guide


programmes-developers:

assets/js/episode-guide.js
```diff
@@ -0,0 +1,108 @@
...
+        var link = findCollapsableControlSeries(series),
+            panel = $('#' + link.attr('aria-controls')),
+            expandedDropdownFlag = link.find(classDropdownExpanded).eq(0),
+            collapsedDropdownFlag = link.find(classDropdownCollapsed).eq(0);
+
+        // expand series
+        series.attr('class', series.attr('data-activeclass'));
+
+        collapsedDropdownFlag.toggle();
+        expandedDropdownFlag.toggle();
```
tatitati: This couple of ```toggles()``` work to display the collapse symbol ```>``` in the proper direction (```/\``` or ```>```). Before we were managing this with css, just changing a class which apply a ```transform:rotate(90deg)```. However now we add this icons with gelicons (SVG html tags), and add, remove classes seems not so easy for old jquery versions:
https://stackoverflow.com/questions/8638621/jquery-svg-why-cant-i-addclass

assets/js/episode-guide.js
```diff
@@ -0,0 +1,108 @@
+define(['jquery-1.9'], function ($) {
```
tatitati: IMPORTANT NOTE: should I include istat already on this PR?

src/Ds2013/Presenters/Pages/EpisodesGuide/Collapsable/CollapsablePresenter.php
```diff
@@ -0,0 +1,69 @@
+<?php
+
+namespace App\Ds2013\Presenters\Pages\EpisodesGuide\Collapsable;
+
+use App\Ds2013\Presenter;
+use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeContainer;
+
+class CollapsablePresenter extends Presenter
```
tatitati: QUESTION: 
not sure if we work in this way. I see that we don't create presenteres only for domain entities. Sometimes for Sub elements in our page. Is this ok?, should be better to use a twig partial?

templates/programme_episodes/guide_partials/listguide.html.twig
```diff
@@ -0,0 +1,26 @@
...
+                    {% else %}
+                        {{ ds2013('collapsable', item, {}) }}
+                    {% endif %}
+                </li>
+            {% endfor %}
+        </ol>
+    {% else %}
+        <p class="note">{{ tr('episodes_none', 0)}}</p>
+    {% endif %}
+</div>
```
tatitati: This partial allow to display Episodes and series (collapsables). If a nested series is found, this template is injected again, allowing to load more nested series and episodes again.

assets/js/rv-bootstrap.js
```diff
@@ -1,4 +1,4 @@
-define(['jquery', 'lazyload', 'istats-tracking', 'respimg', 'lazysizes'],function($, Lazyload, IstatsTracking){
+define(['jquery', 'lazyload', 'istats-tracking', 'respimg', 'lazysizes', 'episode-guide'],function($, Lazyload, IstatsTracking){
```
votemike: Do we need this here? This will load the episode-guide JS on every page won't it?

src/Controller/ProgrammeEpisodes/GuideController.php
```diff
@@ -18,6 +18,10 @@ public function __invoke(
         $page = $this->getPage();
         $limit = 30;
 
+        /**
+         * @var mixed[] $children
```
votemike: Should series and episodes implement a common interface so that it can be type hinted on?

src/Ds2013/Presenters/Pages/EpisodesGuide/Collapsable/CollapsablePresenter.php
```diff
@@ -0,0 +1,69 @@
...
+    {
+        return $this->getParentContainer()->getTitle();
+    }
+
+    /**
+     * get MAIN PAGE specific series
+     */
+    public function getSeriesUrl() :string
+    {
+        return '/programmes/' . $this->getParentContainer()->getPid();
```
votemike: We shouldn't hard-code routes like this should we. We should use the router and a named route.

src/Ds2013/Presenters/Pages/EpisodesGuide/Collapsable/collapsable.html.twig
```diff
@@ -0,0 +1,40 @@
+<div typeof="Season"
+     resource="/programmes/{{ collapsable.getPid() }}"
```
votemike: Use the router. Don't hardcode routes.

assets/js/episode-guide.js
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    /**
+     * "link series" work to load and display the collapsed content (episodes) of a series.
+     */
+    function loadEpisodesFromSelectedSeries(series, redirectIfFail) {
+        var seriesLink = findCollapsableControlSeries(series),
+            href = seriesLink.attr('href');
+
+        series.addClass('js-inprogress');
```
votemike: Does this need removing at some point?

src/Controller/ProgrammeEpisodes/GuideController.php
tatitati: interesting....

felipeparaujo: A new interface seems unecessary. It would complicate an even more complicated data layer. Just use CoreEntity here, should be enough

votemike: Not `Programme`?

felipeparaujo: Yes, `Programme`. Not `CoreEntity`

src/Ds2013/Presenters/Domain/EpisodeGuideList/episode_guide_list.html.twig
```diff
@@ -0,0 +1,74 @@
...
+                     class="js-guideitem highlight-box--list br-keyline br-blocklink-page br-page-linkhover-onbg015--hover"
+                     data-inactiveclass="js-guideitem highlight-box--list br-keyline br-blocklink-page br-page-linkhover-onbg015--hover"
+                     data-activeclass="js-guideitem br-box-subtle island js-active"
+                     data-fetched="false"
+                     data-incpath="{{ path('programme_episodes_guide_partial', {'pid': programme.getPid()}) }}">
+
+                    <span class="loading-spinner pull--right-spaced hidden js-inprogress-showme"></span>
+
+                    {#Title of this season#}
+                    <h2 class="gamma no-margin">
```
felipeparaujo: These headings here should be parametrized, as well as the ones we pass to the episode object. They should increase according to the nesting level of the series/episode.

src/Ds2013/Presenters/Domain/EpisodeGuideList/episode_guide_list.html.twig
```diff
@@ -0,0 +1,74 @@
...
+            {%- else -%}
+                {#EPISODE#}
+                <div class="js-guideitem highlight-box--list block-link block-link--steal br-keyline br-blocklink-page br-page-linkhover-onbg015--hover">
+                    <div class="grid-wrapper">
+                        <div class="grid 2/3@bpw2 2/3@bpe">
+                            {{ ds2013('programme', programme, {
+                                'context_programme': episode_guide_list.getContextProgramme(),
+                                title_options: {
+                                    'title_format': 'item::ancestry',
+                                    'title_tag': 'h3',
```
felipeparaujo: read comment up there ^

src/Ds2013/Presenters/Domain/EpisodeGuideList/episode_guide_list.html.twig
```diff
@@ -0,0 +1,74 @@
...
+                                    'title_classes': 'gamma'
+                                }
+                            }) }}
+                        </div>
+                        <div class="grid 1/3@bpw2 1/3@bpe hidden grid-visible@bpw2 grid-visible@bpe">
+                            <div class="grid-wrapper">
+                                <div class="grid 1/2">
+                                    {%- if episode_guide_list.getUpcomingBroadcastsFromProgramme(programme.getPid()) is not null -%}
+                                        <h3 class="delta no-margin">{{ tr('next_on') }}</h3>
```
felipeparaujo: same thing as the other heading tags

tests/Controller/ProgrammeEpisodes/DomParsers/EpisodesGuideDomParser.php
```diff
@@ -0,0 +1,60 @@
+<?php
```
felipeparaujo: This whole class is a bit of an overkill...

src/Controller/ProgrammeEpisodes/GuidePartialController.php
```diff
@@ -0,0 +1,59 @@
...
+
+        return $this->renderWithChrome('programme_episodes/guide.2013inc.html.twig', [
+            'programme' => $programme,
+            'children' => $children,
+            'totalChildrenCount' => $totalChildrenCount,
+            'upcomingBroadcasts' => $upcomingBroadcasts,
+        ]);
+    }
+
+    public function getUpcomingBroadcastsIndexedByProgrammePid(
```
BjornTwachtmann: This shouldn't be duplicated. It also shouldn't be public.

src/Controller/ProgrammeEpisodes/GuidePartialController.php
```diff
@@ -0,0 +1,59 @@
...
+use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeContainer;
+use BBC\ProgrammesPagesService\Service\CollapsedBroadcastsService;
+use BBC\ProgrammesPagesService\Service\ProgrammesAggregationService;
+use BBC\ProgrammesPagesService\Service\ProgrammesService;
+
+class GuidePartialController extends BaseController
+{
+    public function __invoke(
+        ProgrammeContainer $programme,
+        PresenterFactory $presenterFactory,
```
BjornTwachtmann: Some of your parameters here are unused

src/Controller/ProgrammeEpisodes/GuidePartialController.php
```diff
@@ -0,0 +1,59 @@
...
+
+        $children = $programmesService->findEpisodeGuideChildren($programme, $limit);
+        if (!$children) {
+            throw $this->createNotFoundException('Page does not exist');
+        }
+
+        $totalChildrenCount = $programmesService->countEpisodeGuideChildren($programme);
+        $upcomingBroadcasts = $this->getUpcomingBroadcastsIndexedByProgrammePid($programme, $collapsedBroadcastService);
+
+        $this->setIstatsExtraLabels([
```
BjornTwachtmann: Please check, but I think this probably does nothing.

src/Ds2013/Presenters/Domain/EpisodeGuideList/episode_guide_list.html.twig
```diff
@@ -0,0 +1,74 @@
...
+                    {#Title of this season#}
+                    <h2 class="gamma no-margin">
+                        <a property="url"
+                           resource="{{ path('find_by_pid', {'pid': programme.getPid()}) }}"
+                           class="js-serieslink istats-seriesopen block-link link-complex"
+                           href="{{ path('programme_episodes_guide_partial', {'pid': programme.getPid()}) }}"
+                           aria-expanded="false"
+                           aria-controls="series-{{ programme.getPid() }}">
+
+                            <button class="js-dropdown js-dropdown--collapsed">{{ gelicon('core', 'next', 'gelicon--delta gelicon--trailing') }}</button>
```
BjornTwachtmann: Please remove <button> and use the relevant aria-role on the anchor tag

templates/base_ds2013.html.twig
```diff
@@ -72,9 +72,10 @@
                     'jquery-1.9': "https://static.bbci.co.uk/frameworks/jquery/0.4.1/sharedmodules/jquery-1.9.1",
                     'rv-bootstrap': "{{ asset_js('js/rv-bootstrap.js') }}",
                     'timezone-notification': "{{ asset_js('js/timezone-notification.js') }}",
-                    'comscorews': "{{ asset_js('js/comscorews.js') }}"
+                    'comscorews': "{{ asset_js('js/comscorews.js') }}",
+                    'episode-guide': "{{ asset_js('js/episode-guide.js') }}"
                 },
-                "priority": ['jquery-1.9', 'rv-bootstrap', 'timezone-notification'],
+                "priority": ['jquery-1.9', 'rv-bootstrap', 'timezone-notification', 'episode-guide'],
```
BjornTwachtmann: Why is this here?

assets/js/episode-guide.js
```diff
@@ -0,0 +1,179 @@
...
+        /**
+         * One series can be collapsed (icon: >) or expanded (icon: V)
+         *
+         * @Example 'V SEASON 1'
+         */
+        var showExpandedIcon = function(seriesLinkClicked)
+        {
+            var classesExpanded = seriesLinkClicked.find('[aria-role="button"]').eq(0).attr('data-expandedclass');
+            var gelicon = seriesLinkClicked.find(geliconClass).eq(0);
+            gelicon.attr('class', classesExpanded);
```
tatitati: How to add/remove class to svg elements with jquery.
https://martinwolf.org/blog/2014/12/adding-and-removing-classes-from-svg-elements-with-jquery

src/Controller/ProgrammeEpisodes/GuidePartialController.php
```diff
@@ -0,0 +1,43 @@
...
+declare(strict_types = 1);
+namespace App\Controller\ProgrammeEpisodes;
+
+use App\Ds2013\PresenterFactory;
+use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeContainer;
+use BBC\ProgrammesPagesService\Service\CollapsedBroadcastsService;
+use BBC\ProgrammesPagesService\Service\ProgrammesAggregationService;
+use BBC\ProgrammesPagesService\Service\ProgrammesService;
+
+class GuidePartialController extends GuideController
```
tatitati: To avoid some duplication, this is extending the controller above

src/Ds2013/Presenters/Domain/EpisodeGuideList/EpisodeGuideListPresenter.php
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    public function getQueryParamForNextNestedLevel() :string
+    {
+        return "nestedlevel=" . ($this->nestedLevel + 1);
+    }
+
+    public function getHeadingTag(): string
+    {
+        return 'H' . $this->nestedLevel;
```
BjornTwachtmann: Shouldn't be a capital letter

src/Ds2013/Presenters/Domain/EpisodeGuideList/episode_guide_list.html.twig
```diff
@@ -0,0 +1,76 @@
...
+            {%- else -%}
+                {#EPISODE#}
+                <div class="js-guideitem highlight-box--list block-link block-link--steal br-keyline br-blocklink-page br-page-linkhover-onbg015--hover">
+                    <div class="grid-wrapper">
+                        <div class="grid 2/3@bpw2 2/3@bpe">
+                            {{ ds2013('programme', programme, {
+                                'context_programme': episode_guide_list.getContextProgramme(),
+                                title_options: {
+                                    'title_format': 'item::ancestry',
+                                    'title_tag': episode_guide_list.getHeadingTag(),
```
BjornTwachtmann: Do you want the programme object to have the same heading level as the parent?

src/Ds2013/Presenters/Domain/EpisodeGuideList/episode_guide_list.html.twig
```diff
@@ -0,0 +1,76 @@
...
+                    <{{ episode_guide_list.getHeadingTag() }} class="gamma no-margin">
+                        <a property="url"
+                           resource="{{ path('find_by_pid', {'pid': programme.getPid()}) }}"
+                           class="js-serieslink istats-seriesopen block-link link-complex"
+                           href="{{ path('programme_episodes_guide_partial', {'pid': programme.getPid()}) }}"
+                           aria-expanded="false"
+                           aria-controls="series-{{ programme.getPid() }}">
+
+                            <span aria-role="button"
+                                  data-expandedclass="gelicon gelicon--delta gelicon--trailing ml__button--dropdown ml__button--dropdown-expanded"
```
BjornTwachtmann: The amount of data attributes and classes in them feels like overkill. 
Can't you just make the JS add/remove/toggle a single class here, rather than listing all classes 3 times?

src/Ds2013/Presenters/Domain/EpisodeGuideList/episode_guide_list.html.twig
tatitati: well, actually is not the case:
I would have something like:
```
<h2>Series
        <h3>Series
        <h3>Episode
        <h3>Series
<h2>Episode
......
```
So everything in the same leven keep the same header

src/Ds2013/Presenters/Domain/EpisodeGuideList/episode_guide_list.html.twig
tatitati: https://martinwolf.org/blog/2014/12/adding-and-removing-classes-from-svg-elements-with-jquery

src/Controller/ProgrammeEpisodes/GuidePartialController.php
```diff
@@ -0,0 +1,43 @@
...
+{
+    public function __invoke(
+        ProgrammeContainer $programme,
+        PresenterFactory $presenterFactory,
+        ProgrammesService $programmesService,
+        ProgrammesAggregationService $programmeAggregationService,
+        CollapsedBroadcastsService $collapsedBroadcastService
+    ) {
+
+        $nestedLevel = $this->request()->query->get('nestedlevel');
```
votemike: The request is passed through isn't it. Don't we usually get the parameters from that version of the request?

src/Ds2013/Css/Components/_dropdown.scss
```diff
@@ -0,0 +1,14 @@
+.ml__button--dropdown {
+    margin-left: -$basic-spacing;
+    top: $basic-half-spacing / 2;
+}
+
+.ml__button--dropdown-expanded {
+    transform: rotate(90deg);
+    transition: transform 0.25s linear;
```
votemike: Would the transition be better on `.ml__button--dropdown`?

src/Ds2013/Presenters/Domain/EpisodeGuideList/EpisodeGuideListPresenter.php
```diff
@@ -0,0 +1,78 @@
...
+    {
+        return $this->programmes;
+    }
+
+    public function getContextProgramme(): ProgrammeContainer
+    {
+        return $this->contextProgramme;
+    }
+
+    public function getUpcomingBroadcastsFromProgramme(Pid $pid)
```
votemike: What does this return?

src/Ds2013/Presenters/Domain/EpisodeGuideList/EpisodeGuideListPresenter.php
```diff
@@ -0,0 +1,78 @@
...
+use BBC\ProgrammesPagesService\Domain\Entity\Programme;
+use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeContainer;
+use BBC\ProgrammesPagesService\Domain\ValueObject\Pid;
+
+class EpisodeGuideListPresenter extends Presenter
+{
+    /** @var ProgrammeContainer */
+    private $contextProgramme;
+
+    /** @var Programme[] */
```
votemike: These can be *any* type of Programme? Brand, Clip etc...?

src/Ds2013/Presenters/Domain/EpisodeGuideList/episode_guide_list.html.twig
```diff
@@ -0,0 +1,76 @@
...
+                     data-inactiveclass="js-guideitem highlight-box--list br-keyline br-blocklink-page br-page-linkhover-onbg015--hover"
+                     data-activeclass="js-guideitem br-box-subtle island js-active"
+                     data-fetched="false"
+                     data-incpath="{{ path('programme_episodes_guide_partial', {'pid': programme.getPid() }) }}?{{ episode_guide_list.getQueryParamForNextNestedLevel() }}">
+
+                    <span class="loading-spinner pull--right-spaced hidden js-inprogress-showme"></span>
+
+                    {#Title of this season#}
+                    <{{ episode_guide_list.getHeadingTag() }} class="gamma no-margin">
+                        <a property="url"
```
votemike: This is just a toggle isn't it? If so, it should be a button, not a link.

templates/programme_episodes/guide.2013inc.html.twig
```diff
@@ -1,3 +1,14 @@
...
+{% block body %}
+<h{{ nestedlevel }} class="visually-hidden">Episodes</h{{ nestedlevel }}>
+
+<div>
+    {{ ds2013('episodeGuideList', programme, children, upcomingBroadcasts, nestedlevel) }}
 </div>
+
+{%- if totalChildrenCount > 30 -%}
+    <p class="no-margin">
+        <a class="text--center block-link delta island--squashed br-linkinvert" href="{{ path('programme_episodes_guide', {'pid': programme.getPid()}) }}?page=2">More episodes</a>
```
votemike: Will it never be page 3?

assets/js/episode-guide.js
```diff
@@ -0,0 +1,178 @@
...
+        };
+
+        var isSeriesContentAlreadyLoaded = function(series)
+        {
+            return series.attr('data-fetched') == 'true';
+        };
+
+        var isSeriesContentLoadingNow = function(series)
+        {
+            return series.hasClass('js-inprogress');
```
votemike: Why is the Javascript logic based on classes being added and removed from the DOM? Why is the status not stored as part of an object in the JS?

assets/js/episode-guide.js
```diff
@@ -0,0 +1,178 @@
...
+        };
+
+        var openSeriesIfNeeded = function()
```
votemike: This function name doesn't quite make sense.

assets/js/episode-guide.js
```diff
@@ -0,0 +1,178 @@
...
+        {
+            // special requirement: in case that the first element is a
+            // series, we open it automatically
+            var firstItem = $(classAnyItemInGuide).eq(0);
+            if (firstItem.length > 0 && firstItem.is('[data-incpath]')) {
+                firstItem.find(classDropdownLink).eq(0).trigger('click');
```
votemike: Why are you triggering clicks? Why not just call the method that is triggered by the click?

assets/js/episode-guide.js
```diff
@@ -0,0 +1,178 @@
...
+        };
+
+        /**
+         * One series can be collapsed (icon: >) or expanded (icon: V)
+         *
+         * @Example 'V SEASON 1'
+         */
+        var showExpandedIcon = function(seriesLinkClicked)
+        {
+            var classesExpanded = seriesLinkClicked.find('[aria-role="button"]').eq(0).attr('data-expandedclass');
```
votemike: Why are you finding things based on a[aria-role="button"]a? `aria-role` is for accessibility, **not** for selecting elements.

assets/js/episode-guide.js
```diff
@@ -0,0 +1,178 @@
...
+        };
+
+        /**
+         * We want to show the content of a selected series, so initially we have to load
+         * the content of it to later display it. In case that the content is already loaded,
+         * we go through the happy way (just display it)
+         */
+        var expandSeriesSelected = function(series) {
+            var seriesLinkClicked = series.find(classDropdownLink),
+                panel = $('#' + seriesLinkClicked.attr('aria-controls'));
```
votemike: There should not be any mention of aria in this file.

assets/js/episode-guide.js
```diff
@@ -0,0 +1,178 @@
...
+        /**
+         * Load content (list of episodes or nested series) for the selected series.
+         */
+        var loadEpisodesFromSelectedSeries = function(series, shouldRedirectIfFail) {
+            var url=series.attr('data-incpath');
+
+            $.ajax({
+                url: url,
+                beforeSend: function() {
+                    series.addClass('js-inprogress');
```
votemike: Why set a class? Just store the fact the the ajax is in progress in the JS.

src/Ds2013/Presenters/Domain/EpisodeGuideList/EpisodeGuideListPresenter.php
tatitati: They can be Episodes and Series only.

templates/programme_episodes/guide.2013inc.html.twig
tatitati: Yes, explained with a use case:
We have a brand with 120 series inside. Because this is too much to display in once we display only 30 and a paginator.

Now other case (your question):
One of these series has 200 episodes inside. In this case we display only 30, and after a link at the bottom displaying "More episodes". If clicked, it takes you to the second page of that series because we display only the first one (the 30 I said)

assets/js/episode-guide.js
tatitati: we set the class to display the spinner loading

src/Controller/ProgrammeEpisodes/GuidePartialController.php
tatitati: Yeah. usually they are embeded in the url like /{pid}/. 
In this case I thought that this nested level doesn't affect to the data you receive. It just change the headers such as:
```
?nestedlevel=1 ---> <h1>
?nestedlevel=2 ---> <h2>
....
```
I thought that because is more related with tweaks in the template and can be even optional might be better to add just a query param

templates/programme_episodes/guide.2013inc.html.twig
votemike: Ahh, because on page 1 of that other page, it also displays the same 30.
If we are using this 30 in multiple places, should we make it a constant?

src/Controller/ProgrammeEpisodes/GuidePartialController.php
votemike: I mean coming through the `__invoke` method with `Request $request,` and then access it 
`$request->query->get('nestedlevel');`

tatitati: 👍

assets/js/episode-guide.js
tatitati: We don't make only one call ajax. The user can select quickly three or 20 series that will be opened when finished each request. It was also done in that way in episodes-guide.js on V2

src/Ds2013/Presenters/Domain/EpisodeGuideList/episode_guide_list.html.twig
tatitati: Yesterday we were discussing this @BjornTwachtmann and @felipeparaujo , we decided to keep it as a link and instead add the aria-role="button"

votemike: Tab to it. Press enter.

votemike: Turns out it breaks even if you just click on it....

tatitati: Maybe yarn run gulp?

templates/programme_episodes/guide.2013inc.html.twig
tatitati: yeah , I added it

tests/Controller/ProgrammeEpisodes/DomParsers/EpisodesGuideDomParser.php
tatitati: I really think that it helps to have more expressive tests. 
```php
$this->assertEquals(3, $episodesGuideDom->findGuideItemsInDom()->count());
$this->assertEquals(2, $episodesGuideDom->findSeriesInDom()->count());
$this->assertEquals(1, $episodesGuideDom->findEpisodesInDom()->count());
```
vs
```php
$this->assertEquals(3, $crawler->filter('.js-guideitem')->count());
$this->assertEquals(2, $crawler->filter('[typeof="Season"]')->count());
$this->assertEquals(1, $crawler->filter('.js-guideitem .programme--episode')->count());
```
Instead of a bunch of selectors that don't give you too much information you can really understand what is being parsed and create more complex selectors. For instance one interesting is this:

```php
$this->assertFalse($episodesGuideDom->isPartialTemplate());
```

vs:
```php
// what doesnt mean that there is no footer??
$this->assertFalse($crawler->filter('.footer')->count() === 0);
```

tests/DataFixtures/ORM/ProgrammeEpisodes/ProgrammeItemFixture.php
```diff
@@ -4,21 +4,77 @@
...
+        // standalone episodes (not nested in any series)
+        $this->buildEpisode('p3000000', 'B1-E1', $brand);
+
+        // Series with simple nested episodes
+        $this->buildEpisode('p3000001', 'B1-S1-E1', $series1);
+        $this->buildEpisode('p3000002', 'B1-S1-E2', $series1);
+
+        // Series with more nested series
+        $this->buildEpisode('p3000003', 'B1-S2-S1-E1', $nestedSeries);
+        $this->buildEpisode('p3000004', 'B1-S2-S1-E2', $nestedSeries);
```
tatitati: We can use Composite pattern to create this hierarchies better, something like
```php
$brand->addChild(
         $serie->addChild(
                    $episode
         )
)->flush()
```

src/Ds2013/Presenters/Domain/EpisodeGuideList/episode_guide_list.html.twig
```diff
@@ -0,0 +1,77 @@
...
+                           resource="{{ path('find_by_pid', {'pid': programme.getPid()}) }}"
+                           class="js-serieslink istats-seriesopen block-link link-complex"
+                           href="{{ path('programme_episodes_guide_partial', {'pid': programme.getPid()}) }}"
```
BjornTwachtmann: The href needs to go to the actual episode page. We need another attribute for the lazy load.

src/Ds2013/Presenters/Domain/EpisodeGuideList/episode_guide_list.html.twig
votemike: In any case. If you don't leave the page, this shouldn't be a link.

tests/Controller/ProgrammeEpisodes/DomParsers/EpisodesGuideDomParser.php
felipeparaujo: The problem I see here is that we'll end up with all of these different ideas that could be useful, but are implemented in only a couple of places. This will make the code harder to read in the future, and we'll wonder why this one bit of code is different. 

I'm not saying this is bad, but we need to have standards and stick to them. If we want to do thing this way, fine, but we'll need to update the other tests to follow this standard (in another PR)

src/Ds2013/Css/Components/_dropdown.scss
```diff
@@ -0,0 +1,20 @@
...
+    position: relative;
+    top: $basic-half-spacing / 2;
+    transition: transform 0.25s linear;
+}
+
+.dropdown--icon--expanded {
+    transform: rotate(90deg);
+}
+
+
```
felipeparaujo: Got a bit of extra whitespace here

tests/Controller/ProgrammeEpisodes/DomParsers/EpisodesGuideDomParser.php
tatitati: aaah, ok, I get your point. OK, I will do as we have already.

src/Ds2013/Presenters/Domain/EpisodeGuideList/EpisodeGuideListPresenter.php
```diff
@@ -0,0 +1,65 @@
+<?php
+declare(strict_types=1);
+
+namespace App\Ds2013\Presenters\Domain\EpisodeGuideList;
```
BjornTwachtmann: This is not a domain object

PR #411, opened 2018-02-08
On-now partials without content should still have access-control-allow-origin header

BjornTwachtmann:
This just makes sure that networks without anything currently on them don't put errors in the console.


programmes-developers:

src/Controller/Partial/SchedulesOnNowController.php
```diff
@@ -27,6 +27,7 @@ public function __invoke(
         if (!$network) {
             throw $this->createNotFoundException('No network or service found from network key ' . $networkUrlKey);
         }
+        $this->response()->headers->set('Access-Control-Allow-Origin', '*');
```
votemike: Should this even be before the exception above?

BjornTwachtmann: I tried it, doesn't have any effect there. We shouldn't be linking to this for non-existant networks anyway.

PR #413, opened 2018-02-13
PROGRAMMES-6202 Schedules Structured Data

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6202


programmes-developers:

src/Ds2013/Presenters/Domain/Broadcast/broadcast.html.twig
```diff
@@ -16,11 +16,11 @@
     '3/4 5/6@bpb2 5/6@bpw': not broadcast.getOption('is_stacked'),
 }) -%}
 
-<div class="{{ broadcast_classes }}" vocab="http://schema.org/" typeof="BroadcastEvent">
+<div class="{{ broadcast_classes }}">
     <div class="grid-wrapper">
         <div class="broadcast__info grid {{ info_classes }}">
             {#start time and zone#}
-            <h3 class="broadcast__time gamma" data-timezone="true" property="startDate" datatype="xsd:dateTime" content="{{ broadcast.getStartAt()|date('c') }}">
+            <h3 class="broadcast__time gamma" data-timezone="true" datatype="xsd:dateTime" content="{{ broadcast.getStartAt()|date('c') }}">
```
votemike: Can we get rid of any more of the attributes in this heading?

src/Ds2013/Presenters/Domain/BroadcastEvent/broadcast_event.html.twig
```diff
@@ -24,21 +24,17 @@
...
         </div>
-        <div class="programme__service box-link__elevated micro text--subtle"
-             property="publishedOn"
-             typeof="BroadcastService"
-             resource="{{ broadcast_event.getNetworkUrl(programme_item_network) }}">
+        <div class="programme__service box-link__elevated micro text--subtle">
             {%- for breakdown in broadcast_event.getNetworkBreakdown -%}
                 {%- if breakdown.getNetwork() and breakdown.getNetwork().getUrlKey() -%}
-                    <a property="url" href="{{ broadcast_event.getNetworkUrl(breakdown.getNetwork()) }}"><span property="name">{{ breakdown.getNetworkName() }}</span></a>
+                    <a href="{{ broadcast_event.getNetworkUrl(breakdown.getNetwork()) }}"><span>{{ breakdown.getNetworkName() }}</span></a>
```
votemike: Do we need this `span` any more?

src/Ds2013/Presenters/Domain/Programme/SubPresenters/programme_body.html.twig
```diff
@@ -7,11 +7,11 @@
             }) %}
             {% spaceless %}
                 <abbr title="{{ position_text }}">
-                    <span property="position" datatype="xsd:int">{{ programme_body.getProgramme().getPosition() }}</span>/<span class="programme__groupsize">{{ programme_body.getProgramme().getParent().getExpectedChildCount() }}</span>
+                    <span datatype="xsd:int">{{ programme_body.getProgramme().getPosition() }}</span>/<span class="programme__groupsize">{{ programme_body.getProgramme().getParent().getExpectedChildCount() }}</span>
```
votemike: Can we get rid of any more of the attributes in this span?

src/Ds2013/Presenters/Domain/Programme/SubPresenters/programme_title.html.twig
```diff
@@ -21,11 +21,11 @@
 {% macro render_title(programme, presenter) %}
 {% spaceless %}
     {%- if programme.getPid() != presenter.getProgramme().getPid() -%} {# Do not print RDFa schema for current programme that already has it #}
-        <span property="isPartOf" typeof="{{ rdfa_schema_programme_typeof(programme) }}" resource="http://www.bbc.co.uk/programmes/{{ programme.getPid() }}" >
-            <span property="name">{{ truncate(programme.getTitle(), presenter.getOption('truncation_length'))}}</span>
+        <span>
```
votemike: Do we need this span any more?

src/DsShared/Helpers/SchemaHelper.php
```diff
@@ -0,0 +1,180 @@
...
+    private function getSchemaForEpisode(ProgrammeItem $programmeItem): array
+    {
+        $episode = [
+            '@type' => $programmeItem->isRadio() ? 'RadioEpisode' : 'TVEpisode',
+            'identifier' => $programmeItem->getPid(),
+            'episodeNumber' => $programmeItem->getPosition(),
+            'description' => $programmeItem->getShortSynopsis(),
+            'datePublished' => $programmeItem->getReleaseDate(),
+            'image' => $programmeItem->getImage()->getUrl(480),
+            'name' => $programmeItem->getTitle(), //@TODO This will be JUST the episode name, or so we want the fully qualified name?
```
votemike: I assume this is correct. For the episode name we only want "Episode 1", not "Homes Under The Hammer - Series 87 - Episode 1"?

src/DsShared/Helpers/SchemaHelper.php
```diff
@@ -0,0 +1,180 @@
...
+    private function getSchemaForService(Service $service): array
+    {
+        $bbcContext = $this->getSchemaForOrganisation();
+        $serviceContext = [
+            '@type' => 'BroadcastService',
+            'broadcaster' => $bbcContext,
+            'name' => $service->getName(),
+        ];
+        $network = $service->getNetwork();
+        // @TODO do we always want the network, or not if the name is the same as the service?
```
votemike: If the network and service have the same name, do we still want to add the parent service in?

src/DsShared/Helpers/SchemaHelper.php
```diff
@@ -0,0 +1,180 @@
...
+        if ($network !== null) {
+            $networkContext = [
+                '@type' => 'BroadcastService',
+                'broadcaster' => $bbcContext,
+                'name' => $network->getName(),
+                'logo' => $network->getImage()->getUrl(480),
+                // @TODO Do we want a URL here?
```
votemike: Do we want the URL of the service (e.g. BBC 1)? What would the URL point to?

templates/find_by_pid/tlec.html.twig
```diff
@@ -21,6 +21,12 @@
             {{ script|raw }}
         {%- endfor -%}
     {%- endif -%}
+    {#TODO or should this be in the javascripts block?#}
```
votemike: Is this the correct place in the page for this?

tests/Ds2013/Presenters/Domain/Programme/ProgrammeTemplateTest.php
```diff
@@ -77,7 +76,7 @@ public function testEastendersEpisode()
         $synopsisP = $crawler->filter('.programme__synopsis');
         $this->assertCount(1, $synopsisP, 'Programme has synopsis');
         $this->assertContains('2/5', $synopsisP->text(), 'Synopsis contains x/n parent episode count');
-        $this->assertEquals('Short Synopsis', $synopsisP->filter('span[property=description]')->text(), 'Short synopsis is correct');
+        $this->assertEquals('Short Synopsis', $synopsisP->filter('span')->eq(2)->text(), 'Short synopsis is correct');
```
votemike: Both this and the old way are grim ways of targeting content 😢

src/Builders/BroadcastBuilder.php
```diff
@@ -0,0 +1,31 @@
+<?php
```
felipeparaujo: Shouldn't these builders live in the Test namespace? I don't think we'll use these in the actual app

votemike: I put these Builders alongside the current ones. I guess one day they may be used in the StyleGuide?

src/DsShared/Helpers/SchemaHelper.php
```diff
@@ -0,0 +1,180 @@
+<?php
+declare(strict_types = 1);
+
+namespace App\DsShared\Helpers;
```
BjornTwachtmann: I probably wouldn't put this with the view helpers if it's exclusively used in the controllers.

src/Builders/BroadcastBuilder.php
BjornTwachtmann: I think that's the idea, yeah

src/DsShared/Helpers/SchemaHelper.php
votemike: Where should I put it?

src/Ds2013/Presenters/Domain/Broadcast/broadcast.html.twig
BjornTwachtmann: Probably not. I think they're for the timezone JS

src/Ds2013/Presenters/Domain/Programme/SubPresenters/programme_body.html.twig
BjornTwachtmann: I think they're probably not needed, but check how this displays without htem

src/Ds2013/Presenters/Domain/Programme/SubPresenters/programme_title.html.twig
BjornTwachtmann: doubt it

src/DsShared/Helpers/SchemaHelper.php
BjornTwachtmann: I think that makes sense, it's already linking to context series/brand anyway, right?

src/DsShared/Helpers/SchemaHelper.php
BjornTwachtmann: Generally not. Though the logic for this can get complicated. For the purposes of this, comparing the names shoudl be ok.

templates/find_by_pid/tlec.html.twig
BjornTwachtmann: Do Google or whoever have a preference? The footer seems better to me.

src/DsShared/Helpers/SchemaHelper.php
BjornTwachtmann: There's no natural existing place for it. But maybe create some kind of controller helpers folder or something? The idea of a helper class is right, this just isn't a view helper.

src/DsShared/Helpers/SchemaHelper.php
BjornTwachtmann: You could point to the BBC one homepage, but I'd probably just leave it TBH

src/DsShared/Helpers/SchemaHelper.php
votemike: Right

src/DsShared/Helpers/SchemaHelper.php
votemike: Should it still be loaded via the Shared HelperFactory? Or should I just load it manually? Or create a controller HelperFactory?

BjornTwachtmann: I don't think it needs its own factory. A class that can be autoloaded by symfony DI and injected into the controller that lives somewhere under controller\Helpers shoudl do it.

src/Ds2013/Presenters/Domain/BroadcastEvent/broadcast_event.html.twig
Tasiobg: I would suggest to inspect the HTML elements to see if there is any CSS rules, otherwhise I think there is no much point on using an empty `<span>`

votemike: Any idea where this is rendered?

Tasiobg: bbc.co.uk/programmes/b006m86d/episodes/guide
is the service name under schedule time, example "BBC ONE"

templates/find_by_pid/tlec.html.twig
```diff
@@ -263,4 +263,9 @@
             {{ script|raw }}
         {%- endfor -%}
     {%- endif -%}
+    {% if schema %}
```
BjornTwachtmann: Is it worth moving this to the base templates? It looks like it's being repeated in a few places.

votemike:
Note to self. Look in to adding schema to /schedules home page too.

PR #415, opened 2018-02-13
Remove is_dir checks from the Kernel

BPScott:
Adding the {} means the existence checks are dealt with by the glob
loader, as per https://github.com/symfony/recipes/pull/356/files


programmes-developers:

programmes-developers:

PR #416, opened 2018-02-13
Remove invisible social bar text

BPScott:
Because the type appears in both the title on the <a> tag and in
invisible text it is read out twice by screen readers. Once is
sufficient. The title has more value to sighted users, as you can
mouseover items see what they are if you are unfamilar with the logos so
keep that.

Fixes PROGRAMMES-6180


programmes-developers:

PR #419, opened 2018-02-19
PROGRAMMES-6220 time zone

Tasiobg:
- make time zone code for international services reusable
- move some shared js code into a partial


programmes-developers:

src/Controller/BaseController.php
```diff
@@ -138,6 +141,32 @@ protected function setContextAndPreloadBranding($context)
...
+    {
+        $network = null;
+        if ($context instanceof CoreEntity || $context instanceof Service) {
+            $network = $context->getNetwork();
+        } elseif ($context instanceof Network) {
+            $network = $context;
+        } else {
+            throw new Exception('isInternational method is not implemented by the provided context');
+        }
+        if ($network) {
```
votemike: When will $network not be populated?

src/Controller/FindByPid/TlecController.php
```diff
@@ -61,6 +61,7 @@ public function __invoke(
 
         $this->setIstatsProgsPageType('programmes_container');
         $this->setContextAndPreloadBranding($programme);
+        $this->setInternationalStatusAndTimezoneFromContext($programme);
```
votemike: Should this method be run from inside `setContextAndPreloadBranding()`?

tests/Controller/FindByPid/TlecControllerTest.php
```diff
@@ -147,6 +148,24 @@ public function testTlecIstatsLabels()
...
+        $network = $this->createMock(Network::class);
+        $network->method('isInternational')->willReturn(true);
+        $tlecController = $this->createMock(TlecController::class);
+        $this->invokeMethod($tlecController, 'setInternationalStatusAndTimezoneFromContext', [$network]);
+
+        // check timezone for international services is set to UTC
+        $this->assertSame('UTC', ApplicationTime::getLocalTimeZone()->getName());
+
+        // Set time zone to the previous value to avoid side effect in the remaining test
+        ApplicationTime::setLocalTimeZone($previousTimeZone);
```
votemike: `ApplicationTime::blank()` should be in the `tearDown()` method.

tests/Controller/FindByPid/TlecControllerTest.php
```diff
@@ -147,6 +148,24 @@ public function testTlecIstatsLabels()
         }
     }
 
+    public function testSetInternationalStatusAndTimezoneFromContext()
```
votemike: Is there a test for if a utcoffset is present in the URL?

src/Controller/Schedules/ByWeekController.php
```diff
@@ -38,12 +38,10 @@ public function __invoke(
 
         $this->setIstatsProgsPageType('schedules_week');
         $this->setContextAndPreloadBranding($service);
+        $this->setInternationalStatusAndTimezoneFromContext($service);
```
votemike: Is there any point running this if $utcOffset is true?

src/Controller/BaseController.php
Tasiobg: I think there are some top level services without network, I can't remember, @BjornTwachtmann is best person for this question

src/Controller/FindByPid/TlecController.php
Tasiobg: seems sensible but in that case should `setContextAndPreloadBranding()` be renamed?
any suggestions?  I can only think of `commonProgrammeCode()` but not too convinced

tests/Controller/FindByPid/TlecControllerTest.php
Tasiobg: There are some in `ByDayControllerTest` and `ByWeekControllerTest` do we need them for the TLEC?. I thought there was no utcoffset redirection done from LTEC but I can be wrong.

votemike: Sorry, I didn't read the file name. No need for them here.

src/Controller/BaseController.php
BjornTwachtmann: Not all programmes have a masterbrand. If you look at the relevant domain objects and the DB schema, network/masterbrand is a nullable property.

votemike: So if there is no $network, you could just return early?

src/Controller/Schedules/ByWeekController.php
Tasiobg: Still required to set `$this->isInternational` so it will include the required javascript in the international services. 

Calling `setInternationalStatusAndTimezoneFromContext()` in `ByWeekController.php` and `ByDayController.php` is a bit redundant because the `ApplicationTime` is being set again but in terms of coding this looks better than any of the other ideas I had

src/Controller/BaseController.php
Tasiobg: Yes or move the next condition inside the `if ($network)` but I don't see much of a difference, looks to me more like personal coding style than an improvement.
I don't mind to change it if you prefer is better the early return

src/Controller/FindByPid/TlecController.php
BjornTwachtmann: You could make it a parameter to setContextAndPreloadBranding. Not all pages get the JS timezone reload script (schedule by month/year would break if we stuck it in there for example)

PR #420, opened 2018-02-19
PROGRAMMES-6176 Accessibility urls

tatitati:
Examples URLS:
http://sandbox.bbc.co.uk/programmes/b013pqnm
http://sandbox.bbc.co.uk/programmes/b006q2x0

This fix avoid to navigate by using TAB to the CTA button (which takes to the user to the same url than the title), doing the navigation less confuse from the point of view of the user with problems of accessibility.


programmes-developers:

votemike:
The ticket says the tabindex was going to be set to -1. Why are you instead adding an aria-label?

BjornTwachtmann:
We wanted these not to be read out. Not read out differently. I don't think this PR makes sense, but come and chat to me about it.

votemike:
LGTM. It's a bit hacky but, o well. Everyone else happy?

tests/DsAmen/Presenters/Domain/CoreEntity/Programme/ProgrammeTemplateTest.php
```diff
@@ -97,7 +97,7 @@ public function testEpisodeWithCta()
         $this->assertHasClasses('link-complex', $linkComplex, 'Icon Link classes');
         $this->assertSame('programmeobject_calltoaction', $linkComplex->attr('data-linktrack'));
         $this->assertSame('http://localhost/programmes/clp00001', $linkComplex->attr('href'));
-        $this->assertSame('Play clip', $linkComplex->attr('aria-label'));
+        $this->assertNull($linkComplex->attr('aria-label'));
```
votemike: Would it be better to asset that the tabindex is -1?

tests/DsAmen/Presenters/Domain/CoreEntity/Shared/SubPresenter/StreamableCtaPresenterTest.php
```diff
@@ -102,9 +102,9 @@ public function getPlayTranslationProvider(): array
         $radioEpisode = $this->createMockRadioEpisode();
 
         return [
-            'TV episode shows play episode translation' => [$tvEpisode, 'iplayer_play_episode'],
-            'Clip shows play clip translation' => [$clip, 'iplayer_play_clip'],
-            'Radio episode shows play episode translation' => [$radioEpisode, 'iplayer_play_episode'],
+            'TV episode shows play episode translation' => [$tvEpisode, ''],
```
votemike: I think this is a bit redundant now. But should probably test one that does actually need to have the string.

src/DsAmen/Presenters/Domain/CoreEntity/Base/SubPresenter/base_cta.html.twig
```diff
@@ -1,5 +1,7 @@
 <div class="icon-link {{ cta.getOption('cta_class') }}">
-    <a class="link-complex" href="{{ cta.getUrl() }}" data-linktrack="{{ cta.getLinkLocation() }}" aria-label="{{ tr(cta.getLabelTranslation()) }}">
+    <a class="link-complex" href="{{ cta.getUrl() }}" data-linktrack="{{ cta.getLinkLocation() }}"
+       {% if cta.getLabelTranslation() %}aria-label="{{ tr(cta.getLabelTranslation()) }}" {% else %} tabindex="-1" {% endif %}
```
felipeparaujo: these need to be `{%-` and `-%}`

tests/DsAmen/Presenters/Domain/CoreEntity/Programme/ProgrammeTemplateTest.php
tatitati: actually Both. I added that one as well.

PR #421, opened 2018-02-19
PROGRAMMES-5760 Newest vote test

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5760


programmes-developers:

BjornTwachtmann:
Should this demo iframe even be on develop? It seems a bit dangerous to leave it in the production codebase, if anybody's set the ivote ID somewhere we've not blacklisted then the demo thing shows up.

votemike:
¯\_(ツ)_/¯ I seem to remember asking that question a while ago. But it was fine to go ahead....
Where would you like this instead?

BjornTwachtmann:
I think we should put this on a branch and not render anything for now when an iVote is selected in the programme options. The branch can be built somewhere for testing the vote iframe, that's less risky.

votemike:
Test is on vote-test branch

PR #423, opened 2018-02-19
Document adverts

BPScott:
Fixes PROGRAMMES-5990

---

I've not filled out the "how to test" section about how to appear outside the UK, being based outside the UK already I'll have no idea if my instructions work or not.

### DIFF
--- a/docs/adverts.md
+++ b/docs/adverts.md
@@ -1,0 +1,78 @@
+Adverts
+=======
+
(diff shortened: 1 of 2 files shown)


programmes-developers:

BjornTwachtmann:
I've added a link with some details on how to test adverts. Full details aren't here as I'm not entirely sure if we should be making those public.

PR #427, opened 2018-02-20
PROGRAMMES-6195 upcoming episodes guide

Tasiobg:
highlight changes:
- add upcoming controller and template
- add BaseProgrammeEpisodeController for the shared code, now we only have `getSubNavPresenter()` but soon we should also have `getSchema()`
- remove the temporal routes from 3rd_party.yaml and add the new routes into routes.yaml
- update subnav with the right route
- added "on now" to broadcastEventPresenter
- update unitest
- add "On Air" to the broadcast event
- add "(R)" if is a repeated programme into the programmeBodyPresenter description


programmes-developers:

src/Ds2013/Presenters/Domain/Programme/SubPresenters/programme_body.html.twig
```diff
@@ -11,7 +11,10 @@
                 </abbr>
             {% endspaceless %}
         {% endif %}
-        <span>{{ truncate(programme_body.getProgramme().getShortSynopsis(), programme_body.getSynopsisTruncationLength()) }}</span>
+        <span>
+            {{ truncate(programme_body.getProgramme().getShortSynopsis(), programme_body.getSynopsisTruncationLength()) }}
+            {% if programme_body.getOption('is_repeat') %}<abbr class="repeat" title="Repeat">(R)</abbr>{% endif %}
```
votemike: Should repeat not be translated? And should the (R) be translated?

tests/Ds2013/Presenters/Domain/Programme/ProgrammeTemplateTest.php
```diff
@@ -76,7 +75,7 @@ public function testEastendersEpisode()
         $synopsisP = $crawler->filter('.programme__synopsis');
         $this->assertCount(1, $synopsisP, 'Programme has synopsis');
         $this->assertContains('2/5', $synopsisP->text(), 'Synopsis contains x/n parent episode count');
-        $this->assertEquals('Short Synopsis', $synopsisP->filter('span')->eq(2)->text(), 'Short synopsis is correct');
+        $this->assertEquals('Short Synopsis', trim($synopsisP->filter('span')->eq(2)->text()), 'Short synopsis is correct');
```
Tasiobg: Changes made in programme_body.html.twig add a carriage return to make it more easy to maintain but cause this test to fail, triming the assert fix the issue

src/Controller/ProgrammeEpisodes/UpcomingController.php
```diff
@@ -0,0 +1,56 @@
...
+                if (!$upcoming->isRepeat()) {
+                    $debutUpcomingBroadcast[] = $upcoming;
+                }
+            }
+            $upcomingBroadcast = $debutUpcomingBroadcast;
+        }
+
+        return $this->renderWithChrome('programme_episodes/upcoming.html.twig', [
+            'programme' => $programme,
+            'upcomingBroadcasts' => $upcomingBroadcast,
```
votemike: broadcasts or broadcast?

tests/Ds2013/Presenters/Section/EpisodeSubNav/NavigationItems/NextOnTest.php
```diff
@@ -47,7 +47,7 @@ public function testRouteParams()
     {
         $pid = new Pid('zzzzzzzz');
         $navigationItem = new NextOn($pid, true, 0);
-        $this->assertSame(['pid' => 'zzzzzzzz', 'slice' => 'upcoming'], $navigationItem->getRouteParams());
+        $this->assertSame(['pid' => 'zzzzzzzz'], $navigationItem->getRouteParams());
```
votemike: Does removing the route params not mean this test is redundant?

src/Controller/ProgrammeEpisodes/UpcomingController.php
```diff
@@ -0,0 +1,56 @@
...
+    ) {
+        $this->setContextAndPreloadBranding($programme);
+        $this->setIstatsProgsPageType('broadcast_slice');
+        $this->setInternationalStatusAndTimezoneFromContext($programme);
+
+        $totalAvailableEpisodes = $programmeAggregationService->countStreamableOnDemandEpisodes($programme);
+        $upcomingBroadcast = $collapsedBroadcastsService->findUpcomingByProgrammeWithFullServicesOfNetworksList($programme, 100);
+        $upcomingBroadcastCount = $collapsedBroadcastsService->countUpcomingByProgramme($programme, CacheInterface::MEDIUM);
+
+        $subNavPresenter = $presenterFactory->episodesSubNavPresenter(
```
votemike: Is this the same as in the `GuideController` and `PlayerController`? Is it worth having an abstract controller for the 3 of these to have common methods in?

src/Controller/ProgrammeEpisodes/UpcomingController.php
```diff
@@ -0,0 +1,56 @@
...
+namespace App\Controller\ProgrammeEpisodes;
+
+use App\Controller\BaseController;
+use App\Ds2013\PresenterFactory;
+use BBC\ProgrammesCachingLibrary\CacheInterface;
+use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeContainer;
+use BBC\ProgrammesPagesService\Service\CollapsedBroadcastsService;
+use BBC\ProgrammesPagesService\Service\ProgrammesAggregationService;
+
+class UpcomingController extends BaseController
```
Tasiobg: @votemike Do I need to change anything to make sure the "Structured Data" is working?

votemike: Do you need to set any extra istats labels in here?

src/Controller/ProgrammeEpisodes/UpcomingController.php
```diff
@@ -0,0 +1,56 @@
...
+{
+    public function __invoke(
+        ProgrammeContainer $programme,
+        ?string $debut,
+        CollapsedBroadcastsService $collapsedBroadcastsService,
+        PresenterFactory $presenterFactory,
+        ProgrammesAggregationService $programmeAggregationService
+    ) {
+        $this->setContextAndPreloadBranding($programme);
+        $this->setIstatsProgsPageType('broadcast_slice');
```
votemike: Is "slice" not a V2 concept?

src/Ds2013/Presenters/Domain/BroadcastEvent/broadcast_event.html.twig
```diff
@@ -43,3 +43,8 @@
         </div>
     </div>
 </div>
+{% if (broadcast_event.getCollapsedBroadcast().isOnAir()) %}
```
votemike: Tiny thing, should this be calculated in the presenter?

tests/Ds2013/Presenters/Section/EpisodeSubNav/NavigationItems/NextOnTest.php
votemike: Ignore me, I'm being an idiot.

templates/programme_episodes/upcoming.html.twig
```diff
@@ -0,0 +1,62 @@
...
+
+    <div class="br-box-page programmes-page">
+        {% if upcomingBroadcasts %}
+            <ol class="highlight-box-wrapper">
+                {% for upcoming in upcomingBroadcasts %}
+                    <li>
+                        <div class="broadcast highlight-box--list block-link block-link--steal br-keyline br-blocklink-page br-page-linkhover-onbg015--hover {% if (upcoming.isOnAir()) %} br-box-subtle highlight-box--active{% endif %}" >
+                            <div class="grid-wrapper">
+                                <div class="grid 1/6 1/4@bpb1 1/6@bpb2">
+                                    <div class="grid-wrapper">
```
votemike: I don't think we need grid-wrapper anymore do we?

src/Controller/ProgrammeEpisodes/UpcomingController.php
votemike: I envisioned the SD being done for all the Episode Guide pages in a single separate ticket: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6235 which would mean no work would have to be done as part of this PR.

src/Ds2013/Presenters/Domain/Programme/SubPresenters/programme_body.html.twig
```diff
@@ -11,7 +11,10 @@
                 </abbr>
             {% endspaceless %}
         {% endif %}
-        <span>{{ truncate(programme_body.getProgramme().getShortSynopsis(), programme_body.getSynopsisTruncationLength()) }}</span>
+        <span>
+            {{ truncate(programme_body.getProgramme().getShortSynopsis(), programme_body.getSynopsisTruncationLength()) }}
```
BPScott: If you change `{{` to `{{-` then I think that would remove that extra whitespace that gets generated so you wouldn't need to modify the test.

Same for the `{% endif %}` at the end of the `<span>`

src/Ds2013/Presenters/Domain/Programme/SubPresenters/programme_body.html.twig
Tasiobg: You right!

src/Controller/ProgrammeEpisodes/UpcomingController.php
Tasiobg: This is exactly same value we have in V2. The reason why `setIstatsProgsPageType()` was created is to be able to have same values as in V2. This is a product decision so stats are tracked exactly like they were before.

src/Controller/ProgrammeEpisodes/UpcomingController.php
Tasiobg: We set the default page tipe with `setIstatsProgsPageType()`, other stats are not going to be implemented for now, there is an ongoign discussion about the future of stats tracking

src/Ds2013/Presenters/Domain/BroadcastEvent/broadcast_event.html.twig
Tasiobg: When I was developing that I wasn't sure and I found adding it to the template easy enough to read and maintain.
What other people think?

src/Ds2013/Presenters/Domain/Programme/SubPresenters/programme_body.html.twig
Tasiobg: The issus is the new line (\n) and tabs or spaces before the `{{`
I have double check to make sure and still same
```
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'Short Synopsis'
+'\n
+            Short Synopsis\n
+                    '
```

src/Ds2013/Presenters/Domain/BroadcastEvent/broadcast_event.html.twig
tatitati: As @votemike said is really tiny. But because is being asked..., jeje. 
CollapsedBroadcast is a Domain object, I would say that we don't want to tie too much our views to our model (that is the nice thing about presenters, they decouple our views from the model), otherwise our changes in domain would be propagating around the views all the time.

programmes-developers:

templates/programme_episodes/upcoming.html.twig
Tasiobg: Seems to be using https://github.com/bbc/programmes-frontend/blob/develop/src/Ds2013/Css/Core/Gel/Objects/_grid.scss#L45
If I remove the class it doesn't work fine, please give more details if you think there is something I should change.

votemike: @BPScott Am I getting confused with a different class we no longer need?

src/Controller/ProgrammeEpisodes/BaseProgrammeEpisodeController.php
```diff
@@ -0,0 +1,33 @@
...
+
+abstract class BaseProgrammeEpisodeController extends BaseController
+{
+    protected function getSubNavPresenter(
+        ProgrammesAggregationService $programmeAggregationService,
+        CollapsedBroadcastsService $collapsedBroadcastsService,
+        Programme $programme,
+        PresenterFactory $presenterFactory
+    ): EpisodesSubNavPresenter {
+        $totalAvailableEpisodes = $programmeAggregationService->countStreamableOnDemandEpisodes($programme);
```
BPScott: Use `$programme->getAvailableEpisodesCount()` instead of making a DB query (see #438 where I remove this on master)

templates/programme_episodes/upcoming.html.twig
BPScott: I think you're getting confused with `.grid__inner`, which was the element we no longer needed (that used to go inside the `.grid` element).

templates/programme_episodes/upcoming.html.twig
```diff
@@ -0,0 +1,62 @@
...
+    <div class="br-box-secondary">
+        {{ds2013_presenter(subNavPresenter)}}
+    </div>
+
+    <div class="br-box-page programmes-page">
+        {% if upcomingBroadcasts %}
+            <ol class="highlight-box-wrapper">
+                {% for upcoming in upcomingBroadcasts %}
+                    <li>
+                        <div class="broadcast highlight-box--list block-link block-link--steal br-keyline br-blocklink-page br-page-linkhover-onbg015--hover {% if (upcoming.isOnAir()) %} br-box-subtle highlight-box--active{% endif %}" >
```
BPScott: This is a collapsed broadcast object, which is very similar to what we render on the schedules pages (schedules do broadcasts) but the output is very similar.

Is there a reason that you're recreating the html here instead of leveraging that already existing broadcast object?

It looks like the existing broadcast object requires a broadcast with an optional collapsed broadcast, It seems like it'd be possible to make it so that calling `ds2013('broadcast', ABroadcastOrACollapsedBroadcast)` would result in some output.

src/Controller/ProgrammeEpisodes/BaseProgrammeEpisodeController.php
```diff
@@ -0,0 +1,33 @@
...
+use BBC\ProgrammesPagesService\Service\CollapsedBroadcastsService;
+use BBC\ProgrammesPagesService\Service\ProgrammesAggregationService;
+
+abstract class BaseProgrammeEpisodeController extends BaseController
+{
+    protected function getSubNavPresenter(
+        ProgrammesAggregationService $programmeAggregationService,
+        CollapsedBroadcastsService $collapsedBroadcastsService,
+        Programme $programme,
+        PresenterFactory $presenterFactory
```
BPScott: PresenterFactory and some these classes only seem to be used within the base presenter. It seems a little annoying to have to inject them to every concrete controller and then pass them into this function.

You could specify an `getSubscribedServices` function to pull in those services and have them only accessible within this base controller rather than have to specify those dependencies on every controller that inherits from BaseProgrammeEpisodeController.

e.g.

```php
public static function getSubscribedServices()
    {
        return array_merge(parent::getSubscribedServices(), [
            CollapsedBroadcastsService::class,            
            PresenterFactory::class,
        ]);
    }

protected function getSubNavPresenter(Programme $programme)
{
  $upcomingBroadcastsCount = $this->get(CollapsedBroadcastsService::class)->countUpcomingByProgramme($programme, CacheInterface::MEDIUM);

  return $this->get(PresenterFactory::class)->episodesSubNavPresenter(
    $this->request()->attributes->get('_route'),
    $programme->getNetwork() === null || !$programme->getNetwork()->isInternational(),
    $programme->getFirstBroadcastDate() !== null,
    $programme->getAvailableEpisodesCount(),
    $programme->getPid(),
    $upcomingBroadcastsCount
  );
}
```


This is the same process as how we get access to things in the container in the `BaseController` without having to inject them into every single Controller (see https://github.com/bbc/programmes-frontend/blob/develop/src/Controller/BaseController.php#L62)

programmes-developers:

templates/programme_episodes/upcoming.html.twig
Tasiobg: Thanks for the feedback Ben,
I wans't aware I could use the BroadcastPresenter, I have been investigating that with @bjorn and we reach the conclusion that using the BroadcastPresenter requires some refactorization within the BroadcastPresenter making it more complex and a part from that we are already using same presenters in the other episodes pages so it seems more sensible to keep it like it is now

bjorn: You must have meant somebody else. :)

src/Controller/ProgrammeEpisodes/BaseProgrammeEpisodeController.php
```diff
@@ -0,0 +1,30 @@
...
+namespace App\Controller\ProgrammeEpisodes;
+
+use App\Controller\BaseController;
+use App\Ds2013\PresenterFactory;
+use App\Ds2013\Presenters\Section\EpisodesSubNav\EpisodesSubNavPresenter;
+use BBC\ProgrammesCachingLibrary\CacheInterface;
+use BBC\ProgrammesPagesService\Domain\Entity\Programme;
+use BBC\ProgrammesPagesService\Service\CollapsedBroadcastsService;
+
+abstract class BaseProgrammeEpisodeController extends BaseController
```
votemike: Is this also used for the Episode Controller? If not, should this maybe be `BaseProgrammeEpisodeGuideController `?

src/Controller/ProgrammeEpisodes/UpcomingController.php
```diff
@@ -0,0 +1,44 @@
...
+        $this->setInternationalStatusAndTimezoneFromContext($programme);
+
+        $subNavPresenter = $this->getSubNavPresenter($collapsedBroadcastsService, $programme, $presenterFactory);
+        $upcomingBroadcasts = $collapsedBroadcastsService->findUpcomingByProgrammeWithFullServicesOfNetworksList($programme, 100);
```
votemike: This is limited to 100, does it ever reach that limit? If $debut is set... it may end up with less that 100 results.

src/Controller/ProgrammeEpisodes/UpcomingController.php
```diff
@@ -0,0 +1,44 @@
...
+
+        // display only debut broadcast if $debut is set
+        if ($debut) {
+            $debutUpcomingBroadcasts = [];
+            foreach ($upcomingBroadcasts as $upcoming) {
+                if (!$upcoming->isRepeat()) {
+                    $debutUpcomingBroadcasts[] = $upcoming;
```
votemike: Could probably do `unset($upcomingBroadcasts[$key])` instead. Not sure which is more readable.

templates/programme_episodes/upcoming.html.twig
```diff
@@ -0,0 +1,62 @@
+{% extends 'base_ds2013.html.twig' %}
+
+{% block title %}{{ meta_context.titlePrefix() }} - {{ tr('next_on') }}{% endblock %}
+{% block page_classes %}text-base programmes-page programmes-page--smallpush b-g-p{% endblock %}
+
+{% block body %}
+    <div class="island br-box-secondary">
+        <h1>{{ds_shared('entityContext', programme)}} {{ tr('episodes') }} <span class="visually-hidden">{{ tr('next_on') }}</span></h1>
```
votemike: Can you add a colon just inside the span please?

src/Controller/ProgrammeEpisodes/UpcomingController.php
Tasiobg: already asnwered

src/Controller/ProgrammeEpisodes/UpcomingController.php
Tasiobg: Looks nice, I'll give it a go

PR #430, opened 2018-02-21
PROGRAMMES-6184 - avoid focus on svg (iphone last version)

tatitati:
Jira:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6184

From Aria-hidden doc:
https://developer.paciellogroup.com/blog/2012/05/html5-accessibility-chops-hidden-and-aria-hidden/

> aria-hidden
>
>    Indicates that the element and all of its descendants are not visible or perceivable to any user as implemented by the author.


programmes-developers:

PR #432, opened 2018-02-21
PROGRAMMES-6212 Created specific template for episode

tatitati:
So far the content of the episode.html.twig content is the same we used in example.html.twig.

The rest of logic in the controller was already done (at least the basic)


programmes-developers:

templates/find_by_pid/episode.html.twig
```diff
@@ -0,0 +1,29 @@
+{#
```
votemike: Probably don't need this comment any more.

PR #434, opened 2018-02-21
Episodes tests refactoring

tatitati:
**We had:**
ProgrammeEpisodes\ProgrammeContainerFixture --> It was creating brand and series
ProgrammeEpisodes\ProgrammeItemFixture  -> it was creating episodes

**Now we have:**
ProgrammeEpisodes\BrandFixtures
ProgrammeEpisodes\SeriesFixtures 
ProgrammeEpisodes\EpisodesFixtures

I think is easier to track items in the fixtures in this way, not mixing types.

NOTE: This PR only includes changes in the fixtures structure. Tests and data of the fixture is not mofidied.


programmes-developers:

tests/DataFixtures/ORM/ProgrammeEpisodes/BrandFixtures.php
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Tests\App\DataFixtures\ORM\ProgrammeEpisodes;
+
+use BBC\ProgrammesPagesService\Data\ProgrammesDb\Entity\Brand;
+use Doctrine\Common\DataFixtures\AbstractFixture;
+use Doctrine\Common\Persistence\ObjectManager;
+
+class BrandFixtures extends AbstractFixture
+{
+    public const REF_B006Q2X0 = 'b006q2x0';
```
votemike: Is it useful to set these? They are named after their value.

felipeparaujo: I don't like these. It actually makes it hard for me to read, and you're not really using it everywhere.

tests/DataFixtures/ORM/ProgrammeEpisodes/ProgrammeContainerFixture.php
```diff
@@ -1,55 +0,0 @@
...
-
-namespace Tests\App\DataFixtures\ORM\ProgrammeEpisodes;
-
-use BBC\ProgrammesPagesService\Data\ProgrammesDb\Entity\Brand;
-use BBC\ProgrammesPagesService\Data\ProgrammesDb\Entity\ProgrammeContainer;
-use BBC\ProgrammesPagesService\Data\ProgrammesDb\Entity\Series;
-use Doctrine\Common\DataFixtures\AbstractFixture;
-use Doctrine\Common\Persistence\ObjectManager;
-
-class ProgrammeContainerFixture extends AbstractFixture
```
tatitati: ProgrammeContainerFixture was creating only series, so I changed it to SeriesFixtures

tests/Controller/ProgrammeEpisodes/GuideControllerTest.php
```diff
@@ -14,8 +14,7 @@ class GuideControllerTest extends BaseWebTestCase
     public function setUp()
     {
         $this->loadFixtures([
-            'ProgrammeEpisodes\ProgrammeContainerFixture',
-            'ProgrammeEpisodes\ProgrammeItemFixture',
+            'ProgrammeEpisodes\EpisodesFixtures',
```
tatitati: Not necessary to add these dependencies. Obviously EpisodesFixtures depends on SeriesFixtures and SeriesFixtures depends on BrandsFixtures

tests/DataFixtures/ORM/ProgrammeEpisodes/SeriesFixtures.php
```diff
@@ -0,0 +1,81 @@
...
+     */
+    public function load(ObjectManager $manager)
+    {
+        $this->manager = $manager;
+
+        // a simple series
+        $this->addReference(
+            self::REF_B0000SR1,
+            $this->buildSeries('b0000sr1', 'B1-S1', $this->getReference(BrandFixtures::REF_B006Q2X0))
+        );
```
tatitati: I though this way of creating Elements is more compact because it keeps together the reference of the item created. Othewerise it would be:

```php
$series1 = $this->buildSeries(..., ...., $this->getReference(....));
$this->addReference('.....', $series1);
```

I think is harder to read

tests/DataFixtures/ORM/ProgrammeEpisodes/BrandFixtures.php
tatitati: ok, I will remove the constant, the people seems not happy with that change, cool

tatitati: well, my idea was that instead of REF_<pid>, you can put any better identifier, like REF_SERIES_FINISHED

I se the pid in here just as an example

tatitati: well, my idea was that instead of REF_<pid>, you can put any better identifier, like 
```php
const REF_SERIES_WITH_TWO_SEASONS = '......';
```

THis would be used later as:
```php
$this->buildEpisode(..., ...., $this->getReference(SeriesFixtures::REF_SERIES_WITH_TWO_SEASONS));
```

I set the pid in here just as an example

Tasiobg: Comments also looks good to me
```
$this->getReference('b923423') // Series with two seasons
```

The PR doesn't look bad to me but mixing constant with raw pids can be a bit inconsistent even though I can't think of a better solution.

I think raw values with comments when necesary is ok or constant if the value is repeated is also OK.

I can't see a strong patter to follow because depending on the fixture different solutions may works better

programmes-developers:

PR #435, opened 2018-02-21
Fix width for main page of episode

tatitati:
Original was too wide (same configuration than in Brand page), with this is as the rest of episodes/guide or V2

Also fixed title.


programmes-developers:

PR #436, opened 2018-02-22
Create layout for episode page

BPScott:
Basic columns are present, along with TODO blocks for every module

---

This supersedes #435

### DIFF
--- a/src/Ds2013/Css/Core/Gel/Objects/_grid.scss
+++ b/src/Ds2013/Css/Core/Gel/Objects/_grid.scss
@@ -210,3 +210,30 @@ $massive-pushdown: 50em;
         @include gel-gutterwidths-setup('#{$name}-', $spacing);
     }
 }
+
+// Bound grids
+
+// When a single column looks too wide but two side-by-side columns looks
+// too skinny use a grid--bounded to have a single column with some spacing
+// around it (best used with grid-wrapper--center).
+.grid--bounded {
+    max-width: 624px;
+}
+
+@include mq-range('bpw') {
+    .grow-to-fit\@bpw {
+        float: none;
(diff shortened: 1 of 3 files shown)


programmes-developers:

src/Ds2013/Css/Core/Gel/Objects/_grid.scss
```diff
@@ -210,3 +210,30 @@ $massive-pushdown: 50em;
...
     }
 }
+
+// Bound grids
+
+// When a single column looks too wide but two side-by-side columns looks
+// too skinny use a grid--bounded to have a single column with some spacing
+// around it (best used with grid-wrapper--center).
+.grid--bounded {
+    max-width: 624px;
```
votemike: Do we have some sort of constant/variable for this number?

templates/find_by_pid/episode.html.twig
```diff
@@ -1,24 +1,58 @@
...
+
+{# TODO Episode Panels #}
+
+<div class="prog-layout programmes-page">
+    <div class="grid-wrapper grid-wrapper--center">
+        {# TODO Related Topics Module #}
+
+        <div class="grid grid--bounded 13/24@bpw2 13/24@bpe">
+            <div class="br-box-page prog-box">
+                {# TODO Live Text Stream #}
```
tatitati: Now I catch your idea about TODO, very nice 👍

src/Ds2013/Css/Core/Gel/Objects/_grid.scss
BPScott: No, it isn't used elsewhere / there are no other numbers that hang off it IIRC

PR #437, opened 2018-02-22
PROGRAMMES-6230 Episode Page Clips

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6230

The code for this in V2 is almost unintelligible. I have done my best to replicate it.


templates/find_by_pid/episode.html.twig
```diff
@@ -1,24 +1,91 @@
...
+                    {# TODO Promotions (on the current item) module #}
+
+                    {# TODO MPU at sizes 1 and 2 #}
+
+                    {# TODO Recipes module #}
+
+                    {% if clipsTotal > 0 %}
+                        <div
+                            class="component component--box component--box-flushbody-vertical component--box--primary"
+                            id="clips">
```
votemike: Do we need this id? I have a feeling its only use was from when this was lazy loaded.

programmes-developers:

templates/find_by_pid/episode.html.twig
```diff
@@ -1,24 +1,91 @@
...
+                    {# TODO Live Text Stream #}
+
+                    {# TODO Promotions (on the current item) module #}
+
+                    {# TODO MPU at sizes 1 and 2 #}
+
+                    {# TODO Recipes module #}
+
+                    {% if clipsTotal > 0 %}
+                        <div
```
votemike: I imagine that all this HTML can be refactored once a few more of these sections are added.

src/Controller/FindByPid/EpisodeController.php
```diff
@@ -4,16 +4,25 @@
...
 
 class EpisodeController extends BaseController
 {
-    public function __invoke(Episode $episode)
+    public function __invoke(Episode $episode, ProgrammesAggregationService $aggregationService)
     {
         $this->setIstatsProgsPageType('programmes_episode');
         $this->setContextAndPreloadBranding($episode);
 
+        $totalClips = $aggregationService->countStreamableDescendantClips($episode);
```
BPScott: We already denormalize this value into the programme, so you don't need to make a DB call. Use `$episode->getAvailableClipsCount()`

Same goes for `$programmeAggregationService->countStreamableOnDemandEpisodes` which we seem to use in some other places which can be found from `$programme->getAvailableEpisodesCount()

templates/find_by_pid/episode.html.twig
```diff
@@ -1,24 +1,91 @@
...
+                            class="component component--box component--box-flushbody-vertical component--box--primary"
+                            id="clips">
+                            <div class="component__header br-box-page">
```
BPScott: I think you can remove the `br-box-page`, as this was already set on the wrapping div above

templates/find_by_pid/episode.html.twig
```diff
@@ -1,24 +1,91 @@
...
+                                <h2>Clips</h2>
+                            </div>
+                            <div class="component__body br-box-page">
```
BPScott: I think you can remove the `br-box-page`, as this was already set on the wrapping div above

templates/find_by_pid/episode.html.twig
```diff
@@ -1,24 +1,91 @@
...
+                                                'branding_context': 'page'
+                                            }) }}
+                                        </li>
+                                    {% endfor %}
+                                </ul>
+                            </div>
+                            <div class="component__footer br-box-subtle">
+                                <a class="component__footer__link" href="{{ path('programme_clips', { pid: programme.getPid() }) }}">
+                                    <span class="component__footer__title">{{ tr('see_all_clips_from', {'%1': programme.getTitle()}) }}</span>
+                                    <span class="component__footer__detail">({{ clipsTotal }})</span>
```
BPScott: Use `programme.getAvailableClipsCount()`.

templates/find_by_pid/episode.html.twig
```diff
@@ -1,24 +1,91 @@
...
+                    {# TODO MPU at sizes 1 and 2 #}
+
+                    {# TODO Recipes module #}
+
+                    {% if clipsTotal > 0 %}
+                        <div
+                            class="component component--box component--box-flushbody-vertical component--box--primary"
+                            id="clips">
+                            <div class="component__header br-box-page">
+                                <h2>Clips</h2>
```
BPScott: Translate this. You'll need to pass in the number of clips so it gets pluralised correctly (i.e. it says "clip" if there is only one): `tr('clips', clips|length)`

templates/find_by_pid/episode.html.twig
BPScott: I think it's still useful as a deeplink so you can jump straight to this section.

templates/find_by_pid/episode.html.twig
```diff
@@ -1,24 +1,91 @@
...
+                <div class="br-box-page prog-box">
+                    {# TODO Live Text Stream #}
+
+                    {# TODO Promotions (on the current item) module #}
+
+                    {# TODO MPU at sizes 1 and 2 #}
+
+                    {# TODO Recipes module #}
+
+                    {% if clipsTotal > 0 %}
```
BPScott: Non-0 values and non-empty arrays evaluate to true so `{% if clips %}`, that seems to be the format we're doing in the TLEC controller.

src/Controller/FindByPid/EpisodeController.php
BPScott: We already denormalize this value onto `ProgrammeContainer` and `Episode`, so you don't need to make a DB call. Use `$episode->getAvailableClipsCount()`.

Same goes for `$programmeAggregationService->countStreamableOnDemandEpisodes` which we seem to use in some other places. Use `$programme->getAvailableEpisodesCount()` for that (but that's a different PR).

templates/find_by_pid/episode.html.twig
```diff
@@ -1,24 +1,91 @@
...
+                            id="clips">
+                            <div class="component__header br-box-page">
+                                <h2>Clips</h2>
+                            </div>
+                            <div class="component__body br-box-page">
+                                <ul class="grid-wrapper highlight-box-wrapper highlight-box-wrapper--grid@bpb1 bpw-highlight-box-wrapper--grid">
+                                    {% for clip in clips %}
+                                        <li class="grid 1/2@bpb1 1/2@bpw">
+                                            {{ ds2013('programme', clip, {
+                                                'context_programme': clip,
```
BPScott: Context programme should be the context of the current page (i.e. `programme`), not what you're linking to.

It doesn't really matter for this (as episodes and clips can't have anything between them in the hierarchy) but if the context is a ProgrammeContainer then setting this incorrectly will result in the title hierarch not being generated as expected.

templates/find_by_pid/episode.html.twig
BPScott: I think it's still useful as a deeplink so you can jump straight to this section.

templates/find_by_pid/episode.html.twig
votemike: That makes it feel as though clipsTotal is a bool 👎

src/Ds2013/Presenters/Domain/Programme/SubPresenters/ProgrammeBodyPresenter.php
```diff
@@ -46,7 +46,7 @@ public function getDurationInWords(): string
         if (!$this->programme instanceof ProgrammeItem) {
             throw new InvalidArgumentException("Cannot get duration for non-programmeitem");
         }
-        return $this->playTranslationsHelper->secondsToWords($this->programme->getDuration());
+        return $this->playTranslationsHelper->secondsToWords($this->programme->getDuration()); // @TODO do we want the duration in words or numbers for my clips?
```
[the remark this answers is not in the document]
votemike: @BjornTwachtmann @claudioneves @BPScott ?

votemike:
@BPScott yeah, that component stuff is confusing. I feel it would be better to refactor it after we have the main functionality in V3. Rather than pre-optimising and trying to figure out what V2 was doing and copy it.

src/Ds2013/Presenters/Domain/Programme/SubPresenters/ProgrammeBodyPresenter.php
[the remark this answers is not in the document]
BjornTwachtmann: I'd say the numeric version is probably more readable for clips isn't it? An option should be fine.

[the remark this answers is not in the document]
claudioneves: 00:00, as in the current behaviour. Ex.: https://www.bbc.co.uk/programmes/b06rhv99

templates/find_by_pid/episode.html.twig
```diff
@@ -24,7 +24,44 @@
...
+                                <h2>{{ tr('clips', programme.getAvailableClipsCount()) }}</h2>
+                            </div>
+                            <div class="component__body">
+                                <ul class="grid-wrapper highlight-box-wrapper highlight-box-wrapper--grid@bpb1 bpw-highlight-box-wrapper--grid">
+                                    {% for clip in clips %}
+                                        <li class="grid 1/2@bpb1 1/2@bpw">
+                                            {{ ds2013('programme', clip, {
+                                                'context_programme': programme,
+                                                'highlight_box_classes':'highlight-box--list highlight-box--grid@bpb1 programme--grid@bpb1 highlight-box--grid@bpw programme--grid@bpw',
+                                                'image_options': {
```
BPScott: Do you need to add the image sizes for the responsive images in here?

programmes-developers:

templates/find_by_pid/episode.html.twig
votemike: Added

PR #438, opened 2018-02-22
Remove usage of countStreamableOnDemandEpisodes query

BPScott:
This value is denormalised onto ProgrammeContainers and Episodes already
so we don't need to make this query.


programmes-developers:

PR #440, opened 2018-02-26
PROGRAMMES-6104 Hiding subtitles for Galleries if it's a radio page a…

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6104


programmes-developers:

BPScott:
An easier fix would to be change the  'context_programme'  of the group, to `group.getParent()` when calling the presenter. That way the logic that generates the titling based on the ancestry wouldn't create a subtitle - as it stops looking up the hierarchy of the programme at the `context_programme`

programmes-developers:

votemike:
Are you saying in `templates/find_by_pid/tlec.html.twig` I could have:
```
'context_programme': gallery.isRadio() and gallery.getParent() !==null and (gallery.getParent() instanceOf Episode or gallery.getParent() instanceOf Series) ? gallery.getParent() : programme,
```

Is there any documentation around context_programme? This is the second time I have written a load of load of code just for someone to tell me to adjust the context_programme. I don't quite understand what it's used for.

BPScott:
in `templates/find_by_pid/tlec.html.twig`, the following should work (same as how the `context_programme` is modified for the clips block):

```
'context_programme': programme.isRadio() ? gallery : programme,
```

---

I don't think there's any existing docs but try this on for size...

How we want to display the title hierarchy of a programme/group changes depending upon the page you're on. If you're on a pan-bbc page - such as an a-z listing - you'll want to display the full programme hierarchy in the title and subtitle. However if you're within the context of programme - such as on the Doctor Who brand page, or a specific episode page - then displaying the full programme hierarchy becomes somewhat obnoxious as you don't want to see ", Doctor Who" at the end of every clip title on the Doctor Who brand page. You already know those clips are Doctor Who clips because of the page you're on, you don't need to be told on every programme object within the page.

`context_programme` is how we control that logic. Setting a `context_programme` shall curtail the titles and cut off at that particular programme.

Consider http://www.bbc.co.uk/programmes/p057wxky. If that clip were to be displayed on a pan-BBC page then we would set the`context_programme` to be null (the default) and the titling would be:

```
Tango-Mambo with Missy and the Master
The Doctor Falls, Series 10, Doctor Who
```

Meanwhile, within the context of the Doctor Who brand (e.g. http://www.bbc.co.uk/programmes/b006q2x0 or http://www.bbc.co.uk/programmes/b006q2x0/clips), we don't need to see "Doctor Who" in the titling, so we would set the `context_programme` to be the Doctor Who brand, and the titling would be:

```
Tango-Mambo with Missy and the Master
The Doctor Falls, Series 10
```

You can set the context to be anything in the programme/group's hierarchy. So within the context of Doctor Who Series 10 (e.g. http://www.bbc.co.uk/programmes/b08n5s4l or http://www.bbc.co.uk/programmes/b08n5s4l/clips) the `context_programme` is the programme b08n5s4l and the titling would be:

```
Tango-Mambo with Missy and the Master
The Doctor Falls
```

And for the sake of completeness, the final possible value in the hierarchy the episode. Within the context of The Doctor Falls (e.g. http://www.bbc.co.uk/programmes/b08xhhn4 or http://www.bbc.co.uk/programmes/b08xhhn4/clips), the context_programme is the programme b08xhhn4 and the titling would be:

```
Tango-Mambo with Missy and the Master
```

All of the above applies to groups too, visit /programmes/:pid/galleries instead of /programmes/:pid/clips in the examples above to see that.

If you set the programme and the context_programme to be the same then you'll still get the current programme's title rather than an empty string.

---

In the case of this ticket, we're abusing this facility slightly to hide that hierarchy by pretending the context is the current programme (which shall hide all its parents titles).

votemike:
But what about the logic surrounding the parent being a Series or Episode?
---
Ahh, so in normal cases, you want to set the context to the main Core_Entity of the page? And the title is the _only_ thing it affects?

BjornTwachtmann:
@votemike , the context is currently only used for titling logic. It's basically to say "Don't include this programme in the title you print as that title is already implicit in the page". Setting context to the gallery here should work as ben outlined (i think)

BPScott:
> Ahh, so in normal cases, you want to set the context to the main Core_Entity of the page? And the title is the only thing it affects?

Bingo.

votemike:
Closed in favour of https://github.com/bbc/programmes-frontend/pull/446

PR #441, opened 2018-02-27
PROGRAMMES-6218 Fixing bug with nav bar translation

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6218


programmes-developers:

BjornTwachtmann:
I think this is actually not the right place to put this. I think the issue is likely that the placeholderresolver is called before we call setLanguage in the baseController. There should only be one instance of the TranslateProvider from the DI container. 

The best thing to do is probably to move setting the language from preRender to fulfilledBrandingPromise in BaseController

PR #443, opened 2018-02-27
PROGRAMMES-6213 Logic to show/hide panels in episode Map

tatitati:
Interesting episode pids to test:

With Tx and More panel:
http://sandbox.bbc.co.uk/programmes/b09kkt1m

WIth Tx and Empty panel: 
http://sandbox.bbc.co.uk/programmes/b00744wz

A radio episode:
http://www.bbc.co.uk/programmes/p05yrm66


src/Controller/FindByPid/EpisodeController.php
```diff
@@ -4,16 +4,41 @@
...
 class EpisodeController extends BaseController
 {
-    public function __invoke(Episode $episode)
+    public function __invoke(Episode $episode, CollapsedBroadcastsService $collapsedBroadcastsService)
     {
         $this->setIstatsProgsPageType('programmes_episode');
         $this->setContextAndPreloadBranding($episode);
 
+        $upcomingBroadcasts = $this->getUpcomingBroadcasts($collapsedBroadcastsService, $episode);
+        $lastOn = $this->getLastOn($episode, $collapsedBroadcastsService);
```
tatitati: Not sure if for the next panels we need more data than this.

src/Ds2013/Css/PageSpecfic/Episode/_map.scss
```diff
@@ -0,0 +1,142 @@
+$map-gap : 8px;
```
tatitati: Not used all them right now. But i guess that we will be able to delete the unused one when we finish the map and check what is not used from here

programmes-developers:

src/Controller/FindByPid/EpisodeController.php
```diff
@@ -4,16 +4,41 @@
...
+    ): array {
+        // World News brand pages are the only ones that show the Last On column in the MAP. The Last On column
+        // shows a collapsed broadcast, which has a list of networks and services of the broadcasts, which means
+        // it needs the full list of services for the networks. If something else starts using the Last On
+        // column, this has to be updated, as does the MAP, or else stuff will blow up.
+
+        return $collapsedBroadcastsService->findPastByProgrammeWithFullServicesOfNetworksList($episode, 1);
+    }
+
+    private function getUpcomingBroadcasts($collapsedBroadcastsService, $episode)
```
votemike: If you're limiting to 1, surely this should be `getUpcomingBroadcast()`

src/Ds2013/Presenters/Pages/Episode/Map/EpisodeMapPresenter.php
```diff
@@ -0,0 +1,131 @@
...
+    public function __construct(Episode $episode, array $upcomingBroadcasts, array $lastOnBroadcasts)
+    {
+        // This page has no options
+        parent::__construct([]);
+        $this->episode            = $episode;
+        $this->upcomingBroadcasts = $upcomingBroadcasts;
+        $this->lastOnBroadcasts   = $lastOnBroadcasts;
+    }
+
+    public function getAmountOfSidePanels()
```
votemike: Typehints?

src/Ds2013/Presenters/Pages/Episode/Map/EpisodeMapPresenter.php
```diff
@@ -0,0 +1,131 @@
...
+     * @return AbstractMainPanelMap[]
+     */
+    public function getMainPanelsSubPresenters() :array
+    {
+        if (!is_null($this->mainSubPresenters)) {
+            return $this->mainSubPresenters;
+        }
+
+        $this->mainSubPresenters = [
+            new EpisodeMapPlayoutPresenter($this->episode, $this->upcomingBroadcasts, $this),
```
votemike: Why are you passing `$this` through? Seems like a terrible idea!

src/Ds2013/Presenters/Pages/Episode/Map/Panels/Main/AbstractMainPanelMap.php
```diff
@@ -0,0 +1,21 @@
...
+{
+    protected $episode;
+    protected $upcomingBroadcasts;
+    protected $parentPresenter;
+
+    public function __construct(Episode $episode, array $upcomingBroadcasts, EpisodeMapPresenter $parentPresenter)
+    {
+        $this->episode = $episode;
+        $this->upcomingBroadcasts = $upcomingBroadcasts;
+        $this->parentPresenter = $parentPresenter;
```
votemike: This feels like a bad idea.

src/Ds2013/Presenters/Pages/Episode/Map/Panels/Side/AbstractSidePanelMap.php
```diff
@@ -0,0 +1,27 @@
...
+use App\Ds2013\Presenters\Pages\Episode\Map\EpisodeMapPresenter;
+use BBC\ProgrammesPagesService\Domain\Entity\Episode;
+
+Abstract class AbstractSidePanelMap extends Presenter
+{
+    protected $episode;
+    protected $upcomingBroadcasts;
+    protected $parentPresenter;
+
+    public function __construct(Episode $episode, array $upcomingBroadcasts, EpisodeMapPresenter $parentPresenter)
```
votemike: Do all side panels need upcoming broadcasts?

src/Ds2013/Presenters/Pages/Episode/Map/Panels/Side/EpisodeMapTxPresenter.php
```diff
@@ -0,0 +1,17 @@
...
+namespace App\Ds2013\Presenters\Pages\Episode\Map\Panels\Side;
+
+use App\Ds2013\Presenters\Pages\Episode\Map\EpisodeMapPresenter;
+use BBC\ProgrammesPagesService\Domain\Entity\Episode;
+
+class EpisodeMapTxPresenter extends AbstractSidePanelMap
+{
+    protected $lastOn;
+
+    public function __construct(Episode $episode, array $upcomingBroadcasts, EpisodeMapPresenter $parentPresenter, array $lastOn)
```
votemike: What is `$lastOn` an array of?

src/Ds2013/Presenters/Pages/Episode/Map/EpisodeMapPresenter.php
```diff
@@ -0,0 +1,131 @@
...
+
+        return $this->sideSubPresenters;
+    }
+
+    /**
+     * @return AbstractMainPanelMap[]
+     */
+    public function getMainPanelsSubPresenters() :array
+    {
+        if (!is_null($this->mainSubPresenters)) {
```
votemike: Do you really need to cache them? Will they be called more than once?

src/Ds2013/Presenters/Pages/Episode/Map/episode_map.html.twig
```diff
@@ -0,0 +1,25 @@
+<div class="b-g-p no-margin-vertical">
+    <div class="grid-wrapper grid-wrapper--flush map map--episode map--count-2">
+
+        {#PLAYOUT#}
+        <div class="grid 2/3@bpw2 2/3@bpe map__column map__column--first">
+            <div class="grid-wrapper map__inner grid__inner--flush br-box-secondary">
+                {%- for panel in episode_map.getMainPanelsSubPresenters() -%}
```
votemike: There will always be the same 2 won't there? Would it be better to just hardcode them? Rather than foreach an array. The foreach suggests the number of main panels can vary.

src/Ds2013/Presenters/Pages/Episode/Map/EpisodeMapPresenter.php
```diff
@@ -0,0 +1,131 @@
+<?php
+declare(strict_types=1);
+
+namespace App\Ds2013\Presenters\Pages\Episode\Map;
```
votemike: Should these presenters be under "Section" instead of "Pages"?

src/Controller/FindByPid/EpisodeController.php
tatitati: Make sense, part of my doubt if is in the future we are gonna need for the rest of panels more than one broadcast. For now I'm limiting to 1, but indeed I have to figure out if we need more than one for the rest of logic.

src/Ds2013/Presenters/Pages/Episode/Map/EpisodeMapPresenter.php
tatitati: oh no no. I disagree. 

What you are seeing is mediator pattern (AKA "Double dispatch"). The parent is registering the childs (subPresenters), and these sub-presenters are registering the parent. So they can speak each other.

This is useful when you need info that have in account the whole group of childs but at the same time you don't want to couple your childs each other. 

In this scenario we have some css  depending on the position and the amount of panels that appear in the side-map. As you can imagine is not necessary that each subpresenter know in what position they are and in what order. So the child ask to the parent if is necessary for them these classes. The parent has this knowledge ready and thel them the right classes

src/Ds2013/Presenters/Pages/Episode/Map/EpisodeMapPresenter.php
tatitati: Depends of how is implemented the rest of panels and if they need to call the parent in more than one ocassion. I just cannot say.

src/Ds2013/Presenters/Pages/Episode/Map/EpisodeMapPresenter.php
tatitati: well, I took this option because the episode map is only used in the main page of episode, this section is not reused anywhere else

src/Ds2013/Presenters/Pages/Episode/Map/Panels/Side/AbstractSidePanelMap.php
tatitati: Maybe not, but because I didn't implement them I did it as most generic as I can. We can get rid off this dependency if when developing an specific panel is not necessary.

src/Ds2013/Presenters/Pages/Episode/Map/EpisodeMapPresenter.php
votemike: Surely it should be similar to the Amen MAP for the TLEC page?

src/Controller/FindByPid/EpisodeController.php
```diff
@@ -24,10 +25,33 @@ public function __invoke(Episode $episode, ContributionsService $contributionsSe
...
             'clips' => $clips,
+            'upcomingBroadcasts' => $upcomingBroadcasts,
+            'lastOn' => $lastOn
         ]);
     }
+
+    private function getLastOn(
+        Episode $episode,
+        CollapsedBroadcastsService $collapsedBroadcastsService
+    ): array {
```
felipeparaujo: This and the next method seems like a lot of useless boilerplate to me... You don't need a method that just calss another method

src/Ds2013/Presenters/Pages/Episode/Map/EpisodeMapPresenter.php
felipeparaujo: Yeah, but it's not really a page

src/Ds2013/Presenters/Pages/Episode/Map/Panels/Main/episode_map_details.html.twig
```diff
@@ -0,0 +1,4 @@
+<div class="grid island">
```
felipeparaujo: We know these templates are part of the episode map, we don't need to prefix all of them with `episode_map`. Same goes for the Presenter classes. Also, can we come up with better names for them? I know you just copied this over from v2, but these names are not the best

src/Controller/FindByPid/EpisodeController.php
tatitati: Yeah, my bad. I missed some extra line of that function when solving conflicts.

src/Ds2013/Presenters/Pages/Episode/Map/EpisodeMapPresenter.php
```diff
@@ -0,0 +1,140 @@
...
+     * @return AbstractMainPanelMap[]
+     */
+    public function getMainPanelsSubPresenters() :array
+    {
+        if (!is_null($this->mainSubPresenters)) {
+            return $this->mainSubPresenters;
+        }
+
+        $this->mainSubPresenters = [
+            new PanelPlayoutPresenter($this->episode, $this->upcomingBroadcasts, $this),
```
votemike: Can these just be built as part of constructing this presenter? And saved separately instead of in an array?

src/Ds2013/Presenters/Pages/Episode/Map/Panels/Side/AbstractSidePanelMap.php
votemike: What if the state of the `EpisodeMapPresenter` changes after the SidePanelMap is initialised?

votemike: None of these SidePanelMaps are implemented yet are they? So you probably don't need the constructor yet. Do you?

tatitati: If the mediator state change and this change affect to any of the childs (I cannot think in any example right now), then the mediator must communicate this change to the child/s that care about this change. Is part of the natural process of having a mediator, they communicate through the EpisodeMapPresenter to reduce complex communications among all of them. In that sense is quite similar to Observer, but instead with this pattern the communication is bidirectional.

src/Ds2013/Presenters/Pages/Episode/Map/Panels/Side/AbstractSidePanelMap.php
```diff
@@ -0,0 +1,28 @@
...
+    {
+        parent::__construct([]);
+        $this->episode = $episode;
+        $this->upcomingBroadcasts = $upcomingBroadcasts;
+        $this->parentPresenter = $parentPresenter;
+    }
+
+    public function getPanelCssClasses()
+    {
+        return $this->parentPresenter->getCssClassesBasedOnExistingSubPresenters($this);
```
votemike: AHHHH
AHHHH!!!

tatitati: XDDDDDDDDD

tatitati: I would like to see more constructive feedbacks instead of finding my PRs with pictures like this and shouts, this can be really annoying as well for the rest of reviewers. It distracts and confuse a lot to the rest of members.

### So..
Bidirectional communication between objects is not even unusual, don't confuse with circular dependency which is something completely different and is a real problem, it also looks much more like that picture. 

From the Gang Of Four you have some very extendend and widely examples:

**- Example of Visitor:**
https://sourcemaking.com/design_patterns/visitor/java/2
In java sometimes we don't want to re-compile some classes everytime we add new functionality to a class, so we want to expand the functionality of some classes but without modify them to avoid this problem. This is solved creating an external class that receive the class we try to expand, and adding all the additional functions we wish.  This doesn't use bidirectional communication, but is using double dispatch.

**- Example of observer:**
http://www.fluffycat.com/PHP-Design-Patterns/Observer/
This handle One to Many communications. And event (possibly the change in state of an object) needs to be notified to an observer who is ressponsible of notify to many subscribed objects of the change. This might use bidirectional communication (an object might receive an event or notify it)

**- Example of Mediator:**
https://sourcemaking.com/design_patterns/mediator/php
Sometimes we have relatioships Many to Many. In this case we want to reduce coupling between each component and simplify communications introducing a mediator (considered for some people a kind of facade). This usually use bidirectional.

## And...
I tried some other approach (passing these classes in presenter.options), doesn't look too natural  and it has it problems as well. Also I tried passing the classes in an attribute for the subpresenter, I think is ugly to have a member variable that contains info about css. I will try again this morning.

votemike: Of course. Didn't mean to annoy anyone. It was getting towards the end of the day and I think pictures are fun. I'll avoid them in the future.
The PR has changed now so I can't really make appropriate comments about what it used to be like, but it looked a lot like a circular dependency. Maybe it wasn't.

tatitati: no worries buddy, by the end of the day I was stressed, maybe a little sensitive or tired 🤗
Have a good day working from home! ⛄️

src/Ds2013/Presenters/Pages/Episode/Map/EpisodeMapPresenter.php
BjornTwachtmann: I would put this in section. It can be used in a page that's not the episode page. It just isn't

src/Controller/FindByPid/EpisodeController.php
```diff
@@ -24,10 +25,32 @@ public function __invoke(Episode $episode, ContributionsService $contributionsSe
...
+        // World News brand pages are the only ones that show the Last On column in the MAP. The Last On column
+        // shows a collapsed broadcast, which has a list of networks and services of the broadcasts, which means
+        // it needs the full list of services for the networks. If something else starts using the Last On
+        // column, this has to be updated, as does the MAP, or else stuff will blow up.
+
+        if ($episode->getNetwork() && $episode->getNetwork()->isWorldNews()) {
```
BjornTwachtmann: This is from the Brand page isn't it? Does it have any relevance for an episode page? I'd be rather suprised if it did.

src/Ds2013/Css/PageSpecfic/Episode/_map.scss
```diff
@@ -0,0 +1,156 @@
+$map-gap: 8px;
+$massive-pushdown: 50em;
+
+.map {
+    overflow: hidden;
+}
+
+.map__column {
```
BjornTwachtmann: Where is all this CSS coming from? It has nothing to do with the epiosde page does it? This is from the 2013 brand page map, no?

templates/find_by_pid/episode.html.twig
```diff
@@ -8,7 +8,9 @@
 
     {# TODO Leaderboard advert #}
 
-    {# TODO Episode Panels #}
+    {{
+        ds2013('episodeMap', episode, upcomingBroadcasts, lastOn, {})
```
votemike: @BjornTwachtmann @BPScott @felipeparaujo @Tasiobg When do we do presenters like this? And when do we construct presenters in the Controller?
I was thinking of suggesting that the presenter is constructed in the controller, but then doubted myself. Any views?

src/Ds2013/Presenters/Section/Episode/Map/EpisodeMapPresenter.php
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * @param Episode $episode
+     * @param CollapsedBroadcast[] $upcomingBroadcasts
+     * @param CollapsedBroadcast[] $lastOnBroadcasts
+     */
+    public function __construct(Episode $episode, array $upcomingBroadcasts, array $lastOnBroadcasts)
+    {
+        // This page has no options
+        parent::__construct([]);
+        $this->episode            = $episode;
```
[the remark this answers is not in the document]
tatitati: Episode will be used for instance to Details panel, to display the summary, etc. Is just one example

templates/find_by_pid/episode.html.twig
BjornTwachtmann: It kind of depends on the complexity of the construction. This is sort of on the borderline, you could go either way. I imagine once the episode page is complete we'll want to pull this constructor out of here and do it either in a page presetner or the controller.

src/Ds2013/PresenterFactory.php
```diff
@@ -167,6 +169,18 @@ public function programmePresenter(
...
+    /**
+     * Pages
+     */
+    public function episodeMapPresenter(
+        Episode $programme,
+        ?CollapsedBroadcast $upcomingCollapsedBroadcast,
+        ?CollapsedBroadcast $lastOnCollapsedBroadcast,
+        $options = []
+    ) :EpisodeMapPresenter {
+        return new EpisodeMapPresenter($programme, $upcomingCollapsedBroadcast, $lastOnCollapsedBroadcast);
```
votemike: You're not passing the options through

src/Ds2013/Presenters/Section/Episode/Map/EpisodeMapPresenter.php
```diff
@@ -0,0 +1,106 @@
...
+
+    public function __construct(Episode $episode, ?CollapsedBroadcast $upcomingBroadcasts, ?CollapsedBroadcast $lastOnBroadcasts)
+    {
+        // This page has no options
+        parent::__construct([]);
+        $this->episode            = $episode;
+        $this->upcomingBroadcast = $upcomingBroadcasts;
+        $this->lastOnBroadcast   = $lastOnBroadcasts;
+        $this->sideSubPresenters  = $this->buildSidePanelsSubPresenters();
+        $this->mainSubPresenters  = $this->buildMainPanelsSubPresenters();
```
votemike: I still don't understand why you are doing this. Could you change it? Or explain why it's done this way.

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/PanelDetailsPresenter.php
```diff
@@ -0,0 +1,23 @@
...
+use BBC\ProgrammesPagesService\Domain\Entity\Episode;
+
+class PanelDetailsPresenter extends Presenter
+{
+    /** @var Episode */
+    private $episode;
+
+    public function __construct(Episode $episode)
+    {
+        parent::__construct([]);
```
votemike: Do you need to pass an empty array though?

src/Ds2013/Presenters/Section/Episode/Map/Panels/Side/PanelEmptyPresenter.php
```diff
@@ -0,0 +1,13 @@
+<?php
+declare(strict_types=1);
+namespace App\Ds2013\Presenters\Section\Episode\Map\Panels\Side;
+
+use App\Ds2013\Presenter;
+
+class PanelEmptyPresenter extends Presenter
+{
+    public function __construct()
```
votemike: You could probably delete this constructor.

templates/find_by_pid/episode.html.twig
```diff
@@ -8,7 +8,7 @@
 
     {# TODO Leaderboard advert #}
 
-    {# TODO Episode Panels #}
+    {{ ds2013_presenter(episodeMapPresenter, {}) }}
```
votemike: What are the curly braces for?

src/Controller/FindByPid/EpisodeController.php
```diff
@@ -24,10 +26,21 @@ public function __invoke(Episode $episode, ContributionsService $contributionsSe
...
+        $lastOnBroadcasts = $collapsedBroadcastsService->findPastByProgramme($episode, 1);
+
+        $episodeMapPresenter = new EpisodeMapPresenter(
+            $episode,
+            !empty($upcomingBroadcasts) ? reset($upcomingBroadcasts) : null,
+            !empty($lastOnBroadcasts) ? reset($lastOnBroadcasts) : null
+        );
+
         return $this->renderWithChrome('find_by_pid/episode.html.twig', [
+            'episode' => $episode,
```
votemike: You're not using this in the template.

src/Ds2013/Css/PageSpecfic/Episode/_map.scss
```diff
@@ -0,0 +1,49 @@
+.map__buttons {
```
votemike: Where is this used?

src/Ds2013/Presenters/Section/Episode/Map/Panels/Side/panel_empty.html.twig
```diff
@@ -0,0 +1,3 @@
+<div class="island js-continuous-play-panel">
```
votemike: What's js-continuous-play-panel? Does it need to be on an empty panel?

src/Controller/FindByPid/EpisodeController.php
tatitati: yeah, but will be used in multiple panels. For sure it will be used.

votemike: Panels? But you're already passing the presenter through. It exists in the presenter...

src/Ds2013/Css/PageSpecfic/Episode/_map.scss
tatitati: We imported the css that will be used for the map from V2.
webapp/php/sharedmodules/programmes/views/programmes/episode

Exactly that class is used here:
https://github.com/bbc/programmes/blob/develop/webapp/php/sharedmodules/programmes/views/programmes/episode/details.spv#L48

src/Ds2013/Presenters/Section/Episode/Map/Panels/Side/panel_empty.html.twig
tatitati: https://github.com/bbc/programmes/blob/develop/webapp/php/sharedmodules/programmes/views/programmes/episode/empty.spv#L8

src/Controller/FindByPid/EpisodeController.php
tatitati: ouuu, catch you. True, that bit was refactored

src/Ds2013/Presenters/Section/Episode/Map/Panels/Side/panel_empty.html.twig
votemike: O dear....

BjornTwachtmann: This was part of a continuous playout experiment that dave build aaages ago. Kill it everywhere.

PR #445, opened 2018-02-28
PROGRAMMES-6184 avoid focus on mobiles for cta and recipes

tatitati:
Some stuff still might be focused on safari-iphone

Already tested on test and works as expected.


programmes-developers:

PR #446, opened 2018-02-28
PROGRAMMES-6104 Hiding subtitles for galleries if radio

votemike:
2nd attempt at https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6104

Instead of https://github.com/bbc/programmes-frontend/pull/440


src/DsAmen/Presenters/Domain/CoreEntity/Base/BaseCoreEntityPresenter.php
```diff
@@ -108,7 +107,7 @@ public function getType(): string
 
     protected function validateOptions(array $options): void
     {
-        if (isset($options['context_programme']) && !($options['context_programme'] instanceof Programme)) {
+        if (isset($options['context_programme']) && !($options['context_programme'] instanceof CoreEntity)) {
```
votemike: This is called context_programme.... but it's actually a context_core_entity

programmes-developers:

src/DsAmen/Presenters/Domain/CoreEntity/Base/BaseCoreEntityPresenter.php
BPScott: This is is awkward naming but just about makes sense if you squint at it.

When it comes to these hierarchies, groups cannot have anything under them - they're not valid parents (e.g. you can have "Brand -> Clip -> Gallery" but you can't have "Gallery -> Clip" or "Collection -> Episode" (aside: Programmes & Groups can be members of Collections and Seasons but this is a different concept to the parentage). Note the [return type on CoreEntity->getParent is ?Programme](https://github.com/bbc/programmes-pages-service/blob/1032d7f43ceaef79d5105cafe973fa81173568ab/src/Domain/Entity/CoreEntity.php#L185)

I guess to make things a bit more accepting, and to avoid having to check if a parent exists in some cases it seems we said "you can pass the current entity in as the context and it won't error, you'll get back just the current title - same as if you passed in the parent".

If that affordance seems confusing, we could be stricter with passing correct values into the object rather than letting the error correction do its thing by changing how we do "fake context stuff" on the Radio TLEC page.

```
'context_programme': programme.isRadio() and gallery.getParent() ? gallery.getParent() : programme,
```

PR #448, opened 2018-03-01
Fix instanceof check

votemike:
I'm ~~99%~~ ~~90%~~ 50% sure that this improvement is correct.
@BjornTwachtmann @BPScott could one of you confirm?


programmes-developers:

BjornTwachtmann:
Could a group not also be used as a context? What's the thinking behind restricting it?

votemike:
After Ben's comments on my previous PR, it was my understanding that a context_programme should only be a Programme and not a Group. If you are trying to set a Group, you probably actually mean to set the Group's parent as the context.
At least, that is the impression I got.

BjornTwachtmann:
You probably wouldn't normally set a group as a context, in theory I can't see why it wouldn't work though. Does setting a group here actually work? If not it probably makes sense to disallow it.

votemike:
I tried with my previous PR and I had to update a similar check to allow Core_Entities instead of Programmes. So I guess it wouldn't work.
Only a small thing, we can wait for Ben to confirm. He knows more about this than me.

BPScott:
I don't think you can ever fail this validation currently as [BaseCoreEntityPresenter](https://github.com/bbc/programmes-frontend/blob/46e576240ddaf3b24c6299475d75be83ab39be7f/src/DsAmen/Presenters/Domain/CoreEntity/Base/BaseCoreEntityPresenter.php#L111-L113) ensures that a context is a programme.

I ~~think~~ setting a context to be a group should work in the sense that it shouldn't explode. But it's not a valid data state as programmes and groups can not have a parent that is a group (check the return type on CoreEntity->getParent() per my comments on #446)). I think being able to pass a CoreEntity into this should raise an error, as being strict is preferable IMO.

Given that the name of the property is 'context_programme' and the error message says 'context_programme option must be null or a Programme domain object' I think it makes sense to filter to instances of Programme.

---

Incidentally looking at this class, I don't see a default for 'programme_context' defined in the $options property. We should add that, which would let is remove the isset() check and needing the `?? null` bit  on line 131

votemike:
So.... delete this PR without merging? But create a new PR setting a default context_programme of null?

BPScott:
This PR is good - changing the check to be `instanceof Programme` makes the condition check and the error message it triggers line up.

Can you also add `context_programme => null` to the options array (line 33), as it's an option so it should be defined in there. That will let you change [line 131](https://github.com/bbc/programmes-frontend/blob/c07a3375dc64e78cea026e600bb70fad5789f032/src/DsAmen/Presenters/Domain/CoreEntity/Base/SubPresenter/BaseTitlePresenter.php#L131) from ` $this->options['context_programme'] ?? null,` to ` $this->options['context_programme']`. - the value is always set so you don't need the coalesce check

PR #450, opened 2018-03-01
PROGRAMMES-6207 Cleaning up headings for the episode guide page

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6207

Might be best to view it like this URL: https://github.com/bbc/programmes-frontend/pull/450/files?w=1


programmes-developers:

translations/programmes/en.po
```diff
@@ -1657,6 +1657,10 @@ msgstr ""
 msgid "recommendations_package_title"
 msgstr "Recommendations for you"
 
+#. English: "Related Content"
+msgid "related_content"
+msgstr "Related Content"
```
votemike: I assume this is how you add new strings.

BjornTwachtmann:
The principle of this looks good. I'm struggling to read some of the diff because of the whitespace changes, but I think it's fine. 

Same comment on the translation as the other PR

votemike:
@BjornTwachtmann I put a link in the description to review: https://github.com/bbc/programmes-frontend/pull/450/files?w=1

BjornTwachtmann:
Ah, reading comprehension failure on my part then. But yeah, that's what I thought it was doing :) If you fix the translation files it's good to merge.

votemike:
Test this please

src/Ds2013/Presenters/Pages/EpisodeGuideList/_episode_guide.scss
```diff
@@ -0,0 +1,10 @@
+.js-serieslink {
```
felipeparaujo: @BPScott Is this the best way to set heading styling?

votemike: Because headings are block elements, the "View Episodes" span was being pushed below the Series title. So..... I did this.

BPScott: (Whoops missed this, sorry)

A convention I've found useful to for `.js-blah` classes to be only used as hooks for JS stuff and shouldn't have any styling applied to them.

I'd much rather add a class name to that [specific heading](https://github.com/bbc/programmes-frontend/blob/a6add8ea4f1d1dcb12cb1b2a1f4d340ec79da78d/src/Ds2013/Presenters/Pages/EpisodeGuideList/episode_guide_list.html.twig#L17) rather than do a blanket styling for all heading types.

Also classes specific to a component should be prefixed with the name of the component. That makes it easy to work out where a particular class would be defined. So this class name should be something like `.episode-guide__series-heading`

votemike: Done

PR #451, opened 2018-03-01
PROGRAMMES-6207 Cleaning up headings of TLEC page

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6207

I don't think this breaks anything....


programmes-developers:

BjornTwachtmann:
Changes look good. In terms of the translation, you should run script/translate-updateFromTemplate.sh after adding an item to the .pot file

PR #452, opened 2018-03-02
Fix

tatitati:
Promotions of core entities was left (they are so similar to clip cards.....). 
This fix was just tested by @christian-bell and worked properly.


programmes-developers:

PR #453, opened 2018-03-02
PROGRAMMES-6235 Guide Page Structured Data

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6235

Would appreciate @BPScott's approval before I merge this.

For:
episodes/guide
episodes/player
broadcasts/upcoming


src/Controller/ProgrammeEpisodes/GuidePartialController.php
```diff
@@ -2,20 +2,23 @@
...
 
+use App\Controller\Helpers\SchemaHelper;
 use App\Ds2013\PresenterFactory;
 use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeContainer;
 use BBC\ProgrammesPagesService\Service\CollapsedBroadcastsService;
 use BBC\ProgrammesPagesService\Service\ProgrammesService;
 use Symfony\Component\HttpFoundation\Request;
 
+//@TODO should this stop extending GuideController?
 class GuidePartialController extends GuideController
```
votemike: Does anyone know why this extends the GuideController?

programmes-developers:

votemike:
What do we want to do when a Seaons (Series) is the TELO? E.g. http://sandbox.bbc.co.uk/programmes/b006mgvt/episodes/guide Do we want The Weakest Link to be a Series (Brand) or a Season (Series)?

src/Controller/ProgrammeEpisodes/UpcomingController.php
```diff
@@ -32,11 +37,37 @@ public function __invoke(
...
+            $schemaContext = $schemaHelper->getSchemaForSeason($programme);
+        }
+
+        foreach($upcomingBroadcast as $ub)
+        {
+            $schemaContext['episode'][] = $schemaHelper->getSchemaForCollapsedBroadcast($ub, false);
+        }
+
+        return $schemaHelper->prepare($schemaContext);
+    }
```
tatitati: From my point of view if we put this behaviour in here then we will have to create end to end tests, harder to setup and slower, and working with json structure.

In programmes we use something named Humble Box Pattern (presenters-view). Is part of famous paper written by Robert C. Martin in the book "Clean Architecture" and used widely in many GUI architectures. But the idea behind is wider and is not limited to GUIs/Presenters. The idea behind is "separate what is easy to test from what is hard". That is the origin of our presenters: Presenters are easy to test, and views are hard to test (views are our humble object). 

This said I think that we should move this logic to "maybe" the SchemaHelper o any other place easier to instantiate and write some tests for it, this also would follow better the rule of having "slim" controllers.



https://github.com/bbc/programmes-frontend/pull/413/files

src/Controller/ProgrammeEpisodes/GuideController.php
```diff
@@ -73,4 +83,36 @@ protected function getUpcomingBroadcastsIndexedByProgrammePid(
...
+        if ($programme instanceof Brand) {
+            $schemaContext = $schemaHelper->getSchemaForSeries($programme);
+        } else {
+            /** @var Series $programme */
+            $schemaContext = $schemaHelper->getSchemaForSeason($programme);
+        }
+
+        foreach($children as $child)
+        {
+            if ($child instanceof Season) {
```
BPScott: Should this be Series?

IIRC some of the schema stuff has different names to our data model thanks to the americanism of calling series "seasons"

* Doctor Who = a "Brand" in our models, but a "Series" in americanised schema speak
* That thing that contains a subset of the episodes in doctor Who = a "Series" in our models but a "Season" in americanised schema speak

src/Controller/ProgrammeEpisodes/GuideController.php
```diff
@@ -73,4 +83,36 @@ protected function getUpcomingBroadcastsIndexedByProgrammePid(
...
+     * @param SchemaHelper $schemaHelper
+     * @param ProgrammeContainer $programme
+     * @param Episode[]|Series[] $children
+     * @param CollapsedBroadcast[] $upcomingBroadcast
+     * @return array
+     */
+    private function getSchema(SchemaHelper $schemaHelper, ProgrammeContainer $programme, array $children, array $upcomingBroadcast): array
+    {
+        if ($programme instanceof Brand) {
+            $schemaContext = $schemaHelper->getSchemaForSeries($programme);
```
BPScott: I'd be tempted to have these methods be named after our models, rather the americanised schema versions.

```php
 if ($programme instanceof Brand) {
    $schemaContext = $schemaHelper->getSchemaForBrand($programme);
} else {
    $schemaContext = $schemaHelper->getSchemaForSeries($programme);
}
```

feels a bit more intuitive compared to passing a Brand into `getSchemaForSeries` and passing a Series into `getSchemaForSeason`.

src/Controller/ProgrammeEpisodes/GuidePartialController.php
tatitati: Defined in GuideController we have:
https://github.com/bbc/programmes-frontend/blob/develop/src/Controller/ProgrammeEpisodes/GuideController.php#L63

And is used (obvious) in guideController as well:
https://github.com/bbc/programmes-frontend/blob/develop/src/Controller/ProgrammeEpisodes/GuideController.php#L52

But we saw that the partial produced some duplicated code because was needing the same functionality. So instead of duplicating the code we made GuidePartialController extends GuideController and use the function defined in the parent

https://github.com/bbc/programmes-frontend/blob/develop/src/Controller/ProgrammeEpisodes/GuidePartialController.php#L32

votemike: Do we want to move that method to a trait? Then we don't need the __invoke method signatures to match in each controller.

src/Controller/ProgrammeEpisodes/GuideController.php
```diff
@@ -73,4 +83,36 @@ protected function getUpcomingBroadcastsIndexedByProgrammePid(
...
+            } elseif (isset($upcomingBroadcast[(string)$child->getPid()])) {
+                $schemaContext['episode'][] = $schemaHelper->getSchemaForCollapsedBroadcast($upcomingBroadcast[(string)$child->getPid()], false);
+            } elseif ($child->isStreamable()) {
+                $schemaContext['episode'][] = $schemaHelper->getSchemaForOnDemand($child, false);
+            } else {
+                $schemaContext['episode'][] = $schemaHelper->getSchemaForEpisode($child, false);
+            }
+        }
+
+        return $schemaHelper->prepare($schemaContext);
```
tatitati: my only concern is about this function in the controller. I think if we add in the schema clips, later recipes, series, episodes, franchise, galleries, etc, this function is gonna grow up a lot and this functionality is gonna be hard to test because instantiating a controller is hard. 

Maybe we have tests (I saw them) for some functions of this helper, but the logic integrating them in here can become more complex with all those if's and keys. 

Keeping this logic in here is also creating duplication (i.e. TlecController):
https://github.com/bbc/programmes-frontend/blob/develop/src/Controller/FindByPid/TlecController.php#L317

To solve these problems we can move this logic to the helper. This would make tests easier to create, faster to run than a controller end-to-end test, also we would reduce the knowledge that every controller has about the schemeHelper, and we will keep slim controllers with no duplication of logic.

votemike: I understand what you mean about the testing. How do you envisage that working? I've had a think, but I think it will be a mess and that the controller is the best place for this.
I don't see any duplication in this controller with the TLEC controller with the SD stuff.

tatitati: ah true, cool then.

src/Controller/ProgrammeEpisodes/GuideController.php
Tasiobg:
```
if ($programme instanceof Brand) {
    $schemaContext = $schemaHelper->getSchemaForSeries($programme);
```
This is confusing, I think Ben suggestion could work better and may be a comment inside the function explaning that names are different could be helpful too

votemike: After having a look in to this... the [Schema] series may be one of our Brands *or* one of our Series. So I'm not sure it makes sense for us to rename the methods based on our naming conventions otherwise the logic inside the method is going to get screwy.
The one thing I can think of that _might_ work if we _do_ want to rename the methods is to only include structured data if the Series is a Brand, the Season is a Series and Episode is an Episode, if this is not the case, then just don't produce SD for the item.
What does everyone think?

src/Controller/ProgrammeEpisodes/GuidePartialController.php
```diff
@@ -2,22 +2,25 @@
...
+use App\Controller\BaseController;
+use App\Controller\Traits\IndexerTrait;
 use App\Ds2013\PresenterFactory;
 use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeContainer;
 use BBC\ProgrammesPagesService\Service\CollapsedBroadcastsService;
 use BBC\ProgrammesPagesService\Service\ProgrammesService;
 use Symfony\Component\HttpFoundation\Request;
 
-class GuidePartialController extends GuideController
+class GuidePartialController extends BaseController
```
BjornTwachtmann: Why change what this is extending and pull it out into a trait? You think it's needed elsewhere?

src/Controller/ProgrammeEpisodes/GuideController.php
BjornTwachtmann: It's probably best to keep them as they are then, even if it's confusing. Maybe add some comments as to the differences between our naming and schema.org's naming though?

src/Controller/ProgrammeEpisodes/GuidePartialController.php
votemike: If we keep the `GuidePartialController` extending the `GuideController` we need to make sure each controllers `__invoke()` signatures match. Which seems like more code and a hassle.
I can change it back if you like, but I prefer it this way.

src/Controller/ProgrammeEpisodes/GuideController.php
BPScott: > After having a look in to this... the [Schema] series may be one of our Brands or one of our Series

Oh I'm being dumb, it's not Brand/Series definitions (yay the same reasons why I called it the TLECController rather than a BrandController), its Top Level Containers (i.e. when getParent() returns null) is a schema "Series", otherwise it's a schema "Season".

votemike: I'm re-writing some of this in the sd-renaming branch, it will separate out the BBC terminology from the schema.org terminology.
That branch will also use the immediate parent as a "Season" and the TLEC as the "Series"

src/Controller/ProgrammeEpisodes/GuidePartialController.php
BjornTwachtmann: Didn't think of that, makes sense

src/Controller/ProgrammeEpisodes/GuideController.php
```diff
@@ -51,26 +60,41 @@ public function __invoke(
...
+    /**
+     * @param StructuredDataHelper $structuredDataHelper
+     * @param ProgrammeContainer $programmeContainer
+     * @param Episode[]|Series[] $children
+     * @param CollapsedBroadcast[] $upcomingBroadcast
+     * @return array
+     */
+    private function getSchema(StructuredDataHelper $structuredDataHelper, ProgrammeContainer $programmeContainer, array $children, array $upcomingBroadcast): array
+    {
+        $schemaContext = $structuredDataHelper->getSchemaForProgrammeContainer($programmeContainer);
```
BjornTwachtmann: The /guide page is also applicable on a series, right? Do we need a provision for data on the parent TLEO here if it exists?

src/Controller/ProgrammeEpisodes/PlayerController.php
```diff
@@ -50,12 +53,48 @@ public function __invoke(
...
+        }
+
+        if ($programmeContainer->isTlec()) {
+            return $structuredDataHelper->prepare($schemaContext);
+        }
+
+        // Add in Episodes and parents only
+        // Don't worry about child series
+        $schemaContext = $structuredDataHelper->getSchemaForProgrammeContainer($programmeContainer);
+        $parent = $programmeContainer->getParent();
```
BjornTwachtmann: The parent logic seems to be used in more than one place. Can this be abstracted? Or is it too specific?

programmes-developers:

tests/Controller/ProgrammeEpisodes/GuideControllerTest.php
```diff
@@ -29,7 +29,7 @@ public function testIndexControllerCanLoadGuideItems()
         $this->assertEquals(1, $crawler->filter('.footer')->count(), 'NON Partial-templates should include a footer');
         // assert guide items
         $this->assertEquals(3, $crawler->filter('.js-guideitem')->count());
-        $this->assertEquals(2, $crawler->filter('[typeof="Season"]')->count());
+        $this->assertEquals(2, $crawler->filter('.episode-guide__series-container')->count());
```
tatitati: sorry for this mate. That selector is mine and there was a better way to do it.

votemike:
Test this please

src/Controller/Helpers/StructuredDataHelper.php
```diff
@@ -74,8 +75,8 @@ public function getSchemaForProgrammeContainer(ProgrammeContainer $programmeCont
             return $this->schemaHelper->getSchemaForSeries($programmeContainer);
         }
 
-        // @Todo add schema.org markup for non TLEO containers
-        return [];
+        /** @var Series $programmeContainer */
```
BPScott: This typehint seems wrong, should be `@var ProgrammeContainer`, though I'm unsure why it is need at all as it should be implied by method signature. Series (or Brand) don't add any extra methods beyond those inherited from ProgrammeContainer so there's no point changing the hinting.

src/Controller/ProgrammeEpisodes/BaseProgrammeEpisodesController.php
```diff
@@ -27,4 +28,25 @@ protected function getSubNavPresenter(
...
+
+    protected function getSchemaForProgrammeContainerAndParents(StructuredDataHelper $structuredDataHelper, ProgrammeContainer $programmeContainer): array
+    {
+        $schemaContext = $structuredDataHelper->getSchemaForProgrammeContainer($programmeContainer);
+
+        if ($programmeContainer->isTlec()) {
+            return $schemaContext;
+        }
+
+        // Add in parents only
```
BPScott: Leveraging the logic in getAncestry() so we have a plain loop through an array instead of recursion might make the code more readable, as it doesn't look like we need to build a recursive data structure.


```php
$ancestry = array_slice($programmeContainer->getAncestry(), 1); // First item is the programme itself, we only want the parents
$tleo = array_pop($ancestry); // last item is the TLEO (pop removes this from the ancestry array too)

foreach ($ancestry as $ancestor) {
    $schemaContext['partOfSeason'] = $structuredDataHelper->getSchemaForProgrammeContainer($ancestor);
}
$schemaContext['partOfSeries'] = $structuredDataHelper->getSchemaForProgrammeContainer($tleo);

```

src/Controller/ProgrammeEpisodes/GuideController.php
```diff
@@ -51,26 +60,41 @@ public function __invoke(
...
-    protected function getUpcomingBroadcastsIndexedByProgrammePid(
-        ProgrammeContainer $programme,
-        CollapsedBroadcastsService $collapsedBroadcastsService
-    ): array {
-        $broadcasts = $collapsedBroadcastsService->findUpcomingByProgrammeWithFullServicesOfNetworksList($programme);
-        $upcoming = [];
+    /**
+     * @param StructuredDataHelper $structuredDataHelper
+     * @param ProgrammeContainer $programmeContainer
+     * @param Episode[]|Series[] $children
```
BPScott: Typehint should be `(Series|Episode)[]` as this is an array containing potentially both of those types at the same time, rather than an array of only Episodes or an array of only Series.

src/Controller/ProgrammeEpisodes/GuideController.php
```diff
@@ -51,26 +60,41 @@ public function __invoke(
...
 
-        foreach ($broadcasts as $broadcast) {
-            $upcoming[(string) $broadcast->getProgrammeItem()->getPid()] = $broadcast;
+        foreach ($children as $child) {
+            if ($child instanceof Series) {
+                $schemaContext['containsSeason'][] = $structuredDataHelper->getSchemaForProgrammeContainer($child);
+            } elseif (isset($upcomingBroadcast[(string) $child->getPid()])) {
+                $schemaContext['episode'][] = $structuredDataHelper->getSchemaForCollapsedBroadcast($upcomingBroadcast[(string) $child->getPid()]);
+            } elseif ($child->isStreamable()) {
+                $schemaContext['episode'][] = $structuredDataHelper->getSchemaForOnDemand($child);
```
BPScott: Why do we need two functions for building the schema for an Episode (`getSchemaForOnDemand` and `getSchemaForEpisode`)?

I think it would be more readable to have one method, that that adds the only-relevant-if-it-is-streamable data as required? Then you'd avoid having to do this check every time you want to add schema data for an episode.

I'd even be temped to go further ask why the public interface makes calling locations care about the type, if there was a `$structuredDataHelper->getSchemaForProgramme(Programme $prog)` then we might simplify that.

It seems like a builder pattern might be a useful one, then the logic for adding content to the base schema could be placed there instead of being repeated in every controller.

```php
$builder = new RdfaSchemaBuilder($programmeContainer);
$builder->addChildEntities($children);
$schemaContext = $builder->build();
$structuredDataHelper->prepare($schemaContext);
```

Though I'm spitballing here and don't really understand enough about how schemas get created and what the final formatting is like.

src/Controller/Helpers/StructuredDataHelper.php
votemike: `getSchemaForSeason` only accepts a `Series`. PHPStan was complaining that a `ProgrammeContainer` was being passed in.
Seeing as we are doing the tlec check above, that should mean all remaining ProgrammeContainers are `Series`.

src/Controller/ProgrammeEpisodes/GuideController.php
votemike: I think part of the problem with the logic in this particular method is that it was written before a refactor and then rebased. This method was not fixed up properly. I am about to commit a fixed version.
Getting the OnDemand info for an Episode isn't too difficult as the info can come from the Episode itself. The difficulty is with the Broadcasts for an Episode, which requires either a Broadcast or CollapsedBroadcast. That why I have left there separated out as they are.
I don't think a `getSchemaForProgramme()` would work as different information is required depending on what type of Programme it is, how it relates to other programmes on the page and how it relates to its ancestors.
I like your idea of the builder. I'm not sure that this PR is the place to implement it though. I could raise a new ticket to investigate the possibility if you like.

src/Controller/ProgrammeEpisodes/PlayerController.php
```diff
@@ -50,12 +53,33 @@ public function __invoke(
...
+     * @param ProgrammeContainer $programmeContainer
+     * @param Episode[] $availableEpisodes
+     * @return array
+     */
+    private function getSchema(StructuredDataHelper $structuredDataHelper, ProgrammeContainer $programmeContainer, array $availableEpisodes): array
+    {
+        $schemaContext = $this->getSchemaForProgrammeContainerAndParents($structuredDataHelper, $programmeContainer);
+
+        foreach ($availableEpisodes as $episode) {
+            $episodeSchema = $structuredDataHelper->getSchemaForEpisode($episode, false);
```
BPScott: Do you need the intermediate variable if it is only used once?

```
 $schemaContext['episode'][] = $structuredDataHelper->getSchemaForEpisode($episode, false);
```

src/Controller/ProgrammeEpisodes/UpcomingController.php
```diff
@@ -32,11 +36,33 @@ public function __invoke(
...
+     * @param ProgrammeContainer $programmeContainer
+     * @param CollapsedBroadcast[] $upcomingBroadcasts
+     * @return array
+     */
+    private function getSchema(StructuredDataHelper $structuredDataHelper, ProgrammeContainer $programmeContainer, array $upcomingBroadcasts): array
+    {
+        $schemaContext = $this->getSchemaForProgrammeContainerAndParents($structuredDataHelper, $programmeContainer);
+
+        foreach ($upcomingBroadcasts as $upcomingBroadcast) {
+            $episodeSchema = $structuredDataHelper->getSchemaForEpisode($upcomingBroadcast->getProgrammeItem(), false);
```
BPScott: Do you need the intermediate variable if it is only used once?

```
 $schemaContext['episode'][] = $structuredDataHelper->getSchemaForEpisode($episode, false);
```

src/Controller/ProgrammeEpisodes/GuideController.php
BPScott: > I don't think a getSchemaForProgramme() would work as different information is required depending on what type of Programme it is, how it relates to other programmes on the page and how it relates to its ancestors.

That additional information can be added by additional method calls to the builder, it wouldn't need to be one function with a boatload of arguments. The builder could let you pick and choose what is needed to be added. e.g. some pages have `$builder->addChildEntities($children);`, some pages would also have `$builder->addPublications($broadcasts);`, some would have both or neither.

> I like your idea of the builder. I'm not sure that this PR is the place to implement it though. 

Yep for sure :) a new investigation ticket sounds like the plan if you're up for it (I'm at the "eh do whatever" phase for some stuff given my imminent departure).

src/Controller/ProgrammeEpisodes/UpcomingController.php
votemike: I need to add the publication data on to the episode schema.

src/Controller/ProgrammeEpisodes/PlayerController.php
votemike: I need to add the publication data on to the episode schema.

src/Controller/ProgrammeEpisodes/GuideController.php
votemike: Created https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6288 Feel free to add more thoughts to the ticket if you like.

PR #454, opened 2018-03-02
Avoid superfluous queries on empty TLEC pages

BPScott:
* Don't make a query for clips if we know there aren't any
* Don't make a query for galleries if we know there aren't any


programmes-developers:

PR #455, opened 2018-03-04
Related links

BPScott:
Adding the related links module to the episode page.

---

I'm slightly torn on the 2nd commit - moving this into a partial rather than inlined into the page. This and several other modules (e.g. list of clips, list of galleries) are present on the episode, clip and series page in the same format so it makes sense to deduplicate them IMO. Do you agree?

However I'm not sure what the future state of the Series page is - in the old world before the TV Brand Page it was almost identical to the Series page (in v2 there was just 'container' that encompassed both page types), so recreating the old series page would probably be more work compared to basing it on the new Amen-based brand page. Would it be feasible to get a new design for that page?

### DIFF
--- a/templates/find_by_pid/episode.html.twig
+++ b/templates/find_by_pid/episode.html.twig
@@ -120,21 +120,10 @@
                 {# TODO Vote module #}
 
                 {% if relatedLinks %}
-                    <div class="component component--box component--box--striped component--box--secondary">
-                        <div class="component__header br-box-highlight">
-                            <h2>{{ tr('related_links') }}</h2>
-                        </div>
-                        <div class="component__body br-box-subtle">
-                            <ul class="no-margin-vertical list-spaced">
-                                {%- for link in relatedLinks -%}
-                                    <li>
-                                        <a href="{{ link.getUri() }}">{{ link.getTitle() }}</a>
-                                        {%- if link.isExternal() %} <span class="milli">({{ link.getHost() }})</span>{% endif -%}
-                                    </li>
-                                {%- endfor -%}
-                            </ul>
-                        </div>
-                    </div>
+                    {% include 'find_by_pid/_related_links_module.html.twig' with {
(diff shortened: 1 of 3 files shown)


programmes-developers:

templates/find_by_pid/episode.html.twig
```diff
@@ -119,7 +119,12 @@
 
                 {# TODO Vote module #}
 
-                {# TODO Related Links module #}
+                {% if relatedLinks %}
+                    {% include 'find_by_pid/_related_links_module.html.twig' with {
```
BPScott: Prefixing filenames with an `_` to denote that they're partials to be included, rather than a main template. A convention taken from sass filenames.

As the variable names are the same, this could be `{% include 'find_by_pid/_related_links_module.html.twig' %}`, however I think being explicit about what variables the partial requires is helpful

felipeparaujo: Not sure I like that convention. I'm a bigger fan of having that inside a `partial` folder, for the sake of being explicit and consistent. We already have a `partial` folder, but it might not be the right place for that template.

BPScott: The existing partial folder serves a different purpose - that contains templates that are rendered when you go to a specific URL. They're partials in the sense that they're transcluded into a page within the user's browser via an AJAX request. This related-links template (and things like the list of clips and galleries) however has no URL that you can go to to get just-this-content.

There's two schools of thought here:

* "group by type": where css lives with all the css, partials live with all the partials etc
* "group by component": where we group everything related to find_by_pid together, and everything related to schedules together etc.

I prefer the group by component approach as you tend to be working on "Find by pid stuff" rather than "template partials stuff" so having that as an expanded folder feels more useful. We could do `find_by_pid/partials/related_links_module.html.twig` but that extra depth doesn't feel like it adds anything useful.

PR #456, opened 2018-03-05
Bugfixes for clips list

BPScott:
* Use new style per-breakpoint class
* When there is only one clip make it full width to avoid whitespace
* When there is only one clip hide the component footer

---

FAO @votemike as I think he did the the original clip list work

Example page to demo a single clip: http://www.bbc.co.uk/programmes/b01jxym4

### DIFF
--- a/templates/find_by_pid/episode.html.twig
+++ b/templates/find_by_pid/episode.html.twig
@@ -47,9 +47,9 @@
                                 <h2>{{ tr('clips', programme.getAvailableClipsCount()) }}</h2>
                             </div>
                             <div class="component__body">
-                                <ul class="grid-wrapper highlight-box-wrapper highlight-box-wrapper--grid@bpb1 bpw-highlight-box-wrapper--grid">
+                                <ul class="grid-wrapper highlight-box-wrapper highlight-box-wrapper--grid@bpb1 highlight-box-wrapper--grid@bpw">
                                     {% for clip in clips %}
-                                        <li class="grid 1/2@bpb1 1/2@bpw">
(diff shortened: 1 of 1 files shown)


programmes-developers:

templates/find_by_pid/episode.html.twig
```diff
@@ -68,12 +68,14 @@
                                     {% endfor %}
                                 </ul>
                             </div>
-                            <div class="component__footer br-box-subtle">
-                                <a class="component__footer__link" href="{{ path('programme_clips', { pid: programme.getPid() }) }}">
-                                    <span class="component__footer__title">{{ tr('see_all_clips_from', {'%1': programme.getTitle()}) }}</span>
-                                    <span class="component__footer__detail">({{ programme.getAvailableClipsCount() }})</span>
-                                </a>
-                            </div>
+                            {% if programme.getAvailableClipsCount() > 1 %}
```
votemike: Should this be there all the time? Or when it's `> 4`?

BPScott: >1 according to the v2 code https://github.com/bbc/programmes/blob/72d3a73e453e8b1a94dfbfca9e2bab57bb76de2b/webapp/php/sharedmodules/programmes/views/partials/programmes/clips.spv#L44

votemike: Odd

BjornTwachtmann: It may be in the V2 code, is it correct though? It seems weird to me.

BPScott: It's the same for the gallery listing too. Gimmie a consensus on what this should do and I'll change it :)

PR #458, opened 2018-03-07
Move partials into more relevant namespaces

BPScott:
Stop conflating two concepts:

* Partials are parts of a page that get included in others with a Twig
  {% include %} tag, resulting in a single HTML document.
* Lazyloads are parts of a page that get transcluded into a page at load
  time via an AJAX request. The chunk of HTML that gets trancluded has
  its own URL, thus there are two HTML documents.

Both the recipe and on now concepts are lazyloads rather than partials.

I'm strongly in favour of grouping by responsibility, not by type. Thus I dislike the idea of having a "lazyloads" folder at the root of `src/Controllers` and `templates`.

---

I made these two follow the same convention as Recommendations for template naming, where the template name and the "page extension" (e.g. `.ameninc`) is separated by a dot.

 * `App\Controller\Partial\SchedulesOnNowController` is now ` App\Controller\Schedules\Lazyload\OnNowController `
* `templates/partial/on_now_*.html.twig` is now `templates/schedules/lazyload/on_now.*.html.twig`
*  `App\Controller\Partial\RecipesController` is now `App\Controller\RecipesController` as this controller can pull double duty for both `:pid/recipes` and `:pid/recipes.ameninc`:
* `templates/partial/recipes.html.twig` is now `templates/recipes/show.ameninc.html.twig`


programmes-developers:

BPScott:
As a preemptive avoiding of a "Hey Ben": The change to the Recipes routing does mean that `/programmes/:pid/recipes` is no longer forcibly redirected to http. But given that we've just made the findbypid pages and some of their transcluded routes, including https://www.bbc.co.uk/programmes/b013pqnm/recipes available over https, and are planning to make them HTTPS only imminently I think that's perfectly fine.

votemike:
LGTM

PR #459, opened 2018-03-07
Schema data structure for clips

tatitati:
[jira ticket](https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6234)

Next PRs will include Gallery and promotions information into the schema.


programmes-developers:

src/Controller/Helpers/SchemaHelper.php
```diff
@@ -105,6 +107,42 @@ public function prepare($schemaToPrepare, $isArrayOfContexts = false): array
         return $schemaToPrepare;
     }
 
+    /**
+     * @param Programme $programme
+     * @param Episode|null $onDemandEpisode
+     * @param CollapsedBroadcast|null $upcomingBroadcast
+     * @param Clip[] $clips
+     */
+    public function getSchemaForTleoController(
```
tatitati: this is the logic I put out of the controller to make it easier to test.

src/Controller/Helpers/ClipToSchemaTranslator.php
```diff
@@ -0,0 +1,110 @@
...
+        $clipSchema = [];
+        $clipSchema['@type'] = $clip->isRadio() ? 'RadioClip' : 'TVClip';
+        $clipSchema['identifier'] = (string) $clip->getPid();
+
+        if (!is_null($clip->getReleaseDate())) {
+            $clipSchema['datePublished'] = $clip->getReleaseDate()->asDateTime()->format(DATE_ATOM);
+        }
+
+        if (!is_null($clip->getPosition())) {
+            $clipSchema['position'] = $clip->getPosition();
```
votemike: Do clips have positions? And would clipNumber be a better thing to use in this case?

src/Controller/Helpers/ClipToSchemaTranslator.php
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        $clipSchema['url'] = $this->router->generate('find_by_pid', ['pid' => $clip->getPid()], UrlGeneratorInterface::ABSOLUTE_URL);
+        $clipSchema['duration'] = (string) new ChronosInterval(null, null, null, null, null, null, $clip->getDuration());
+        $clipSchema['image'] = $clip->getImage()->getUrl(480);
```
votemike: Do all clips have images?

src/Builders/GenreBuilder.php
```diff
@@ -0,0 +1,24 @@
+<?php
+declare(strict_types = 1);
+namespace App\Builders;
+
+use BBC\ProgrammesPagesService\Domain\Entity\Genre;
+use Faker\Factory;
+
+class GenreBuilder extends AbstractBuilder
```
tatitati: after some refactoring In the end is not used in this PR. but still is useful for the future

src/Controller/Helpers/ClipToSchemaTranslator.php
```diff
@@ -0,0 +1,110 @@
...
+    private function collectClipHierarchySchema(Clip $clip, array &$clipSchema)
+    {
+        $parent = $clip->getParent();
+
+        if (!$parent instanceof Episode && !$parent instanceof Series) {
+            return null;
+        }
+
+        if ($parent instanceof Episode) {
+            $clipSchema['partOfEpisode'] = [
```
votemike: This mostly duplicates the method `getSchemaForEpisode()`. Can you use that method instead?

src/Controller/Helpers/ClipToSchemaTranslator.php
```diff
@@ -0,0 +1,110 @@
...
+            $clipSchema['partOfSeries'] = [
+                '@type' => $parent->isRadio() ? 'RadioSeries' : 'TVSeries',
+                'name' => $parent->getTitle(),
+                'url' => $this->router->generate('find_by_pid', ['pid' => $parent->getPid()], UrlGeneratorInterface::ABSOLUTE_URL),
+            ];
+
+            $parent = $parent->getParent();
+        }
+
+        if ($parent instanceof Series) {
```
votemike: This is the same `if` statement as above.

src/Controller/Helpers/SchemaHelper.php
votemike: Are you planning on having a method for each controller type in this file?

src/Controller/Helpers/ClipToSchemaTranslator.php
```diff
@@ -0,0 +1,110 @@
...
+        $this->collectClipHierarchySchema($clip, $clipSchema);
+
+        return $clipSchema;
+    }
+
+    private function collectClipHierarchySchema(Clip $clip, array &$clipSchema)
+    {
+        $parent = $clip->getParent();
+
+        if (!$parent instanceof Episode && !$parent instanceof Series) {
```
votemike: Why do you not allow parents that are Brands?

src/Controller/Helpers/SchemaHelper.php
```diff
@@ -105,6 +107,42 @@ public function prepare($schemaToPrepare, $isArrayOfContexts = false): array
...
+            $episode = $this->getSchemaForCollapsedBroadcast($upcomingBroadcast);
+        }
+        if ($episode) {
+            $schemaContext['episode'] = $episode;
+        }
+
+        // clips
+        $clipSchema = $this->collectSchemaDataForClips($clips);
+        if (!empty($clipSchema)) {
+            $schemaContext['hasPart'] = $clipSchema;
```
votemike: What's "hasPart"?

src/Controller/Helpers/ClipToSchemaTranslator.php
tatitati: yeah, my way of thinking (maybe wrong) was something like:
```
// is $parent a series?, then:
             // add series schema
             $parent = $parent->getParent(); // we scale up to the next parent
// now $parent is a series, so we speak about a season series (series nesting series)
             // add season schema
```

As I said this might be wrong. I will check this process with you in my desk later.

src/Controller/Helpers/SchemaHelper.php
tatitati: Not sure. Maybe we just need one, but maybe we want to customize the schema. For example, for a recipe page I guess that we don't want season in the schema. But not sure, I might be wrong.

src/Controller/Helpers/ClipToSchemaTranslator.php
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        return $allClipsSchemas;
+    }
+
+    /**
+     * @return array
+     * @see SchemaHelperClipsTest::testAddMultipleClips()
+     */
+    private function buildSchema(Clip $clip) :array
```
votemike: Clips have titles don't they? Or something similar. I don't see any mention of that here.

src/Controller/Helpers/SchemaHelper.php
felipeparaujo: Also, shouldn't this be TlecController?

votemike: What about the SD for the Schedules page? And the Episode Guide pages? And the Episode page?

src/Controller/Helpers/SchemaHelper.php
tatitati: http://schema.org/hasPart

Also I saw an example of it being used.

src/Controller/Helpers/SchemaHelper.php
tatitati: True. thanks for reminding me

tests/Controller/Helpers/SchemaHelperClipsTest.php
```diff
@@ -0,0 +1,260 @@
...
+use App\Builders\EpisodeBuilder;
+use App\Builders\SeriesBuilder;
+use App\Controller\Helpers\ClipToSchemaTranslator;
+use App\Controller\Helpers\SchemaHelper;
+use BBC\ProgrammesPagesService\Domain\ValueObject\PartialDate;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+
+/**
+ * @group seo_schema
```
felipeparaujo: what does this do?

tests/Controller/Helpers/SchemaHelperClipsTest.php
```diff
@@ -0,0 +1,260 @@
...
+
+    public function testAddMultipleClips()
+    {
+        $schema = $this->createSchemaWhenClips([
+            ClipBuilder::any()->with(['parent' => SeriesBuilder::any()->build()])->build(),
+            ClipBuilder::any()->with(['parent' => SeriesBuilder::any()->build()])->build(),
+        ]);
+
+        $expectedSchemaForClip = [
+                0 => [
```
felipeparaujo: Indexing with integers is not really necessary, is it?

tests/Controller/Helpers/SchemaHelperClipsTest.php
```diff
@@ -0,0 +1,260 @@
...
+                'partOfSeries' => '[...]',
+            ],
+        ];
+
+        $this->assertPartialArray($expectedSchemaForClip, $schema);
+    }
+
+    /**
+     *
+     * [ Clip is PartOf a hierarchy ]
```
felipeparaujo: What does this comment mean exactly?

src/Controller/Helpers/ClipToSchemaTranslator.php
tatitati: added

tests/Controller/Helpers/SchemaHelperClipsTest.php
votemike: It allows you to run a set of tests:
https://phpunit.de/manual/current/en/appendixes.annotations.html#appendixes.annotations.group

src/Controller/Helpers/ClipToSchemaTranslator.php
tatitati: yes
Mandatory field of type Image in constructor and getImage() defined in CoreEntity.
https://github.com/bbc/programmes-pages-service/blob/master/src/Domain/Entity/CoreEntity.php#L141-L144

tests/Controller/Helpers/SchemaHelperClipsTest.php
tatitati: im doing my tickets with TDD, so I need to run my suite often. In this way you can create your group of tests and run only the one you wish like:

```vendor/bin/phpunit --group seo_schema``` 

which is covering all the seo schema generation code we have.

The other one is seo_schema_clip (you can add multiple groups), which cover only the code that produce the schema for clips. This allow me to test only the new stuff im creating while developing it. Instead, the seo_schema allows me to know if I'm breaking something of the schema that we already have and the integration of all the pieces.

File: tests/Controller/Helpers/SchemaHelperClipsTest.php (hunk 12 above)
Comment (tatitati) [reply tests/Controller/Helpers/SchemaHelperClipsTest.php]: is the subject of the group of test that come next. The next two, three tests are focus on hierarchy of clips. There is no an stardard way to do this, some people use tabs like a checklist, some use @groups again, this is just my way.

File: tests/Controller/Helpers/SchemaHelperClipsTest.php (hunk 11 above)
Comment (tatitati) [reply tests/Controller/Helpers/SchemaHelperClipsTest.php]: not at all. We just append our schema one by one. When the schema is converted to json this index dissapear completely.

File: tests/Controller/Helpers/SchemaHelperClipsTest.php (hunk 12 above)
Comment (tatitati) [reply tests/Controller/Helpers/SchemaHelperClipsTest.php]: Actuall i changed it, I think is a nice trick. Usually people write checklist when writing tdd on paper or in code. So changed the notation to looks more like a checklist, it makes the trick nicely I think:

image

File: src/Controller/Helpers/ClipToSchemaTranslator.php (hunk 5 above)
Comment (tatitati) [reply src/Controller/Helpers/ClipToSchemaTranslator.php]: yes indeed. I thought about it, but i think is not too clean. I tried a few times, so I would like to check with you the process. Maybe is easier after the refactoring we have in mind.

File: src/Controller/Helpers/ClipToSchemaTranslator.php (hunk 6 above)
Comment (BPScott) [reply src/Controller/Helpers/ClipToSchemaTranslator.php]: The number of sub series may be arbitrarily deep (e.g. https://www.bbc.co.uk/programmes/b06zyl30 is Brand -> Series -> Series -> Episode) You'll need to recurse up the full hierarchy rather than doing one parent level check.

File: src/Controller/Helpers/ClipToSchemaTranslator.php (hunk 3 above)
Comment (BPScott) [reply src/Controller/Helpers/ClipToSchemaTranslator.php]: @votemike getImage() always returns an Image as Francisco said, so we don't need to do any null checking. It *might be* that the image is the default "bbc blocks" image, if a clip (or any of its parents or the master brand it belongs to) has no image associated with it in PIPS. See [CoreEntityMapper->getImageModel()](https://github.com/bbc/programmes-pages-service/blob/0a3c19e87ff334022991f8a8d1491d5666f4d7e4/src/Mapper/ProgrammesDbToDomain/CoreEntityMapper.php#L399) for the gritty details.

File: src/Builders/ImageBuilder.php [hunk 13]
```diff
@@ -23,4 +23,11 @@ protected function __construct()
             'extension' => 'jpg',
         ];
     }
+
+    public static function anyWithPid($pid)
```
votemike: Would this be better on the base builder?

src/Controller/FindByPid/TlecController.php
```diff
@@ -314,10 +315,30 @@ private function setIstatsLabelsForTlec(
...
+        ProgrammeContainer $programme,
+        ?Episode $onDemandEpisode,
+        ?CollapsedBroadcast $upcomingBroadcast,
+        array $clips = []
+    ): array {
         $schemaContext = $structuredDataHelper->getSchemaForProgrammeContainer($programme);
 
+        $clipSchema = $structuredDataHelper->buildSchemaForClips($clips);
+        if (!is_null($clipSchema)) {
+            $schemaContext['hasPart'] = $clipSchema;
```
votemike: Is this correct? Should all clips go under 1 `hasPart`? Or should each clip be in it's own `hasPart`?

src/Controller/Helpers/SchemaHelper.php
```diff
@@ -136,4 +137,39 @@ public function getSchemaForSeason(Series $season): array
...
     }
+
+    /**
+     * @param Clip[] $clips
+     */
+    public function buildSchemaForClips(array $clips) :?array
+    {
+        $schemas = [];
+        foreach ($clips as $clip) {
+            if (!$clip instanceof Clip) {
```
votemike: When would this happen?

src/Controller/Helpers/SchemaHelper.php
```diff
@@ -136,4 +137,39 @@ public function getSchemaForSeason(Series $season): array
...
+            }
+
+            if (!is_null($clip->getPosition())) {
+                $clipSchema['clipNumber'] = $clip->getPosition();
+            }
+
+            $clipSchema['url'] = $this->router->generate('find_by_pid', ['pid' => $clip->getPid()], UrlGeneratorInterface::ABSOLUTE_URL);
+            $clipSchema['image'] = $clip->getImage()->getUrl(480);
+            $clipSchema['description'] = $clip->getShortSynopsis();
+
```
votemike: TOO MANY BLANK LINES!!!!!!

src/Controller/Helpers/StructuredDataHelper.php
```diff
@@ -78,6 +79,14 @@ public function getSchemaForProgrammeContainer(ProgrammeContainer $programmeCont
         return [];
     }
 
+    /**
+     * @param Clip[] $clips
+     */
+    public function buildSchemaForClips(array $clips) :?array
```
votemike: All the other methods in this class act upon a single object. Do you think it would be better to act upon a single Clip at a time?

src/Controller/Helpers/SchemaHelper.php
```diff
@@ -136,4 +137,39 @@ public function getSchemaForSeason(Series $season): array
             'url' => $this->router->generate('find_by_pid', ['pid' => $season->getPid()], UrlGeneratorInterface::ABSOLUTE_URL),
         ];
     }
+
+    /**
+     * @param Clip[] $clips
+     */
+    public function buildSchemaForClips(array $clips) :?array
```
BjornTwachtmann: All the other methods in this class provide schema.org data for a single item. This one operates on a list. That seems weird to me.

src/Controller/Helpers/StructuredDataHelper.php
BjornTwachtmann: Indeed, same as in the other class.

src/Controller/Helpers/SchemaHelper.php
tatitati: I just try to avoid an unproper use of this functionality from the client. We are just passing an array of things. We know that must be clips, but my mistake someone might pass an array of something that is different to a list of clips.

src/Controller/Helpers/StructuredDataHelper.php
tatitati: I thought about it, but in TlecController we get a list of clips. So for now I thought it was a good solution. If we need to create the schema for an individual clip then we can move this block easily to a function that handle only one clip. But for now sounds a little YAGNI ("you are not gonna need it")

tatitati: ok, I saw some comments similar below, I will move it to handle only one

src/Controller/Helpers/SchemaHelper.php
tatitati: changed

src/Controller/FindByPid/TlecController.php
tatitati: If you see testing tool with the resulting schema I think that they consider that each element in the list inside of hasPart is a hasPart item. You can see that each item has that key

src/Controller/Helpers/SchemaHelper.php
tatitati: changed

src/Builders/ImageBuilder.php
votemike: Or would it not just be easy enough to use the normal builder and add a PID?

src/Controller/Helpers/SchemaHelper.php
```diff
@@ -136,4 +137,26 @@ public function getSchemaForSeason(Series $season): array
...
+
+        if (!is_null($clip->getReleaseDate())) {
+            $clipSchema['datePublished'] = (string) $clip->getReleaseDate();
+        }
+
+        if (!is_null($clip->getPosition())) {
+            $clipSchema['clipNumber'] = $clip->getPosition();
+        }
+
+        $clipSchema['url'] = $this->router->generate('find_by_pid', ['pid' => $clip->getPid()], UrlGeneratorInterface::ABSOLUTE_URL);
```
votemike: Small point, but I would probably keep all the keys that are definitely going to exist next to each other in the code. After them, put the conditional ones.

tests/Controller/Helpers/SchemaHelperTest.php
```diff
@@ -151,4 +152,52 @@ public function testGetSchemaForSeason()
...
+
+        $schema = $this->helper->buildSchemaForClip($clip);
+
+        $this->assertEquals([
+            '@type' => 'RadioClip',
+            'identifier' => 'b00818rm',
+            'name' => 'a clip title',
+            'datePublished' => '2030-02-02',
+            'clipNumber' => 12,
+            'url' => null,
```
votemike: The URL shouldn't be null.

src/Controller/FindByPid/TlecController.php
```diff
@@ -314,10 +315,35 @@ private function setIstatsLabelsForTlec(
...
+        ProgrammeContainer $programme,
+        ?Episode $onDemandEpisode,
+        ?CollapsedBroadcast $upcomingBroadcast,
+        array $clips = []
+    ): array {
         $schemaContext = $structuredDataHelper->getSchemaForProgrammeContainer($programme);
 
+        // clips
+        $clipSchemas = [];
+        foreach ($clips as $clip) {
```
votemike:
```
foreach ($clips as $clip) {
    $schemaContext['hasPart'][] = $structuredDataHelper->buildSchemaForClip($clip);
}
```
???

tatitati: really nice. Yep

src/Controller/Helpers/SchemaHelper.php
```diff
@@ -136,4 +137,26 @@ public function getSchemaForSeason(Series $season): array
...
+        $clipSchema['@type'] = $clip->isRadio() ? 'RadioClip' : 'TVClip';
+        $clipSchema['identifier'] = (string) $clip->getPid();
+        $clipSchema['name'] = $clip->getTitle();
+
+        if (!is_null($clip->getReleaseDate())) {
+            $clipSchema['datePublished'] = (string) $clip->getReleaseDate();
+        }
+
+        if (!is_null($clip->getPosition())) {
+            $clipSchema['clipNumber'] = $clip->getPosition();
```
BjornTwachtmann: I wouldn't bother with this. It's not really a thing on clips that would make any sense.

PR #460, opened 2018-03-08
Episode page gallery list

BPScott:
This adds a Group Presenter and renders it on the episode page.

It's still a bit WIP as it needs some polish and tests.

The GroupPresenter currently uses ProgrammeTitlePresenter for rendering the titling. I should rename that to CoreEntityTitlePresenter.

Following on from pondering that update I spotted that how we organise the Group / Programme presenters in Ds2013 vs DsAmen is a bit different. @BjornTwachtmann have you got any thoughts on how you'd want to structure ProgrammePresenters vs GroupPresenters in Ds2013?


programmes-developers:

BjornTwachtmann:
@BPScott the group and programme presenters are all subclasses of baseCoreEntityPresenter in dsAmen. I'm not sure there's enough shared logic here to make that worthwhile, but I haven't looked into it closely. Depends how complete you think this Group presenter is I guess. Maybe the easiest thing for the moment is for the subPresenters to be shared. 

Off the top of my head, you could Create a folder structure that looks like...
`Presenters->Domain->CoreEntity->Programme->subPresenters
                                                         ->Group
                                                         ->SharedSubPresenters
`
and wang the title presenter in there. Dunno if that much else needs to change, but I'm open to suggestions.

BPScott:
I've added tests for the GroupTemplates.

As per a phone conversation with Bjorn, I'm off on holiday till a week on Tuesday so letting somebody else deal with the refactoring of the sub presenters object sounds like a useful thing. I won't get it working by the end of the day and having this linger for a week will be somewhat annoying.

Tasiobg:
Please do not review yet, I'm working on this PR feedback

Tasiobg:
I have refactored the programme subpresenters to be 'core entity subpresenters' but I didn't  manage to refator the tests yet, is the only remaning work. I'm going on holidays and I'll continue when I'm back.

src/Ds2013/Presenters/Domain/CoreEntity/Programme/BroadcastSubPresenters/BroadcastProgrammeTitlePresenter.php
```diff
@@ -43,6 +43,6 @@ public function getAriaTitle(): string
 
     public function getTemplateVariableName(): string
     {
-        return 'programme_title';
+        return 'core_entity_title';
```
Tasiobg: This is named `core_entity_title` instead of `programme_title` which would make more sense because in `broadcast_programme_title.html.twig` we extend `core_entity_title.html.twig` and we need the template variable to be same as the template we extend.
I have tried different options like moving the `aria_title` into the `core_entity_title.html.twig` but it doesn't look better on my point of view.
Please feel free to share you opinion about it

programmes-developers:

Tasiobg:
Ready for review

BPScott:
Looks reasonable to me, though I haven't delved too far into the Programme->CoreEntity refactor stuff, as Bjorn knows more about that than I.

PR #462, opened 2018-03-08
Extract clips and promotions listings into partials

BPScott:
These components shall be used within the clip page too, so extract them
into a reusable location in preparation.


programmes-developers:

BPScott:
Actually, no, this isn't helpful, ignore me.

PR #463, opened 2018-03-08
Remove inactive facebook page IDs

BPScott:
Fixes PROGRAMMES-6132


programmes-developers:

PR #464, opened 2018-03-09
Fix on-air link for Radio Cymru 2

BPScott:
Fixes PROGRAMMES-6155

---

The playspace player uses sids now, so update the variable names in the route to reflect that.


programmes-developers:

PR #465, opened 2018-03-12
Sd renaming

tatitati:


programmes-developers:

src/Controller/FindByPid/TlecController.php
```diff
@@ -314,23 +315,37 @@ private function setIstatsLabelsForTlec(
...
-            $episode = $schemaHelper->getSchemaForCollapsedBroadcast($upcomingBroadcast);
+        $schemaContext = $structuredDataHelper->getSchemaForProgrammeContainer($programme);
+
+        if (!$onDemandEpisode && !$upcomingBroadcast) {
+            return $structuredDataHelper->prepare($schemaContext);
         }
-        if ($episode) {
-            $schemaContext['episode'] = $episode;
+
+        if ($onDemandEpisode && $upcomingBroadcast && (string) $onDemandEpisode->getPid() !== (string) $upcomingBroadcast->getProgrammeItem()->getPid()) {
```
votemike: Would this `if` be better as a named variable or private method?

src/Controller/FindByPid/TlecController.php
```diff
@@ -314,23 +315,37 @@ private function setIstatsLabelsForTlec(
...
+        $publications = [];
+        if ($onDemandEpisode) {
+            $publications[] = $structuredDataHelper->getSchemaForOnDemand($onDemandEpisode);
+        }
+
+        if ($upcomingBroadcast) {
+            $publications[] = $structuredDataHelper->getSchemaForCollapsedBroadcast($upcomingBroadcast);
+        }
+
+        $schemaContext['episode']['publication'] = count($publications) === 1 ? reset($publications) : $publications;
```
votemike: We refactored this incorrectly. If there is an on-demand episode AND a upcoming broadcast, then there should be 2 episodes with 1 publication each. This code is not producing that result.

src/Controller/Helpers/SchemaHelper.php
```diff
@@ -105,44 +56,43 @@ public function prepare($schemaToPrepare, $isArrayOfContexts = false): array
         return $schemaToPrepare;
     }
 
-    private function getSchemaForEpisode(ProgrammeItem $programmeItem): array
+    public function getSchemaForEpisode(Episode $programmeItem): array
```
votemike: `$programmeItem` should be `$episode`

src/Controller/Helpers/StructuredDataHelper.php
```diff
@@ -0,0 +1,89 @@
...
+namespace App\Controller\Helpers;
+
+use BBC\ProgrammesPagesService\Domain\Entity\Broadcast;
+use BBC\ProgrammesPagesService\Domain\Entity\CollapsedBroadcast;
+use BBC\ProgrammesPagesService\Domain\Entity\Episode;
+use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeContainer;
+use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeItem;
+use BBC\ProgrammesPagesService\Domain\Entity\Service;
+use Cake\Chronos\ChronosInterval;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
```
votemike: Is this used?

src/Controller/Helpers/StructuredDataHelper.php
```diff
@@ -0,0 +1,89 @@
...
+                $episode['partOfSeries'] = $this->schemaHelper->getSchemaForSeries($parent->getTleo());
+                $episode['partOfSeason'] = $this->schemaHelper->getSchemaForSeason($parent);
+            }
+        }
+        return $episode;
+    }
+
+    public function getSchemaForProgrammeContainer(ProgrammeContainer $programmeContainer): array
+    {
+        if ($programmeContainer->isTlec()) {
```
votemike: else?

src/Controller/Schedules/ByDayController.php
```diff
@@ -4,6 +4,7 @@
 
 use App\Controller\BaseController;
 use App\Controller\Helpers\SchemaHelper;
```
votemike: Is this used any more?

src/Controller/Schedules/ByDayController.php
```diff
@@ -38,7 +39,7 @@ public function __invoke(
         CollapsedBroadcastsService $collapsedBroadcastsService,
         HelperFactory $helperFactory,
         UrlGeneratorInterface $router,
-        SchemaHelper $schemaHelper
+        StructuredDataHelper $structuredSchemaHelper
```
votemike: `$structuredSchemaHelper` probably shouldn't contain the word "schema" as we're trying to disconnect the schema.org domain language from ours. Could you change this to `$structuredDataHelper` like in the TLECController?

tests/Controller/Helpers/SchemaHelperTest.php
```diff
@@ -1,392 +0,0 @@
-<?php
```
votemike: This file should probably still contain _some_ tests shouldn't it?

tests/Controller/Helpers/StructuredDataHelperTest.php
```diff
@@ -0,0 +1,266 @@
...
+    use CustomAssertsTrait;
+
+    /**
+     *
+     * [x] - Basic braodcasts and collapsedBroadcasts has correct info
+     *
+     */
+    public function testOnDemandEpisodeIncludeCorrectFields()
+    {
+        $onDemanEpisode = EpisodeBuilder::any()->build();
```
votemike: `$onDemanEpisode` -> `$onDemandEpisode`

tests/Controller/Helpers/StructuredDataHelperTest.php
```diff
@@ -0,0 +1,266 @@
...
+
+        $schema = $this->helper()->getSchemaForOnDemand($onDemanEpisode);
+
+        $this->assertPartialArray([
```
votemike: This is testing the SchemaHelper. It should be testing the StructuredDataHelper. So this test should just ensure that calling `$this->helper()->getSchemaForOnDemand($onDemandEpisode);` calls `getSchemaForOnDemandEvent()` in the SchemaHelper shouldn't it?

tests/Controller/Helpers/StructuredDataHelperTest.php
```diff
@@ -0,0 +1,266 @@
...
+        ], $schema);
+    }
+
+    /**
+     * Helpers test
+     */
+    private function helper()
+    {
+        $router = $this->createMock(UrlGeneratorInterface::class);
+        $helper = new StructuredDataHelper(new SchemaHelper($router));
```
votemike: You will expect methods of the `SchemaHelper` to be called. So it should probably be a mock.

tests/Controller/Helpers/StructuredDataHelperTest.php
```diff
@@ -0,0 +1,266 @@
...
+            ])->build(),
+        ])->build();
+
+        $schema = $this->helper()->getSchemaForEpisode($episode, $includeParent = true);
+
+        $this->assertArrayHasKey('partOfSeries', $schema);
+        $this->assertArrayHasKey('partOfSeason', $schema);
+    }
+
+    public function testEpisodeSchemaCanAvoidToInjectParentsInformationIfIndicated()
```
votemike: Is this testing the case of a super nested Episode? I'm having a bit of trouble reading the test and I'm vaguely aware of what's going on. I imagine someone that isn't aware would have an even harder time.
Would you be able to refactor or add docblocks or turn this in to multiple tests in order to make it clearer what you are testing? This is one of the trickiest bits, so I think it's important to get right.

tests/Controller/Helpers/StructuredDataHelperTest.php
tatitati: is doing so. I only added tests for the StructuredDataHelper class thinking that would be better to think the one that integrates everything about the schema.

```php
private function helper()
{
        $router = $this->createMock(UrlGeneratorInterface::class);
        $helper = new StructuredDataHelper(new SchemaHelper($router));

        return $helper;
}
```

tests/Controller/Helpers/StructuredDataHelperTest.php
tatitati: well I choose the option of not mocking if really necessary. Is my personal choice. It allows me to focus on results instead of implementation details. This allows easier refactoring, reason why we broke all the schema tests after refactor it and without using mocks our tests are less couple to the production code, so in the end we have less brittle tests. The only think that make me a little sad about it is break a rule (the "dont mock what you dont own"), but it was impossible an elegant workaround for the URLGeneratorInterface, which is not a component we can control, but instead is part of Symfony.

tests/Controller/Helpers/StructuredDataHelperTest.php
tatitati: I aggree and I changed it. But I have also to recognise that part of the confusion is that this class is creating many different schemas for different type of objects, so is different to track the topic of each test when everything is mixed,  as result we have in the same test-class tests for schema Broadcasts, Series, Episodes, TLec......that creates confusion about the topic of each test.

src/Controller/Helpers/StructuredDataHelper.php
tatitati: phpstan having fun with that else, jeje

BjornTwachtmann: you still need to return something if it's not a tlec, this is typehinted to an array

tests/Controller/Helpers/CustomAssertsTrait.php
```diff
@@ -0,0 +1,42 @@
...
+trait CustomAssertsTrait
+{
+    /**
+     * This function allow to assert ony what you care about an array. Values and/or structure. But you can
+     * skip internal arrays by using '[...]', and non-important values with '...'
+     * This allow to focus better in what is being tested.
+     *
+     * @see ClipToSchemaTranslatorTest
+     */
+    private function assertPartialArray($expected, $b)
```
BjornTwachtmann: I really don't understand what this function is doing, but I think it's probably more readable to make assertions on an array within a test than to insert "...", which seems a bit magical.

tests/Controller/Helpers/SchemaHelperTest.php
tatitati: There is very little behaviour (to not say anything) on it and all the tests I wrote are covering these functions

All code in that class looks like a bunch of getters:
```php
public function getSchemaForEpisode(Episode $episode): array
    {
        return [
            '@type' => $episode->isRadio() ? 'RadioEpisode' : 'TVEpisode',
            'identifier' => $episode->getPid(),
            'episodeNumber' => $episode->getPosition(),
            'description' => $episode->getShortSynopsis(),
            ....
```

I can add a bunch of tests but for me they would duplicate what is being tested in the other class, the other class is the interesting one with rich behaviour.

tests/Controller/Helpers/StructuredDataHelperTest.php
votemike: Do you think that's the correct way to test this? It makes it more of an integration test than a unit test.

tests/Controller/Helpers/CustomAssertsTrait.php
tatitati: OK, I understand is confuse (and maybe even wrong), but I will explain my point at least:

My point when creating this function was to make more visible the structure of our schema array, sometimes with nested arrays in some fields.

```php
$a = [
    'key1' => 'value1',
    'key2' => [
          'key2.1' => 'value2.1',
          'key2.2' => 'value2.2',
];
```

I thought that if my interests was in the structure of this array, then going with ```assert_array_has_key``` would be less readable than just use something that really represent the real structure or section in the array I'm interested on. The magic '...' / [...] can be completely replaced (if we wish) for wathever we want. We might have something like:

```php
$a = [
    'key1' => '<string>',
    'key2' => [
          'key2.1' => '<int>',
          'key2.2' => '<any>,
          'key2.2' => '<array>,
];
```

Assert agains that kind of representation I think produce more readable tests as we understand better the structure of complex arrays when testing. IT WAS REALLY EFFECTIVE when developing the tests as it was more intuitive.

src/Controller/Helpers/StructuredDataHelper.php
tatitati: true!

tests/Controller/Helpers/CustomAssertsTrait.php
BjornTwachtmann: Why do we need to assert whether an array contains "any" or an integer or another array or whatever? This is a unit test, shouldn't we know exactly what it contains?

tatitati: Real example:
the function getSchemaForOnDemand($episode) return this array:

```php

'@type' => 'BroadcastEvent'
'startDate' => '2018-03-12T15:17:43+00:00'
'endDate' => '2018-03-12T15:17:43+00:00'
'publishedOn' => [
    '@type' => 'BroadcastService',
    'broadcaster' => [
        '@type' => 'Organization',
        'legalName' => 'British Broadcasting Corporation',
        'logo' => 'http://ichef.bbci.co.uk/images/ic/1200x675/p01tqv8z.png',
        'name' => 'BBC',
        'url' => 'https://www.bbc.co.uk/'
    ],
    'name' => 'Repellendus quod sunt ipsa'
]
```
I can assert that:
```php
 $schema['publishedOn']['broadcaster']['legalName'] = 'British Broadcasting Corporation'
```

but the true is that this assert says very little because doesn't provide any context. Instead I tried to do something like:
```php
$this->assertPartialArray([
        '@type' => '...',
        'legalName' => 'British Broadcasting Corporation',
        'logo' => '...',
        'name' => '...',
        'url' => '...'
], $schema['publishedOn']['broadcaster'])
```

Anyway I will delete it, I thought it would make more readable our tests.

src/Controller/FindByPid/TlecController.php
```diff
@@ -5,6 +5,7 @@
 
 use App\Controller\BaseController;
 use App\Controller\Helpers\SchemaHelper;
```
votemike: This is no longer used. Is it?

src/Controller/FindByPid/TlecController.php
```diff
@@ -314,23 +315,49 @@ private function setIstatsLabelsForTlec(
...
+            $episodeSchema['publication'] = $structuredDataHelper->getSchemaForCollapsedBroadcast($upcomingBroadcast);
         }
 
-        return $schemaHelper->prepare($schemaContext);
+        return $structuredDataHelper->prepare($schemaContext);
+    }
+
+    private function isSamePublication(Episode $onDemandEpisode, CollapsedBroadcast $upcomingBroadcast)
+    {
+        return $onDemandEpisode->getPid() == (string) $upcomingBroadcast->getProgrammeItem()->getPid();
```
votemike: Why is only one of these being cast to a string? This will compare a `Pid` against a `string` won't it?

tests/Controller/Helpers/StructuredDataHelperTest.php
```diff
@@ -0,0 +1,308 @@
...
+     *
+     * [x] - Schema for episode know when is part of a series or a season
+     *
+     */
+    public function testEpisodeWithTlecParentIsConsideredPartOfSeries()
+    {
+        $tlecSeriesFather = SeriesBuilder::any()->build();
+        $nestedEpisode = EpisodeBuilder::any()->with(['parent' => $tlecSeriesFather])->build();
+
+        $episodeSchema = $this->helper()->getSchemaForEpisode($nestedEpisode, $includeParent = true);
```
votemike: Do you need $includeParent? Can you not just pass true through?

src/Controller/Helpers/StructuredDataHelper.php
```diff
@@ -0,0 +1,88 @@
...
+    {
+        return $this->schemaHelper->getSchemaForOnDemandEvent($episode);
+    }
+
+    public function prepare($schemaToPrepare, $isArrayOfContexts = false): array
+    {
+        return $this->schemaHelper->prepare($schemaToPrepare, $isArrayOfContexts);
+    }
+
+    public function getSchemaForEpisode(Episode $programmeItem, bool $includeParent): array
```
votemike: Is it in this PR that we are going to deal with episode -> parent -> parent -> parent -> tlec? Or a different PR?

tatitati: that is your PR, episodes guide. This PR was only to introduce the concept of season. Also there is not recursion going through the hierarchy anywhere in this PR

src/Controller/Helpers/StructuredDataHelper.php
```diff
@@ -0,0 +1,93 @@
...
+        return $episode;
+    }
+
+    public function getSchemaForProgrammeContainer(ProgrammeContainer $programmeContainer): array
+    {
+        if ($programmeContainer->isTlec()) {
+            return $this->schemaHelper->getSchemaForSeries($programmeContainer);
+        }
+
+        return [];
```
votemike: Do we need a `@TODO` here?

src/Controller/Helpers/StructuredDataHelper.php
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    /**
+     * @param Service|Network $service
```
votemike: I _think_ this should only allow a `Service` (you should also typehint it). Whereas the `getSchemaForService()` method in the schemaHelper CAN accept a service or network.

tests/Controller/Helpers/SchemaHelperTest.php
```diff
@@ -1,392 +1,129 @@
...
-        if ($isRadio) {
-            $this->assertSame('RadioEpisode', $schemaContext['@type']);
-        } else {
-            $this->assertSame('TVEpisode', $schemaContext['@type']);
-        }
-
-        $this->assertInternalType('array', $schemaContext['publication']);
-        $this->assertBroadcastEvent($schemaContext['publication'], $serviceHasNetwork, false);
+        $series = SeriesBuilder::anyRadioSeries()->build();
+        $schema = $this->helper()->getSchemaForSeries($series);
```
votemike: Should you also have a test where you pass a Brand in?

tests/Controller/Helpers/SchemaHelperTest.php
```diff
@@ -1,392 +1,129 @@
...
-        }
-
-        $this->assertInternalType('array', $schemaContext['publication']);
-        $this->assertBroadcastEvent($schemaContext['publication'], $serviceHasNetwork, true);
+        $episode = EpisodeBuilder::anyRadioEpisode()->build();
+        $schema = $this->helper()->getSchemaForEpisode($episode);
+
+        $this->assertEquals([
+            '@type' => 'RadioEpisode',
+            'identifier' => $episode->getPid(),
```
votemike: This doesn't quite feel like a unit test to me. I think you should be testing a hardcoded string here rather than a method such as `getPid()`.
Would it be better if this test just asserted that that array contains the correct keys? And also test the the `@type` is "RadioEpisode"?
I see similar tests in this file which I also think could do with changing.

tests/Controller/Helpers/SchemaHelperTest.php
```diff
@@ -1,392 +1,129 @@
...
-        if ($multipleBroadcasts) {
-            $this->assertInternalType('array', $publishedOn);
-            foreach ($publishedOn as $pOn) {
-                $this->assertInternalType('array', $pOn);
-                $this->assertBroadcastService($pOn, $serviceHasNetwork);
-            }
-        } else {
-            $this->assertBroadcastService($publication['publishedOn'], $serviceHasNetwork);
-        }
+        return new SchemaHelper($router);
```
votemike: You could create this in a `__setup()` method couldn't you? To avoid making it new in each test?

tests/Controller/Helpers/StructuredDataHelperTest.php
```diff
@@ -0,0 +1,292 @@
...
+        $this->assertKeys([
+            '@type',
+            'startDate',
+            'endDate',
+            'publishedOn',
+        ], $collapsedSchema);
+
+        $this->assertEquals('BroadcastEvent', $collapsedSchema['@type']);
+        
+        $this->assertKeys([0], $collapsedSchema['publishedOn']); // guard assertion
```
votemike: What's this? What is guard assertion?

tests/Controller/Helpers/StructuredDataHelperTest.php
```diff
@@ -0,0 +1,292 @@
...
+            'description',
+            'identifier',
+            'name',
+            'url',
+        ], $containerSchema);
+
+        $this->assertEquals('RadioSeries', $containerSchema['@type']);
+    }
+
+    public function testEmptyArrayIsReturnedForContainerNotTlec()
```
votemike: I don't think this needs a test. It's unimplemented functionality. We don't want to break tests when we come to implement it.

tests/Controller/Helpers/StructuredDataHelperTest.php
tatitati: Is not a concept of mine, so you can google it

tests/Controller/Helpers/StructuredDataHelperTest.php
tatitati: it doesnt hurt I think. Is good to know that until that functionaility is implemented it will return an empty array.

tests/Controller/Helpers/StructuredDataHelperTest.php
tatitati: why?

src/Controller/Helpers/StructuredDataHelper.php
tatitati: yeah, it would be nice, any idea? (that next step is not something I'm involved, so if you have any idea to put in there I will be happy to do it)

tests/Controller/Helpers/StructuredDataHelperTest.php
votemike: Ahh. I see what you're doing. 👍 Would it be clearer if you did something like `$this->assertCount(1, $collapsedSchema['publishedOn']);`?
This way also makes the test stronger as it ensures there is 1 and only 1 element in publishedOn

tests/Controller/Helpers/StructuredDataHelperTest.php
votemike: Not a problem at all. It just seems odd.
Why not `$this->helper()->getSchemaForEpisode($nestedEpisode, true);`?

src/Controller/Helpers/StructuredDataHelper.php
votemike: I think something as simple as `// @TODO` will be good enough. Just so that we know that this is not the expected behaviour in the finished product.

src/Controller/Helpers/StructuredDataHelper.php
tatitati: If I put it for only services, then I have some end-to-end tests failing because they are passing networks.
I just follow the design is already implemented.

Error example:
```
Tests\App\Controller\Schedules\ByWeekControllerTest::testNextWeekGivesProperNextWeekLink with data set #2 ('/2011/w52', '/2012/w01')
TypeError: Argument 1 passed to App\Controller\Helpers\SchemaHelper::getSchemaForService() must be an instance of BBC\ProgrammesPagesService\Domain\Entity\Service, instance of BBC\ProgrammesPagesService\Domain\Entity\Network given
```

tests/Controller/Helpers/SchemaHelperTest.php
tatitati: No. Is a pitfall in testing: try to cover all the possibilities and all the paths with all the possible permutations and combinations just create an explosion of tests impossible to understand and impossible to finish of coding. 

At some point tests has to stop. This point happens when we feel confident with the code. Nothing can test 100% anything.

src/Controller/Helpers/StructuredDataHelper.php
tatitati: ```// @TODO``` IS REALLY NOTHING.

BjornTwachtmann: Should this function be called getSchemaForTlec or something instead?  This seems a little mesleading. Otherwise just "@TODO , add schema.org markup for non TLEO containers" is probably okay.

tests/Controller/Helpers/SchemaHelperTest.php
tatitati: Would it be better if this test just asserted that that array contains the correct keys?
-  Yes, the array structure must be tested, and is already tested

test the the @type is "RadioEpisode"?
- Is already tested

I think you should be testing a hardcoded string here rather than a method such as getPid():
- Ok, this can be part of a discussion. I took this option to avoid duplication of code in the test. The other alternative is this:

```php
$episode = EpisodeBuilder::anyRadioEpisode()->with([
   'name' => 'any name',
   'duration' => 'any duration',
])->build();
...
$this->assertEquals([
    'name' => 'any name',   // duplication in test code
    'duration' => 'any duration', // dupliaction in test code
    ...

I can save it also in a different variable, but my fixture would be harder to understand. So I still think this option is not a bad choice.
```

tests/Controller/Helpers/StructuredDataHelperTest.php
tatitati: I considered that option, BUT THEN what about the key?, it might be different to a simple number index. This option make it more explicit

tests/Controller/Helpers/StructuredDataHelperTest.php
tatitati: So you prefer this?
```
$this->helper()->getSchemaForEpisode($nestedEpisode, true)
```
Is a pitfall to use magic number, magic constants, magic booleans, etc in testing.

src/Controller/Helpers/StructuredDataHelper.php
tatitati: thanks @BjornTwachtmann

tests/Controller/Helpers/StructuredDataHelperTest.php
votemike: That's what we do in all our other tests isn't it?
If it's on purpose, I'm happy leaving it. But we should probably do that all the time rather than just sometimes.

tatitati: Cool

tests/Controller/Helpers/SchemaHelperTest.php
tatitati: > To avoid making it new in each test?
No. That concept is wrong. From PHPUNIT doc:
```Before a test method is run, a template method called setUp() is invoked```

This means phpunit creates a new one as I was doing so far (But instead of in _setUP in my custom function.)

File: src/Controller/Helpers/SchemaHelper.php [hunk 25]
```diff
@@ -105,44 +56,43 @@ public function prepare($schemaToPrepare, $isArrayOfContexts = false): array
...
+    {
+        $event = [
+            '@type' => 'OnDemandEvent',
+            'publishedOn' => [
+                '@type' => 'BroadcastService',
+                'broadcaster' => $this->getSchemaForOrganisation(),
+                'name' => 'iPlayer',
+            ],
+            'duration' => (string) new ChronosInterval(null, null, null, null, null, null, $episode->getDuration()),
+            'url' => $this->router->generate('iplayer_play', ['pid' => $episode->getPid()], UrlGeneratorInterface::ABSOLUTE_URL),
```
BjornTwachtmann: There seems to be an assumption here that these are always TV episodes? Am I reading this right? Do we need to correct that?

tatitati: Yes. The only uses of this function appear in tlecController.php
https://github.com/bbc/programmes-frontend/blob/develop/src/Controller/FindByPid/TlecController.php#L317

tests/Controller/Helpers/SchemaHelperTest.php
votemike: Ahh. My mistake. Any reason it's not in __setup() though?

tatitati: Actually no, I would say that I just thought that was not critical to left that function in its current place. 

I usually use fixtures-custom-functions because in there you can pass arguments and customize your fixtures, something that you cannot do with _setup(). After finishing I thought that was not creating either any problem that call in our test even if it can be moved to the _setup(). 

If I see that in future tests it gives problems of maintain I will do what is necessary (maybe add arguments, maybe move to the _setup(), ....)

src/Controller/Helpers/SchemaHelper.php
tatitati: It was investigated, and solved the issue working with @votemike

BjornTwachtmann: The on-demand URL for radio is... annoying. Will take this offline...

tatitati: From the discussion, ticket created:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6268

tests/Controller/Helpers/StructuredDataHelperTest.php
votemike: I think testing if there are 2 `publishedOn`s when we're only expecting 1 is more important than testing a auto-incrementing key.

tests/Controller/Helpers/StructuredDataHelperTest.php
felipeparaujo: I mean, that's not really magic, is it? You're just setting the value of the parameter. I don't see any aggregated value for setting that value to a variable and immediately using it. If I were reading this without any one to ask about, I would wonder what's going on here and wether this is getting used anywhere else.

felipeparaujo: https://betterexplained.com/articles/understanding-quakes-fast-inverse-square-root/ this would be a magic number, not true/false

tatitati: Then the poing would be that we prefer this option, right?
```php
$includeParent = true;
$this->helper()->getSchemaForEpisode($nestedEpisode, $includeParent);
```

Just this I think is a mistake:

```php
$this->helper()->getSchemaForEpisode($nestedEpisode, true);
```

felipeparaujo: Why do you think that's a mistake exactly? You just have to read the method signature to understand what it's doing

tatitati: but in testing you want to explain hot it works. You should avoid to make the people dig into the code and just show the real behaviour, the interesting thing.

felipeparaujo: This is not a coding standard. If anything, the oposite would be a coding standard - not assigning a value to a variable that will be used immediately and in only place. Reading the method signature is also one thing that's extremely easy given that we're all using IntelliJ.

I'm insisting on this issue because this exemplifies one of the problems we've having lately: introduction of concepts that get used once and then dropped. Doing this sort of thing makes our codebase inconsistent, and in the future someone will come along, read all of this stuff and wonder why this specific bit of code is different from everything else.

tatitati: Changed even if I disagree.

tests/Controller/Helpers/StructuredDataHelperTest.php
tatitati: EI @votemike, sorry, I didn't spot this comment. I understand your question.

A little ago, after trying and reading everything about testing I saw that there are two branches about how to TDD. One is focus on behaviour (using mock), the other one is called "state verification", also called "classical" (mocking just if is annoying depedency or this one is dangerous). 

Some people speak of two schools: The london one ("mockists"), and the american one ("classicals").

Both has Pros and Cons:

**- Mockist style:**
They are really really unit as you said. You change something that change a behaviour and only one test class will be broken. Easy to detect what broke it, fast and easy, and faster because they run in isolation.
They use mocks, mocks are based on expectations, so basically you code thinking of implementation expecting that some communications occurs. This is fine, but doesn't mean that this communications occurs in the correct order. (for example you can expect to call Doctrine->flush(), but you forgot to call openConnection() before), is just a class example, your tests still will pass. 
The problem you were listening to me these days was about coupling. This happens because you describe the protocol of communication, so in the end, your mocks (your tests) know too much about how a task is developed, this doesn't need to be too bad, but means that changes will be followed by many broken tests and for that reason any refactoring will be harder, even when the general behaviour is still the same your tests will break, they will be as we say "brittle".
People say that TDD start writing the client code, which describes how is gonna be used, some people refer to this process "interface discovering". Your first test describes what you need to implement in the next layer. But with mockist style this is quite different. They first step is not close at all to the client code, is much less descriptive becuase is using mocks, noisy. 
I think that mockist style is very powerful when you develop stuff that appear in different layers (services with no state calling to repositories, etc). Because of that the people call to this style "out-in" (you TDD from the UI to the Model).

-About classical style:
This style focuses on state, your interest is not about how a task is developed, instead you think about the result you want to have, because of that is less coupled to the code, it knows less about the code, and refactoring are easier. You can change details in your code but still they will tell you that the desired output is still the same. 
But they have a problem, if you break a class then you will break all the tests that are using that class as a collaborator instead of one. This is usually easy to solve, because there is always a root reason behind of all them. And this is the dilema: Stable tests that break in groups (classical), or brittle tests that break in isolation (mockist).
They are usually used in the oposite direction (in-out), going from the model to the UI because is easier to use existing components in this direction. But also can be done in Out-in as mockist, just hardcoding as you go all the responses you need from your real classes.
Classicals consider that only is necessary to mock an object if this is annoying, difficult, slow, or dangerous. In normal scenarios, it should be easy to create this collaborators as a real one.
They think that mockist style is quite confuse, because multiple mocks means that is difficult to understand what is the expectation of interest, making the test what some people call "obscure". 


After a while I realized that TDD for me was easier using classical approach, and the resulting tests were more useful as documentation, because there is no so much noise in the code as mocks create.  Is just more descriptive because let me focus on the result I want to have instead of how I'm gonna make it.

Sorry for the long text, but I think this is impossible to explain with something shorter. 

Some refs about it:
Developer testing book:
https://www.amazon.co.uk/Developer-Testing-Building-Addison-Wesley-Signature/dp/0134291069/ref=sr_1_1?ie=UTF8&qid=1521038811&sr=8-1&keywords=developer+testing

Growing Object-Oriented Software, Guided by Tests
https://www.amazon.co.uk/Growing-Object-Oriented-Software-Guided-Signature/dp/0321503627/ref=sr_1_10?s=books&ie=UTF8&qid=1521038831&sr=1-10&keywords=tdd

Interesting article from Martin Fowler (author of the book Domain Driven Design):
https://martinfowler.com/articles/mocksArentStubs.html

tests/Controller/Helpers/SchemaHelperTest.php
BjornTwachtmann: This makes more sense to me as part of setUp(). The reason being, it's the same in every test. 
If there were parameters that needed to change in the invocation between tests, I would make it a helper function as you have done. 

I do think we should try and be somewhat consistent, we use setUp mostly. If you've got ideas about how to structure unit tests better, we should discuss them as a group and implement them everywhere, not overload this PR.

tests/Controller/Helpers/StructuredDataHelperTest.php
BjornTwachtmann: I would tend to agree with Mike that this isn't testing behaviour we actually want, therefore doesn't belong in the test suite. This will just need to be changed when we implement the real functionality.

tests/Controller/Helpers/StructuredDataHelperTest.php
BjornTwachtmann: It depends if you want this to be a "pure" unit test or more of an integration test. As these two classes don't really have a purpose independent of each other, I'm fine with this being an integration test in this way really.

PR #466, opened 2018-03-13
Improved frontend performance

Tasiobg:
Bases on the frontend perforace traininig @votemike and my self did the other day we identify some improvements

Commits:
- Reorder headers to follow this optimal order: Char encoding, then Title tag, then inline scripts, then stylesheets, then inline style tags, then all JS, then meta tags (apart from viewport, charset and equiv)
- Added dns-prefetch

Those improvements decrase the amount of bloking times when loading a page, especially notizable by users who first visit our pages in a slow connection.


programmes-developers:

BjornTwachtmann:
Assuming stuff has just been moved around in the <head> I can't see this causing any major issues.

BPScott:
Sounds like we'd want to do this in the base_amen template too?

felipeparaujo:
As @BPScott said, should probably do that for amen pages. What sort of improvement are we seeing with these changes?

BjornTwachtmann:
@felipeparaujo @BPScott , we created a ticket for that, https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6267

PR #468, opened 2018-03-14
PROGRAMMES-6214 Episode Details Panel

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6214

I haven't added the Podcast download button yet. I think it would be best to do that in a separate PR for several reasons.


src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/PanelDetailsPresenter.php
```diff
@@ -3,21 +3,110 @@
...
+
+    /** @var Version[] */
+    private $versions;
+
+    public function __construct(PlayTranslationsHelper $playTranslationsHelper, Episode $episode, array $versions, bool $hasPreviousBroadcast)
     {
         parent::__construct();
+
         $this->episode = $episode;
+        $this->hasPreviousBroadcast = $hasPreviousBroadcast; //@TODO is this available from elsewhere?
```
votemike: Does anyone know if there's a better way to check if an episode has had a previous broadcast?

programmes-developers:

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/panel_details.html.twig
```diff
@@ -1,7 +1,149 @@
...
+
+        {% if presenter.getEpisode().getDuration() %}
+            <p class="episode-panel__meta">
+                {{ gelicon('core', 'duration', 'gelicon--centi gelicon--leading') }}{{ presenter.getDuration() }}
+            </p>
+        {% endif %}
+        {% if presenter.getEpisode().isStreamableAlternatate() %}
+            <p class="episode-panel__meta delta">
+                {% if presenter.hasAvailableSignedVersion() %}
+                    {#TODO Does this work with radio episodes?#}
```
votemike: I assume radio episodes don't have alternative versions? So this is OK? Right?

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/panel_details.html.twig
```diff
@@ -1,7 +1,149 @@
...
+                    <a class="text--no-ul" title="{{ tr('signed') }}" href="{{ path('iplayer_play', {'pid': presenter.getEpisode().getPid(), 'version': 'sign'}) }}">
+                        {#TODO not sure about how to size this GelIcon...#}
```
votemike: These icons look pants. I'm not 100% sure about the gelicon stuff. Does anyone else know what I should be putting here?

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/panel_details.html.twig
```diff
@@ -1,7 +1,149 @@
...
+
+{% macro intro(episode) %}
+    <div class="episode-panel__intro">
+        {% if episode.getLongestSynopsis() %}
+            <div class="map__intro__synopsis centi">{{ ds_shared('synopsis', episode.getSynopses(), 300) }}</div>
+        {% endif %}
+        {% if episode.isRadio() %}
+            <div class="favourites-button buttons--favourites-small"></div>
+        {% endif %}
+        {#{% if episode.getDownloadableVersion() %}#}
```
votemike: Will implement this is a new PR.

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/PanelDetailsPresenter.php
BjornTwachtmann: You can check $episode->getFirstBroadcastDate() . Just be aware that that can potentially be in the future, so you may have to do a comparision if that's significant.

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/panel_details.html.twig
BjornTwachtmann: The main issue here seems to be that the <path> is hardcoded to white in the gelicon. Not quite sure what to do about that.

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/panel_details.html.twig
BjornTwachtmann: They *shouldn't*, but this is PIPs data, so expect insanity. I think we should be explicit.

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/panel_details.html.twig
BjornTwachtmann: This is fixed in the latest version of the gelicons. Update bbc/gel-iconography-assets to v2.0 in composer.json. Do a composer update of it. Locally, you'll need to do rm -rf assets/gelicons before the composer install so your changes show up.

BjornTwachtmann: The changes to gel-iconography-assets will require a regression check on all icons we use. Please make sure to note that in the test notes for your ticket.

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/PanelDetailsPresenter.php
votemike: Would that be a nicer way to do it?
I think either would work. It depends on what people prefer.

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/panel_details.html.twig
votemike: Add an extra check next to `isStreamableAlternatate()` to make sure it's TV?

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/panel_details.html.twig
votemike: Looks a bit better

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/panel_details.html.twig
BjornTwachtmann: The thing is... It's technically valid for a streamable video episode with audio description to sit under a radio masterbrand. Weird, but valid. 

To me, this is more a presentation issue of how we're linking to playout for these things. Currently if this is a radio programme we're not linking to playout, so the logic can probably go here or in the presenter.

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/PanelDetailsPresenter.php
BjornTwachtmann: I think it's probably cleaner to look at the firstBroadcastDate myself. But I don't mind really.

votemike: @felipeparaujo @Tasiobg @tatitati Any thoughts?

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/panel_details.html.twig
BjornTwachtmann: Ignore the above. Yes, a check next to isStreamableAlternate makes perfect sense. 

/me goes to get new glassess

votemike: Is "Alternatate" a word?

BjornTwachtmann: Uhh. Alternatate is a perfectly cromulent spelling. That's why I deliberately put it in there... Shit.

votemike: Raise a GitHub issue and let open source sort it out?

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/PanelDetailsPresenter.php
felipeparaujo: using firstBroadcastDate would contain the logic in the presenter. I think that would be my favorite way of doing it, but as Bjorn, I don't mind

src/Controller/FindByPid/EpisodeController.php
```diff
@@ -46,23 +53,38 @@ public function __invoke(
...
+        $streamableVersions = [];
+        if ($episode->isStreamableAlternatate()) {
+            $streamableVersions = $versionsService->findStreamableByProgrammeItem($episode);
+        }
+
         // TODO check $episode->getPromotionsCount() once it is populated in
         // Faucet to potentially save on a DB query
         $promotions = $promotionsService->findActivePromotionsByEntityGroupedByType($episode);
 
         $episodeMapPresenter = new EpisodeMapPresenter(
```
BjornTwachtmann: I think it's probably better to use the PresenterFactory than pass all the deps into the controller. Any reason that would be tricky here?

src/Ds2013/PresenterFactory.php
```diff
@@ -63,11 +64,15 @@ class PresenterFactory
...
-    public function __construct(TranslateProvider $translateProvider, UrlGeneratorInterface $router, HelperFactory $helperFactory)
+    /** @var PlayTranslationsHelper */
+    private $playTranslationsHelper;
+
+    public function __construct(TranslateProvider $translateProvider, UrlGeneratorInterface $router, HelperFactory $helperFactory, PlayTranslationsHelper $playTranslationsHelper)
     {
         $this->translateProvider = $translateProvider;
         $this->router = $router;
         $this->helperFactory = $helperFactory;
+        $this->playTranslationsHelper = $playTranslationsHelper;
```
BjornTwachtmann: The HelperFactory should already provide this, no need to pass into the constructor.

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/PanelDetailsPresenter.php
```diff
@@ -3,21 +3,107 @@
...
+
+    public function hasPreviousBroadcast(): bool
+    {
+        return $this->episode->getFirstBroadcastDate() && $this->episode->getFirstBroadcastDate()->isPast();
+    }
+
+    public function getReleaseDate(): ?Date
+    {
+        if ($this->episode->getReleaseDate()) {
+            return new Date((string) $this->episode->getReleaseDate());
```
BjornTwachtmann: The PartialDate object returned by the getReleaseDate method already has a "asDateTime" method. Any reason not to use that here? Release dates can get a bit fruity with zeros for month/day which that already takes account of.

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/panel_details.html.twig
```diff
@@ -1,7 +1,90 @@
...
+                            {{ tr('episode_number', {'%1': panel_details.getEpisode().getPosition(), '%2': panel_details.getEpisode().getParent().getExpectedChildCount()}) }}
+                        </span>
+                    {% endif %}
+                {% endif %}
+            </div>
+        {% endif %}
+    </div>
+    <div class="grid-wrapper">
+        <div class="grid 2/3@bpw 3/4@bpe">
+            {{ self.intro(panel_details.getEpisode()) }}
```
BjornTwachtmann: Do these bits need to be re-usable? Or why are they macros?

votemike: They don't need to be reuse-able. They are macros purely for readability. Otherwise it's just a massive splat of HTML.

BjornTwachtmann: fair enough, makes sense

PR #469, opened 2018-03-14
override presenter path

Tasiobg:
At the moment if a presenter needs a differnet twig template we have to override `getTemplatePath()`. 
example https://github.com/bbc/programmes-frontend/blob/develop/src/Ds2013/Presenters/Domain/Programme/CollapsedBroadcastProgrammePresenter.php#L96

This way to set a path name is potentially dangeros becuase the hardcode value could get unsycronised with other changes we do, for this reason I have decided to implement a function to chose a different "Presenter Class" to generate the twig template path name, for the above example instead of using `getTemplatePath()` we could `$this->setTemplatePathClassOverride(ProgrammePresenter::class);` in the constructorl, this way presenters extending other presenter would be more easy to maintain


programmes-developers:

src/DsShared/BasePresenter.php
```diff
@@ -103,22 +115,24 @@ protected function validateOptions(array $options): void
      * the name of the variable to be passed to that template.
      * e.g. App\DsShared\Organism\ExampleThingPresenter becomes example_thing
      *
+     * @param string $class
+     * @return string
      */
-    protected static function snakeCasePresenterName(): string
+    private function snakeCasePresenterName(string $class): string
```
Tasiobg: Not sure why this was static, I think private makes more sense or at least now with those changes

tests/Ds2013/PresenterTest.php
```diff
@@ -67,4 +68,12 @@ public function testGetUniqueId()
         $secondPresenter = $this->getMockForAbstractClass(Presenter::class, [], 'TestDemoObjectPresenter');
         $this->assertNotEquals($initialId, $secondPresenter->getUniqueId());
     }
+
+    public function testTemplatePathClassOverride()
```
Tasiobg: Does this test make sense here? I think could be good to have it but not too sure if this place is the best option

src/DsShared/BasePresenter.php
```diff
@@ -21,6 +21,9 @@
     /** @var string */
     protected $uniqueId;
 
+    /** @var string */
```
votemike: `/** @var string|null */`

src/DsShared/BasePresenter.php
```diff
@@ -32,17 +35,20 @@ public function __construct(array $options = [])
      */
     public function getTemplatePath(): string
     {
-        if (!isset(self::$templatePathCache[static::class])) {
+        // set the class we want to use to generate the twig template path
+        $class = ($this->templatePathClassOverride) ? $this->templatePathClassOverride : static::class;
```
votemike: `$class = $this->getTemplateVariableName()`?

src/DsShared/BasePresenter.php
```diff
@@ -53,7 +59,8 @@ public function getTemplatePath(): string
      */
     public function getTemplateVariableName(): string
     {
-        return static::snakeCasePresenterName();
+        $class = ($this->templatePathClassOverride) ? $this->templatePathClassOverride : static::class;
```
votemike: Do you need the parentheses?

src/DsShared/BasePresenter.php
BjornTwachtmann: This is static because the presenter name is per class, not per instance, and there is a high overhead to running these calculations when creating 100s of presenters in a page.

src/DsShared/BasePresenter.php
```diff
@@ -84,6 +91,11 @@ public function getUniqueId(): string
         return $this->uniqueId;
     }
 
+    public function setTemplatePathClassOverride($class)
```
votemike: Typehint?

src/DsShared/BasePresenter.php
```diff
@@ -21,6 +21,9 @@
     /** @var string */
     protected $uniqueId;
 
+    /** @var string */
+    private $templatePathClassOverride;
```
votemike: Would you be able to make this protected? Or can you not do `protected $templatePathClassOverride = static::class`?

src/DsShared/BasePresenter.php
BjornTwachtmann: I don't really like this being alterable on a per-instance basis. It opens up a can of worms. Would this work with templatePathClassOverride as a constant on the child class?

src/DsShared/BasePresenter.php
Tasiobg: So `static::snakeCasePresenterName()` is faster than calling `$this->snakeCasePresenterName()` even when this function is only call from `getTemplatePath()` and `getTemplateVariableName()`? I don't understand why but I haven't benchmatch it, if you think is worth I'll move it back to static

BjornTwachtmann: The function itself probably doesn't have to be static, just the cache TBH.

src/DsShared/BasePresenter.php
felipeparaujo: That wouldn't work. We sometimes override that value to be something less long e.g. https://github.com/bbc/programmes-frontend/blob/develop/src/Ds2013/Presenters/Domain/Programme/CollapsedBroadcastSubPresenters/CollapsedBroadcastProgrammeBodyPresenter.php#L76. This way is safer.

src/DsShared/BasePresenter.php
```diff
@@ -21,6 +21,8 @@
     /** @var string */
     protected $uniqueId;
 
+    const TEMPLATE_PATH_CLASS_OVERRIDE = null;
```
BjornTwachtmann: Can you add a comment saying what this does and how and when to override it?

votemike: So this should be something like `TEMPLATE_PATH_CLASS_OVERRIDE = Mike::class`?

Tasiobg: A real working example would be to add this 
```
const TEMPLATE_PATH_CLASS_OVERRIDE = ProgrammePresenter::class;
```
to BroadcastProgrammePresenter

PR #470, opened 2018-03-15
Fixing SD after refactor

votemike:
The Schedule pages were no longer showing Episode information after the refactor. I have fixed this.

1 or 2 other small changes too.


programmes-developers:

src/Controller/Schedules/ByDayController.php
```diff
@@ -282,4 +278,21 @@ private function isValidDate(?string $date): bool
 
         return checkdate((int) $month, (int) $day, (int) $year);
     }
+
+    /**
+     * @param StructuredDataHelper $structuredDataHelper
+     * @param Broadcast[] $broadcasts
+     * @return array|null
+     */
+    private function getSchema(StructuredDataHelper $structuredDataHelper, array $broadcasts): ?array
```
felipeparaujo: Is this going to be used in more places? Should we extract this to a trait or something?

votemike: The methods in the controllers will probably all be named `getSchema()`. However, they will mainly be different. It just happens that the one for the ByDay and ByWeek are exactly the same.
http://wiki.c2.com/?ThreeStrikesAndYouRefactor

felipeparaujo: I see. Leave it as it is then.

tests/Controller/Helpers/StructuredDataHelperTest.php
```diff
@@ -269,8 +269,10 @@ private function getEpisodeNestedInMultipleSeries()
         return $episode;
     }
 
-    private function assertKeys($expectedKeys, $schema)
+    private function assertKeys(array $expectedKeys, array $schema)
     {
-        $this->assertEquals($expectedKeys, array_keys($schema));
+        foreach ($expectedKeys as $expectedKey) {
```
tatitati: I prefer not to have any kind of logic in tests as If, switch, foreach, etc. Is one of those thing that can bring more errors. The reason I have to think that is:
Any xUnit framework throw an exception when an assert/expectations fails (the framework then carry on with the next test). 
In this case the current failing test stop at the first assert failing, so the rest of asserts are not ran. If you put this in a loop then is difficult to know what assert failed exactly. 

As example I just tried this:
```php
        $episodeSchema = $this->helper->getSchemaForEpisode($nestedEpisode, false);

        $this->assertArrayHasKey('aaaaaaaaaaaa', $episodeSchema); // 'aaaaaaaaaaaa' doesnt exist
        $this->assertArrayHasKey('bbbbbbbbbbbb', $episodeSchema); // 'bbbbbbbbbbbb' doesnt exist
```
The resulting output was:
```
1) Tests\App\Controller\Helpers\StructuredDataHelperTest::testEpisodeSchemaCanAvoidToInjectParentsInformationIfIndicated
Failed asserting that an array has the key 'aaaaaaaaaaaa'.

....tests/Controller/Helpers/StructuredDataHelperTest.php:240
```

We don't get info about the second assert.

votemike: I changed the old way was also checking the order of the keys. Which we don't care about. This way purely checks that the keys exist in the array.
Can you think of a better way to improve the check?

tatitati: is ok for me

PR #471, opened 2018-03-15
Sandbox - running tests with no coverage

tatitati:


programmes-developers:

BPScott:
Yo @tatitati, I'm not convinced this is the best way to do this. It means you've got to edit a tracked file when you want to test with coverage which means there's a small chance of people accidentally reverting this.

Until now I've been using an approach that leverages per-user phpunit.xml files and it's worked really well for me. I've copied phpunit.xml.dist into phpunit.xml then commented out the `<logging>` section. As `phpunit.xml` is in the gitignore  list then there's no chance of accidentally committing it. The down side is you might occassionally need to update your phpunit.xml to reflect new changes in phpunit.xml.dist but those changes shall be few and far between.

BjornTwachtmann:
How often do we want test coverage reports on the sandbox though? A parameter to the script to enable them might be better.

BPScott:
It's pretty rare but I find it handy occasionally if I'm adding new bits and what what I'm missing in my tests.

Parameter on `script/test` could work, but writing option parsing stuff in bash is uuuugh.

PR #472, opened 2018-03-15
add google-site-verification for /schedules

Tasiobg:
In order to be able to use google search console we need the google-site-verification, we already do that for `/programmes` so this PR is to add the same to` /schedules`
Adding the meta tag into the parent URL is enough


programmes-developers:

templates/schedules/home.html.twig
```diff
@@ -3,6 +3,9 @@
 
 {% block title %}BBC - {{tr('schedules')}}{% endblock %}
 
+{% block inline_head %}
+    <meta name="google-site-verification" content="hV6KPZarfCnRMLwC5CGKEw8-rjA4WGSanUzaGvGkf7s" />
```
felipeparaujo: where did the content value come from?

BjornTwachtmann:
Do we need this only on schedule pages? What about the /programmes URLs?

programmes-developers:

votemike:
Because the only correct way to code is directly on the server.... would it be better to just SSH in to LIVE. Verify the site, then remove line?
If you do want to go the PR route... should this not be aimed at master so that we can release it faster?

templates/schedules/home.html.twig
BPScott: It's the same that is on the [/programmes homepage](https://github.com/bbc/programmes-frontend/blob/b256dc6cc61a5897d1a91b518a82194b52bd7606/templates/home/show.html.twig#L6).

BPScott:
@votemike, i /think/ this needs to hang around rather than be a one-off verification.

Tasiobg:
Google site verification value is provided by google and seems to be the same value for the whole programmes team account.
The verification should stay in the HTML, is not a one-off verification.
We already have this for /programmes https://github.com/bbc/programmes-frontend/blob/develop/templates/home/show.html.twig#L6
but because /schedules is different path it requires its own one

The plan was to get this out in the next release, if we are not going to merge develop into master again I would consider to merge this branch into master.

PR #476, opened 2018-03-16
TX panel

tatitati:


programmes-developers:

src/Builders/CollapsedBroadcastBuilder.php
```diff
@@ -42,4 +44,35 @@ public static function anyOfEpisode()
             'programmeItem' => EpisodeBuilder::any()->build(),
         ]);
     }
+
+    public static function anyLive()
+    {
+        $self = new self();
```
votemike: Should this not be self::any()

src/Ds2013/Presenters/Domain/BroadcastEvent/BroadcastEventPresenter.php
```diff
@@ -86,6 +96,30 @@ public function getOnAirMessage(): string
...
+        return null;
+    }
+
+    public function getTitle() :?string
+    {
+        if ($this->collapsedBroadcast->isOnAir()) {
+            return $this->collapsedBroadcast->getProgrammeItem()->isRadio() ? 'on_air' : 'on_now';
+        }
+
+        if ($this->collapsedBroadcast->getEndAt() < ApplicationTime::getTime()) {
```
votemike: Can you do something like `$this->collapsedBroadcast->getEndAt()->isPast()`?

src/Ds2013/Presenters/Domain/BroadcastEvent/broadcast_event.html.twig
```diff
@@ -1,3 +1,7 @@
+{%- if broadcast_event.getOption('show_logo') -%}
+    <h2>{{ tr(broadcast_event.getTitle()) }}</h2>
```
votemike: Should this definitely be an H2? Could you check the page with that Chrome add-on.

src/Ds2013/Presenters/Domain/BroadcastEvent/broadcast_event.html.twig
```diff
@@ -43,7 +47,17 @@
         </div>
     </div>
 </div>
-{% if (broadcast_event.isOnAir()) %}
+
+{%- if broadcast_event.getOption('show_watch_from_start') and broadcast_event.rewindUrl() -%}
+    <p class="programme__live hidden visible@bpw">
```
votemike: What's this hidden class?

tests/Ds2013/Presenters/Domain/BroadcastEvent/BroadcastEventPresenterTest.php
```diff
@@ -58,4 +66,151 @@ public function getNetworkUrlProvider(): array
             'null returns empty url' => [null, ''],
         ];
     }
+    /**
+     * [x] - Broadcast titles are based on when is being broadcasted and the type of programmeType
+     *
+     * @dataProvider outputsBasedOnEpisodeTypeForBroadcastsOnNow
+     */
+    public function testOnNowForTvEpisode($whenBroadcastOn, $withEpisode, $expectedPanelTitle)
```
votemike: Typehints?

src/Ds2013/Presenters/Domain/BroadcastEvent/BroadcastEventPresenter.php
```diff
@@ -86,6 +96,30 @@ public function getOnAirMessage(): string
         return $this->collapsedBroadcast->getProgrammeItem()->isRadio() ? 'on_air' : 'on_now';
     }
 
+    public function getRewindUrl(): ?string
+    {
+        if ($this->collapsedBroadcast->getProgrammeItem()->isTv() && $this->liveBroadcastHelper->isWatchableLive($this->collapsedBroadcast)) {
+            $url = $this->liveBroadcastHelper->simulcastUrl($this->collapsedBroadcast);
+            $url .= ((strpos($url, '?') !== false) ? '&' : '?') . 'rewindTo=current';
```
felipeparaujo: https://github.com/bbc/programmes-frontend/blob/develop/src/DsAmen/Presenters/Domain/CoreEntity/CollapsedBroadcast/SubPresenter/LiveCtaPresenter.php#L92-L102 This is how you should be building the query string

src/Ds2013/Presenters/Domain/BroadcastEvent/BroadcastEventPresenter.php
```diff
@@ -86,6 +96,30 @@ public function getOnAirMessage(): string
         return $this->collapsedBroadcast->getProgrammeItem()->isRadio() ? 'on_air' : 'on_now';
     }
 
+    public function getRewindUrl(): ?string
```
felipeparaujo: What happens to episode guide pages? Are they also showing the rewind button?

src/Ds2013/Presenters/Section/Episode/Map/EpisodeMapPresenter.php
```diff
@@ -76,7 +74,7 @@ private function buildSidePanelsSubPresenters() :array
     {
         $sidePanels = [];
         if ($this->mustDisplayTxPanel()) {
-            $sidePanels[] = new PanelTxPresenter($this->episode, $this->upcomingBroadcast, $this->lastOnBroadcast);
+            $sidePanels[] = new PanelTxPresenter($this->upcomingBroadcast ?: $this->lastOnBroadcast);
```
felipeparaujo: This should be `??` instead of `?:`

tests/Ds2013/Presenters/Domain/BroadcastEvent/BroadcastEventPresenterTest.php
```diff
@@ -58,4 +66,151 @@ public function getNetworkUrlProvider(): array
...
+     *
+     */
+    public function testWeCanCreateWatcheableBroadcasts()
+    {
+        $dummy = $this->createMock(UrlGeneratorInterface::class);
+        $helper = new LiveBroadcastHelper($dummy);
+
+        $cBroadcast = $this->buildBroadcastsWith('a TV episode', ['service1' => 'bbc_two_wales_digital', 'service2' => 'bbc_one_scotland']);
+
+        $this->assertTrue($helper->isWatchableLive($cBroadcast), 'is should be watchable');
```
votemike: Are we going to start adding custom messages for assertions? I don't think we do currently. Is everyone happy with that?
If we are going to start doing that, could we make the message descriptive? This one doesn't seem that useful to me. Something along the lines of "A CollapsedBroadcast should be watchable live when x, y and z."

votemike: In fact, that should probably be the name of the test instead of the custom message on the assertion.

tests/Ds2013/Presenters/Domain/BroadcastEvent/BroadcastEventPresenterTest.php
```diff
@@ -58,4 +66,151 @@ public function getNetworkUrlProvider(): array
...
+        }
+
+        $dummy = $this->createMock(UrlGeneratorInterface::class);
+        return $this->buildPresenterForBroadcast($broadcast, $dummy);
+    }
+
+    /**
+     * Build watchable broadcast to exercise simulcast feature.
+     */
+    private function buildBroadcastsWith(string $episodeType, array $sids)
```
votemike: `buildBroadcastsWith ` => `buildLiveBroadcastsWith`?

tests/Ds2013/Presenters/Domain/BroadcastEvent/BroadcastEventPresenterTest.php
tatitati: inded, to be honest I wrote a 💩  message in there, I will try to put something with more value

src/Ds2013/Presenters/Section/Episode/Map/EpisodeMapPresenter.php
BjornTwachtmann: I'm not sure I'm a big fan of conflating the upcoming with the last on broadcast here. Could we pass both in and make the decision as to which to use in the TxPresenter?

src/Ds2013/Presenters/Domain/BroadcastEvent/broadcast_event.html.twig
```diff
@@ -43,7 +47,17 @@
         </div>
     </div>
 </div>
-{% if (broadcast_event.isOnAir()) %}
+
+{%- if broadcast_event.getOption('show_watch_from_start') and broadcast_event.rewindUrl() -%}
```
BjornTwachtmann: broadcast_event.getRewindUrl(), I know twig will let you use this but it slows things down

src/Ds2013/Presenters/Domain/BroadcastEvent/broadcast_event.html.twig
BjornTwachtmann: it's only visible above bpw (600px I think). TBH I don't know why we have that, it should work fine at much smaller sizes

src/Ds2013/Presenters/Section/Episode/Map/EpisodeMapPresenter.php
felipeparaujo: Would probably be better, [it's what I'm doing in the PlayoutPresenter](https://github.com/bbc/programmes-frontend/blob/playout/src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/PlayoutPresenter.php#L39)

src/Ds2013/Presenters/Domain/BroadcastEvent/BroadcastEventPresenter.php
tatitati: Just checked, no.

src/Ds2013/Presenters/Domain/BroadcastEvent/broadcast_event.html.twig
tatitati: I confirm that is the correct one. Thanks for the idea

src/Ds2013/Presenters/Domain/BroadcastEvent/broadcast_event.html.twig
felipeparaujo: Missed this one, I think @tatitati

src/Ds2013/Presenters/Section/Episode/Map/EpisodeMapPresenter.php
felipeparaujo: missed this one as well, it looks like @tatitati

src/Ds2013/Presenters/Section/Episode/Map/Panels/Side/PanelTxPresenter.php
```diff
@@ -4,24 +4,25 @@
...
-    /** @var Episode */
-    protected $episode;
+    /** @var CollapsedBroadcast */
+    private $collapsedBroadcast;
 
-    /** @var CollapsedBroadcast|null */
-    protected $upcomingBroadcast;
+    public function __construct(?CollapsedBroadcast $upcomming, ?CollapsedBroadcast $lastOn)
+    {
+        if (!$upcomming && !$lastOn) {
```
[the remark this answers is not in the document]
BjornTwachtmann: It depends. As long as you only try and create the TX panel presenter when one of these exists (is that what you're doing?) it's okay. You just have to be careful.

src/Ds2013/Presenters/Domain/BroadcastEvent/BroadcastEventPresenter.php
BjornTwachtmann: I'm amazed you managed to add the "show_watch_from_start" parameter to prevent exactly this case without thinking about it mate ;-)

src/Ds2013/Presenters/Domain/BroadcastEvent/BroadcastEventPresenter.php
```diff
@@ -86,6 +96,30 @@ public function getOnAirMessage(): string
...
+    {
+        if ($this->collapsedBroadcast->getProgrammeItem()->isTv() && $this->liveBroadcastHelper->isWatchableLive($this->collapsedBroadcast)) {
+            $url = $this->liveBroadcastHelper->simulcastUrl($this->collapsedBroadcast, null, ['rewindTo' => 'current']);
```
BjornTwachtmann: Petty, but do you need a variable here, just return seems enough

src/Controller/FindByPid/EpisodeController.php
```diff
@@ -25,7 +26,8 @@ public function __invoke(
         CollapsedBroadcastsService $collapsedBroadcastsService,
         FavouritesButtonService $favouritesButtonService,
         VersionsService $versionsService,
-        PresenterFactory $presenterFactory
+        PresenterFactory $presenterFactory,
+        LiveBroadcastHelper $liveBroadcastHelper
```
votemike: Is this used?

src/Ds2013/Presenters/Domain/BroadcastEvent/BroadcastEventPresenter.php
```diff
@@ -86,6 +95,28 @@ public function getOnAirMessage(): string
...
+    public function getRewindUrl(): ?string
+    {
+        if ($this->collapsedBroadcast->getProgrammeItem()->isTv() && $this->liveBroadcastHelper->isWatchableLive($this->collapsedBroadcast)) {
+            return $this->liveBroadcastHelper->simulcastUrl($this->collapsedBroadcast, null, ['rewindTo' => 'current']);
+        }
+
+        return null;
+    }
+
+    public function getTitle() :?string
```
votemike: Where is this displayed?

tests/Ds2013/Presenters/Domain/BroadcastEvent/BroadcastEventPresenterTest.php
```diff
@@ -58,4 +66,151 @@ public function getNetworkUrlProvider(): array
             'null returns empty url' => [null, ''],
         ];
     }
+    /**
+     * [ Title ]. They are based on when is being broadcasted and the type of programmeType
+     *
+     * @dataProvider outputsBasedOnEpisodeTypeForBroadcastsOnNow
+     */
+    public function testOnNowForTvEpisode(string $whenBroadcastOn, Episode $withEpisode, string $expectedPanelTitle)
```
votemike: This test is called `testOnNowForTvEpisode()`, but the data provider is supplying radio episodes sometimes.

tests/Ds2013/Presenters/Domain/BroadcastEvent/BroadcastEventPresenterTest.php
```diff
@@ -58,4 +66,151 @@ public function getNetworkUrlProvider(): array
...
+        ];
+    }
+
+    /**
+     *
+     * [ watch from the start ]. Understand when is watchable a cBroadcast
+     *
+     * @dataProvider servicesTypeProvider
+     */
+    public function testWeCanCreateWatcheableBroadcasts(string $typeServices, bool $expectedIsWatcheable)
```
votemike: `testWeCanCreateWatcheableBroadcasts()` -> `testWeCanDisplayWatcheableBroadcasts()`?

votemike: In what way is this testing the presenter?

tests/Ds2013/Presenters/Domain/BroadcastEvent/BroadcastEventPresenterTest.php
```diff
@@ -58,4 +66,151 @@ public function getNetworkUrlProvider(): array
...
+    }
+
+    /**
+     * test helpers
+     */
+    private function buildBroadcastOnTime(string $when, Episode $episode): BroadcastEventPresenter
+    {
+        $broadcast = null;
+        switch ($when) {
+            case 'broadcast on past':
```
votemike: Not keen on these magic strings. Is there some way you can change it to not be magic?

src/Ds2013/Presenters/Domain/BroadcastEvent/BroadcastEventPresenter.php
tatitati: you can find it in this PR, here:
src/Ds2013/Presenters/Domain/BroadcastEvent/broadcast_event.html.twig

src/Ds2013/Presenters/Domain/BroadcastEvent/broadcast_event.html.twig
votemike: Why is the option to display this called "show_logo"?

tatitati: ouss...I was passing show_head_title=true, but the template was checking show_logo (by default true as well). That is a bug indeed. I didn't see it. Thanks @votemike

tatitati:
@BjornTwachtmann, question for you. @votemike had an interesting question. The tx Panel displays titles like "last ON", 'On now'. I've put that logic in BroadcastEventPresenter object. I thought that Keep only that logic in TxPanel was weird because they are strongly related. But Also this way of thought might be wrong. So what do you think?, is it good in that place?, should I put back that logic in PanelTxPresenter?

felipeparaujo:
@tatitati I think the title logic should live in the TxPresenter. None of the other places use that titling logic.

src/Controller/FindByPid/EpisodeController.php
```diff
@@ -5,6 +5,7 @@
 use App\Controller\BaseController;
 use App\Ds2013\PresenterFactory;
 use App\ExternalApi\FavouritesButton\Service\FavouritesButtonService;
+use App\DsShared\Helpers\LiveBroadcastHelper;
```
votemike: Is this used?

tests/Ds2013/Presenters/Domain/BroadcastEvent/BroadcastEventPresenterTest.php
```diff
@@ -58,4 +65,102 @@ public function getNetworkUrlProvider(): array
...
+        if ($episodeType == 'tv episode') {
+            $episode = EpisodeBuilder::anyTVEpisode()->build();
+        } elseif ($episodeType == 'radio episode') {
+            $episode = EpisodeBuilder::anyRadioEpisode()->build();
+        } else {
+            $episode = EpisodeBuilder::any()->build();
+        }
+
+        /** @see LiveBroadcastHelper::LIVE_SERVICE_URLS */
+        if ($typeOfServices == 'services doing live broadcasts') {
```
votemike: Did we decide we were moving away from magic strings?

src/Ds2013/Presenters/Section/Episode/Map/Panels/Side/tx.html.twig
```diff
@@ -1,5 +1,11 @@
 <div class="island">
-    PANEL: [SIDE/TX]
+    <h2>{{ tr(tx.getTitle()) }}</h2>
 
-    Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Aenean commodo ligula eget dolor.
+    {{-
+        ds2013('broadcastEvent', tx.getCollapsedBroadcast(), {
+            'show_watch_from_start':true,
+            'show_on_air_message': false,
+            'show_head_title' : true
```
BjornTwachtmann: show_head_title is gone now, right?

tests/Ds2013/Presenters/Domain/BroadcastEvent/BroadcastEventPresenterTest.php
tatitati: ok, i see was not the previous ones only. I changed all of them in the same way I did with the previous one with you.

PR #478, opened 2018-03-19
fix episodes_none string

tatitati:
The problem
We were displaying the text: "episodes_none" instead of "There are no episodes"

You can test the fix here:
http://sandbox.bbc.co.uk/programmes/p01x8j6n/episodes/guide?__flush_cache=1


programmes-developers:

PR #480, opened 2018-03-19
PROGRAMMES-6257 fix promo-priority

tatitati:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6257

----

Based on this we can see that programmes-pages-service map the property (in our DB data structure) 'cascadesToDescendants' to the attribute '$isSuperPromotion' in our entity Promotion model.

https://github.com/bbc/programmes-pages-service/blob/master/src/Mapper/ProgrammesDbToDomain/PromotionMapper.php#L56


programmes-developers:

BPScott:
You'll need to do something with [this](https://github.com/bbc/programmes-frontend/blob/a613f28ca77bb73ea5370c8971032db1d7f53e66/src/Controller/FindByPid/TlecController.php#L158-L162) too, so that the first "regular" promo doesn't appear in the list in the main page.

Consider the following list of promos when in "promo priority mode":

* super promo 1
* regular promo 1
* super promo 2
* regular promo 2

With your current code, you'll display "regular promo 1" in the MAP, then "regular promo 1, super promo 2, regular promo 2" in the page, which is wrong as you see "regular promo 1"  twice and "super promo 1" never.

src/Controller/FindByPid/TlecController.php
```diff
@@ -380,4 +380,18 @@ private function isSamePublication(Episode $onDemandEpisode, CollapsedBroadcast
     {
         return (string) $onDemandEpisode->getPid() == (string) $upcomingBroadcast->getProgrammeItem()->getPid();
     }
+
+    /**
+     * @param Promotion[] $promotions
+     */
+    private function filterPromoPriority(array $promotions) :?Promotion
```
votemike: filterByWhat?

votemike:
And a test for this?

PR #481, opened 2018-03-20
PROGRAMMES-6215 episode more map column

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6215

I _think_ everything else is OK. I don't think I've missed any important niche conditions or classes.


programmes-developers:

src/Controller/FindByPid/EpisodeController.php
```diff
@@ -61,11 +63,22 @@ public function __invoke(
         // Faucet to potentially save on a DB query
         $promotions = $promotionsService->findActivePromotionsByEntityGroupedByType($episode);
 
+        $nextEpisode = null;
+        $previousEpisode = null;
+        if (!$episode->isTleo()) {
+            /** @var Episode $nextEpisode|null */
```
felipeparaujo: shouldn't these docs be defined where the variable is first declared?

src/Ds2013/Presenters/Domain/Programme/CollapsedBroadcastProgrammePresenter.php
```diff
@@ -56,7 +56,7 @@ public function __construct(
         $this->collapsedBroadcast = $collapsedBroadcast;
     }
 
-    public function getProgrammeImagePresenter(): ProgrammeImagePresenter
+    public function getProgrammeImagePresenter(): ProgrammeOverlayPresenter
```
felipeparaujo: Why is this getting renamed?

src/Ds2013/Presenters/Domain/Programme/SubPresenters/programme_overlay.html.twig
```diff
@@ -0,0 +1,87 @@
...
+                            <meta property="thumbnailUrl" content="{{ programme_overlay.getImageUrl(1920, 1080) }}" />
+                        {% endif %}
+                    {% endif %}
+                </div>
+            {% endif %}
+        {% endblock %}
+
+    </div>
+
+    {% if programme_overlay.getOption('show_secondary_cta') and programme_overlay.isAvailable() %}
```
BjornTwachtmann: Question: Do you think this could be done without duplicating the logic in the overlay? Or is that not possible/more complicated?

src/Ds2013/Presenters/Domain/Programme/CollapsedBroadcastProgrammePresenter.php
votemike: Because it's not just an image presenter any more. It also has this new overlay thing. See V2 for reference.

src/Ds2013/Presenters/Domain/Programme/SubPresenters/programme_overlay.html.twig
```diff
@@ -0,0 +1,87 @@
...
+                                <span>
+                                    <meta property="endDate" datatype="xsd:dateTime" content="{{ programme_overlay.getProgramme().getStreamableUntil().format(constant('DateTime::ATOM')) }}" />
+                                </span>
+                            {% endif %}
+                        {% else %}
+                            {# If our available programme is NOT an episode, e.g. a clip #}
+                            <div class="{{ programme_icon_classes }}">
+                                {{ gelicon('audio-visual', programme_overlay.getMediaIconName(), 'gelicon--half-container') }}
+                            </div>
+                            <meta property="description" content="{{ programme_overlay.getProgramme().getShortSynopsis() }}" />
```
BjornTwachtmann: Realise this isn't part of this PR, but aren't these meta tags schema.org stuff?

src/Ds2013/Presenters/Domain/Programme/CollapsedBroadcastProgrammePresenter.php
BjornTwachtmann: basically because it's presenting things when show_image = false. Wasn't sure whether to do it this way or go the whole hog and create a CTA presenter. This seemed like less complexity.

BjornTwachtmann: Shouldn't this be renamed to getProgrammeOverlayPresenter though?

src/Ds2013/Presenters/Domain/Programme/SubPresenters/programme_overlay.html.twig
votemike: Possibly/Probably. When I was reading up about schema.org stuff, I don't remember it saying anything about meta tags. But the two attributes on it do look like schema.org stuff.
This only looks like a new file as it has been renamed.
We can remove these tags in a new PR at some point. They aren't doing any harm at the moment.

src/Ds2013/Presenters/Domain/Programme/SubPresenters/programme_overlay.html.twig
votemike: I think that would be more complicated.
The image part of this file already has plenty of ifs and class building.
I think I'd rather keep this part separate.
I'm not sure if this is the final thing that needs to be added to the programme presenter. It may be better to keep refactoring until everything that is needed is present in the presenter.

src/Ds2013/Presenters/Section/Episode/Map/Panels/Side/more.html.twig
```diff
@@ -1,5 +1,87 @@
...
+                            'show_image': false,
+                            'show_secondary_cta': true,
+                        },
+                        title_options: {
+                            'title_format': 'item::ancestry',
+                            'title_tag': 'h4',
+                            'title_classes': 'gamma'
+                        }
+                    }) }}
+                    {#<? self::view('/partials/generic/programme-object.spv', array(#}
```
BjornTwachtmann: Is there a reason this commented out code is still here?

Tasiobg:
Looks like `link_location` for previous or next episode is tracked as the default value `programmes_programmeobjectlink`. I think we want a different meanful value, I would suggest to double check with Claudio

votemike:
I think https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6282 will deal with that.

src/Ds2013/Presenters/Domain/Programme/SubPresenters/programme_overlay.html.twig
Tasiobg: Should a ticket be created soon to avoid confusion?
I found one of them http://schema.org/uploadDate , I didn't look for the others.
I think it may be worth to investiage how much search engines use this meta info because some of our pages have a big amount of meta data

votemike: @Tasiobg https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6239
Notice the line "As part of this, ensure that you get rid of any legacy Microdata or RDFa attributes on the episode page."

PR #484, opened 2018-03-21
Playout column iStats tracking

felipeparaujo:


programmes-developers:

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/playout.html.twig
```diff
@@ -6,7 +6,7 @@
             {{ self.image(playout) }}
         {% else %}
             {% if playout.isAvailableForStreaming() %}
-                <a class="playout__blocklink" href="{{ playout.getUrl() }}">
+                <a class="playout__blocklink" data-linktrack="programmes_iplayer_episodepage_playcurrent_calltoaction" href="{{ playout.getUrl() }}">
```
votemike: Should probably just be iplayer_episodepage_playcurrent_calltoaction. Could you double check?

PR #487, opened 2018-03-21
Fix empty panel classes

tatitati:
This is a fix for something that was not breaking anything, even crawler was passing the tests and in the browsers was looking good, but indeed was wrong. 

Before fix was (BAD):

```html
<div class="br-box-secondary  fauxcolumn  map__column" ">
```

After fix is (GOOD):
```html
<div class="br-box-secondary fauxcolumn map__column">
```


programmes-developers:

BPScott:
It might be a bit more readable to use the build_css_classes view helper, rather than have a bunch of if clauses:

```
{{ build_css_classes({ 'br-box-secondary': true, 'fauxcolumn': loop.last, 'map__column': loop.length > 1, }) }}
```

PR #490, opened 2018-03-22
PROGRAMMES-6276 Episode Broadcast List

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6276

I think I deciphered V2 correctly.

I'm not sure if we want all this code to live in episode.html.twig. Should I refactor it? To where?


programmes-developers:

templates/find_by_pid/episode.html.twig
```diff
@@ -105,7 +105,50 @@
...
+                                                {% if key < 3 %}
+                                                    <li class="grid 1/2 1/3@bpb1 1/3@bpb2 1/4@bpw 1/3@bpw2 1/3@bpe">
+                                                        {{- ds2013('broadcastEvent', broadcast, {
+                                                            'show_on_air_message': false,
+                                                            'container_classes': 'programme--grid highlight-box--grid'
+                                                        }) -}}
+                                                    </li>
+                                                {% else %}
+                                                    <li class="grid 1/2 1/3@bpb1 1/3@bpb2 1/4@bpw 1/3@bpw2 1/3@bpe ml__hidden">
+                                                        {{- ds2013('broadcastEvent', broadcast, {
```
BjornTwachtmann: Rather than repeating the presenter construction here, could you just set the classes differently, that's the only thing that changes, right?

BjornTwachtmann:
I don't really mind having this in the episode page template, it's not really re-used anywhere. It's probably more testable if you write it as a presenter, but there isn't so much logic here that a single controller test asserting that there is a broadcast list of N items that have X classes wouldn't be enough tbh

src/Controller/FindByPid/EpisodeController.php
```diff
@@ -48,10 +48,12 @@ public function __invoke(
         }
         $upcomingBroadcasts = [];
         $lastOnBroadcasts = [];
+        $allBroadcasts = [];
         if ($episode->getFirstBroadcastDate()) {
             // Only search for broadcasts if a programme has them
             $upcomingBroadcasts = $collapsedBroadcastsService->findUpcomingByProgrammeWithFullServicesOfNetworksList($episode, 1);
             $lastOnBroadcasts = $collapsedBroadcastsService->findPastByProgrammeWithFullServicesOfNetworksList($episode, 1);
+            $allBroadcasts = $collapsedBroadcastsService->findByProgrammeWithFullServicesOfNetworksList($episode, 100);
```
BPScott: This max number should be reduced, i reckon, showing 100 things in a 3 item wide list is going to look rather rubbish.

It was high in v2 because we had to request a list of Broadcasts, which we then converted into CollapsedBroadcasts (or, well, the v2 versions of those concepts) within code. This was because CollapsedBroadcasts are one of our concepts rather than something that is modelled within Nitro. Thus we'd request 100 Broadcasts, but we'd only end up displaying 10 or so CollapasedBroadcasts as e.g. if an episode was shown on BBC One there would have been 22 Broadcasts (one for each service) fetched that got collapsed into 1 CollapsedBroadcast.

src/Ds2013/Css/Components/_moreless.scss
```diff
@@ -0,0 +1,106 @@
...
+
+.ml__button--faq__answer {
+    padding: 0 $faq-answer-indent; // width of the icon + 8px;
+}
+
+// swapped (so undo-visually-hidden doesn't have to be used)
+// checked = hidden (also means no need for unsupported browser hacks)
+@mixin moreless-setup($namespace: '') {
+
+    .#{$namespace}ml .ml__status,
```
BPScott: The namespace should be after the ml, as the new convention is to have per-breakpoint variants as suffixes rather than prefixes: We should have `.ml@bpw` rather than `.bpw-ml`

BPScott:
As for testing, at some point we probably want a Episode Page Controller tests that sets up a load of stuff in the DB and asserts that various parts of the page are present.

templates/find_by_pid/episode.html.twig
votemike: I can. The classes are so ugly and unreadable that way though.

src/Controller/FindByPid/EpisodeController.php
votemike: Cathal says no

src/Ds2013/Css/Components/_moreless.scss
votemike: Nothing seems to work now. Have I done it wrong?

votemike: Fixed. I think.

votemike:
Test this please

PR #492, opened 2018-03-23
PROGRAMMES-6262 - Segments List Container And Groups

felipeparaujo:
Tests will come. Eventually.


programmes-developers:

src/Ds2013/Css/Components/_moreless.scss
```diff
@@ -0,0 +1,108 @@
...
+$faq-answer-indent: 23px;
+.ml__button--faq__answer {
+    padding: 0 $faq-answer-indent; // width of the icon + 8px;
+}
+
+// swapped (so undo-visually-hidden doesn't have to be used)
+// checked = hidden (also means no need for unsupported browser hacks)
+@mixin moreless-setup($namespace: '') {
+
+    .#{$namespace}ml .ml__status,
```
votemike: This should be `.ml\@#{$namespace}`

src/Ds2013/PresenterFactory.php
```diff
@@ -204,6 +206,23 @@ public function episodeMapPresenter(
         );
     }
 
+    public function segmentItemsPresenter(
+        ProgrammeItem $programmeItem,
+        array $segmentEvents,
```
votemike: What is this an array of? Would it be worth having a docblock?

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
```diff
@@ -0,0 +1,215 @@
...
+    private $context;
+
+    /** @var CollapsedBroadcast|null */
+    private $collapsedBroadcast;
+
+    /** @var bool|null */
+    private $isLive = null;
+
+    /** @var bool|null */
+    private $isReversed = false;
```
votemike: Should the docblock for this have a bit of an explanation?

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
```diff
@@ -0,0 +1,215 @@
...
+
+    public function getTitle(): string
+    {
+        $hasChapterSegments = false;
+        $hasMusicSegments = false;
+
+        foreach ($this->segmentEvents as $segmentEvent) {
+            if ($segmentEvent->getSegment() instanceof MusicSegment) {
+                $hasMusicSegments = true;
+            } elseif ($segmentEvent->getSegment()->getType() === 'chapter') {
```
votemike: Why elseif and not just if? Is it never both for the same segmentEvent? Can it be both per context?

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
```diff
@@ -0,0 +1,215 @@
...
+        $this->collapsedBroadcast = $upcoming ?? $lastOn;
+        $this->segmentEvents = $segmentEvents;
+    }
+
+    public function getTitle(): string
+    {
+        $hasChapterSegments = false;
+        $hasMusicSegments = false;
+
+        foreach ($this->segmentEvents as $segmentEvent) {
```
votemike: It's a waste foreaching the rest if the first one sets the bools to be true.

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
```diff
@@ -0,0 +1,215 @@
...
+                }
+
+                // things within the offset range, or that don't have offsets get displayed
+                if (!$segmentEvent->getOffset() || $currentOffset <= $segmentEvent->getOffset()) {
+                    $filteredSegmentEvents[] = $segmentEvent;
+                }
+            }
+
+            if ($this->isReversed) {
+                $this->isReversed = true;
```
votemike: Why do you need to set this? You just checked it above.

src/Ds2013/Presenters/Section/Segments/Types/AbstractSegmentItemPresenter.php
```diff
@@ -0,0 +1,20 @@
...
+
+use App\Ds2013\Presenter;
+
+abstract class AbstractSegmentItemPresenter extends Presenter
+{
+    protected $options = [
+        'moreless_class' => '',
+    ];
+
+    public function __construct(array $options = [])
```
votemike: Do you really need the constructor?

src/Ds2013/Presenters/Section/Segments/segments_list.html.twig
```diff
@@ -0,0 +1,57 @@
...
+            </div>
+        </div>
+    </div>
+{% endif %}
+
+{% macro segment_items(self, segments_list) %}
+    {% for segment_item in segments_list.getSegmentItemsPresenters() %}
+        {% if segment_item|length > 1 %}
+            <li class="segments-list__item segments-list__item--group br-keyline {{ segments_list.getMorelessClass() }}">
+                <h3>GROUP</h3>
```
votemike: This should be translated shouldn't it?

src/DsShared/Helpers/CanonicalVersionHelper.php
```diff
@@ -0,0 +1,67 @@
...
+
+            if (array_key_exists($type, self::VERSION_PREFERENCE_WEIGHTINGS)) {
+                $weighting = self::VERSION_PREFERENCE_WEIGHTINGS[$type];
+            }
+
+            if (is_null($highest)) {
+                $highest = $weighting;
+            }
+
+            if ($weighting > $highest) {
```
votemike: This could be an elseif couldn't it?

src/Ds2013/Presenters/Section/Segments/segments_list.html.twig
```diff
@@ -0,0 +1,57 @@
...
+                            <span class="ml__label--sep"> / </span>
+                            <span class="ml__label--less">{{ tr('show_less') }}</span>
+                        </span>
+                    </label>
+                {% else %}
+                    <ul class="list-unstyled segments-list__items">
+                        {{ self.segment_items(self, segments_list) }}
+                    </ul>
+                {% endif %}
+                {# TODO playlister #}
```
votemike: Are we not getting rid of that? https://github.com/bbc/programmes/pull/846

src/Controller/FindByPid/EpisodeController.php
```diff
@@ -27,11 +30,32 @@ public function __invoke(
...
+        $versions = $versionsService->findByProgrammeItem($episode);
+
+        if ($versions) {
+            $canonicalVersion = $canonicalVersionHelper->getCanonicalVersion($versions);
+            $segmentEvents = $segmentEventsService->findByVersionWithContributions($canonicalVersion);
+            if ($segmentEvents) {
+                $segmentsListPresenter = $presenterFactory->segmentsListPresenter(
+                    $episode,
+                    $segmentEvents,
+                    !empty($upcomingBroadcasts) ? reset($upcomingBroadcasts) : null,
```
votemike: Where is `$upcomingBroadcasts` set?

src/Controller/FindByPid/EpisodeController.php
```diff
@@ -27,11 +30,32 @@ public function __invoke(
...
+                    !empty($lastOnBroadcasts) ? reset($lastOnBroadcasts) : null,
```
votemike: Where is `$lastOnBroadcasts ` set?

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
felipeparaujo: I'm checking if there are any `MusicSegment`s or `chapter` segments in the list of segment events. There could be both types of segments in that list. There are even more types of segments than this one, but for the purposes of the segment list, we only care about these two.

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
felipeparaujo: Not really. We could have both `$hasMusicSegments` and `$hasChapterSegments` set to `true`. We won't know until we reach the end of the array. I could add a break when both are set to `true`, but more often than not that won't happen, so the check would probably make the whole thing less performant

src/Ds2013/Presenters/Section/Segments/Types/AbstractSegmentItemPresenter.php
felipeparaujo: These are just placeholders for now. I'll fix all of these in the next PR

src/Ds2013/Presenters/Section/Segments/segments_list.html.twig
felipeparaujo: Just placeholders, again

src/DsShared/Helpers/CanonicalVersionHelper.php
felipeparaujo: Probably. I just copied this over from Faucet.

src/Ds2013/Presenters/Section/Segments/segments_list.html.twig
felipeparaujo: I'm not sure. @BPScott ?

src/Ds2013/Css/Components/_moreless.scss
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    .ml#{$suffix} .ml__status:checked ~ .ml__button .ml__label--more {
+        display: inline;
+    }
+}
+
+@include moreless-setup();
+
+@each $name in $gel-widths-breakpoints {
```
felipeparaujo: @BPScott : This generates the moreless classes for all breakpoints. There's a problem though: anything below 320px doesn't get covered by the breakpoints. How do I fix it, or indeed, should I even try fixing this?

src/Controller/FindByPid/EpisodeController.php
felipeparaujo: Moved that to use the `$upcomingBroadcasts` fetched for the episode MAP

src/Controller/FindByPid/EpisodeController.php
felipeparaujo: same as above ^

src/Controller/FindByPid/EpisodeController.php
```diff
@@ -27,11 +30,16 @@ public function __invoke(
...
         VersionsService $versionsService,
-        PresenterFactory $presenterFactory
+        SegmentEventsService $segmentEventsService,
+        PresenterFactory $presenterFactory,
+        CanonicalVersionHelper $canonicalVersionHelper
     ) {
         $this->setIstatsProgsPageType('programmes_episode');
         $this->setContextAndPreloadBranding($episode);
 
+        $versions = $versionsService->findByProgrammeItem($episode);
```
BjornTwachtmann: Do we always need the versions? There is a segment_event_count on core_entity

src/Controller/FindByPid/EpisodeController.php
```diff
@@ -86,6 +91,21 @@ public function __invoke(
             $previousEpisode
         );
 
+        $segmentsListPresenter = null;
+        if ($versions) {
+            $canonicalVersion = $canonicalVersionHelper->getCanonicalVersion($versions);
+            $segmentEvents = $segmentEventsService->findByVersionWithContributions($canonicalVersion);
```
BjornTwachtmann: There is a segment_event_count on version that you should check before calling this

src/Controller/FindByPid/EpisodeController.php
```diff
@@ -27,11 +30,17 @@ public function __invoke(
         CollapsedBroadcastsService $collapsedBroadcastsService,
         FavouritesButtonService $favouritesButtonService,
         VersionsService $versionsService,
-        PresenterFactory $presenterFactory
+        SegmentEventsService $segmentEventsService,
+        PresenterFactory $presenterFactory,
+        CanonicalVersionHelper $canonicalVersionHelper
```
votemike: Do we want to start a standard on the way to order the params coming in to our `__invoke` methods?

src/Controller/FindByPid/EpisodeController.php
```diff
@@ -86,6 +91,21 @@ public function __invoke(
...
 
+        $segmentsListPresenter = null;
+        if ($versions) {
+            $canonicalVersion = $canonicalVersionHelper->getCanonicalVersion($versions);
+            $segmentEvents = $segmentEventsService->findByVersionWithContributions($canonicalVersion);
+            if ($segmentEvents) {
+                $segmentsListPresenter = $presenterFactory->segmentsListPresenter(
+                    $episode,
+                    $segmentEvents,
+                    !empty($upcomingBroadcasts) ? reset($upcomingBroadcasts) : null,
```
votemike: This logic is duplicated when constucting the MAP Presenter above. Is it worth refactoring it?

src/Ds2013/Presenters/Section/Segments/segments_list.html.twig
BPScott: I'm unsure what playlister here refers to, is sit some kind of favourite button?

We're removing the "playlister popularity module" which is a box that sits at the top of the right-hand panel which is unrelated to anything in the segments area.

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
```diff
@@ -0,0 +1,228 @@
...
+
+    private function filterSegmentEvents(): array
+    {
+        if ($this->context->getOption('show_tracklist_inadvance')) {
+            return $this->segmentEvents;
+        }
+
+        // if the programme item is currently being broadcast for the first time, filter to only the segment events that
+        // have already started
+        if ($this->collapsedBroadcast && !$this->collapsedBroadcast->isRepeat() && $this->isLive($this->collapsedBroadcast)) {
```
votemike: If you reverse this check, you could return early and move all of this a tab left.

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
```diff
@@ -0,0 +1,228 @@
...
+
+    private function createSegmentItem(
+        SegmentEvent $segmentEvent,
+        int $relativeOffset,
+        int $totalCount
+    ): AbstractSegmentItemPresenter {
+        $options = [];
+
+        // If there are more than 6 segments and we're past the third one, we start hiding things
+        if ($relativeOffset > 2 && $totalCount >= 6) {
```
votemike: Should this be `$totalCount > 6`?

src/DsShared/Helpers/CanonicalVersionHelper.php
```diff
@@ -0,0 +1,63 @@
...
+            if ($a->isStreamable()) {
+                $weightingA += self::STREAMABLE_WEIGHT;
+            }
+
+            if ($b->isStreamable()) {
+                $weightingB += self::STREAMABLE_WEIGHT;
+            }
+
+            // should be ordered by descending, so B first
+            return $weightingB <=> $weightingA;
```
votemike: 👾

src/DsShared/Helpers/CanonicalVersionHelper.php
```diff
@@ -0,0 +1,63 @@
...
+        });
+
+        return reset($versions);
+    }
+
+    private function typeWeighting(Version $version)
```
votemike: Return typehint?

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
```diff
@@ -0,0 +1,228 @@
...
+
+        $groups = [];
+        $totalCount = count($this->segmentEvents);
+
+        $group = [];
+        $previousTitle = reset($this->segmentEvents)->getTitle();
+
+        // We use the index and call it 'relative offset' here because we could
+        // have reversed the array in case of music segments for live programme debuts.
+        foreach ($segmentEvents as $relativeOffset => $segmentEvent) {
```
votemike: This whole method is a bit grim and hard to understand. Is it possible to make it simpler? If not, is it possible to refactor with some well named methods in to something a bit more understandable?

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    public function getMorelessClass(): string
+    {
+        return $this->hasMoreless() ? 'ml@bpb1' : '';
+    }
+
+    public function hasMoreless(): bool
+    {
+        return count($this->segmentEvents) >= 6;
```
votemike: Is this correct?

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
```diff
@@ -0,0 +1,273 @@
...
+            } else {
+                $presenters[] = new SpeechPresenter($segmentEvent->getSegment(), $options);
+            }
+
+            $current += 1;
+        }
+
+        if ($isGroup) {
+            $options = [];
+            if ($start > 3 && $total >= 6) {
```
votemike: This is the 3rd time 3 and 6 have been mentioned. Should they be constant or something?

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
felipeparaujo: yes.

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
felipeparaujo: I don't see how the code would benefit much from it... But I don't care that much. If you think I should abstract this, then I will.

votemike: So that if in the future it's decided that the number should be 9. We don't miss a part of the code when we make the change.

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
felipeparaujo: actually, it wasn't. I fixed it.

src/Ds2013/Presenters/Section/Segments/SegmentItem/GroupPresenter.php
```diff
@@ -0,0 +1,31 @@
+<?php
+declare(strict_types=1);
+
+namespace App\Ds2013\Presenters\Section\Segments\SegmentItem;
+
+class GroupPresenter extends AbstractSegmentItemPresenter
+{
+    /** @var array */
```
BjornTwachtmann: AbstractSegmentItemPresenter[]

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
```diff
@@ -0,0 +1,271 @@
...
+            $this->collapsedBroadcast->isRepeat() ||
+            !$this->isLive($this->collapsedBroadcast)
+        ) {
+            return $this->segmentEvents;
+        }
+
+        // if the programme item is currently being broadcast for the first time, filter to only the segment events that
+        // have already started
+        $filteredSegmentEvents = [];
+        $currentOffset = $this->collapsedBroadcast->getStartAt()->getTimestamp() - (ApplicationTime::getTime())->getTimestamp();
```
BjornTwachtmann: This is the wrong way round, will always be negative

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
```diff
@@ -0,0 +1,271 @@
...
+
+    /** @var bool|null */
+    private $isLive = null;
+
+    /**
+     * Has the list of segment events been reversed?
+     *
+     * @var bool
+     */
+    private $isReversed = false;
```
BjornTwachtmann: Does this need to be a class property, it's only used in one method as far as I can see

felipeparaujo: This is in preparation for Music segments. We need to know if the segments array has been reversed to decide the timing type.

src/Ds2013/Presenters/Section/Segments/SegmentItem/chapter.html.twig
```diff
@@ -0,0 +1,6 @@
+<li class="segments-list__item segments-list__item--music {{ chapter.getOption('moreless_class') }}">
+    <{{ chapter.getOption('h_tag') }} class="gamma no-margin">
+    {{ chapter.getTitle() }}
+</{{ chapter.getOption('h_tag') }}>
```
votemike: The indenting looks a bit off here.

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
```diff
@@ -0,0 +1,271 @@
...
+            $this->collapsedBroadcast->isRepeat() ||
+            !$this->isLive($this->collapsedBroadcast)
+        ) {
+            return $this->segmentEvents;
+        }
+
+        // if the programme item is currently being broadcast for the first time, filter to only the segment events that
+        // have already started
+        $filteredSegmentEvents = [];
+        $currentOffset = (ApplicationTime::getTime())->getTimestamp() - $this->collapsedBroadcast->getStartAt()->getTimestamp();
```
votemike: Would something like `$this->collapsedBroadcast()->getStartAt()->diffInSeconds(ApplicationTime::getTime())` be more readable? Or maybe the other way around.

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
```diff
@@ -0,0 +1,271 @@
...
+    private function createSegmentItem(array $segmentEvents, int $start): AbstractSegmentItemPresenter
+    {
+        $current = $start;
+        $presenters = [];
+
+        // If the segment event title is non-empty, we're dealing with a group
+        $isGroup = !empty(reset($segmentEvents)->getTitle());
+
+        // If a segment item is part of a group, the h tag has to be one level deeper
+        $hTag = $isGroup ? 'h4' : 'h3';
```
votemike: Is the level not an option?

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
felipeparaujo: yes. I didn't know about that helper.

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
```diff
@@ -41,6 +42,10 @@ class SegmentsListPresenter extends Presenter
      */
     private $isReversed = false;
 
+    protected $options = [
+        'h_tag_level' => 1,
```
votemike: Will the h_tag ever be an h1? Is that a sensible default?

PR #493, opened 2018-03-23
fix edge-case promo priority

tatitati:
Jatin found this edge case related with promo priority. 

We found a bug for an edge case. This bug is met when (using the original logic in tlecController):
```php
if ($isPromoPriority) { // true
    $firstRegularPromotion = $this->extractFirstRegularPromotionAndUpdateList($promotions); // null
}
```

tlecController Ref code:
https://github.com/bbc/programmes-frontend/blob/develop/src/Controller/FindByPid/TlecController.php#L146-L148

This true & null values are a problem because $isPromoPriority is the flag that indicates that we have to create a PromoPriorityPresenter with a $firstRegularPromotion. But instead, the presenter has no promotion to create, throwing an exception.

MapPresenter ref code:
https://github.com/bbc/programmes-frontend/blob/develop/src/DsAmen/Presenters/Section/Map/MapPresenter.php#L238-L239

----

Related to this ticket:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6257


programmes-developers:

src/DsAmen/Presenters/Section/Map/MapPresenter.php
```diff
@@ -235,7 +235,7 @@ private function constructTwoColumnMap(): void
     private function constructLeftColumn(): void
     {
         $leftColumnOptions = ['is_three_column' => $this->countTotalColumns() === 3];
-        if ($this->isPromoPriority) {
+        if ($this->isPromoPriority && $this->firstPromo) {
```
votemike: If the is no firstPromo, then the MAP shouldn't be in PromoPriority mode.

BjornTwachtmann: yeah, if there's a bug here it's in whatever is setting isPromoPriority

tests/DsAmen/Presenters/Section/Map/MapPresenterTest.php
```diff
@@ -95,6 +97,37 @@ public function testDefaultColumns()
         $this->assertColumns($presenter, [OnDemandPresenter::class]);
     }
 
+    /**
+     * [ edge case ]. A promo priority presenter should be created but also need a promotion for it.
```
votemike: Is this method of commenting a standard of some sort? What is in the square brackets?

tatitati: is not an standard. I will remove it. Was just a kind of intention to separate a little more group of tests.

src/DsAmen/Presenters/Section/Map/MapPresenter.php
tatitati: I thought about it, but I was shy with my solution, thanks guys

tests/DsAmen/Presenters/Section/Map/MapPresenterTest.php
votemike: We could use `@group` for that, couldn't we?

tatitati: but my guess is that we will finish with @groups everywhere.....I will delete my format with squares, is gonna be simpler

src/Controller/FindByPid/TlecController.php
```diff
@@ -140,12 +140,8 @@ public function __invoke(
         // Map parameters
         $isVotePriority = $this->isVotePriority($programme);
         $showMiniMap = $this->showMiniMap($request, $programme, $isVotePriority, isset($resolvedPromises['lxPromo']));
-        $isPromoPriority = $this->isPromoPriority($programme, $showMiniMap, !empty($promotions));
 
-        $firstRegularPromotion = null;
-        if ($isPromoPriority) {
-            $firstRegularPromotion = $this->extractFirstRegularPromotionAndUpdateList($promotions);
-        }
+        $firstRegularPromotion = $this->extractFirstRegularPromotionAndUpdateList($programme, $showMiniMap, $promotions);
```
BjornTwachtmann: I see what you're doing here, if you name it this way, nobody else will though :)

How about calling the variable $promoPriorityPromotion, here and in the MapPresenter, removing the $isPromoPriority variable from the MapPresenter, and renaming the method extractPromoPriorityPromotionAndUpdateList

tests/DsAmen/Presenters/Section/Map/MapPresenterTest.php
```diff
@@ -111,7 +148,7 @@ private function assertColumns(MapPresenter $presenter, array $columns)
         }
     }
 
-    private function createMapPresenter($programmeContainer, ?CollapsedBroadcast $upcomingBroadcasts = null, ?Promotion $firstPromo = null): MapPresenter
+    private function createMapPresenter($programmeContainer = null, ?CollapsedBroadcast $upcomingBroadcasts = null, ?Promotion $firstPromo = null): MapPresenter
```
votemike: Why have you made this nullable? The MAPPresenter doesn't accept null as the 4th argument.

tests/DsAmen/Presenters/Section/Map/MapPresenterTest.php
```diff
@@ -77,22 +79,57 @@ public function testWorldNewsColumns()
...
+        $leftC = $pr->getLeftColumn();
+
+        $this->assertNotNull($leftC);
+        $this->assertInstanceOf(PromoPriorityPresenter::class, $leftC);
+    }
+
+    /**
+     * if there is no PriorityPromotion, then we dont create a PromoPriorityPresenter
+     */
+    public function testAssertDontCreateLeftColumn()
```
votemike: When is there not a left column? There's always a left column, isn't there?

tests/Controller/FindByPid/TlecControllerTest.php
```diff
@@ -108,19 +108,6 @@ public function showMiniMapDataProvider(): array
         return $cases;
     }
 
-    public function testShowPromoPriority()
```
votemike: Is this no longer a relevant test?

src/Controller/FindByPid/TlecController.php
```diff
@@ -140,23 +140,18 @@ public function __invoke(
         // Map parameters
         $isVotePriority = $this->isVotePriority($programme);
         $showMiniMap = $this->showMiniMap($request, $programme, $isVotePriority, isset($resolvedPromises['lxPromo']));
-        $isPromoPriority = $this->isPromoPriority($programme, $showMiniMap, !empty($promotions));
 
-        $firstRegularPromotion = null;
-        if ($isPromoPriority) {
-            $firstRegularPromotion = $this->extractFirstRegularPromotionAndUpdateList($promotions);
-        }
+        $promoPriorityPromotion = $this->extractPromoPriorityPromotionAndUpdateList($programme, $showMiniMap, $promotions);
```
votemike: This method is now doing a lot. I'd probably leave it as it was, but surround it with the `if` statement that you've added to it. e.g.:
```
$promoPriorityPromotion = null;
if ($programme->getOption('brand_layout') === 'promo' && !empty($promotions) && $programme->isTlec() && !$showMiniMap) {
    $promoPriorityPromotion = $this->extractFirstRegularPromotionAndUpdateList($promotions);
}
```
Or something like that.

tests/Controller/FindByPid/TlecControllerTest.php
tatitati: No anymore. The reasons are:
1. The trigger action of the test doesn't exist any more. TlecController::isPromoPriority doesn't exist anymore. Instead we created  a new TlecController:extractPromoPriorityPromotionAndUpdateList
```php
$this->assertTrue($this->invokeMethod($controller, 'isPromoPriority', [$programmeContainer, false, true]));
```

2. That test was also using mocked entities when instead we can create real entities with our builders. You can see more tests with the same setup exercising different aspects of the behaviour like:
- testCanSetFlagsProperly
- testCannotRemoveAnythingFromArray
- testCanProduceAnEmptyArrayOfPromotions
- testReturnNullWhenNoPromotions

tests/DsAmen/Presenters/Section/Map/MapPresenterTest.php
tatitati: true

tests/DsAmen/Presenters/Section/Map/MapPresenterTest.php
tatitati: true. Maybe a better name is testAssertDontCreatePromoPriorityPanel

votemike: I'm not sure you need both the words 'test' and 'assert' in the name.
How about `testLeftColumnHasProgrammeInfo()`?

src/Controller/FindByPid/TlecController.php
```diff
@@ -140,23 +140,22 @@ public function __invoke(
         // Map parameters
         $isVotePriority = $this->isVotePriority($programme);
         $showMiniMap = $this->showMiniMap($request, $programme, $isVotePriority, isset($resolvedPromises['lxPromo']));
-        $isPromoPriority = $this->isPromoPriority($programme, $showMiniMap, !empty($promotions));
 
-        $firstRegularPromotion = null;
-        if ($isPromoPriority) {
-            $firstRegularPromotion = $this->extractFirstRegularPromotionAndUpdateList($promotions);
+        $promoPriorityPromotion = null;
```
votemike: Tiny thing, `$promoPriorityPromotion` sounds repetitive. What do people thing of something like `$priorityPromotion`

src/Controller/FindByPid/TlecController.php
```diff
@@ -140,23 +140,22 @@ public function __invoke(
...
+        if ($programme->getOption('brand_layout') === 'promo' && !empty($promotions) && $programme->isTlec() && !$showMiniMap) {
+            $promoPriorityPromotion = $this->extractPromoPriorityPromotionAndUpdateList($promotions);
```
votemike: `extractPromoPriorityPromotionAndUpdateList()` -> `extractPriorityPromotionAndUpdateList()`?

src/DsAmen/Presenters/Section/Map/MapPresenter.php
```diff
@@ -34,14 +34,11 @@ class MapPresenter extends Presenter
     private $debutsCount;
 
     /** @var Promotion|null */
-    private $firstPromo;
+    private $promoPriorityPromotion;
```
votemike: Same comment about repetitiveness.

src/Controller/FindByPid/TlecController.php
```diff
@@ -381,9 +375,10 @@ private function isSamePublication(Episode $onDemandEpisode, CollapsedBroadcast
     }
 
     /**
-     * @param Promotion[] $promotions
+     * @param array $promotions
```
votemike: Why has this changed?

src/DsAmen/PresenterFactory.php
```diff
@@ -99,12 +99,11 @@ public function mapPresenter(
         ProgrammeContainer $programme,
         ?CollapsedBroadcast $upcomingBroadcast,
         ?CollapsedBroadcast $lastOn,
-        ?Promotion $firstPromo,
+        ?Promotion $promoPriorityPromotion,
```
votemike: Again, `$promoPriorityPromotion` seem repetitive.

tests/Controller/FindByPid/Tlec/TlecControllerPromoPriorityTest.php
```diff
@@ -17,21 +17,7 @@ public function setUp()
...
-            PromotionBuilder::any()->with(['isSuperPromotion' => false])->build(),
-        ];
-
-        $foundRegularPromotion = $this->invokeMethodInController('extractFirstRegularPromotionAndUpdateList', [&$promotions]);
-
-        $this->assertEquals($firstRegularPromotion, $foundRegularPromotion);
-    }
-
-    public function testCanRemoveSpecifiedPromotionFromList()
+    public function testCanSetFlagsProperly()
```
votemike: Which flags? Do we have flags?

tests/Controller/FindByPid/Tlec/TlecControllerPromoPriorityTest.php
```diff
@@ -53,24 +40,34 @@ public function testCannotRemoveAnythingFromArray()
...
-
     /**
-     * [ Edge case ]. Remove the first regular promotion from a list with only one promotion
+     * Edge case. Remove the first regular promotion from a list with only one promotion
      */
     public function testCanProduceAnEmptyArrayOfPromotions()
     {
         $promotions = [
-            $promo2 = PromotionBuilder::any()->with(['isSuperPromotion' => false])->build(),
+            $promo0 = PromotionBuilder::any()->with(['isSuperPromotion' => false])->build(),
```
votemike: Assigning something to a variable while constructing an array seems a bit odd. Did you mean to do that?

tests/DsAmen/Presenters/Section/Map/MapPresenterTest.php
```diff
@@ -77,22 +79,57 @@ public function testWorldNewsColumns()
...
         $programmeContainer = $this->createProgrammeWithEpisodes();
         $presenter = $this->createMapPresenter($programmeContainer);
-        $this->assertColumns($presenter, [OnDemandPresenter::class]);
+        $this->assertRightColumns($presenter, [OnDemandPresenter::class]);
+    }
+
+    /**
+     * If we pass a promo priority, then we create a PromoPriorityPresenter
+     */
+    public function testAssertLeftColumns()
```
votemike: I don't think test names need both "test" and "assert" in them. Also, the name seems incomplete... testAssertLeftColumns...... what? Have a promo priority presenter in them?

PR #495, opened 2018-03-26
PROGRAMMES-6277 Stream Url Helper

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6277


programmes-developers:

BjornTwachtmann:
I think this looks good. One point... I wonder if it might be more flexible for the helper to just generate the full URL itself, what's the reason you decided to have it return the route name instead?

votemike:
@BjornTwachtmann many reasons.

- We don't need to pass a UrlGeneratorInterface in to the class
- We don't need to worry about seldom used logic such as `force_iplayer_linking` (we can make the calling class worry about that)
- We don't have to worry about `_fragment`s such as `#play` (we can make the calling class worry about that)

I could change it to generate the whole URL if you like.

BjornTwachtmann:
Fair enough. I think returning the route name should be enough when we switch this to the new thing later on anyway.

PR #496, opened 2018-03-26
Fix ie11 links anchors schedules

tatitati:
## To test it you need JAWS on IE-11
URL example to test:
https://programmes-frontend.int.api.bbc.co.uk/schedules/p00fzl7k

Ticket:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6091


programmes-developers:

votemike:
Does this also fix issue 2?

votemike:
The issue looks to be something to do with lazy-loading. As the problem doesn't occur if you first scroll the page and load all the lazy images.
The page also reacts in this way when clicking on the links as well as when you tab and press enter.
@BjornTwachtmann I feel this fix is more of a hack which accidentally fixes the problem. Are we happy with that?

BjornTwachtmann:
Hiding these links from screen readers doesn't seem like the best way to tackle this issue TBH. We should discuss this...

tatitati:
im closing for now the PR, I'm trying experiments and I consider im gonna create just useless noise, thanks guys

PR #497, opened 2018-03-27
PROGRAMMES-6244 Fixing Chrome+Voiceover bug where first tab into role…

votemike:
…=main read out the whole div

https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6244


programmes-developers:

PR #500, opened 2018-03-28
Fix CSS class naming

BjornTwachtmann:
* This fixes a bug with clip display at small breakpoints on pages like /programmes/p005xm30


programmes-developers:

PR #501, opened 2018-03-28
PROGRAMMES-6260 & PROGRAMMES-6261 - Speech (and Chapter) Segments

felipeparaujo:
We're removing playout from Episode pages, and future systems won't support deep linking, so Chapter and Speech segments are the same.


programmes-developers:

src/Ds2013/Presenters/Section/Segments/SegmentItem/GroupPresenter.php
```diff
@@ -8,14 +8,10 @@ class GroupPresenter extends AbstractSegmentItemPresenter
     /** @var AbstractSegmentItemPresenter[] */
     private $presenters;
 
-    /** @var string */
-    private $title;
-
     public function __construct(string $title, array $presenters, array $options)
```
BjornTwachtmann: Don't you need to change this if you're passing in a SegmentEvent now?

src/Ds2013/Presenters/Section/Segments/SegmentItem/SpeechPresenter.php
```diff
@@ -3,26 +3,26 @@
...
+    public function getDuration(): string
     {
-        return $this->segment->getTitle();
+        $format = 'i:s';
+
+        if ($this->segmentEvent->getSegment()->getDuration() >= 3600) {
+            $format = 'h:i:s';
+        }
+
+        return gmdate($format, $this->segmentEvent->getSegment()->getDuration());
```
BjornTwachtmann: That's going to fail if the programme is >24h long. There are things that long in the DB, although they're mostly "barkers" which won't have segments. I can't think of a "nice" way to do this either though tbh.

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
```diff
@@ -251,15 +250,9 @@ private function createSegmentItem(array $segmentEvents, int $start): AbstractSe
...
-            } elseif ($segmentEvent->isChapter() &&
-                $segmentEvent->getOffset() &&
-                ($this->context->isStreamable() || $this->context->isDownloadable()) &&
-                !$this->isLive
-            ) {
-                $presenters[] = new ChapterPresenter($segmentEvent->getSegment(), $options);
+                $presenters[] = new MusicPresenter($segmentEvent, $options);
             } else {
-                $presenters[] = new SpeechPresenter($segmentEvent->getSegment(), $options);
+                $presenters[] = new SpeechPresenter($segmentEvent, $options);
```
BjornTwachtmann: Shouldn't we still be checking segment type here, what if new ones get added or something? We could just ignore the ones we don't understand. Or am I missing something?

src/Ds2013/Presenters/Section/Segments/SegmentItem/speech.html.twig
```diff
@@ -1,6 +1,13 @@
-<li class="segments-list__item segments-list__item--speech {{ speech.getOption('moreless_class') }}">
-    <{{ speech.getOption('h_tag') }} class="gamma no-margin">
-        {{ speech.getTitle() }}
-    </{{ speech.getOption('h_tag') }}>
-    <p>Speech Segment</p>
+<li class="segments-list__item segments-list__item--music {{ speech.getOption('moreless_class') }}">
+    <div class="segment text--prose">
+        <div class="segment__content">
+            {{ speech.getPid() }}
```
BjornTwachtmann: This method doesn't exist, also why?

felipeparaujo: Just debugging stuff that got leftover, I'll delete it

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
felipeparaujo: V2 defaults to speech when it doesn't recognize the type of the segment. I thought we should be doing the same here.

felipeparaujo: https://github.com/bbc/programmes/blob/develop/webapp/php/lib/BBC/Programmes/View/SegmentItems.php#L263-L279

src/Ds2013/Presenters/Section/Segments/SegmentItem/GroupPresenter.php
felipeparaujo: Yes.

src/Ds2013/Presenters/Section/Segments/SegmentItem/SpeechPresenter.php
felipeparaujo: https://github.com/bbc/programmes-frontend/blob/develop/src/Ds2013/Presenters/Domain/CoreEntity/Programme/SubPresenters/ProgrammeBodyPresenter.php#L53-L69

The programme object does this. How about extracting it to a helper?

BjornTwachtmann: That seems like a good approach, yeah

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
BjornTwachtmann: Fair enough, leave it this way for now. I think there's a ticket somewhere for a new type anyway.

PR #502, opened 2018-03-28
PROGRAMMES-6278: Recipes panel

tatitati:
Example url to test (don't forget to run ```yarn run gulp```):
https://sandbox.bbc.co.uk/programmes/b0876646?__flush_cache=1

### Important facts about recipes-urls:
1. How we load recipes module with ajax:
Episode page is loading recipes using the route ```programme_recipes_ds2013```
Tlec page instead uses ```programme_recipes_amen```

2. When loaded, there is a “see all recipes” link in the module.
For episode page and tlec page this is pointing to ```food_recipes``` (example: /food/programme/{pid})

3. The visible banner displayed when the user disable javascript show a link as well similar to the previous one: “see all recipes”
For episode page and tlec page this is pointing to ```food_recipes``` as well like the previous one


programmes-developers:

templates/find_by_pid/episode.html.twig
```diff
@@ -46,6 +46,11 @@
                     {# TODO MPU at sizes 1 and 2 #}
 
                     {# TODO Recipes module #}
+                    <div class="component component--box component--box-flushbody-vertical component--box--primary lazy-module"
```
BjornTwachtmann: This should only be inserted where the relevant programme option is set. Check the TLEC page stuff for details. Calling this on every episode page would bring down the recipes API.

src/Ds2013/Presenters/Domain/Recipe/recipe.html.twig
```diff
@@ -0,0 +1,41 @@
+<div class="grid__inner">
```
BjornTwachtmann: Starting a presenter with grid__inner makes it very specific to where you're inserting it. Also, looking at the classes below, probably some of the highlight box stuff should be passed in rather than assumed to be correct anywhere this is used. 

I realise this isn't likely to be used anywhere other than the episode page, but it's good practice to try and keep presenters reusable.

BjornTwachtmann: also, we don't even have grid__inner in V3, do we?

src/Controller/RecipesController.php
```diff
@@ -38,7 +38,7 @@ public function __invoke(
         $this->response()->setMaxAge(300);
 
         // render does not automatically pick up the headers from the response
-        return $this->render('recipes/show.ameninc.html.twig', [
+        return $this->render('recipes/show' . $extension . '.html.twig', [
```
BjornTwachtmann: I know that technically $extension is filtered by Symfony's routing, but passing stuff directly into file paths always makes me nervous. I would do something like

`$templatePath = ($extension === 'ameninc') ? 'recipes/show.ameninc.html.twig' : 'recipes/show.inc.html.twig';`

That way if somebody makes changes in the routing file it's not going to open any vulnerabilities.

src/Controller/FindByPid/EpisodeController.php
```diff
@@ -34,7 +35,8 @@ public function __invoke(
         SegmentEventsService $segmentEventsService,
         GroupsService $groupsService,
         PresenterFactory $presenterFactory,
-        CanonicalVersionHelper $canonicalVersionHelper
+        CanonicalVersionHelper $canonicalVersionHelper,
+        RecipesService $recipesService
```
BjornTwachtmann: No point passing this in if you're not using it.

src/Ds2013/PresenterFactory.php
```diff
@@ -294,6 +296,12 @@ public function promotionPresenter(
         );
     }
 
+    public function recipePresenter(
+        Recipe $recipesApiResult
```
BjornTwachtmann: $recipe ?

templates/recipes/show.inc.html.twig
```diff
@@ -0,0 +1,21 @@
...
+{%- block body -%}
+    <div id="recipes" class="component component--box component--box-flushbody-vertical component--box--primary lazy-module--complete">
+        <div class="component__header br-box-page">
+            <h2>{{ tr('recipes') }}</h2>
+        </div>
+        <div class="component__body br-box-page">
+            <ul class="grid-wrapper highlight-box-wrapper highlight-box-wrapper--grid@bpb1 highlight-box-wrapper--grid@bpw">
+                {%- for recipe in recipes -%}
+                    <li class="grid 1/2@bpb1 1/2@bpw">
+                        <div class="grid__inner">
```
BjornTwachtmann: Again, no more grid_inner

src/Controller/RecipesController.php
BPScott: This pattern crops up a few times in the codebase. If that's the way we want to do it, then we should fix it everywhere

src/Ds2013/Presenters/Domain/Recipe/recipe.html.twig
```diff
@@ -0,0 +1,41 @@
...
+                            <img class="circular-img rsp-img" alt="" src="{{ recipe.getChefImageUrl() }}"/>
+                        </div>
+                    {%- endif -%}
+
+                    <span>{{ recipe.getChefName() }}</span>
+                </div>
+            {% endif %}
+        </div>
+
+        <a class="block-link__overlay-link" href="{{ recipe.getRecipeUrl() }}" aria-hidden="true" tabindex="-1">
```
BPScott: I believe we've moved to a new style of blocklinks in ds2013 style layouts too so you no longer need the link that sits at the bottom of the object for the fake giant clickable areas, instead you put the link in the heading with a specific class to make the link expand to the full area. Check the programme or group object to see the usage.

src/Ds2013/Presenters/Domain/Recipe/recipe.html.twig
BPScott: No grid__inner in v3 in either design system.

templates/find_by_pid/episode.html.twig
BPScott: If this is only appearing on pages that have the option set, is it worth including this directly in the page rather than having a lazyload? It'd help reduce complexity. Or is the API that slow that we need a lazyload?

src/Ds2013/Presenters/Domain/Recipe/recipe.html.twig
```diff
@@ -0,0 +1,41 @@
...
+        </div>
+
+        <div class="programme__body">
+            <h4 class="programme__titles">
+                <a href="{{ recipe.getRecipeUrl() }}" class="br-blocklink__link programme__title">
+                    {{ recipe.getTitle() }}
+                </a>
+            </h4>
+
+            {%- if recipe.hasChef() -%}
```
BPScott: The convention in a lot of our existing objects is to roll the check and the output into one. E.g. in the Programme object instead of having "hasImage" and "getImage", "getImage" shall return null if there is no image.

 So instead of having  "hasChef" and "getChefName" methods, have a single "getChefName" method that returns an empty string if the name is absent.

Then here you can test `if recipe.getChefName`.

You can do the same with the chefImage too - removing the `hasChefImage` method

src/Builders/ExternalApi/Recipes/RecipeImageBuilder.php
```diff
@@ -0,0 +1,24 @@
...
+    protected function __construct()
+    {
+        $faker = Factory::create();
+
+        $this->classTarget = RecipeImage::class;
+        $this->blueprintConstructorTarget = [
+            'id' => $faker->word,
+            'title' => $faker->sentence(3),
+            'description' => $faker->sentence(5),
+            'image' => null,
```
votemike: Why would a RecipeImage have an image?

src/Ds2013/Presenters/Domain/Recipe/RecipePresenter.php
```diff
@@ -0,0 +1,77 @@
...
+use Exception;
+
+class RecipePresenter extends Presenter
+{
+    /** @var Recipe */
+    private $recipe;
+
+    public function __construct(Recipe $recipe)
+    {
+        $options = [
```
votemike: Are these options fixed everywhere Recipes are used?

src/Ds2013/Presenters/Domain/Recipe/RecipePresenter.php
```diff
@@ -0,0 +1,77 @@
...
+            ],
+        ];
+
+        $this->recipe = $recipe;
+        parent::__construct($options);
+    }
+
+    public function hasChef() :bool
+    {
+        if (!$this->recipe->getChef()) {
```
votemike: `return $this->recipe->getChef();`?

src/Ds2013/Presenters/Domain/Recipe/recipe.html.twig
```diff
@@ -0,0 +1,41 @@
...
+                <div class="recipe__chef">
+                    {%- if recipe.hasChefImage() -%}
+                        <div class="recipe__chef-img 1/5">
```
votemike: Is 1/5 a class that we have?

src/Ds2013/Presenters/Domain/Recipe/recipe.html.twig
```diff
@@ -0,0 +1,41 @@
...
+                        </div>
+                    {%- endif -%}
+
+                    <span>{{ recipe.getChefName() }}</span>
+                </div>
+            {% endif %}
+        </div>
+
+        <a class="block-link__overlay-link" href="{{ recipe.getRecipeUrl() }}" aria-hidden="true" tabindex="-1">
+            View full recipe
```
votemike: Should this be translated?

src/Ds2013/Presenters/Domain/Recipe/recipe.html.twig
```diff
@@ -0,0 +1,41 @@
...
+                        recipe.getOption('image_options')['default_width'],
+                        recipe.getOption('image_options')['sizes'],
+                        {'is_lazy_loaded': recipe.getOption('image_options')['is_lazy_loaded']}
+                    )
+                -}}
+            {%- endif -%}
+        </div>
+
+        <div class="programme__body">
+            <h4 class="programme__titles">
```
votemike: Is there an h3 above these?

tests/Builders/ExternalApi/Recipes/ChefBuilderTest.php
```diff
@@ -0,0 +1,20 @@
+<?php
```
votemike: Do we need tests for the builders?

templates/find_by_pid/episode.html.twig
BjornTwachtmann: I think when Felipe looked at this last time he said this was slow enough to warrant lazy-loading, that's the way the amen stuff was done anyway

felipeparaujo: It was slow enough to warrant lazyloading

src/Controller/RecipesController.php
felipeparaujo: I don't like having a controller serving two different routes. If it were up to me, we would have an abstract controller that abstracts the whole data fetching logic and two controllers that serve the different extensions.

tatitati: is another option. In my case I followed the same way we did in OnNowController.

BjornTwachtmann: Felipe's suggestion works. I'm happy with that, or with making the change I outlined above in both places. I don't have a strong preference.

tatitati: cool. I will go for felipe's way then, seems fine for me too,  thanks guys

templates/find_by_pid/episode.html.twig
BjornTwachtmann: Another point on this. You need to add in the icon-box from V2. Currently you see nothing with JS disabled, you should see a box with an icon and a link to a page displaying the recipes.

Look at the recipes box here http://www.bbc.co.uk/programmes/b080z0hs with JS disabled for an example.

templates/recipes/show.ameninc.html.twig
```diff
@@ -15,7 +15,7 @@
             {%- if loop.last and total > 4 -%}
                 <div class="trailing-link br-box-secondary">
                     <div class="media__footer br-keyline">
-                        <a href="{{ path('programme_recipes', { pid: pid }) }}" class="icon-link link-complex 1/1 gel-pica-bold">
+                        <a href="{{ path('programme_recipes_amen', { pid: pid }) }}" class="icon-link link-complex 1/1 gel-pica-bold">
```
votemike: Won't this link to a partial? We want to link to an entire page. Right?

config/routes.yaml
```diff
@@ -45,10 +45,15 @@ programme_recommendations:
     defaults: { _controller: App\Controller\RecommendationsController, extension: '' }
     requirements: { pid: '[0-9b-df-hj-np-tv-z]{8,15}', extension: '$|\.2013inc' }
 
-programme_recipes:
-    path: /programmes/{pid}/recipes{extension}
-    defaults: { _controller: App\Controller\RecipesController, extension: '' }
```
votemike: What about if no extension is specified?

templates/recipes/show.ameninc.html.twig
tatitati: yeah. Is a question I have.
For instance, in the original we have (in the amen version of recipes):
https://sandbox.bbc.co.uk/programmes/b0876646/recipes.ameninc?__flush_cache=1

THe link ```see all recipes``` is pointing to:
https://sandbox.bbc.co.uk/programmes/b0876646/recipes

I'm a little confuse about it.

config/routes.yaml
tatitati: Original code:
```php
if ($extension === '') {
            // We haven't got around to rendering the main route yet,
            // replace this with a call to $this->renderWithChrome later.
            throw $this->createNotFoundException('No such page');
        }
```

src/Ds2013/Presenters/Domain/Recipe/RecipePresenter.php
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    public function getTitle() :string
+    {
+        return $this->recipe->getTitle();
+    }
+
+    public function getImage() :?RecipeImage
+    {
+        return $this->recipe->getImage() ?? null;
```
votemike: What does `$this->recipe->getImage()` return if there is no image?

config/routes.yaml
```diff
@@ -45,10 +45,15 @@ programme_recommendations:
...
-    requirements: { pid: '[0-9b-df-hj-np-tv-z]{8,15}', extension: '$|\.ameninc' }
+programme_recipes_amen:
+    path: /programmes/{pid}/recipes.ameninc
+    defaults: { _controller: App\Controller\Recipes\RecipesAmenController }
+    requirements: { pid: '[0-9b-df-hj-np-tv-z]{8,15}' }
+
+programme_recipes_ds2013:
+    path: /programmes/{pid}/recipes.inc
```
felipeparaujo: This should be `.2013inc`

templates/recipes/show.inc.html.twig
```diff
@@ -0,0 +1,25 @@
...
+            <ul class="grid-wrapper highlight-box-wrapper highlight-box-wrapper--grid@bpb1 highlight-box-wrapper--grid@bpw">
+                {%- for recipe in recipes -%}
+                    <li class="grid 1/2@bpb1 1/2@bpw">
+                        {{- ds2013('recipe', recipe) -}}
+                    </li>
+                {%- endfor -%}
+            </ul>
+        </div>
+        <div class="component__footer br-box-subtle">
+            <a class="component__footer__link" href="/food/programmes/{{ pid }}">
```
votemike: This link shouldn't be hardcoded.

src/Controller/Recipes/RecipesDs2013Controller.php
```diff
@@ -0,0 +1,14 @@
+<?php
+namespace App\Controller\Recipes;
+
+use App\ExternalApi\Recipes\Service\RecipesService;
+
+class RecipesDs2013Controller extends AbstractRecipesController
+{
+    public function __invoke(string $pid, RecipesService $recipesService)
+    {
+        $this->templatePath = 'recipes/show.inc.html.twig';
```
felipeparaujo: this should be show.2013inc.html.twig

src/Controller/Recipes/RecipesAmenController.php
```diff
@@ -0,0 +1,14 @@
+<?php
+namespace App\Controller\Recipes;
+
+use App\ExternalApi\Recipes\Service\RecipesService;
+
+class RecipesAmenController extends AbstractRecipesController
+{
+    public function __invoke(string $pid, RecipesService $recipesService)
```
votemike: Could this just be something like:
```
class RecipesAmenController extends AbstractRecipesController
{
    $this->templatePath = 'recipes/show.ameninc.html.twig';
}
```
If you rename the `execute` method to `__invoke`

src/Ds2013/Presenters/Domain/Recipe/RecipePresenter.php
felipeparaujo: Yeah, this should definetely be outside the constructor

templates/recipes/show.ameninc.html.twig
votemike: This line never specified the extension ameninc.

config/routes.yaml
votemike: But we will still want a route for recipes.

src/Ds2013/Presenters/Domain/Recipe/RecipePresenter.php
tatitati:
```php
public function testKnowNoExistChefImage()
{
    $recipe = RecipeBuilder::anyWithChef()->build(); // Recipe has no image
   
    $hasChefImage = (new RecipePresenter($recipe))->getChefImageUrl();

    $this->assertNull($hasChefImage);
}

File: templates/recipes/show.ameninc.html.twig (hunk 16 above)
Comment (felipeparaujo) [reply templates/recipes/show.ameninc.html.twig]: This should be linking to :pid/recipes. You don't have that route (or controller) anymore, so you should add those back. But keep the different includes and the full page in separate controllers as you have now, I think that's good.

File: templates/recipes/show.inc.html.twig [hunk 23]
```diff
@@ -0,0 +1,25 @@
+{% extends 'base_ds2013_partial.html.twig' %}
+
```
felipeparaujo: Rename `show.inc.html.twig` to `show.2013inc.html.twig` to keep in line with the rest of the includes

src/Ds2013/Presenters/Domain/Recipe/RecipePresenter.php
votemike: So. You could just do:
```
public function getImage(): ?RecipeImage
{
    return $this->recipe->getImage();
}
```

templates/recipes/show.ameninc.html.twig
BjornTwachtmann: Right... I can see why you're confused.

Actually looking at this, this code seems wrong in both the V2 and V3 brand page implementation, the correct impementation is on the episode page, see http://www.bbc.co.uk/programmes/b080z0hs. The "see all recipes" link should go to /food/programmes/pid. 

Unless I'm missing something,  @felipeparaujo , can you remeber why this was linking there?

templates/recipes/show.2013inc.html.twig
```diff
@@ -0,0 +1,25 @@
...
+            <ul class="grid-wrapper highlight-box-wrapper highlight-box-wrapper--grid@bpb1 highlight-box-wrapper--grid@bpw">
+                {%- for recipe in recipes -%}
+                    <li class="grid 1/2@bpb1 1/2@bpw">
+                        {{- ds2013('recipe', recipe) -}}
+                    </li>
+                {%- endfor -%}
+            </ul>
+        </div>
+        <div class="component__footer br-box-subtle">
+            <a class="component__footer__link" href="/food/programmes/{{ pid }}">
```
tatitati: ok, just a reminder, in show.ameinc.html.twig we have:
```href="{{ path('programme_recipes', { pid: pid }) }}"     ---> pointing to the full page``` 
But as bjorn  said it should be something like (no hardcoded):
```href="/food/programmes/{{ pid }}"   ----> pointing to an external page```

@felipeparaujo can you confirm something about this doubt?

File: templates/recipes/show.ameninc.html.twig (hunk 16 above)
Comment (felipeparaujo) [reply templates/recipes/show.ameninc.html.twig]: I just copied this over from v2. But we do have a {pid}/recipes page. If we're not linking to it from anywhere except the nav, should we keep it around at all?

File: src/Controller/Recipes/RecipesController.php [hunk 25]
```diff
@@ -0,0 +1,12 @@
+<?php
+namespace App\Controller\Recipes;
+
+class RecipesController extends AbstractRecipesController
+{
+    protected $templatePath = 'recipes/show.html.twig';
```
BjornTwachtmann: Is there any point in having this as a variable? It's basically hardcoded for each class, might as well be clear about that.

src/Ds2013/Presenters/Domain/Recipe/recipe.html.twig
```diff
@@ -0,0 +1,34 @@
+ <div class="recipe programme block-link highlight-box--list highlight-box--grid@bpb1 programme--grid@bpb1 highlight-box--grid@bpw programme--grid@bpw br-keyline br-blocklink-subtle br-subtle-linkhover-onbg015--hover">
+    <div class="programme__img 1@bpb1 1@bpw programme__img--hasimage recipe__img block-link br-blocklink-page">
+        {%- if recipe.getImage() -%}
+            {{-
+                ds_shared('image',
+                    recipe.getImage(),
+                    recipe.getOption('image_options')['default_width'],
```
BjornTwachtmann: You don't seem to be setting defaults for these in the recipe presenter. This will crash if the calling code does not set those options.

templates/find_by_pid/episode.html.twig
```diff
@@ -42,10 +42,28 @@
...
-                    {# TODO Recipes module #}
+                    {# Recipes module #}
+                    <div class="component component--box component--box-flushbody-vertical component--box--primary lazy-module"
+                         data-lazyload-inc="{{ path('programme_recipes_ds2013', {pid: programme.getPid() }) }}"
+                         data-lazyload-threshold="768"
+                         data-lazyload-delay="true">
+
+                        <div class="icon-box br-box-page">
+                            <a class="icon-box__link br-box-page__link br-page-link-onbg015 br-page-linkhover-onbg015--hover"
+                               href="{{ path('programme_recipes_ds2013', {pid: programme.getPid() }) }}">
```
BjornTwachtmann: Why are you linking the non-js user to partial mate? That's clearly wrong. This link is for non-JS users and bots so that they can see a page, it needs to link to /pid/recipes

templates/recipes/show.html.twig
```diff
@@ -0,0 +1,5 @@
+{% extends 'base_ds2013.html.twig' %}
+
+{% block body %}
+    <h1>Recipes</h1>
```
BjornTwachtmann: Do we have a separate ticket for creating the full page? Please check. If not, either create a ticket and put it in next sprint or do the work as part of this PR.

tests/Builders/ExternalApi/Recipes/ChefBuilderTest.php
BjornTwachtmann: Tests for the tests? How deep do we want this rabbit hole to go :D

BjornTwachtmann: Yeah, this looks weird to me

templates/recipes/show.ameninc.html.twig
```diff
@@ -15,7 +15,7 @@
             {%- if loop.last and total > 4 -%}
                 <div class="trailing-link br-box-secondary">
                     <div class="media__footer br-keyline">
-                        <a href="{{ path('programme_recipes', { pid: pid }) }}" class="icon-link link-complex 1/1 gel-pica-bold">
+                        <a href="/food/programmes/{{ pid }}" class="icon-link link-complex 1/1 gel-pica-bold">
```
BjornTwachtmann: Shouldn't this go in the third party routes file?

templates/find_by_pid/episode.html.twig
```diff
@@ -42,10 +42,28 @@
...
+
+                                <i class="icon-box__icon gelicon gelicon--info"></i>
```
BjornTwachtmann: <i> is from V2 and this will not render an icon in V3. Look around for how icons are done with the gelicon twig function .

templates/recipes/show.2013inc.html.twig
```diff
@@ -0,0 +1,37 @@
...
+        <div class="component__body br-box-page">
+            <ul class="grid-wrapper highlight-box-wrapper highlight-box-wrapper--grid@bpb1 highlight-box-wrapper--grid@bpw">
+                {%- for recipe in recipes -%}
+                    <li class="grid 1/2@bpb1 1/2@bpw">
+                        {{- ds2013('recipe', recipe, {
+                            'image_options' : {
+                                'classes' : '1/4@bpb1 1/4@bpb2 1/4@bpw hidden visible@bpb2 visible@bpw',
+                                'default_width' : 280,
+                                'is_lazy_loaded' : false,
+                                'sizes' : {
```
BjornTwachtmann: where did you get these sizes from? They look wrong, and blurrier than V2. 
V2 sizes are here, https://github.com/bbc/programmes/blob/develop/webapp/php/sharedmodules/programmes/views/partials/generic/recipe.spv#L9-L14

src/Controller/Recipes/RecipesController.php
```diff
@@ -0,0 +1,12 @@
...
+
+    protected function renderRecipes($templatePath, $dataForTemplate, $response)
```
votemike: What is `$response` for?

src/Ds2013/Presenters/Domain/Recipe/recipe.html.twig
```diff
@@ -0,0 +1,37 @@
+ <div class="recipe programme block-link highlight-box--list highlight-box--grid@bpb1 programme--grid@bpb1 highlight-box--grid@bpw programme--grid@bpw br-keyline br-blocklink-subtle br-subtle-linkhover-onbg015--hover">
+    <div class="programme__img 1@bpb1 1@bpw programme__img--hasimage recipe__img block-link br-blocklink-page">
+        {%- if recipe.getImage() -%}
+            {{-
+                ds_shared('image',
+                    recipe.getImage(),
+                    280,
+                    { 0: '50vw', 768: '25vw', 1008: '237px' },
```
votemike: Is this correct? It looks like the old way of doing image sizes.

templates/find_by_pid/episode.html.twig
```diff
@@ -42,10 +42,28 @@
...
+                    <div class="component component--box component--box-flushbody-vertical component--box--primary lazy-module"
+                         data-lazyload-inc="{{ path('programme_recipes_ds2013', {pid: programme.getPid() }) }}"
+                         data-lazyload-threshold="768"
+                         data-lazyload-delay="true">
+
+                        <div class="icon-box br-box-page">
+                            <a class="icon-box__link br-box-page__link br-page-link-onbg015 br-page-linkhover-onbg015--hover"
+                               href="{{ path('food_recipes', {pid: programme.getPid() }) }}">
+
+                                <i class="icon-box__icon gelicon gelicon--info"></i>
```
votemike: This seems like the old way of doing icons.

templates/find_by_pid/episode.html.twig
```diff
@@ -42,10 +42,28 @@
...
+                                <div class="icon-box__hgroup">
+                                    <h3 class="icon-box__title gamma">{{ tr('recipes') }}</h3>
+                                </div>
+                                <p class="icon-box__note micro">
+                                    See all recipes from {{ programme.getTitle() }}
```
votemike: Should this be translated?

src/Builders/ExternalApi/Recipes/ChefBuilder.php
```diff
@@ -0,0 +1,22 @@
...
+class ChefBuilder extends AbstractBuilder
+{
+    protected function __construct()
+    {
+        $faker = Factory::create();
+
+        $this->classTarget = Chef::class;
+        $this->blueprintConstructorTarget = [
+            'id' => $faker->word,
+            'name' => $faker->sentence(3),
```
votemike: Doesn't sentence put a full stop at the end? That would look weird for a name.

src/Ds2013/Presenters/Domain/Recipe/recipe.html.twig
```diff
@@ -0,0 +1,37 @@
+ <div class="recipe programme block-link highlight-box--list highlight-box--grid@bpb1 programme--grid@bpb1 highlight-box--grid@bpw programme--grid@bpw br-keyline br-blocklink-subtle br-subtle-linkhover-onbg015--hover">
+    <div class="programme__img 1@bpb1 1@bpw programme__img--hasimage recipe__img block-link br-blocklink-page">
```
votemike: Are `1@bpb1` and `1@bpw` real classes?

tests/Controller/RecipesControllerTest.php
```diff
@@ -49,7 +76,24 @@ public function testProgrammeHasNoRecipesInFoodApi()
...
+    public function testControllerAlsoWorkForD2013Design()
+    {
+        $json = file_get_contents(__DIR__ . '/../DataFixtures/JSON/bakeoff.json');
+        $response = new Response(200, [], $json);
+        $client = $this->createClientWithMockedGuzzleResponse($response);
+
+        $crawler = $client->request('GET', 'programmes/b013pqnm/recipes.2013inc');
+
+        $this->assertResponseStatusCode($client, 200);
+        $this->assertCount(4, $crawler->filter('li'), 'Expected 4 recipes');
```
votemike: Is check for li's a specific enough test?

votemike:
The gap between the recipe title and the chef image seems too large.

programmes-developers:

votemike:
Test me please

PR #503, opened 2018-03-28
First pass at adding Segment find-by-pid pages

BPScott:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6309

Adds the details section at the top of the page, and requests all data
required in the Controller.

We still need to format the output of programmes the Segement appeared
on, and work out how to deal with the case when a Segment appears on
multiple programmes within the same Network.


programmes-developers:

votemike:
@felipeparaujo are you already doing this page?

templates/find_by_pid/segment.html.twig
```diff
@@ -0,0 +1,55 @@
+{% extends 'base_ds2013.html.twig' %}
+
+{% block page_classes %}programmes-page text-base programmes-page--flush{% endblock %}
+{% block title %}{{ meta_context.titlePrefix() ? meta_context.titlePrefix() ~ ' - ' }}{{ segment.getTitle() }}{% endblock %}
+
+{% block body %}
+<div class="b-g-p">
+<div class="map__column br-box-secondary">
```
votemike: Why is this div on the same level as the one above? This cause a weird PR if someone comes to auto format this file at a later date.

templates/find_by_pid/segment.html.twig
```diff
@@ -0,0 +1,55 @@
...
+    <div class="grid-wrapper">
+        <div class="grid 2/3@bpw2 3/4@bpe">
+            <div class="island">
+                <h1>{{ segment.getTitle() }}</h1>
```
votemike: Will the branding not have an H1 in it?

templates/find_by_pid/segment.html.twig
BPScott: Will Fix.

templates/find_by_pid/segment.html.twig
BPScott: No, the branding has no headers tags in it - as just the programme title would be a poor heading for all pages that hang off the a findbypid page e.g. for :pid/episodes you'd want "Episode Guide" in the heading.

Brands and Episodes have their <h1> in their MAP panels.

BPScott:
AFAIK @felipeparaujo is working on "on the episode page, show a list of segments on this episode", this PR is for "on the segment page, give me info about the segment and the programmes it appears on".

programmes-developers:

votemike:
Test me please

PR #505, opened 2018-03-29
Fix

votemike:
Looks like it was a rebase error


programmes-developers:

PR #506, opened 2018-03-29
PROGRAMMES-6294 Remove DOM node if partial is empty

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6294

Change recipies partial to return a [204](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#2xx_Success) instead of 404 if there are no recipes. This stops the console or network tabs showing errors.
Once I had made this change, the previous DOM element was no longer hidden. So I changed the JS to remove the DOM node if the ajax call returns no data.

Good to look at this PR with [?w=1](https://github.com/bbc/programmes-frontend/pull/506/files?w=1) on the end of the URL.

To test, check out branch. Run GULP. Visit https://production.test.bbc.co.uk/isite2-xforms/fr/progs-factuallondon/programmes-options/edit/102b4b12-11c0-36eb-896d-341df8d7b4e7 and make sure there are no recipes, but "Display Recipes Widget?" is set to "Yes". Then visit http://sandbox.bbc.co.uk/programmes/b00cyj3f.


programmes-developers:

PR #507, opened 2018-03-29
PROGRAMMES-6231 Adding Episode Page supporting content

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6231

Can test on http://sandbox.bbc.co.uk/programmes/b071vl2l or http://sandbox.bbc.co.uk/programmes/b09tnfj7


templates/find_by_pid/episode.html.twig
```diff
@@ -91,7 +91,29 @@
...
-                    {# TODO Features module #}
+                    {% if supportingContentItems %}
+                        <div id="features" class="feature-list">
+                            {% for item in supportingContentItems %}
+                                <div class="component component--box component--box-flushbody-vertical component--box--primary component--feature">
+                                    <div class="component__header br-box-page">
+                                        <h2>{{ item.getTitle() }}</h2>
+                                    </div>
+                                    <div class="component__body br-box-page">
+                                        <div class="feature br-keyline">
```
votemike: Would we like some of this HTML to be pulled out in to a presenter?

programmes-developers:

templates/find_by_pid/episode.html.twig
BjornTwachtmann: I'd probably pull some of this out into a SupportingContentPresenter, that's what I did for the dsamen stuff. I think this file is getting a bit too big anyway.

BjornTwachtmann:
Wouldn't it make more sense for the supporting content presenter to present a single item? That's how pretty much all the other presenters do it.

votemike:
Doing that will mean some of the HTML classes are outside of the feature. Should I spilt out the CSS for those classes in to a separate CSS file? Or should I just move the whole CSS file back in to _features?

PR #509, opened 2018-04-03
PROGRAMMES-6307 fix css

tatitati:
This fix exactly this problem in the ticket (there are a few more):
```when the page is >769px the Last On column has a little bit of extra margin at the top (http://bbc.co.uk/programmes/b09pjgly)```


programmes-developers:

PR #510, opened 2018-04-03
Fix ie11 placeholders schedules

tatitati:
# dont review please

<details>
<summary>EXPANDED EXPLANATION</summary>
After a little more research I really open the same PR than before. The reasons are as follow:

I could isolate the scenario with a very simple html and replicate the problem with just a few simple divs and links and the problem in JAWS was replicated successfully.

# About navigation using only TAB.
As a reminder this is the list of elements that can be focusable:
```
  a[href]:not([tabindex='-1']),
  area[href]:not([tabindex='-1']),
  input:not([disabled]):not([tabindex='-1']),
  select:not([disabled]):not([tabindex='-1']),
  textarea:not([disabled]):not([tabindex='-1']),
  button:not([disabled]):not([tabindex='-1']),
  iframe:not([tabindex='-1']),
  [tabindex]:not([tabindex='-1']),
```

From here we can see that our nav links in shcedules (morning, evening, late...) are pointing to ```<li>``` elements. We can say then that these elements are not focusable using TABs. However all the focusable elements nested inside are still focusable. This said, you cannot focus a whole period, but instead, we focus programmes-links, etc. All this using TAB navigation.

# How can we make a no-focusable element into a focusable?

There are two ways: 
1. We can force a no-focusable element to behave like a normal focusable when using tabs with tabindex="0". Because these elements will be as any normal focusable elements, they can be focused as well by using #links or JS.

2. We can force a no-focusable element to focus when using JS or #links. We won't be able to focus it using TAB as a normal focusable element. This focus is named "programatically". This is done with tabindex="-1". As you can see this is similar to the previous one, but more restricted to only #links and JS.
</details>

----

# Outcome:
I created this HTML to isolate the problem: http://pasted.co/3d7b1e1f
As I could see IE-11 and Chrome work properly, but there is something a little hackie on both. The inconsistencty is that they can focus divs (a no focusable element by default) when using #links to focus them

To visualize this weird behaviour I added the next css rule to the example:
```css
:focus {
  border: 5px solid red;
}
```

What I saw was that the border was visible around divs forced to be focusables with tabindex=0 or -1, but the border was not visible at all when the div has not any tabindex, however the #link took me to the right section of the page, but not focusing (BUG?, inconsistent?, if is not focusing it, why is taking me there?).

JAWS seems to be more rigid, and detect this kind of inconsistency, which makes sense, for JAWS a no-focusable element is a no-focusable element. Finish. 
Cause of that JAWS takes you back to the original place again.

# why tabindex="-1"
The reason is that tabindex="-1" make focusable any no-focusable element by using #links or JS, and this happens to our ```<li>``` elements, they are not focusable by tab navigation and we want them to be focusable only by clicking in the #periodlink  we want . So we keep the original behaviour. 

Tabindex="-1" still allows to focus any focusable element nested in the container. The only one that is not focusable is the whole container.

Feel free to try the example I created. Is easy to test in chrome and IE-11


programmes-developers:

PR #511, opened 2018-04-03
Add a replacement text when there is no MAP. Fix favourites button be…

BjornTwachtmann:
…ing wrongly rendered when no MAP. Make sure recipes render even if there are no promos/galleries/clips/etc

See https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6315 .


programmes-developers:

PR #512, opened 2018-04-04
PROGRAMMES-6307: Fix map episode css height

tatitati:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6307

This fix this issue in the ticket:
```when the page is >769px the columns on the right could be longer than the ones on the left (http://bbc.co.uk/programmes/b09pjgly)```

**Image in the ticket: bug2.png**

### CODE REVIEW

**Comment by programmes-developers** [comment]

File: src/Ds2013/Presenters/Section/Episode/Map/episode_map.html.twig [hunk 1]
```diff
@@ -3,7 +3,7 @@
 
         {#PLAYOUT#}
         <div class="grid 2/3@bpw2 2/3@bpe map__column map__column--first">
-            <div class="map__inner grid__inner--flush br-box-secondary">
+            <div class="map__inner grid__inner--flush br-box-secondary fauxcolumn">
```
BjornTwachtmann: Can you remove the grid__inner here as well please, it doesn't exist in V3. The fauxcolumn thing is fine, that's the right way to solve it I think.

PR #514, opened 2018-04-05
PROGRAMMES-6259 - Music Segments

felipeparaujo:
Snippet playback will be done in a separate PR


programmes-developers:

src/Controller/FindByPid/EpisodeController.php
```diff
@@ -116,12 +116,7 @@ public function __invoke(
             }
         }
 
-        $favouritesButtonPromise = new FulfilledPromise(null);
-        if ($episode->isRadio()) {
```
votemike: Does this mean the favourite button will appear in the MAP for TV episodes now?

src/Controller/AncillaryPages/SegmentsController.php
```diff
@@ -0,0 +1,77 @@
...
+        ProgrammesService $programmesService,
+        SegmentEventsService $segmentEventsService,
+        CollapsedBroadcastsService $collapsedBroadcastsService,
+        VersionsService $versionsService,
+        FavouritesButtonService $favouritesButtonService,
+        PresenterFactory $presenterFactory,
+        CanonicalVersionHelper $canonicalVersionHelper
+    ) {
+        /** @var ProgrammeItem $programmeItem */
+        $programmeItem = $programmesService->findByPidFull(new Pid($pid), 'ProgrammeItem');
```
votemike: Why does the argument resolver not just give you a programme?

src/Controller/AncillaryPages/SegmentsController.php
```diff
@@ -0,0 +1,77 @@
...
+                    return $this->createNotFoundException('Segments not found');
+                }
+            }
+        }
+
+        $resolvedPromises = $this->resolvePromises(['favouritesButton' => $favouritesButtonService->getContent()]);
+
+        return $this->renderWithChrome('ancillary_pages/segments.html.twig', [
+            'programme' => $programmeItem,
+            'segmentsListPresenter' => $segmentsListPresenter,
```
votemike: What is displayed if this is null?

src/Controller/AncillaryPages/SegmentsController.php
```diff
@@ -0,0 +1,77 @@
...
+                $segmentEvents = $segmentEventsService->findByVersionWithContributions($canonicalVersion);
+                if ($segmentEvents) {
+                    $segmentsListPresenter = $presenterFactory->segmentsListPresenter(
+                        $programmeItem,
+                        $segmentEvents,
+                        $upcomingBroadcast,
+                        $lastOnBroadcast
+                    );
+                } else {
+                    return $this->createNotFoundException('Segments not found');
```
votemike: Do you return Exceptions? Don't you throw them?

src/Controller/Styleguide/Ds2013/Domain/MusicSegmentsController.php
```diff
@@ -0,0 +1,87 @@
...
+use App\Builders\ContributorBuilder;
+use App\Builders\EpisodeBuilder;
+use App\Builders\MusicSegmentBuilder;
+use App\Builders\SegmentEventBuilder;
+use App\Controller\BaseController;
+use App\Ds2013\Presenters\Section\Segments\SegmentsListPresenter;
+use App\DsShared\Helpers\LiveBroadcastHelper;
+use App\DsShared\Helpers\PlayTranslationsHelper;
+
+class MusicSegmentsController extends BaseController
```
votemike: Going to do any more with this?

src/Controller/Styleguide/Ds2013/Domain/SegmentsController.php
```diff
@@ -0,0 +1,16 @@
...
+
+namespace App\Controller\Styleguide\Ds2013\Domain;
+
+use App\Controller\BaseController;
+
+class SegmentsController extends BaseController
+{
+    public function __invoke()
+    {
+        return $this->renderWithChrome('styleguide/ds2013/domain/segments.html.twig', [
```
votemike: Going to do any more with this?

Tasiobg:
This page is redirecting http://sandbox.bbc.co.uk/programmes/b093sdk0
but this one is throwing a 500 http://sandbox.bbc.co.uk/programmes/b093sdk0/segments
Can someone else confirm this is not just me?

src/Ds2013/Presenters/Section/Segments/SegmentItem/PopularMusicPresenter.php
```diff
@@ -0,0 +1,31 @@
...
+                $this->primaryContributions[] = $contribution;
+            } elseif ($role === 'featured artist') {
+                $this->featuredContributions[] = $contribution;
+            } elseif ($role === 'vs artist') {
+                $this->versusContributions[] = $contribution;
+            } else {
+                $this->otherContributions[] = $contribution;
+            }
+        }
+
```
BjornTwachtmann: The lists created above need to be de-duplicated by contributor PID. Have a look at /programmes/b00tfsgc and scroll down to "Hit Me With Your Rhythm Stick", you'll see the same name 3 times.

src/Ds2013/Presenters/Section/Segments/SegmentItem/ClassicalMusicPresenter.php
```diff
@@ -0,0 +1,23 @@
...
+        /** @var Contribution $contribution */
+        foreach ($this->segment->getContributions() as $contribution) {
+            if (empty($this->primaryContributions) && strtolower($contribution->getCreditRole()) === 'composer') {
+                $this->primaryContributions[] = $contribution;
+            } else {
+                $this->otherContributions[] = $contribution;
+            }
+        }
+
+        \assert(!empty($this->primaryContributions));
```
BjornTwachtmann: If there's no composer, the first "other" contribution should be moved into the primary. Also, what's this assert() about, looks weird. Should we be throwing 500s if there are no contributions to a segment?

src/Controller/FindByPid/EpisodeController.php
BjornTwachtmann: If you're changing this here, you chould change this https://github.com/bbc/programmes-frontend/blob/develop/templates/find_by_pid/episode.html.twig#L236 to check for isRadio , we don't want the JS loaded where it isn't used.

src/Ds2013/Presenters/Section/Segments/SegmentItem/AbstractMusicSegmentItemPresenter.php
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    public function hasRecordDetails(): bool
+    {
+        return $this->segment->getReleaseTitle() || $this->segment->getRecordLabel() || $this->segment->getCatalogueNumber() || $this->segment->getTrackNumber();
+    }
+
+    public function getTemplatePath(): string
+    {
+        return '@Ds2013/Presenters/Section/Segments/SegmentItem/music.html.twig';
```
BjornTwachtmann: Tasio added a thing that overrides the template path and variable to that used by another presenter, see https://github.com/bbc/programmes-frontend/blob/develop/src/Ds2013/Presenters/Domain/CoreEntity/Programme/BroadcastProgrammePresenter.php#L25 . Would that work here? Might as well try and use it everywhere.

programmes-developers:

src/Ds2013/Presenters/Section/Segments/SegmentItem/AbstractMusicSegmentItemPresenter.php
Tasiobg: Not sure if this could work with a non existing class.
May be overriding `snakeCasePresenterName()` and returning something like "music" could be a good solution.
Please keep in mid that refactorizations could break the code if we use hardoded URIs, I suggest to add unitest to check if URI exist on those cases.

src/Ds2013/Presenters/Section/Segments/SegmentItem/AbstractMusicSegmentItemPresenter.php
```diff
@@ -0,0 +1,162 @@
...
+        $segment = $this->segmentEvent->getSegment();
+        $this->segment = $segment;
+
+        $this->setupContributions();
+    }
+
+    public function getImageUrl(): string
+    {
+        if ($this->segment->getMusicRecordId()) {
+            return 'https://www.bbc.co.uk/music/images/records/96x96/' . $this->segment->getMusicRecordId();
```
Tasiobg: Two suggestion:
1.-
if possible, add unitest just to make sure this resource exist, hardcoded URI are not easy to maintain.

2.-
What do you think about adding a configuration in the parameters.yaml or somewhere else with the different extrenal URL we use across the project? example: base_url: bbc.co.uk
This has worked fine for me in the past making code more easy to maintain

src/Ds2013/Presenters/Section/Segments/SegmentItem/_segment.scss
```diff
@@ -0,0 +1,201 @@
+$play-box-size: 40px;
+$image-width: 88px;
+
+.segment {
```
Tasiobg: Do you think it would make sense to wrap all this CSS inside a parent element to decrease the chances of conflicting with other element styles?

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
```diff
@@ -257,8 +269,19 @@ private function createSegmentItem(array $segmentEvents, int $start): AbstractSe
...
+                $this->hasMusicSegmentItems = true;
+                $options['context_pid'] = (string) $this->context->getPid();
+
+                if ($this->isClassicalMusicSegment($segmentEvent->getSegment())) {
+                    $presenters[] = new ClassicalMusicPresenter($segmentEvent, $this->getTimingType(), $this->collapsedBroadcast, $options);
+                } else {
+                    $presenters[] = new PopularMusicPresenter($segmentEvent, $this->getTimingType(), $this->collapsedBroadcast, $options);
+                }
             } else {
+                if ($segmentEvent->getSegment()->getType() === 'chapter') {
```
Tasiobg: would it make sense to have the types as a class constants?

src/Controller/FindByPid/EpisodeController.php
votemike: I believe we need to load the Javascript all the time there is a favourites button now incase the page has MusicSegments.

votemike:
@Tasiobg /programmes/b093sdk0/segments redirects for me....

src/Ds2013/Presenters/Section/Segments/SegmentItem/AbstractMusicSegmentItemPresenter.php
votemike: Overriding `snakeCasePresenterName()` seems a bit hacky. It is also private, probably because it would be hacky to override it.

src/Ds2013/Presenters/Section/Segments/SegmentItem/AbstractMusicSegmentItemPresenter.php
votemike: 1. What do you mean by add a unit test?
2. That would be best, but then they'd need to be passed down in to this file. So I think leaving them as-is should be fine.

src/Ds2013/Presenters/Section/Segments/SegmentItem/_segment.scss
votemike: Is segment not specific enough?

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
votemike: Felipe and I wondered about this. And decided that there was very little benefit.

src/Controller/FindByPid/EpisodeController.php
BjornTwachtmann: That JS can fuck up the page layout when it's run without the correct elements being present in the page, the favourites button ends up in the whitespace below the bottom of the page sometimes. We can either get them to fix that, or only load the JS when we know it's needed.

src/Ds2013/Presenters/Section/Segments/SegmentItem/AbstractMusicSegmentItemPresenter.php
Tasiobg: I think hardcoded value is fine if we have a mechanims that checks if this template can be loaded by twig, example using unitest for this presenter or unitest in a controller who is using this resource.
Would it be worth to create a generic funtion in tests to verify if all twig resources can be loaded by twig?

src/Ds2013/Presenters/Section/Segments/SegmentItem/ClassicalMusicPresenter.php
```diff
@@ -0,0 +1,30 @@
...
+            if (empty($this->primaryContributions) && strtolower($contribution->getCreditRole()) === 'composer') {
+                $this->primaryContributions[] = $contribution;
+            } else {
+                $this->otherContributions[] = $contribution;
+            }
+        }
+
+        if (empty($this->primaryContributions)) {
+            if (empty($this->otherContributions)) {
+                throw new Exception('There must be a primary contributor for music segments.');
```
BjornTwachtmann: This should not be an exception, having segments without contributions is something that actually happens. In this case, V2 falls back to displaying the segment title only, see /programmes/b00tfsf6 .

src/Ds2013/Presenters/Section/Segments/SegmentItem/PopularMusicPresenter.php
```diff
@@ -0,0 +1,54 @@
...
+                if (!isset($otherContributors[$contributorPid])) {
+                    $otherContributors[$contributorPid] = true;
+                    $this->otherContributions[] = $contribution;
+                }
+            }
+        }
+
+        if (empty($this->primaryContributions)) {
+            if (empty($this->otherContributions)) {
+                throw new Exception('There must be a primary contributor for music segments.');
```
BjornTwachtmann: Same comment as above, should not be an exception.

src/Ds2013/Presenters/Section/Segments/SegmentItem/_segment.scss
BjornTwachtmann: I think segment is specific enough. We tend to avoid nesting styles more deeply than is necessary, it just bulks up the CSS and makes it harder to re-use.

tests/Ds2013/Presenters/Section/Segments/SegmentItem/PopularMusicPresenterTest.php
```diff
@@ -0,0 +1,82 @@
...
+        $performer = $this->createConfiguredMock(Contribution::class, [
+            'getCreditRole' => 'performer',
+            'getContributor' => $this->createConfiguredMock(Contributor::class, ['getPid' => new Pid('cnt000002')]),
+        ]);
+        $transcriber = $this->createConfiguredMock(Contribution::class, [
+            'getCreditRole' => 'transcriber',
+            'getContributor' => $this->createConfiguredMock(Contributor::class, ['getPid' => new Pid('cnt000003')]),
+        ]);
+
+        return [
```
BjornTwachtmann: Unless I'm reading this wrong, it doesn't look like any of these test duplicate contributions from the same contributor are being deduped, can you add a test for that?

src/Controller/FindByPid/EpisodeController.php
votemike: Added

src/Ds2013/Presenters/Section/Segments/SegmentItem/AbstractMusicSegmentItemPresenter.php
votemike: Done

tests/Ds2013/Presenters/Section/Segments/SegmentItem/PopularMusicPresenterTest.php
votemike: Added

src/Ds2013/Presenters/Section/Segments/SegmentItem/ClassicalMusicPresenter.php
```diff
@@ -0,0 +1,26 @@
...
+class ClassicalMusicPresenter extends AbstractMusicSegmentItemPresenter
+{
+    public function setupContributions(): void
+    {
+        /** @var Contribution $contribution */
+        foreach ($this->segment->getContributions() as $contribution) {
+            if (empty($this->primaryContributions) && strtolower($contribution->getCreditRole()) === 'composer') {
+                $this->primaryContributions[] = $contribution;
+            } else {
+                $this->otherContributions[] = $contribution;
```
BjornTwachtmann: This list still needs to be deduplicated by contribuor PID like the Popular contributions are

PR #515, opened 2018-04-05
Fix for brittle test

tatitati:
Working in my new ticket I saw that EpisodeBuilder was creating episodes of the same type "audio" always. I fixed it. After the fix we were creating Episodes with different mediay types, randomly chosen. That was good but suddenly I saw that a test about displaying the duration of CTA was brittle, breaking for some cases (now we had to be more explicit about what was being tested because the media_type was changing). The behaviour of CTA is quite complex, we cannot have brittle tests on this


programmes-developers:

tests/DsAmen/Presenters/Domain/CoreEntity/Programme/ProgrammeTemplateTest.php
```diff
@@ -211,29 +211,37 @@ public function testEpisodeWithStandaloneCta()
         $this->assertSame('30:00', $iconLabel->children()->eq(1)->text());
     }
 
-
-    public function testTVEpisodeDisplayDuration()
+    /**
+     * @dataProvider mediaTypeProvider
+     */
+    public function testRadioEpisodeNotDisplayDuration($programmeBuilder, $mediaType, $expectDurationBeingDisplayed)
```
votemike: This mentions radio. But accept TV media types.

PR #517, opened 2018-04-06
PROGRAMMES-6317 Fixing height bug

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6317

The longest synopses's create divs around 1600px high.  
However, if there are a lot of carriage returns, shorter synopses's can cause even larger divs. The example in the ticket makes the div 2216px high.  
So bumping this up to be even higher.


programmes-developers:

PR #518, opened 2018-04-06
restore change. Tx column css height issue

tatitati:
I did a fix that was looking good, but today I saw that still some ede cases were failing. I prefer to restore everything (if I go for holidays this will make clearer that still has not been fixed).

Original PR:
https://github.com/bbc/programmes-frontend/pull/512


programmes-developers:

votemike:
Have you marked the ticket as not fixed?
Which PR was this from?

BjornTwachtmann:
what's failing here @tatitati , reverting it seems weird

votemike:
What is a fauxcolumn anyway?

tatitati:
After the "fix" the height of the columns were correct, but when you reduce the width happen this:

## Without fauxcolumn
withou-fauxcolumn



## With:
withfauxcolumn

PR #429, opened 2018-02-21
PROGRAMMES-6204 Adding tabindex of -1 to skip-to-content target to fi…

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6204

If in doubt...... add a tabindex of -1.

PR #524, opened 2018-04-18
PROGRAMMES-6244 Fixing accessibility of skip to content link

votemike:
A second attempt at https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6244

The Skip to Content link points to #programmes-content.  
Clicking it moves the page to the element with the id "programmes-content".  
[In IE (with JAWS), it does something iffy with the focus as tabbing moves you to the next link after the link for #programmes-content rather than the link after the element with id "programmes-content".](https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6204) To remedy this I [added a tabindex of -1 to the element](https://github.com/bbc/programmes-frontend/pull/429/files).
Now, it seems if you navigate to an element with a tabindex of -1 in Chrome, the entire contents of that element is read out.
So I have created an empty div to navigate to. This should work in both Chrome (as there is nothing to read out) and IE (as the element still has a tabindex of -1).

FYI, this also seems to be how BBC News deal with the problem (although their element is an H1).


programmes-developers:

PR #528, opened 2018-04-23
PROGRAMMES-6287 related topics and programmes

benolot:
Ticket: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6287

@tatitati, I removed the fauxcolumn class as it didn't seem to be doing anything, but broke mobile styling. I also removed the related tests.


tests/ExternalApi/Ada/Mapper/AdaProgrammeMapperTest.php
```diff
@@ -0,0 +1,59 @@
...
+    {
+        $adaEntity = [
+            'id' => 'Fellows_of_the_Royal_Society',
+            'type' => 'category',
+            'title' => 'Fellows of the Royal Society',
+            'image' => 'ichef.bbci.co.uk/images/ic/$recipe/p01l7xx5.jpg',
+            'programme_items_count' => 16,
+        ];
+
+        $mapper = new AdaClassMapper();
```
votemike: This class/file is called AdaProgrammeMapperTest, but you're testing a AdaClassMapper

src/ExternalApi/Ada/Service/AdaClassService.php
```diff
@@ -169,4 +170,33 @@ private function deduplicateClasses(array $classes): array
         }
         return $uniqueClasses;
     }
+
+    private function requestRelatedProgrammeItems(string $pid, $scope = null, Pid $countContextPid = null, $threshold = null, $order = null, $orderDirection = null, $limit = null, $page = 1)
```
votemike: Should this method be in this file?

BjornTwachtmann: This looks like it's been added here and to AdaProgrammesService. It shouldn't be here.

src/ExternalApi/Ada/Service/AdaClassService.php
```diff
@@ -64,21 +65,21 @@ public function __construct(
 
     /**
      * @param Programme $programme
+     * @param int $limit
      * @param bool $countWithinTleo
      * @return PromiseInterface (Promise return AdaClass[] when unwrapped)
      */
     public function findRelatedClassesByContainer(
         Programme $programme,
+        int $limit,
```
BjornTwachtmann: If you're adding a new parameter like this, make it the last one and give it a default. Default should be 5, as that's what the previous hardcoded limit was. Note also that your changed invocation for this in TlecController has the limit as the last parameter.

src/ExternalApi/Ada/Service/AdaProgrammeService.php
```diff
@@ -0,0 +1,194 @@
...
+        $this->mapper = $mapper;
+        $this->programmesService = $programmesService;
+    }
+
+    /**
+     * @param Programme $programme
+     * @param bool $countWithinTleo
+     * @return PromiseInterface (Promise return AdaClass[] when unwrapped)
+     */
+    public function findRelatedClassesByContainer(
```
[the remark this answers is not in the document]
BjornTwachtmann: This doesn't belong in here, it looks copy/paste from AdaClassService

src/ExternalApi/Ada/Service/AdaProgrammeService.php
```diff
@@ -0,0 +1,194 @@
...
+        );
+
+        $uniqueProgrammes = array_unique(array_merge($resolvedPromises['relatedByTag'], $resolvedPromises['relatedByBrand'], $resolvedPromises['relatedByCategory']), SORT_REGULAR);
+
+        return \array_slice($uniqueProgrammes, 0, 3);
+    }
+
+    private function makeClient(Programme $programme, bool $countWithinTleo): HttpApiClient
+    {
+        $limit = 5;
```
BjornTwachtmann: Again,  looks like copy/paste from AdaClassService.

src/ExternalApi/Ada/Service/AdaProgrammeService.php
```diff
@@ -0,0 +1,194 @@
...
+    {
+        $page = intval($page);
+        $limit = is_numeric($limit) ? intval($limit) : '10';
+
+        $url = $this->baseUrl . '/programme_items/' . $pid . '/related?page=' . $page . '&page_size=' . $limit .
+            ($countContextPid ? '&programme=' . $countContextPid : '') .
+            ($scope ? '&scope=' . $scope : '') .
+            ($threshold ? '&threshold=' . $threshold : '') .
+            ($order ? '&order=' . $order : '');
+        ($orderDirection ? '&direction=' . $orderDirection : '');
```
BjornTwachtmann: This line doesn't do anything, you've terminated the string concatenation on the last line.

src/ExternalApi/Ada/Service/AdaProgrammeService.php
```diff
@@ -0,0 +1,194 @@
...
+        Programme $programme,
+        bool $countWithinTleo = true
+    ): PromiseInterface {
+        $client = $this->makeClient($programme, $countWithinTleo);
+        return $client->makeCachedPromise();
+    }
+
+    public function findSuggestedByProgrammeItem(Programme $programme): array
+    {
+        $resolvedPromises = \GuzzleHttp\Promise\unwrap(
```
BjornTwachtmann: This isn't really optimal. In the callback for each of these requests, you're making a DB query. Ultimately, there only needs to be one DB query with the 3 programmes we end up using. 

Guzzle promises are tricky fuckers, and not that clearly documented. I will attempt to set up a branch that does this and put in a link.

BjornTwachtmann: Okay, have a look at https://github.com/bbc/programmes-frontend/compare/PROGRAMMES-6287-related-topics-and-programmes...more-of-bjorns-broken-code?w=1 . 

That was more annoying than I thought. I do not regret being such a smartarse about this. However I do both rue and lament it.

src/Ds2013/Presenters/Domain/CoreEntity/Programme/SubPresenters/programme_body.html.twig
```diff
@@ -26,3 +26,7 @@
 {% endif %}
 {# This exists solely to be extended. @TODO revisit once this is more fleshed out #}
 {% block body_extra %}{% endblock %}
+
+{% if programme_body.getOption('body_override') %}
```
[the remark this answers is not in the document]
benolot: Would be blank without the override, so I guess adding to it?

src/Ds2013/PresenterFactory.php
```diff
@@ -15,6 +15,8 @@
 use App\Ds2013\Presenters\Section\Episode\Map\EpisodeMapPresenter;
 use App\Ds2013\Presenters\Section\EpisodesSubNav\EpisodesSubNavPresenter;
 use App\Ds2013\Presenters\Section\Footer\FooterPresenter;
+use App\Ds2013\Presenters\Section\RelatedProgrammes\RelatedProgrammesPresenter;
```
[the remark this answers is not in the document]
benolot: Removed in next commit

src/ExternalApi/Ada/Domain/AdaProgrammeItem.php
```diff
@@ -0,0 +1,73 @@
+<?php
+declare(strict_types=1);
+
+namespace App\ExternalApi\Ada\Domain;
+
+use BBC\ProgrammesPagesService\Domain\Entity\Image;
```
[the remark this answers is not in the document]
benolot: Yes, it's used for getImage()

src/ExternalApi/Ada/Domain/AdaProgrammeItem.php
```diff
@@ -0,0 +1,73 @@
...
+use BBC\ProgrammesPagesService\Domain\Entity\Programme;
+use BBC\ProgrammesPagesService\Domain\ValueObject\Pid;
```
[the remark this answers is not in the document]
benolot: No - removed.

src/ExternalApi/Ada/Service/AdaProgrammeService.php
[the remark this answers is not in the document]
benolot: Removed in next commit

src/ExternalApi/Ada/Mapper/AdaClassMapper.php
```diff
@@ -9,7 +9,7 @@
 
 class AdaClassMapper
 {
-    public function mapItem(array $adaClass, ?string $countContextPid): AdaClass
+    public function mapItem(array $adaClass, ?string $countContextPid = null): AdaClass
```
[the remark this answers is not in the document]
benolot: I couldn't remember the reason, I think it was a hacky fix to something that should have been fixed properly. Have now fixed the issue properly in next commit.

benolot:
Test this please

src/ExternalApi/Ada/Service/AdaClassService.php
benolot: I had put it as second parameter as it didn't have a default (and in PSR2, parameters with default's go to the bottom), but I've updated this now.

tests/ExternalApi/Ada/Mapper/AdaProgrammeMapperTest.php
benolot: Good spot, it was this way for all the new AdaProgramme* tests, so have fixed

tests/ExternalApi/Ada/Mapper/AdaProgrammeMapperTest.php
```diff
@@ -0,0 +1,65 @@
...
+        $this->assertEquals($expectedEntity, $mapper->mapItem($adaEntity, 'b0000001'));
+    }
+
+    public function testMapItemWithNullContext()
+    {
+        $programme = $this->createMock(Programme::class);
+
+        $adaEntity = [
+            'pid' => '01l7xx5',
+            'type' => 'category',
```
votemike: Is this used?

tests/ExternalApi/Ada/Mapper/AdaProgrammeMapperTest.php
```diff
@@ -0,0 +1,65 @@
...
+            'title' => 'Fellows of the Royal Society',
+            'image' => 'ichef.bbci.co.uk/images/ic/$recipe/p01l7xx5.jpg',
```
votemike: Is this used?

tests/ExternalApi/Ada/Service/AdaProgrammeServiceTest.php
```diff
@@ -0,0 +1,161 @@
...
+
+        // Assert an empty result was stored in the cache
+        $validCacheValues = $this->validCacheValues();
+        $this->assertCount(1, $validCacheValues);
+        $this->assertContains([], $validCacheValues);
+    }
+
+    private function service($client)
+    {
+        return new AdaClassService(
```
votemike: This file is supposed to test the AdaProgrammeService, not the AdaClassService. Right?

src/ExternalApi/Ada/Service/AdaProgrammeService.php
```diff
@@ -0,0 +1,143 @@
...
+        $relatedProgrammes = [];
+        foreach ($uniqueProgrammes as $key => $item) {
+            $relatedProgrammes[] = $this->mapper->mapItem($programmeItems[$key], $item);
+        }
+
+        $this->cache->setItem($cacheItem, $relatedProgrammes, CacheInterface::MEDIUM);
+        return $relatedProgrammes;
+    }
+
+    private function requestRelatedProgrammeItems(string $pid, $scope = null, Pid $countContextPid = null, $threshold = null, $order = null, $orderDirection = null, $limit = null, $page = 1)
```
votemike: It seems this is always called with a threshold of 2.

votemike: It seems this is always called with an order of 'random'.

votemike: It seems this is always called with an orderDirection of null.

src/ExternalApi/Ada/Service/AdaProgrammeService.php
```diff
@@ -0,0 +1,143 @@
...
+    {
+        $page = intval($page);
+        $limit = is_numeric($limit) ? intval($limit) : '10';
+
+        $url = $this->baseUrl . '/programme_items/' . $pid . '/related?page=' . $page . '&page_size=' . $limit .
```
votemike: This seems to built put together in much the same way as the call in AdaClassService. Is it worth creating an abstract class for these two and moving a modified version of the URL in to that base class?

src/ExternalApi/Ada/Service/AdaProgrammeService.php
votemike: It's a bit weird that the main PID is a string, but the countContextPid is a Pid. Maybe change it so that they're both Pids?

src/ExternalApi/Ada/Service/AdaClassStubService.php
```diff
@@ -16,7 +16,8 @@ class AdaClassStubService extends AdaClassService
 {
     public function findRelatedClassesByContainer(
         Programme $programme,
-        bool $countWithinTleo = true
+        bool $countWithinTleo = true,
```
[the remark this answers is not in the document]
benolot: No, it's only used in testing and not needed for any AdaProgramme tests.

src/ExternalApi/Ada/Service/AdaProgrammeService.php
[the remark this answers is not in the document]
benolot: Would be cleaner after the typehinting change for sure

src/ExternalApi/Ada/Service/AdaProgrammeService.php
benolot: Are you recommending that we set the default Threshold of 2 or just leave it nullable and let it be passed in as 2 whenever needed

benolot: Setting the default order to Random doesn't seem like a great idea to me, maybe Alphabetical by default if we have to have one?

votemike: I was thinking, just hardcode it.

src/ExternalApi/Ada/Service/AdaProgrammeService.php
benolot: It's different compared to Ada Class as they use different endpoints. (One needs the PID as a query string, the other needs it in the URL (/pid/stuff vs /stuff?pid=pid)

src/ExternalApi/Ada/Service/AdaProgrammeService.php
votemike: Where in the code are we using anything other than 'random'?

benolot: Okay, fair point, we're not, it just seems very counter-intuitive to default to random. Like the default play format on music being shuffle play...

benolot: I think it would be better to set a default and allow it as an option? That way if the customization is ever needed, it's there.

votemike: I thought it would simplify a 8 parameter method....

benolot: Fixed in next commit

src/ExternalApi/Ada/Service/AdaProgrammeService.php
```diff
@@ -0,0 +1,149 @@
...
+        $relatedProgrammes = [];
+        foreach ($uniqueProgrammes as $key => $item) {
+            $relatedProgrammes[] = $this->mapper->mapItem($programmeItems[$key], $item);
+        }
+
+        $this->cache->setItem($cacheItem, $relatedProgrammes, CacheInterface::MEDIUM);
+        return $relatedProgrammes;
+    }
+
+    private function requestRelatedProgrammeItems(Pid $pid, ?string $scope = null, ?Pid $countContextPid = null, int $limit = 10, int $page = 1)
```
votemike: None of the calling code ever seems to pass in a page... Is it a necessary var? Can we just hardcode it?

src/ExternalApi/Ada/Service/AdaProgrammeService.php
```diff
@@ -0,0 +1,149 @@
...
+    {
+        $order = 'random';
```
votemike: These 3 probably don't need to be variables. You could probably hardcode them in the http_build_query bit.

src/ExternalApi/Ada/Service/AdaProgrammeService.php
```diff
@@ -0,0 +1,149 @@
...
+            [
+                'page' => $page,
+                'page_size' => $limit,
+                'scope' => $scope,
+                'count_context' => $countContextPid,
+                'threshold' => $threshold,
+                'order' => $order,
+            ]
+        );
+        $url =  $this->baseUrl . '/programme_items/' . $pid->__toString() . '/related?' . $params;
```
votemike: `__toString()` is a magic method. If you try and concatenated an object with a `__toString()` method, it will magically convert. So you can change this line to `$url =  $this->baseUrl . '/programme_items/' . $pid . '/related?' . $params;`.

src/ExternalApi/Ada/Service/AdaProgrammeService.php
```diff
@@ -0,0 +1,149 @@
...
+
+        $cacheKey = $this->clientFactory->keyHelper(__CLASS__, __FUNCTION__, $pid, $scope, $countContextPid, $threshold, $order, $orderDirection, $limit, $page);
```
votemike: If you're hardcoding things like `$threshold`, there's not a lot of point in using it as part of the cache key.

tests/ExternalApi/Ada/Mapper/AdaProgrammeMapperTest.php
```diff
@@ -0,0 +1,92 @@
...
+
+class AdaProgrammeMapperTest extends TestCase
+{
+    public function testMapItemWithClassModels()
+    {
+        $programme = $this->createMock(Programme::class);
+
+        $adaEntity = [
+            'pid' => 'p03knff4',
+            'type' => 'category',
```
votemike: I don't think type is used.

tests/ExternalApi/Ada/Mapper/AdaProgrammeMapperTest.php
```diff
@@ -0,0 +1,92 @@
...
+            'title' => 'Fellows of the Royal Society',
+            'image' => 'ichef.bbci.co.uk/images/ic/$recipe/p01l7xx5.jpg',
```
votemike: I don't think image is used.

tests/ExternalApi/Ada/Mapper/AdaProgrammeMapperTest.php
```diff
@@ -0,0 +1,92 @@
...
+        $this->assertEquals($expectedEntity, $mapper->mapItem($programme, $adaEntity));
+    }
+
+    public function testMapItemWithoutClassModels()
+    {
+        $programme = $this->createMock(Programme::class);
+
+        $adaEntity = [
+            'pid' => 'b071vl2l',
+            'type' => 'category',
```
votemike: I don't think type is used.

tests/ExternalApi/Ada/Mapper/AdaProgrammeMapperTest.php
```diff
@@ -0,0 +1,92 @@
...
+            'title' => 'Fellows of the Royal Society',
+            'image' => 'ichef.bbci.co.uk/images/ic/$recipe/p01l7xx5.jpg',
```
votemike: I don't think image is used.

tests/ExternalApi/Ada/Mapper/AdaProgrammeMapperTest.php
```diff
@@ -0,0 +1,92 @@
...
+            'class_count' => 16,
+        ];
+
+        $mapper = new AdaProgrammeMapper(new AdaClassMapper());
+
+
```
votemike: TOO MANY NEW LINES!!!!

src/ExternalApi/Ada/Service/AdaProgrammeService.php
benolot: I did at first, but I noticed they are used a couple times in the function, so it's probably easier to make them as variables to make it easier to update

src/Ds2013/Presenters/Section/Episode/Map/episode_map.html.twig
```diff
@@ -13,7 +13,7 @@
         <div class="grid 1/3@bpw2 1/3@bpe map__column map__column--2 map__column--last">
             <div class="map__inner">
                 {%- for panel in episode_map.getSidePanelsSubPresenters() -%}
-                    <div class="{{ build_css_classes({ 'br-box-secondary': true, 'fauxcolumn': loop.last, 'map__column': loop.last and loop.length > 1 }) }}">
+                    <div class="{{ build_css_classes({ 'br-box-secondary': true, 'map__column': loop.last and loop.length > 1 }) }}">
```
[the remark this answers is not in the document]
tatitati: Sure!, the fix is here:
https://github.com/bbc/programmes-frontend/pull/535

PR #529, opened 2018-04-24
PROGRAMMES-6331 Adding priority telescope vote

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6331

2 things to note:

- I don't think the LIVE URL works... so.... what should we do about that? I guess we don't want to merge this yet...
- In PPS is this line: https://github.com/bbc/programmes-pages-service/blob/master/src/Mapper/ProgrammesDbToDomain/OptionsMapper.php#L38 I don't know if that needs to be updated, or if anything needs to be added to PPS as part of this.


programmes-developers:

config/services.yaml
```diff
@@ -250,6 +253,11 @@ services:
         arguments:
             - '@null_cache_service_factory'
 
+    App\Controller\FindByPid\TlecController:
+        public: true
+        arguments:
+            - '%app.vote.accountwrapperurl%'
```
BjornTwachtmann: Rather than passing this through the controller, would it be cleaner to make it directly available to twig via a global? It will be used in more than one template.
https://symfony.com/doc/current/templating/global_variables.html

BjornTwachtmann:
In anwser to the questions above:
1. This isn't enabled in live iSite yet. We just need to be sure their live env works before we push those changes to live iSite.
2. Yeah, you should add a telescope_block to PPS that looks the same as isite_block. 

Once 2 is done and merged, do a composer update in this branch and merge this

tests/DataFixtures/PagesService/OptionsFixture.php
```diff
@@ -64,7 +64,7 @@ public static function radioFour()
         return new Options(
             [
                 'twitter_block' => '',
-                'ivote_block' => '',
```
votemike: Is it OK to have gotten rid of this?

PR #532, opened 2018-04-25
PROGRAMMES-6340 Adding TLEC non-priority telescope vote

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6340

I think this is correct. However locally I have to mess with my hosts file to get the vote to display.
Not sure if this is a problem my end or somewhere else along the chain.


programmes-developers:

PR #534, opened 2018-04-26
PROGRAMMES-6325-Added translation strings for the home nations

benolot:
Welsh, Scottish Gaelic, Irish Gaelic translations


programmes-developers:

PR #535, opened 2018-04-26
PROGRAMMES-6307 Fix panels fauxcolumn

tatitati:
Fauxcolumn class was giving some issues:
1- The last panel on the episode map was not using 100% of the height in the map, so was looking shorter

2- It was happening collapsing margins. So if the user squash the screen the bottom margin of the map and the top margin of the next element below the map collapse.

This fix both issues by using flexbox.

----

Some urls to test the fix:
http://sandbox.bbc.co.uk/programmes/b0b10zdp?__flush_cache=1
https://sandbox.bbc.co.uk/programmes/b0876646?__flush_cache=1
http://sandbox.bbc.co.uk/programmes/b09pjgly?__flush_cache=1


programmes-developers:

PR #536, opened 2018-04-26
Fix bug where recipe does not have a chef

BjornTwachtmann:


programmes-developers:

PR #541, opened 2018-04-27
Added getSymfonyRouter

Tasiobg:
Added a static function to be able to use the symfony router across other test


programmes-developers:

PR #542, opened 2018-04-27
Programmes 6339 move radio episodes and audio clip playout URLs to Sounds

Tasiobg:
PROGRAMMES-6339 move radio episodes and audio clip playout URLs to Sounds
- Updated StreamableHelper, CoreEntityTitlePresenter::getTitleLinkUrl() and BaseTitlePresenter::getUrl() to generate the right link
- Renamed force_iplayer_linking to force_playout_linking to make it more clear
- Unitest


src/DsShared/Helpers/StreamableHelper.php
```diff
@@ -9,12 +9,18 @@
 
 class StreamableHelper
 {
+    private $playspaceNetworsList = ['bbc_radio_four_extra', 'bbc_radio_three', 'bbc_radio_scotland'];
```
BjornTwachtmann: should be spelt playspaceNetworksList

PR #543, opened 2018-04-30
Update IndexerTrait to take account of broadcast ordering and use the…

BjornTwachtmann:
… first, not last broadcast

This fixes a bug on /programmes/b006qykl/episodes/guide where the last upcoming broadcast is shown, when the first should be. Compare live to V3 test to view.


programmes-developers:

programmes-developers:

programmes-developers:

PR #550, opened 2018-05-14
[Programmes-6375] [fix] [UI] - Recommendations not showing images on IE-11

tatitati:
We were setting a default size of 148 four our recommendation images, so the resulting image html:

```
<img class="image" 

src="https://ichef.bbci.co.uk/images/ic/148x83/p066rywx.jpg"    ---> the only one broken

srcset="	
        https://ichef.bbci.co.uk/images/ic/80x45/p066rywx.jpg 80w, 
	https://ichef.bbci.co.uk/images/ic/160x90/p066rywx.jpg 160w, 
	https://ichef.bbci.co.uk/images/ic/320x180/p066rywx.jpg 320w, 
	https://ichef.bbci.co.uk/images/ic/480x270/p066rywx.jpg 480w, 
	https://ichef.bbci.co.uk/images/ic/640x360/p066rywx.jpg 640w, 
	https://ichef.bbci.co.uk/images/ic/768x432/p066rywx.jpg 768w, 
	https://ichef.bbci.co.uk/images/ic/896x504/p066rywx.jpg 896w, 
	https://ichef.bbci.co.uk/images/ic/1008x567/p066rywx.jpg 1008w
 	" 

sizes=" (min-width: 63em) 178px,  ......."
>
```
As you can see this is not a lazyload  image as there is no  ```class=lazyload```, so our ```lazysizes JS library``` doesn't touch this image. This means that the only thing that can touch this image in order to make it responsive is the browser. The problem with this is that IE-11 doesn't support srcset/sizes as you can see here:
https://caniuse.com/#search=srcset

So IE-11 only see an equivalent image to this, which has the url broken:

All urls in the field "srcset" works, but our "src" url is broken. 
```
<img class="image" 

src="https://ichef.bbci.co.uk/images/ic/148x83/p066rywx.jpg"    ---> the only one broken
>
```


programmes-developers:

PR #551, opened 2018-05-18
Composer update comscore to fix https mixed content warning on episod…

BjornTwachtmann:
…e page


programmes-developers:

BjornTwachtmann:
I thought it's a sort-of-potentially breaking change. It changes behaviour rather than just fixing a bug. Changes it to be better, but still.

PR #552, opened 2018-05-18
[NO TICKET][refactor] - Deleted unused dependency in constructor

tatitati:
I saw that this is not being used. As a reference you can see the html.twig for this presenter:

https://github.com/bbc/programmes-frontend/blob/develop/src/Ds2013/Presenters/Pages/Schedules/ByDayPage/schedules_by_day_page.html.twig

There is no tests for this presenter, so I couldn't update any test


programmes-developers:

benolot:
Test this please

PR #553, opened 2018-05-18
[PROGRAMMES-6353] [fix][usr] - Tx column date shows "zero" digit for day field

tatitati:
This change in ```LocalisedDaysAndMonthsHelper``` affects to:

CollapasedBroadcastPresenter (DSamen) and BroadcastEventPresenter (Ds2013).

CollapasedBroadcastPresenter is used in **Brand page** in the Txcolumn. As reference:
https://github.com/bbc/programmes-frontend/blob/develop/src/DsAmen/Presenters/Domain/CoreEntity/CollapsedBroadcast/SubPresenter/details.html.twig

BroadcastEventPresenter is used in the new **Episode guide** and new **Episode page** in the txt column:
https://github.com/bbc/programmes-frontend/blob/develop/src/Ds2013/Presenters/Domain/BroadcastEvent/broadcast_event.html.twig


programmes-developers:

PR #554, opened 2018-05-21
[No Ticket] - Fix rmspodcast panel 

tatitati:
Working in the "is uk only" ticket I saw that there was an error in the previous one.


programmes-developers:

PR #556, opened 2018-05-22
[PROGRAMMES-6395][change][usr] - Button CTA too small at small device sizes

tatitati:
Example url:
http://sandbox.bbc.co.uk/programmes/b05qqk5c/episodes/player


programmes-developers:

PR #558, opened 2018-05-23
PROGRAMMES-6167-Fixing a comscore bug with episode guides

benolot:
Ticket link: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6167

Possibly better to write it like:

```
$this->setIstatsExtraLabels(['has_available_items' => 'false']);
if ($programme->getAvailableEpisodesCount() >= 1) {
    $this->setIstatsExtraLabels(['has_available_items' => 'true']);
}
```
But for me having it as an if/else is cleaner and clearer.

This PR fixes the bug Jatin raised, about the `has_available_items` parameter not being passed to iSite


programmes-developers:

PR #559, opened 2018-05-23
[Programmes-6289][change][usr] - Fix problem parsing date in some browsers

tatitati:
### What:
Safari and IE-11 cannot parse dates in some dates formats from [ISO-8061, Wikipedia](https://en.wikipedia.org/wiki/ISO_8601)

### Why:
It happens the next depending on the browser:

Case when date is:```2018-11-05T20:13:40+0702```:
Is a proper format, but Safari and IE-11 fails when parsing it.

Case when date is:```2018-11-05T20:13:40+07:02```:
Is a proper format, but instead it includes ```":"``` in the utc offset. It works in all browsers (even if the other format is also correct)

You can check also this thread: [Stackoverflow comment related](https://stackoverflow.com/a/49138448/1447456)

### To test the fix:
```yarn run gulp```
http://sandbox.bbc.co.uk/programmes/p016tklr/broadcasts/upcoming

<details>
  <summary>Click to expand info</summary>
   Some highglights texts from the official doc of Date.parse() javascript:
   
>  There are still many differences in how different hosts parse date strings, therefore date strings should be manually parsed 

> Because of the variances in parsing of date strings, it is recommended to always manually parse strings as results are inconsistent, especially across different ECMAScript implementations where strings like "2015-10-12 12:00:00" may be parsed to as NaN, UTC or local timezone. 

</details>


programmes-developers:

PR #560, opened 2018-05-24
[Programmes 6396][change][fix] - Broadcast dates issue with time

tatitati:
### What:
A user can see that the moment of start for a programme in the same day is displayed with different labels such as: "Saturday, Next Saturday" when the different between both starts are just a few minutes or seconds.

### Why:
To translate dates like "2013-12-04 23:22:12" to wordy/human format like "Next Saturday" we compare the date with an interval of 5 or 8 days. And if it is inside of any of these intervals, then we make the proper translation.

The problem is that this have in accounts hours and minutes. So is possible to have a "Saturday", and 1 minute later a "Last Saturday".

### The fix:
Compare dates not having in account hours/minutes/seconds.

Example URL:
https://programmes-frontend.test.api.bbc.co.uk/programmes/b007rn05/broadcasts/upcoming


programmes-developers:

src/DsShared/Helpers/LocalisedDaysAndMonthsHelper.php
```diff
@@ -104,15 +107,16 @@ public function getFormatedDay(Chronos $date): string
             return $translate->translate('schedules_new_years_day') . ' ' . $date->format('Y');
         }
 
-        if ($date->isWithinNext('5 days') || $date->wasWithinLast('5 days')) {
+        $diffInDays = $now->diffInDays($date, false);
+        if ((0 < $diffInDays && $diffInDays <= 5) || (-5 <= $diffInDays && $diffInDays < 0)) {
```
BjornTwachtmann: `if ($diffInDays >= -5 && $diffInDays <= 5)`
or am I reading that wrong?

PR #561, opened 2018-05-25
[PROGRAMMES-6263] Changed solent_west_dorset to point at playspace

benolot:
Ticket can be found here: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6263


programmes-developers:

BjornTwachtmann:
I *think* that this is the only use of the popup_player route, so that can probably be removed from the routing YAML

benolot:
V2 instance of the same thing: https://github.com/bbc/programmes/pull/862

PR #563, opened 2018-05-29
[No Ticket][fix] - fix to display also an image even without JS

tatitati:
## What
Trying to investigate the low resolution issue I realized of this other issue: whatever we tag the image as "lazyload image", if the user disable javascript, then no image at all appear. 

| URL | javascript | are lazyload images visible? |
| --- | --- | --- |
| (v2) https://www.bbc.co.uk/programmes/b0876646/recipes | ON | YES |
|  | OFF | YES |
| (v3) http://sandbox.bbc.co.uk/programmes/b0876646/recipes | ON | YES |
|  | OFF | **NO (BUG)** |

You can try more URLs apart of the recipes and see the differences between V2 and V3 toggling JS:
(Episode page) https://www.bbc.co.uk/programmes/b0876646
(schedules page) http://www.bbc.co.uk/schedules/p00fzl6p

From [lazysizes doc](https://github.com/aFarkas/lazysizes/blob/gh-pages/README.md):
> lazysizes does not need any JS configuration: Add the class "lazyload" to your images/iframes in conjunction with a data-src and/or data-srcset attribute. Optionally you can also add a src attribute with a low quality image

> Note: In case you are using either srcset/data-srcset or picture, we recommend to extend this pattern with either a data-src (see next pattern: "Combine data-srcset with data-src") or with a suitable src attribute.

## WHY
If we tag an image as "lazyload" then our third-party JS library "lazysizes" try to lazyload and make the images responsive. The problem is that if the user has not JS enabled then this script cannot run and cannot create any "src" attribute on the fly for each image. The resulting effect is the proper one from an image without any url attached (no image is displayed).


programmes-developers:

benolot:
Does this not disable lazy loading? (Given the browser will read the src and load it when it parses the document)

tatitati:
@benolot I can run a demo to show you that lazy load still work as expected. Also is documented in the doc:
> lazysizes does not need any JS configuration: Add the class "lazyload" to your images/iframes in conjunction with a data-src and/or data-srcset attribute. Optionally you can also add a src attribute with a low quality image

programmes-developers:

PR #565, opened 2018-06-01
Update gulp-sass to function with newest node

BjornTwachtmann:
This basically updates the yarn.lock file to pull in a newer grunt-sass and some other dependencies which enable our build pipeline to work with node 10, I can't see any significant differences in the CSS generated (yep, I checked)


programmes-developers:

PR #568, opened 2018-06-04
[PROGRAMMES-6411][fix] - Fix for episodes guide redirecting to an empty series

tatitati:
## What
In case of loading the content (the partial) for a series with no episodes we were redirecting the user to that specific series.

## Why
From the Partial controller we return 404 instead of the original (v2) 200 when the series has no episodes as you can see here:
https://github.com/bbc/programmes-frontend/blob/develop/src/Controller/ProgrammeEpisodes/GuidePartialController.php#L31

This is considered by ajax an error, and in the callback of this ajax we redirect (this time doing it right as in V2).
V3 javascript:
https://github.com/bbc/programmes-frontend/blob/develop/assets/js/episode-guide.js#L151

V2 javascript:
https://github.com/bbc/programmes/blob/develop/webapp/static-versioned/js-modules/episode-guide.js#L86

## The fix:
this is fixed just returning the same response code from the partial (which is in V3 different from V2)

## URL example:
http://sandbox.bbc.co.uk/programmes/b09n14fm/episodes/guide?__flush_cache=1


programmes-developers:

PR #570, opened 2018-06-05
Update to use picturefill when support is not present.

BjornTwachtmann:
Adds a full polyfill for responsive images to the project that is only loaded when a browser does not support responsive images natively.


programmes-developers:

PR #572, opened 2018-06-05
[No Ticket][refactor] - Improved end to end tests for episodes guide

tatitati:
Doing the fix for a previous ticket in Episodes guide I saw that our end to end tests (made by myself) were quite bad, so took me a little more to understand how nested level works in there. 

In order to solve these un-readable tests I refactor them

URL to test that everything keep as expected:
http://sandbox.bbc.co.uk/programmes/b006qftk/episodes/guide


programmes-developers:

PR #575, opened 2018-06-06
Fix various accessibility issues

BjornTwachtmann:
* Fix accessibility issue: aria lables with broadcast times in episode guide pages are misleading as they do not feature the date, and timezone is incorrect
* Fix accessibility issue: Brand titles being read out in inapproriate places on episode guide pages
* Fix accessibility issue: Episode guide pages not showing aria-label with broadcast time on programme title where other pages are
* Fix accessibility issue: Paginator titles not reading correctly in screen readers


programmes-developers:

PR #576, opened 2018-06-07
[PROGRAMMES-6389] Added page load iStats to clip pages

benolot:
Ticket: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6389
Example URL: https://sandbox.bbc.co.uk/programmes/p04yxr0l / https://www.bbc.co.uk/programmes/p04yxr0l


programmes-developers:

src/Controller/FindByPid/ClipController.php
```diff
@@ -10,10 +10,40 @@ class ClipController extends BaseController
...
+            $this->setIstatsExtraLabels(['clip_release_year' => $clip->getStreamableFrom()->format('Y')]);
+        }
+    }
+
+    private function setParentIstats(Clip $clip): void
+    {
+        $parent = $clip->getParent();
+        if ($parent) {
+            $this->setIstatsExtraLabels(['parent_available' => $parent->isStreamable() ? 'true' : 'false']);
+            $this->setIstatsExtraLabels(['parent_entity_type' => $parent->getType()]);
```
BjornTwachtmann: Looking at the ticket, I would probably only set this value if the parent is a ProgrammeItem. ProgrammeContainers will be marked as streamable if any of the episodes within them are streamable, which probably isn't the intention of that feature request.

src/Controller/FindByPid/ClipController.php
```diff
@@ -41,7 +41,7 @@ private function setIstatsReleaseYear(Clip $clip): void
     private function setParentIstats(Clip $clip): void
     {
         $parent = $clip->getParent();
-        if ($parent) {
+        if ($parent instanceof \BBC\ProgrammesPagesService\Domain\Entity\ProgrammeItem) {
```
BjornTwachtmann: Can you add a use statement for that rather than use the fully qualified class here, sorry :)

PR #579, opened 2018-06-07
Fix bug with "and $num more" not being displayed due to wrong typehin…

BjornTwachtmann:
…t in BroadcastNetworkBreakdown


programmes-developers:

PR #580, opened 2018-06-08
Snippet music player

Tasiobg:
PROGRAMMES-6394
PROGRAMMES-6392

Backend:
- Implementation of TUPAC API so we can get all the snippet players info

Frontend:
- Refactor the original snippet player into v3

Please review the backend only, I can walk you throug the frontend but the JS is almost the same we had before and the only big change I did is in transform.js


programmes-developers:

src/Controller/SnippetController.php
```diff
@@ -0,0 +1,45 @@
...
+
+class SnippetController extends BaseController
+{
+    public function __invoke(Request $request, string $recordsIds, TupacService $tupacService)
+    {
+        $recordsIdsArray = $this->getRecordsIdsAsArray($recordsIds);
+        $records = [];
+        if (!empty($recordsIdsArray)) {
+            // check if is UK IP
+            $isUk = ($request->headers->get('X-Ip_is_uk_combined', 'no') === 'yes') ? true : false;
```
votemike: Do you need ` ? true : false`? Won't `$request->headers->get('X-Ip_is_uk_combined', 'no') === 'yes'` give you a boolean anyway?

src/Ds2013/Presenters/Section/Segments/segments_list.html.twig
```diff
@@ -1,6 +1,16 @@
 {% import _self as self %}
 
 {% if segments_list.getSegmentEvents() %}
+    <link rel="stylesheet" href="{{ asset('css/snippet/snippets.css') }}" />
```
votemike: Shouldn't CSS and JS be in the head of the page?

src/Ds2013/Presenters/Section/Segments/segments_list.html.twig
```diff
@@ -1,6 +1,16 @@
...
+    <script>
+        require(["snippets", "bump-3", "istats-1"], function(snippets, bump, istats) {
+            snippets.init({
+                base_url: '/programmes/snippet',
```
votemike: Should this be hardcoded like this?

src/ExternalApi/Tupac/Service/TupacService.php
```diff
@@ -0,0 +1,93 @@
...
+    public function fetchRecordsByIds(array $recordsIds, bool $isUk = false): PromiseInterface
+    {
+        if (empty($recordsIds)) {
+            return new FulfilledPromise([]);
+        }
+        $client = $this->makeHttpApiClient($recordsIds, $isUk);
+        return $client->makeCachedPromise();
+    }
+
+    private function makeHttpApiClient(array $recordsIds, bool $isUk = false): HttpApiClient
```
votemike: $recordIds instead of $recordsIds?

src/ExternalApi/Tupac/Service/TupacService.php
```diff
@@ -0,0 +1,93 @@
...
+    {
+        $recordsIdsHash = md5(implode('_', $recordsIds));
+        $cacheKey = $this->clientFactory->keyHelper(__CLASS__, __FUNCTION__, $recordsIdsHash);
+
+        $recordsIdsAsQueryParameters = '';
+        foreach ($recordsIds as $recordsId) {
+            $recordsIdsAsQueryParameters .= '&id=' . urlencode($recordsId);
+        }
+
+        $ukQueryParameter = '0';
```
votemike: `$ukQueryParameter = $isUk ? 1 : 0;`?

src/Ds2013/Presenters/Section/Segments/segments_list.html.twig
BjornTwachtmann: Yeah, looking at it the path() function won't generate this URL correctly.

PR #581, opened 2018-06-08
[PROGRAMMES-6382] Added related links to clip pages

benolot:
Ticket: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6382

I also had to update the Episode page to allow the related links partial to be more reusable.


programmes-developers:

benolot:
Test this please

templates/find_by_pid/clip.html.twig
```diff
@@ -8,4 +8,20 @@
...
             <h1>THIS IS CLIP</h1>
         </div>
+
+        {% if relatedLinks %}
+        <div class="component component--box--primary">
+            <div class="component__header br-box-page">
+                <h2>{{ tr('related_links') }}</h2>
+            </div>
+            <div class="component__body br-box-page">
+                {% if relatedLinks %}
```
tatitati: I think this ```{% if relatedLinks %}``` is not necessary here as it is inside (nested) in the same ```{% if relatedLinks %}``` conditional (on line 12)

PR #585, opened 2018-06-12
PROGRAMMES-6418 Fixing ampersand link bug

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6418


programmes-developers:

BjornTwachtmann:
I think changing the logic in the BroadcastNetworksHelper like this will also have knockon effects in the other places this is used. If you check the TX column on the brand page for a brand for e.g. /programmes/b09c12lj , it looks broken as a result of this change. 

I'm not entirely sure I like the service list having its punctuation constructed differently than the networks list in this way, it feels like it could get messy, but I could be talking shite on that, I'm honestly not sure I understand this code at all. 

At minimum the brand page thing needs to be fixed though.

votemike:
I think I've fixed the brand page now. Had a look and I _think_ that's the only 2 places that list is used.

If you're not keen on this solution, do you have any ideas for a different solution?

BjornTwachtmann:
You want me to actually be constructive? Jesus. I don't really have a better idea either, this looks like it should work.

PR #586, opened 2018-06-12
PROGRAMMES-6304 Adding hover state to image links

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6304

So the default focus styling for anchor tags on this page is a thin dotted line. So I just changed the anchor tag to be a block level element so that this border is visible.
If people prefer I could change the links to have the standard browser focus state instead.


programmes-developers:

PR #587, opened 2018-06-12
[PROGRAMMES-6276] World service pages now show local time on episodes

benolot:
Fixed the bug of the episode pages for world service radio episodes not showing in local times


programmes-developers:

src/Controller/FindByPid/EpisodeController.php
```diff
@@ -52,6 +52,8 @@ public function __invoke(
     ) {
         $this->setIstatsProgsPageType('programmes_episode');
         $this->setContextAndPreloadBranding($episode);
+        // If it's international such as World Service, show local time's
```
votemike: I think the method is probably clear enough that we don't need the comment.

votemike:
Is this working on all other pages? Or does that line need to be added elsewhere too?

benolot:
The ticket related to episode pages, perhaps brand pages would need it too? I haven't checked, will take a look

benolot:
Have checked, it works on brand page already and cannot see anything else that displays a broadcastEvent

PR #588, opened 2018-06-12
Fix timezone for lazy-loads on episode guide pages. Add GMT to timezo…

BjornTwachtmann:
…ne on non-JS page variants. Make episode guide partial not load unncessary branding.

See https://www.test.bbc.co.uk/programmes/w27vq0vt/episodes/guide for example. Ticket: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6439


programmes-developers:

src/Controller/ProgrammeEpisodes/GuidePartialController.php
```diff
@@ -34,12 +32,12 @@ public function __invoke(
 
         $upcomingBroadcasts = $this->getUpcomingBroadcastsIndexedByProgrammePid($programme, $collapsedBroadcastService);
 
-        return $this->renderWithChrome('programme_episodes/guide.2013inc.html.twig', [
+        return $this->render('programme_episodes/guide.2013inc.html.twig', [
```
votemike: Will the set the fact that it's international correctly?

BjornTwachtmann: Yeah, renderWithChrome just deals with branding/orb. The international JS stuff is included on the parent page anyway, all we should need here is https://github.com/bbc/programmes-frontend/pull/588/files#diff-5e9a3abed6f22e92f2eb1cb56fe7557eR25 that bit to make sure the DateTimeZone is set to GMT correctly

votemike: But it's `renderWithChrome()` that passes is_international through to the view: https://github.com/bbc/programmes-frontend/blob/develop/src/Controller/BaseController.php#L211

BjornTwachtmann: yes, it is, but is_international is only used in the base template to decide whether to include timezone JS and adverts JS stuff, which isn't needed on a partial anyway. The setInternationalStatus thing sets the global timezone for the app, which is what's used in the date presentation to make this work

votemike: Ahh, OK then.

PR #589, opened 2018-06-12
[Not ticket] - Fix for [PROGRAMMES-6274]

tatitati:
Original PR:
https://github.com/bbc/programmes-frontend/pull/555/files

RmsPodcast might be null, and is defined as such but I forgot it in one place. I will try to create a good end to end to ensure that this doesn't explode anymore for this special case.


programmes-developers:

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/DetailsPresenter.php
```diff
@@ -137,7 +137,11 @@ public function getPodcastFileName(): string
 
     public function isUkOnlyPodcast(): bool
     {
-        return $this->rmsPodcast->isOnlyInUk();
+        if ($this->rmsPodcast) {
```
votemike: In the file docblocks, it says it can't be null...

tatitati: Done

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/DetailsPresenter.php
```diff
@@ -28,7 +28,7 @@ class DetailsPresenter extends Presenter
     /** @var UrlGeneratorInterface */
     private $router;
 
-    /** @var RmsPodcast */
+    /** @var RmsPodcast|null */
```
benolot: So if it's not a 200, it passed null into the presenter right?

tatitati: Yes, as you can see here:
https://github.com/bbc/programmes-frontend/blob/develop/src/ExternalApi/RmsPodcast/Service/RmsPodcastService.php#L54

PR #596, opened 2018-06-14
[PROGRAMMES-6438] Fix collapsing padding

benolot:
Ticket: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6438

Okay so I know I'm being very specific with the selector, but this should ensure it's only that one column I target, if you think it's better to go with a less specific selector let me know.


programmes-developers:

tatitati:
not sure if we can avoid that specificity issue, otherwise seems fine to me

src/Ds2013/Presenters/Section/Episode/Map/_episode-map.scss
```diff
@@ -23,6 +23,10 @@
     padding-bottom: 0;
 }
 
+.map__column--last > .map__inner > div.br-box-secondary.map__column {
```
BjornTwachtmann: .map__column--last > .map__inner > .map__column {

src/Ds2013/Presenters/Section/Episode/Map/episode_map.html.twig
```diff
@@ -13,7 +13,7 @@
         <div class="grid 1/3@bpw2 1/3@bpe map__column map__column--2 map__column--last">
             <div class="map__inner">
                 {%- for panel in episode_map.getSidePanelsSubPresenters() -%}
-                    <div class="{{ build_css_classes({ 'br-box-secondary': true, 'fauxcolumn': loop.last, 'map__column': loop.last and loop.length > 1 }) }}">
```
votemike: Is fauxcolumn being used elsewhere?

src/Ds2013/Presenters/Section/Episode/Map/_episode-map.scss
```diff
@@ -23,6 +23,10 @@
     padding-bottom: 0;
 }
 
+.map__column--last > .map__inner > .map__column {
+    margin-bottom: 0;
```
votemike: Is this needed? What is the margin and padding of this usually?

src/Ds2013/Presenters/Section/Episode/Map/episode_map.html.twig
benolot: No, this just puts it on the last column (Which was the more episodes column) so removing it does not affect other columns

PR #597, opened 2018-06-14
[Expand for PROGRAMMES-6316][improvement]: Footer of recipes panel is always displayed

tatitati:
## Why this improvement:
- As @bassani2016  pointed, is weird to have in the recipes panel 3 recipes and a link below saying ```See all recipes from .... (3) ```. However this make sense when we display 4 and exist more apart of those 4.
- Is an improvement because in pal we display this footer always as you can see here:
http://pal.test.bbc.co.uk/programmes/b0111461/recipes

## Url examples
- With 3 recipes
http://sandbox.bbc.co.uk/programmes/b0111461/recipes
- With 397 recipes:
http://sandbox.bbc.co.uk/programmes/b013pqnm/recipes


programmes-developers:

votemike:
This isn't actually a link to all the recipes, it's a link to the food website. If you remove this, is there still a link somewhere on the page to allow them to get to the food website?

PR #598, opened 2018-06-14
PROGRAMMES-6381 Adding Clip Segments

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6381

I _think_ I have fulfilled all the ACs.


src/Controller/FindByPid/ClipController.php
```diff
@@ -41,7 +51,26 @@ public function __invoke(
...
+        $segmentsListPresenter = null;
+        if ($versions) {
+            $canonicalVersion = $canonicalVersionHelper->getCanonicalVersion($versions);
+            if ($canonicalVersion->getSegmentEventCount()) {
+                $segmentEvents = $segmentEventsService->findByVersionWithContributions($canonicalVersion);
+                if ($segmentEvents) {
+                    $segmentsListPresenter = $presenterFactory->segmentsListPresenter(
+                        $clip,
+                        $segmentEvents,
+                        null,
```
votemike: @BjornTwachtmann  I think the upcoming and lastOn variables are used to say how long since/until a segment is being played, but I assume that doesn't matter on Clips. Right?

programmes-developers:

src/Controller/FindByPid/ClipController.php
BjornTwachtmann: It looks like it's being used to show timings relative to a live/upcoming broadcast or something. I think it should work without, but I couldn't swear to it.

PR #601, opened 2018-06-15
PROGRAMMES-6384 Adding parents clips to clip page

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6384

Review, but do not approve yet, there are a few questions.

Also, the clips seem to be in a different order to LIVE. Any ideas why?


src/Controller/FindByPid/ClipController.php
```diff
@@ -121,4 +138,20 @@ private function getSchema(
...
+     * @param ProgrammesAggregationService $aggregationService
+     * @param Programme $programme
+     * @param Pid $pid
+     * @return Clip[]
+     */
+    private function getClipsExcept(ProgrammesAggregationService $aggregationService, Programme $programme, Pid $pid): array
+    {
+        $clips = $aggregationService->findStreamableDescendantClips($programme, 5);
+        $filteredClips = array_filter($clips, function (Clip $clip) use ($pid) {
+            return (string) $clip->getPid() !== (string) $pid;
```
votemike: Is there somewhere we could extract comparing `Pid`s to?

programmes-developers:

templates/find_by_pid/clip.html.twig
```diff
@@ -81,6 +83,60 @@
...
+                        {{ ds2013('programme', clip, {
+                            'context_programme': programme,
+                            'highlight_box_classes':'highlight-box--list highlight-box--grid@bpb1 programme--grid@bpb1 highlight-box--grid@bpw programme--grid@bpw',
+                            'image_options': {
+                                'classes': '1/1@bpb1 1/1@bpw',
+                                'default_width': 240,
+                                'sizes': { 0: 1/2, 768: 1/4, 1008: '240px'},
+                            },
+                            'title_options': {
+                                'link_location_track': 'component_clips_carousel',
```
votemike: What should the link_location_track be here? Is this correct?

templates/find_by_pid/clip.html.twig
```diff
@@ -81,6 +83,60 @@
...
+                        }) }}
+                    </li>
+                {% endfor %}
+            </ul>
+        </div>
+        {% if clips|length > 1 %}
+            <div class="component__footer br-box-subtle">
+                <a class="component__footer__link"
+                   href="{{ path('programme_clips', { pid: programme.getPid() }) }}"
+                   data-linktrack="component_clips_summary_footer">
```
votemike: Is this data-linktrack correct?

src/Controller/FindByPid/ClipController.php
```diff
@@ -46,6 +50,21 @@ public function __invoke(
             $relatedLinks = $relatedLinksService->findByRelatedToProgramme($clip, ['related_site', 'miscellaneous']);
         }
 
+        $parentClips = [];
+        $tleoClips = [];
+        if ($clip->getParent()) {
+            /** @var ProgrammeContainer|Episode $parent */
```
votemike: This is right. Isn't it?

tatitati: Yes.

templates/find_by_pid/clip.html.twig
benolot: There's a different ticket open for istats stuff, PROGS-6390

votemike: Reckon I should leave this as it is? Delete it? Change it to "PROGRAMMES-6390"?

templates/find_by_pid/clip.html.twig
benolot: See earlier reply about istats

templates/find_by_pid/clip.html.twig
benolot: I think leave it as it is or change it, but either way we will be coming back to it later so I don't think it's going to save anytime to do it now.

src/Controller/FindByPid/ClipController.php
BjornTwachtmann: I think it might be better to do this filtering at the query stage than in code. Create a query with a WHERE c.pid NOT IN (...) or something?

src/Controller/FindByPid/ClipController.php
```diff
@@ -46,6 +50,21 @@ public function __invoke(
...
+            $parent = $clip->getParent();
+            if ($parent->getAvailableClipsCount() > 0) {
+                $parentClips = $this->getClipsExcept($aggregationService, $parent, $clip->getPid());
+            }
+            /** @var ProgrammeContainer|Episode $tleo */
```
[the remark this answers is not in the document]
BjornTwachtmann: should be in practice, yeah

src/Controller/FindByPid/ClipController.php
votemike: Should I write a new function in PPS? Or just updated the one that's currently there?
This is the way it's done in V2. And I thought it would be nicer to do this way than to write a whole new function. Up to you though.

PR #602, opened 2018-06-18
PROGRAMMES-6434 Swapping in canonical version instead of using helper

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6434

@BjornTwachtmann Do we want to get rid of `findByVersionWithContributions()` from PPS?


programmes-developers:

PR #604, opened 2018-06-18
Adding column attributes

votemike:
Adding data attributes for the testers.


programmes-developers:

PR #605, opened 2018-06-18
PROGRAMMES-6295 Fixing headings

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6295

See comments against the code.


src/DsAmen/Presenters/Section/Footer/footer.html.twig
```diff
@@ -1,5 +1,6 @@
 <div class="br-container br-box-subtle">
     <div class="p-g-p footer">
+        <h2 class="visually-hidden">{{ tr('related_content') }}</h2>
```
votemike: This keeps the Amen markup inline with the 2013 markup.
Is it work extracting them to a common base to avoid divergences like this?

programmes-developers:

templates/programme_episodes/player.html.twig
```diff
@@ -21,6 +21,8 @@
                     'highlight_box_classes': 'highlight-box--list',
                     title_options: {
                         'title_format': 'item::ancestry',
+                        'title_tag': 'h2',
```
votemike: Should we extract the programme presenter options out in to the `BaseProgrammeEpisodesController`? Should they all have the same options?

src/DsAmen/Presenters/Section/Footer/footer.html.twig
BjornTwachtmann: Dunno, my suspicion would be that the HTML is too different, but I haven't really looked...

templates/programme_episodes/player.html.twig
BjornTwachtmann: Doesn't look like they're using the same options, so probably not a good idea, they are presented differently between the pages.

votemike: Should they be using the same options? The title options were missing. Are the others?

PR #606, opened 2018-06-19
PROGRAMMES-6442 Fixing aria-expanded attributes

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6442


programmes-developers:

votemike:
When the parent series was expanded, all child .js-serieslink were marked as expanded. Now, just the first one is. Which should be correct. (It seems to work when testing.)

PR #608, opened 2018-06-19
[PROGRAMMES-6348][new] - Partial implementation of Clip Map-Details panel 

tatitati:
**This PR includes:**
-Basic skeleton for Clip map. This map includes Details panel(alsmost complete) and Playout panel (empty as a placeholder to populate in the future).

-From details panel ONLY left buttons to download and add to favourites, and credits. They will be created in a next PR to reduce the size of this.

**Examples URL**: 
https://www.bbc.co.uk/programmes/p01vfrv3
https://www.bbc.co.uk/programmes/p069d7vn
https://www.bbc.co.uk/programmes/p064bw8d

**Other tests:**
- [x] Responsive
- [x] headers tags changed to favor accessibility


programmes-developers:

src/Ds2013/PresenterFactory.php
```diff
@@ -310,6 +312,20 @@ public function recipePresenter(
         return new RecipePresenter($recipe, $options);
     }
 
+    /**
+     * @group MapClip
```
votemike: What's this?

src/Ds2013/Presenters/Section/Clip/Map/ClipMapPresenter.php
```diff
@@ -0,0 +1,34 @@
...
+    /** @var DetailsPresenter  */
+    private $detailsPresenter;
+
+    /** @var PlayoutPresenter */
+    private $playoutPresenter;
+
+    public function __construct(Clip $clip, PlayTranslationsHelper $playTranslationsHelper, array $options = [])
+    {
+        parent::__construct($options);
+        $this->detailsPresenter = new DetailsPresenter($clip, $playTranslationsHelper, $options);
```
votemike: Should the options be being passed through in this way?

src/Ds2013/Presenters/Section/Clip/Map/Panels/DetailsPresenter.php
```diff
@@ -0,0 +1,69 @@
...
+    public function getReleaseDate(): ?DateTime
+    {
+        if ($this->clip->getReleaseDate()) {
+            return $this->clip->getReleaseDate()->asDateTime();
+        }
+
+        return null;
+    }
+
+    public function getStreamableFrom(): ?DateTimeImmutable
```
votemike: Is this not a Chronos object?

src/Ds2013/Presenters/Section/Clip/Map/Panels/details.html.twig
```diff
@@ -0,0 +1,57 @@
+<div class="clip__details grid-wrapper">
+    <div class="grid 2/3@bpw2 3/4@bpe">
+        <div class="island">
+            <h1 class="details__title">{{ details.getTitle() }}</h1>
+
+            {%- if details.getLongestSynopsis() -%}
+                <div class="details__description" property="description">
```
votemike: Things like `property="description"` should be in the JSON Schema, not directly in the HTML.

src/Ds2013/Presenters/Section/Clip/Map/Panels/details.html.twig
```diff
@@ -0,0 +1,57 @@
...
+                    <p>{{- details.getLongestSynopsis() -}}</p>
+                </div>
+            {%- endif -%}
+
+            <div class="cf component" data-section="metadata">
+                {%- if details.getReleaseDate() -%}
+                    <div class="episode-panel__meta" property="datePublished" content="<?=$this->data->programme->release_date_iso?>">
+                        <h2 class="visually-hidden">{{ tr('release_date', {'%1': ''}) }}</h2>
```
votemike: Is this really an H2? Where is it in the page?

src/Ds2013/PresenterFactory.php
tatitati: phpunit allow to assign tags to tests, and make groups. Is good because when developing a component with multiple classes the most likely is that you want to run all those test-classes, test-functions related with the component. For instance it might exist a function in PresenterFactoryTest, and later DetailsPresenterTest, PlayoutPresenterTest, etc.

When this is done you can run your group quickly doing:
```vendor/bin/phpunit --no-coverage --group <my_new_component_name>```

This helps when doing TDD as you can just focus in your component and run tests quickly

Comment (votemike) [reply src/Ds2013/PresenterFactory.php]: I thought you put those tags in the code for the tests. Not in the actual production code.

File: src/Ds2013/Presenters/Section/Clip/Map/Panels/DetailsPresenter.php (hunk 3 above)
Comment (tatitati) [reply src/Ds2013/Presenters/Section/Clip/Map/Panels/DetailsPresenter.php]: No:
https://github.com/bbc/programmes-pages-service/blob/master/src/Domain/Entity/ProgrammeItem.php#L124-L126

File: src/Ds2013/Presenters/Section/Clip/Map/Panels/details.html.twig (hunk 5 above)
Comment (tatitati) [reply src/Ds2013/Presenters/Section/Clip/Map/Panels/details.html.twig]: http://sandbox.bbc.co.uk/programmes/p01vfrv3

File: src/Ds2013/PresenterFactory.php (hunk 1 above)
Comment (tatitati) [reply src/Ds2013/PresenterFactory.php]: ouch!, didn't see. True

File: src/Ds2013/Presenters/Section/Clip/Map/Panels/DetailsPresenter.php (hunk 3 above)
Comment (votemike) [reply src/Ds2013/Presenters/Section/Clip/Map/Panels/DetailsPresenter.php]: Have you tried dumping it? I think it's a Chronos object. I could be wrong though... https://github.com/bbc/programmes-pages-service/blob/866803fafec3e4ae2b8a421ee6730a3810b82c74/src/Data/ProgrammesDb/Type/UtcDateTimeType.php

**Comment by votemike** [comment]
On LIVE, when you hover over the "This Clip Is From" box, it highlights in gray. Which doesn't seem to happen in your branch.

**Comment by votemike** [comment]
On small screen sizes, the margin of the "This Clip Is From" box seems wrong.

File: src/Ds2013/Presenters/Section/Clip/Map/Panels/_details.scss [hunk 6]
```diff
@@ -0,0 +1,3 @@
+.clip__details .details__times {
```
votemike: I don't think you need this. At least, you shouldn't need it in this form. Comparing with live it looks like there should be a div with the classes clip-panel__buttons component and also ANOTHER div after that. They should sort out the spacing you are trying to create.

src/Ds2013/Presenters/Section/Clip/Map/Panels/details.html.twig
```diff
@@ -0,0 +1,58 @@
...
+                        'title_tag': 'h3',
+                        'title_classes': 'delta'
+                    },
+                    'body_options': {
+                        'highlight_box_classes': 'highlight-box--list-borderless programme--grid@bpw2 highlight-box--grid@bpw2 programme--grid@bpe highlight-box--grid@bpe',
+                        'show_synopsis': false,
+                        'show_duration': false
+                    },
+                    'image_options': {
+                        'classes': '1/3 1/1@bpe 1/1@bpw2'
```
votemike: Is 1/3 a class?

tatitati: yes:
https://github.com/bbc/programmes-frontend/blob/develop/src/Ds2013/Css/Core/Gel/Objects/_widths.scss#L37

As you can see we use Unicode to make possible to use classes starting by numbers, something that allows html, but not css.
So an example would be: 
```"<b class="10">" would be .\31 0 {}```

For reference:
https://benfrain.com/when-and-where-you-can-use-numbers-in-id-and-class-names/

File: src/Ds2013/Presenters/Section/Clip/Map/Panels/DetailsPresenter.php [hunk 8]
```diff
@@ -0,0 +1,69 @@
+<?php
+namespace App\Ds2013\Presenters\Section\Clip\Map\Panels;
```
votemike: Last thing. This presenter is no longer in a MAP. And not really a panel any more. Do we want to change the folder structure?

src/Ds2013/Presenters/Section/Clip/Map/Panels/DetailsPresenter.php
```diff
@@ -0,0 +1,102 @@
...
+    public function getReleaseDate(): ?DateTime
+    {
+        if ($this->clip->getReleaseDate()) {
+            return $this->clip->getReleaseDate()->asDateTime();
+        }
+
+        return null;
+    }
+
+    public function getStreamableFrom(): ?Chronos
```
BjornTwachtmann: I'm not really sure that all these wrapper functions around the domain object are helpful. Why not simply call that function on the domain object directly from the template, this way makes it harder to modify if we ever change the API of the domain objects, but no easer to use really.

src/Ds2013/Presenters/Section/Clip/Map/Panels/DetailsPresenter.php
BjornTwachtmann: This is not a MAP, or a Panel, please don't call it one. Section\Clip\Details seems better.

src/Ds2013/Presenters/Section/Clip/Map/Panels/DetailsPresenter.php
```diff
@@ -0,0 +1,102 @@
...
+namespace App\Ds2013\Presenters\Section\Clip\Map\Panels;
+
+use App\Ds2013\Presenter;
+use App\DsShared\Helpers\PlayTranslationsHelper;
+use BBC\ProgrammesPagesService\Domain\Entity\Clip;
+use BBC\ProgrammesPagesService\Domain\Entity\Programme;
+use Cake\Chronos\Chronos;
+use DateTime;
+
+class DetailsPresenter extends Presenter
```
BjornTwachtmann: ClipDetailsPresenter seems more descriptive?

src/Ds2013/Presenters/Section/Clip/Map/Panels/DetailsPresenter.php
```diff
@@ -0,0 +1,102 @@
...
+    public function isAvailableIndefinitely(): bool
+    {
+        if (!$this->clip->getStreamableUntil()) {
+            return true;
+        }
+
+        return !$this->clip->getStreamableUntil()->isWithinNext('1 year');
+    }
+
+    public function getStreamableTimeRemaining(): string
```
BjornTwachtmann: Is this function doing the same as translateAvailableUntilToWords in the PlayTranslationsHelper? It looks rather similar? How is it different?

src/Ds2013/Presenters/Section/Clip/Map/Panels/details.html.twig
```diff
@@ -0,0 +1,83 @@
+<div class="clip__details grid-wrapper">
+    <div class="grid 2/3@bpw2 3/4@bpe">
```
BjornTwachtmann: Presenters shouldn't really decide their own grid spacings normally, it's up to the calling template to put them into a grid. I know in this case we'll never reuse it, so it doens't matter, but just for general reference.

src/Ds2013/Presenters/Section/Clip/Map/Panels/details.html.twig
```diff
@@ -0,0 +1,83 @@
...
+            {%- if details.getLongestSynopsis() -%}
+                <div class="details__description">
+                    <p>{{- details.getLongestSynopsis() -}}</p>
+                </div>
+            {%- endif -%}
+
+
+            <div class="details__times cf component" data-section="metadata">
+                {%- if details.getReleaseDate() -%}
+                    <div class="episode-panel__meta" content="{{ details.getReleaseDate()|local_date('c') }}">
```
BjornTwachtmann: All these content attributes are schema.org. They are unused and should be removed.

src/Ds2013/Presenters/Section/Clip/Map/Panels/DetailsPresenter.php
tatitati: Is the same. In the next PR I will refactor both (details panel in episode and clip maps) to avoid duplication of code. I didn't want to do it in this one to avoid noise introducing changes in DetailsPresenter from Episode map as well.

src/Ds2013/Presenters/Section/Clip/Map/Panels/details.html.twig
```diff
@@ -0,0 +1,83 @@
...
+    <div class="grid 2/3@bpw2 3/4@bpe">
+        <div class="island">
+
+
+            <h1 class="details__title">{{ details.getTitle() }}</h1>
+
+
+            {%- if details.getLongestSynopsis() -%}
+                <div class="details__description">
+                    <p>{{- details.getLongestSynopsis() -}}</p>
```
BjornTwachtmann: V2 is doing truncation here. We probably need to do the same. There is a SynopsisPresenter that handles this, see the Episode page Details column for how it's used

src/Ds2013/Presenters/Section/Clip/Map/Panels/details.html.twig
```diff
@@ -0,0 +1,83 @@
...
+                {%- elseif details.getStreamableFrom() -%}
+                    <div class="episode-panel__meta" content="{{ details.getStreamableFrom()|local_date('Y-m-d') }}">
+                        <h2 class="visually-hidden">{{ tr('release_date', {'%1': ''}) }}</h2>
+                        <time class="details__streamablefrom" datetime="{{ details.getStreamableFrom()|local_date('c') }}">{{ details.getStreamableFrom()|local_date_intl('dd MMMM Y') }}</time>
+                    </div>
+                {%- endif -%}
+
+
+                {%- if details.isAvailableIndefinitely() == false -%}
+                    <p class="episode-panel__meta" typeof="OnDemandEvent">
```
BjornTwachtmann: All this typeof stuff is schema.org, please remove it.

src/Ds2013/Presenters/Section/Clip/Map/Panels/details.html.twig
```diff
@@ -0,0 +1,83 @@
...
+                    </div>
+                {%- elseif details.getStreamableFrom() -%}
+                    <div class="episode-panel__meta" content="{{ details.getStreamableFrom()|local_date('Y-m-d') }}">
+                        <h2 class="visually-hidden">{{ tr('release_date', {'%1': ''}) }}</h2>
+                        <time class="details__streamablefrom" datetime="{{ details.getStreamableFrom()|local_date('c') }}">{{ details.getStreamableFrom()|local_date_intl('dd MMMM Y') }}</time>
+                    </div>
+                {%- endif -%}
+
+
+                {%- if details.isAvailableIndefinitely() == false -%}
```
BjornTwachtmann: You also need to check for availability here, clips can be unavailble and the date logic below will probably crash in that case.

src/Ds2013/Presenters/Section/Clip/Map/Panels/details.html.twig
```diff
@@ -0,0 +1,83 @@
...
+                    },
+                    'body_options': {
+                        'highlight_box_classes': 'highlight-box--list-borderless programme--grid@bpw2 highlight-box--grid@bpw2 programme--grid@bpe highlight-box--grid@bpe',
+                        'show_synopsis': false,
+                        'show_duration': false
+                    },
+                    'image_options': {
+                        'classes': '1/3 1/1@bpe 1/1@bpw2'
+                    },
+                    'container_classes': 'highlight-box--list-borderless programme--grid@bpw2 highlight-box--grid@bpw2 programme--grid@bpe highlight-box--grid@bpe br-keyline br-blocklink-secondary',
```
BjornTwachtmann: Should both the body and the container have all these classes? That seems a bit weird to me. Does it work without the same classes on the body?

PR #609, opened 2018-06-19
Tidy up 2013 programme object standalone CTA

BjornTwachtmann:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6452


programmes-developers:

src/Ds2013/Presenters/Section/Episode/Map/Panels/Side/more.html.twig
```diff
@@ -43,8 +44,9 @@
                             },
                             image_options: {
                                 'show_image': false,
-                                'show_secondary_cta': true,
+                                'show_standalone_cta': true,
                                 'cta_link_location_track': 'upcoming_episode_map_calltoaction',
+                                'classes': '',
```
votemike: Is this needed? Seems.... icky.

BjornTwachtmann: Possbily not, I initially used this, then added the width: inherit instead, let me check...

PR #612, opened 2018-06-20
[PROGRAMMES-6463][fix] - Episode side-panels broken css

tatitati:
## Examples episodes with different panels combinations:

Table of truth + examples:

| Tx | More | Empty | URL example:
|--- | --- | --- |--- |
| 0 | 0 | 1 | http://sandbox.bbc.co.uk/programmes/b0074ggc | 
| 0 | 1 | 0 | http://sandbox.bbc.co.uk/programmes/p06467mn | 
| 1 | 0 | 1 | http://sandbox.bbc.co.uk/programmes/b00744w1|
| 1 | 1 | 0 | http://sandbox.bbc.co.uk/programmes/b0b6v9qt

# What
The problem affects to any combination of the map episode in which only one panel appear. These scenarios are:
  - Only appears More panel
  - Only appears Empty panel

# Why
The problem was introduced here:
https://github.com/bbc/programmes-frontend/commit/75f8ee678fa28703a3e10dfa581cfe68729894d1

As you can see ```.fauxcolumn``` was adding some special margins to fill the whole vertical space always in the last panel. However when we delete fauxcolumn we kept still the condition of ```loop.length > 1```, and this is wrong because sometimes appear only one panel (only More panel or only Empty panel) as you can see in the table of truth above and fill the whole vertical space


programmes-developers:

votemike:
On http://sandbox.bbc.co.uk/programmes/b0074ggc?__flush_cache=1 you can see that between 600px and 770px, there is an empty bar under the main programme info.

PR #613, opened 2018-06-21
Added visually hidden class to Amen

votemike:


programmes-developers:

BjornTwachtmann:
I think this just duplicates the existing .invisible class in DSAmen. Why they were named differently I don't know, but this concept already exists. There's also a .hidden for display:none, as well as the .invisible for "hidden visually but accessible to screen readers"

PR #614, opened 2018-06-21
PROGRAMMES-6463: Fix empty side panel 

BjornTwachtmann:
Move fauxcolumn into a separate class, rather than having everything in map__column


programmes-developers:

src/Ds2013/Presenters/Section/Episode/Map/_episode-map.scss
```diff
@@ -32,6 +32,10 @@
         margin-bottom: -50em;
         padding-bottom: 50em;
     }
+    .map__fauxcolumn {
+        margin-bottom: -50em;
```
votemike: This is duplicated above.

BjornTwachtmann: Fixed

src/Ds2013/Presenters/Section/Episode/Map/_episode-map.scss
```diff
@@ -28,7 +28,8 @@
 }
 
 @include mq-range('bpw2') {
-    .map--episode .map__column .map__inner {
+    .map--episode .map__column .map__inner,
```
votemike: Is this declaration still needed for map__inner?

BjornTwachtmann: yeah, it's on more than one thing in V2 also, christ knows why, but we've gone far enough down this rabbit hole

PR #615, opened 2018-06-21
PROGRAMMES-6394 Update TUPAC API to use HttpApiMultiClient

Tasiobg:
TUPAC API is limited to 100 results, this PR is to update the TupacService to use the new HttpApiMultiClient


programmes-developers:

PR #618, opened 2018-06-21
[PROGRAMMES-6462][fix] - Return signature not matching with real return type

tatitati:
This function ```getPrimaryContribution()```  is only used in:
1. this presenter:
https://github.com/bbc/programmes-frontend/blob/develop/src/Ds2013/Presenters/Section/Segments/SegmentItem/AbstractMusicSegmentItemPresenter.php#L62

2. And in this twig template:
https://github.com/bbc/programmes-frontend/blob/develop/src/Ds2013/Presenters/Section/Segments/SegmentItem/music.html.twig

## Url example to test:
Before fix:
http://test.bbc.co.uk/programmes/b04l04x5

After fix
http://sandbox.bbc.co.uk/programmes/b04l04x5


programmes-developers:

programmes-developers:

PR #619, opened 2018-06-21
[PROGRAMMES-6348][new] - Added favourites button to clip-details

tatitati:
## Url example to test:
http://sandbox.bbc.co.uk/programmes/p06bq4k6

You can also check the dev-tool in the browser and see the ajax calls to:
```
POST:  http://......../my/favourites/add
POST:  http://......../my/favourites/remove
```


programmes-developers:

src/Ds2013/Presenters/Section/Clip/Details/clip_details.html.twig
```diff
@@ -26,6 +26,7 @@
                     </div>
                 {%- endif -%}
 
+
```
votemike: 😠 😡 😠

tatitati: is not casual. I woudn't do this in a PHP code, but I think is not a bad idea in a css or twig code. And some other people from bbc also think in the same way:
https://cssguidelin.es/

I really think that adding two/three empty lines between independent blocks really make sense and help to make the logic blocks more visuals and readable.

votemike: I think if we're going to introduce things like this, there should be a discussion about it and the reason behind it should be explained rather than it just appearing out of the blue. I see the benefit of the idea, but I think 1 blank line is enough. But if other people overrule me, just as happy to go with them.

tatitati: OK, I will move back to 1 empty line then

votemike: Maybe ask about it in Slack? Or in person? Or in retro?

programmes-developers:

PR #621, opened 2018-06-22
Programmes 6431 fix double counting again

BjornTwachtmann:
Set BBCDotCom to comscoreDisabled on all pages with comscore enabled, not just pages with ads enabled.


programmes-developers:

PR #626, opened 2018-06-26
PROGRAMMES-6473 Adding links to networks on CollapsedBroadcast details

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6473


programmes-developers:

votemike:
I was wondering about that. But the classes on the elements are different. Which shouldn't be a problem to deal with now.... but if we want to use this in a 3rd place in the future and the markup for that is different, then we're knackered. What do you think?

PR #628, opened 2018-07-02
Update symfony/cache to 3.4.12 to fix lack of try/catch around redis …

BjornTwachtmann:
…exceptions on fetch

This fixes the issue that took programmes down on live


programmes-developers:

PR #629, opened 2018-07-03
[NO TICKET][refactor] - Renamed variables to avoid confusions

tatitati:
The last time I was working with something related with podcats I found these variables names very confusing. So I saw necessary to rename them


programmes-developers:

PR #630, opened 2018-07-03
Log warnings from Symfony\Cache (we usually only log errors). Remove …

BjornTwachtmann:
…app_access as it is unused.

Without this patch, errors caused by a redis CLUSTERDOWN condition will not be logged (it attempts to log them at DEBUG level, which is ridiculous)


programmes-developers:

PR #631, opened 2018-07-03
[NO TICKET][new] - Added some ids episode page DOM for QA

tatitati:
Specs coming from @jatinkaswa


programmes-developers:

PR #632, opened 2018-07-04
[PROGRAMMES-6268][change] - Schema update to include more broadcasts services

tatitati:
So far we have hardcoded the broadcast service as you can see here:
https://github.com/bbc/programmes-frontend/blob/develop/src/Controller/Helpers/SchemaHelper.php#L98

However, we have another one, playspace. This PR add the possibility of having 'iPlayer' or 'playspace' as our broadcastServices in schema data structure.


programmes-developers:

src/Builders/EpisodeBuilder.php
```diff
@@ -69,4 +70,41 @@ public static function anyTVEpisode()
...
+    {
+        return self::any()->with([
+            'mediaType' => MediaTypeEnum::AUDIO,
+            'masterBrand' => MasterBrandBuilder::any()->with([
+                'network' => NetworkBuilder::any()->build(),
+            ])->build(),
+        ]);
+    }
+
+    public static function anyInPlayspace()
```
votemike: I'm not sure Playspace and IPlayer should be mentioned in the Episode builder. Neither of them are concepts that Episodes are aware of. I think that there should be built in the tests you are using them in.
What do you think?

tests/Controller/Helpers/SchemaHelperTest.php
```diff
@@ -31,7 +35,7 @@ public function setUp()
         $router = $this->createConfiguredMock(UrlGeneratorInterface::class, [
             'generate' => 'this/url/was/stubbed',
         ]);
-        $this->helper = new SchemaHelper($router, $this->createMock(StreamableHelper::class));
+        $this->helper = new SchemaHelper($router, new StreamableHelper());
```
votemike: Why has this changed? Doesn't this mean you're now testing the `SchemaHelper` **AND** the `StreamableHelper` in this SchemaHelperTest?

src/Controller/Helpers/SchemaHelper.php
```diff
@@ -90,16 +90,32 @@ public function getSchemaForEpisode(Episode $episode): array
 
     public function getSchemaForOnDemandEvent(Episode $episode): array
     {
+        $name = null;
+        $urlBroadcastService = null;
+
+        if ($this->streamableHelper->shouldStreamViaPlayspace($episode)) {
+            $name = 'playspace';
```
votemike: Is it called playspace to the outside world? Is it not "iPlayer Radio"? Or "Sounds"?

src/Controller/Helpers/SchemaHelper.php
```diff
@@ -90,16 +90,32 @@ public function getSchemaForEpisode(Episode $episode): array
 
     public function getSchemaForOnDemandEvent(Episode $episode): array
     {
+        $name = null;
+        $urlBroadcastService = null;
```
votemike: This means the broadcast is on the /programmes page doesn't it? Would it be useful to point it to the correct page?

tatitati: Done

tests/Controller/Helpers/SchemaHelperTest.php
tatitati: True. The reason to use it as it is is that StreamableHelper is that is a very easy dependency to create (is not using any other dependency). 

This seems something trivial, but is really magic. Testing is not only about making sure that is not gonna break in the future, it also tells you if a component is becoming difficult to initialize because is starting to have many dependencies (which means that the class is doing too much). If you cannot use easily a component, then there is something wrong. Is in this way (one of the ways) how testing helps to improve the quality of the code.

Also there is some debate about what the term "UNIT" means. For some people a unit means a class, but there is also many professionals that think that a "UNIT" means "a component". 
Both of them are correct and people adapt this term to the problem they are working on (if you are crossing layers usually a UNIT means a class. However, if you are working with an individual component the term UNIT means a component). 

The funny thing about the term of unit testing is that the origin is not even related to software, but related to Mother Boards testing, in which a unit test meant "testing how the board works, as a whole". They didn't test screws, resistors, etc, but the board behaviour.

PR #634, opened 2018-07-05
PROGRAMMES-6391 Adding adverts to clip page

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6391

Can compare:
https://www.bbc.co.uk/programmes/p06c24f9?ads=big&site=test&ads-debug=1&ads-debug-trace=1&cb=1
https://sandbox.bbc.co.uk/programmes/p06c24f9?ads=big&site=test&ads-debug=1&ads-debug-trace=1&cb=1


programmes-developers:

templates/find_by_pid/clip.html.twig
```diff
@@ -102,6 +104,9 @@
             {% if tleoClips %}
                 {{ self.more_clips('tleoClips', programme.getTleo(), tleoClips) }}
             {% endif %}
+            {% if hasAdverts %}
```
BjornTwachtmann: Looking at it, I don't think you should actually need this wrapper, the BbcDotComExtension code should deal with it, plus you haven't wrapped the first one anyway. Probably test that though, I could be talking rubbish.

votemike: The other adverts in V3 work the same way... I can bin all of the hasAdverts stuff if you like.

BjornTwachtmann: The TLEC page needs that wrapper becuase it actually outputs weird bespoke wrapper divs conditionally. The episode page and clip page *shouldn't* need it, and I would remove it, but do check, like I said.

PR #637, opened 2018-07-06
[PROGRAMMES-6216][change] - Fix wrong title in tx panel for broadcast events

tatitati:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6216

# What:
If you go to this episode (in develop): http://sandbox.bbc.co.uk/programmes/b0bbmh5s?__flush_cache=1 and observe the title in TX-PANEL you will see the title "On Radio". However, going here: https://www.bbc.co.uk/programmes/b0bbmh5s?__flush_cache=1 you will see "On TV".

# Why:
This problem happens because we choose the title based on the programme item:
https://github.com/bbc/programmes-frontend/blob/develop/src/Ds2013/Presenters/Section/Episode/Map/Panels/Side/TxPresenter.php#L34

However V2 make the decisision based on the Main network service as you can see here:
https://github.com/bbc/programmes/blob/develop/webapp/php/lib/BBC/Programmes/View/Episode/Tx.php#L49

📝 If happy with the solution I will update tests (right now not passing).


programmes-developers:

src/Ds2013/Presenters/Section/Episode/Map/Panels/Side/tx.html.twig
```diff
@@ -1,10 +1,9 @@
 <div class="island">
-    <h2>{{ tr(tx.getTitle()) }}</h2>
-
     {{-
         ds2013('broadcastEvent', tx.getCollapsedBroadcast(), {
             'show_watch_from_start':true,
-            'show_on_air_message': false
+            'show_on_air_message': false,
+            'show_mainbroadcast_network_title': true
```
votemike: I don't think the title is related enough to the broadcast event to be part of the broadcastEvent presenter. Was there a reason you moved it inside?

src/Ds2013/Presenters/Domain/BroadcastEvent/BroadcastEventPresenter.php
```diff
@@ -140,4 +141,23 @@ protected function validateOptions(array $options): void
             throw new InvalidOptionException("show_on_air_message must a boolean");
         }
     }
+
+    public function getTitle() :?string
+    {
+        $mainNetworkForBroadcast = $this->getMainBroadcastNetwork();
+
+        if (!$mainNetworkForBroadcast) {
```
votemike: If there is no `$mainNetworkForBroadcast` should it not default to the `isRadio()` method for the ProgrammeItem?

src/Ds2013/Presenters/Domain/BroadcastEvent/BroadcastEventPresenter.php
```diff
@@ -140,4 +141,23 @@ protected function validateOptions(array $options): void
             throw new InvalidOptionException("show_on_air_message must a boolean");
         }
     }
+
+    public function getTitle() :?string
```
votemike: A test for this?

src/Ds2013/Presenters/Section/Episode/Map/Panels/Side/tx.html.twig
tatitati: I had a reason. However not sure if is good enough for this and I should try to find another solution, I wanted to ask you all. 

The reason is that this presenter uses a heavy operation that help to decide what LOGO network display inside the [broadcast_event.html.twig](https://github.com/bbc/programmes-frontend/blob/develop/src/Ds2013/Presenters/Domain/BroadcastEvent/broadcast_event.html.twig#L12) . You can see this operation in this presenter [BroadcastEventPresenter.php#L52](https://github.com/bbc/programmes-frontend/blob/develop/src/Ds2013/Presenters/Domain/BroadcastEvent/BroadcastEventPresenter.php#L52)

The content of this ```getNetworksAndServicesDetails()```  function can be found here:
[BroadcastNetworksHelper.php#L32](https://github.com/bbc/programmes-frontend/blob/develop/src/DsShared/Helpers/BroadcastNetworksHelper.php#L32)

As you can see the steps needed to decide the logo and the title of the broadcast-event are based on the same function. Cause of that I tried to avoid a second call to this function from txPresenter, moving this function to BroadcastEventPresenter. However, that is my reason....

**What do you think?**

votemike: I don't think that's a good enough reason to move it in.
If you want to avoid the heavy operation from running twice, you should move that operation out of the presenter and pass the result it somehow.
That's my opinion anyway.

tatitati: Yeah, I agree. I will do that. @BjornTwachtmann any other thought?

PR #642, opened 2018-07-09
Optimize Imports

votemike:


programmes-developers:

src/Fixture/Doctrine/Entity/HttpFixture.php
```diff
@@ -5,7 +5,6 @@
 
 use DateTime;
 use Gedmo\Timestampable\Traits\TimestampableEntity;
-use Doctrine\ORM\Mapping as ORM;
```
BjornTwachtmann: This will break the entire fixturing mechanism, or at least it did the last time you raised this PR :)

src/Fixture/Doctrine/Entity/Scenario.php
```diff
@@ -5,7 +5,6 @@
 
 use DateTimeInterface;
 use Gedmo\Timestampable\Traits\TimestampableEntity;
-use Doctrine\ORM\Mapping as ORM;
```
BjornTwachtmann: Again, breaks fixtures, doctrine and magic annotations...

PR #643, opened 2018-07-09
Music segments fixes for future broadcasts

BjornTwachtmann:
* Prevent segments list rendering for broadcasts in the past without show_tracklist_inadvance enabled.

* Modify SegmentsListPresenter to behave more like V2. Using the first broadcast instead of an upcoming/last-on combo


programmes-developers:

src/Controller/FindByPid/EpisodeController.php
```diff
@@ -98,7 +98,8 @@ public function __invoke(
         $previousEpisode = null;
 
         $upcomingBroadcast = !empty($upcomingBroadcasts) ? reset($upcomingBroadcasts) : null;
-        $lastOnBroadcast =  !empty($lastOnBroadcasts) ? reset($lastOnBroadcasts) : null;
+        $lastOnBroadcast = !empty($lastOnBroadcasts) ? reset($lastOnBroadcasts) : null;
```
votemike: Is this used any more?

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
```diff
@@ -217,17 +215,23 @@ protected function validateOptions(array $options): void
...
     {
         if ($this->context->getOption('show_tracklist_inadvance') ||
-            !$this->collapsedBroadcast ||
-            $this->collapsedBroadcast->isRepeat() ||
-            !$this->liveBroadcastHelper->isOnNowIsh($this->collapsedBroadcast, true)
+            !$this->firstBroadcast ||
+            $this->context instanceof Clip ||
+            $this->firstBroadcast->getEndAt()->isPast() // Episode last broadcast in the past
         ) {
+            // In all these cases we want to show the full track list in chronological order
```
votemike: If there's the need to add these comments, should we instead refactor?
E.g. `if ($this->showFullTrackListInChronilogicalOrder($blah))` and `if ($this->beingBroadcastForFirstTime($blah))`?

src/Controller/FindByPid/EpisodeController.php
BjornTwachtmann: Yeah, in a few places, the iStats stuff and episode map

src/Ds2013/Presenters/Section/Segments/SegmentsListPresenter.php
BjornTwachtmann: I can do if you think that's clearer, I thought keeping the logic together might be less confusing, but happy to change it if people think otherwise.

votemike: Long comments seems like code smell to me.
Happy to go with the majority though.

BjornTwachtmann: They are a code smell, this code is horrible, however the problem it solves is also horrible :) @Tasiobg @tatitati , any opinions?

PR #645, opened 2018-07-10
PROGRAMMES-6216: Take account of broadcast's radio/tv status when displaying TX column title

BjornTwachtmann:


programmes-developers:

votemike:
If *any* of the CollapsedBroadcasts services are radio, it should be treated as radio?

BjornTwachtmann:
@votemike , yeah, that's the idea. A Collapsed broadcast is a list of everywhere the same Episode is being broadcast at the same time. The same thing is never really broadcast on radio/tv simultaneously. The same event might be (WC final etc.), but it's always a different episode in those cases.

Which, if you think about it makes sense, an audio_video episode on radio is nonsence.

PR #646, opened 2018-07-10
Prevent snippet player JS from doing empty request

Tasiobg:
Chapter segment pages may include the snippet JS librarty and this could generate an empty request to the snippet player controller, example: /programmes/snippet/.json

This small improvement double check that there is at least one record Id before calling the snippet endpoint.


programmes-developers:

PR #648, opened 2018-07-10
Bump up the size of our AD/SL icons because they're too small

BjornTwachtmann:


programmes-developers:

PR #649, opened 2018-07-11
[PROGRAMMES-6514] Fix Previous and Next buttons being misaligned

benolot:
Ticket: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6514

This brings it in line with the live (V2) behavior.


programmes-developers:

src/Ds2013/Presenters/Domain/CoreEntity/Programme/_programme.scss
```diff
@@ -43,6 +43,10 @@ $programme-icon-bpw-box-size: gel-typography('bpw', 'delta', 'line-height') + 18
     }
 }
 
+.programme__img--standalone-cta:not(.programme__img--available) {
+    height: 32px;
```
tatitati: Seems ok to me. I think it might improve a bit if we use a variable we have to avoid inconsistent numbers in the future. We have $wide-spacing = 16px. We might use it here like : ```height: $wide-spacing * 2```

benolot:
Have added support for wider screensizes, and there is no variables that can be used in this situation, and the other approach of adding classes (EG: 1/3@bpw) doesn't work either because of where the classes are applied. Happy to explain why if anyone needs.

benolot:
Test this please

benolot:
@votemike Happy with changes made?

PR #650, opened 2018-07-12
PROGRAMMES-6448 add new GuidanceWarningHelper

Tasiobg:
Implements guidance warning text helper, the class expect the database field value of version.guidance_warning_codes to generates the text. All available codes and text are attached in a CSV

An example of how this message would be diplayed could be seen in v2 pages https://www.bbc.co.uk/programmes/p04bzrtr

This work is based on ticket PROGRAMMES-6448 but we have decided to do a different implementation, instead of doing it through Faucet ingest I'm doing it directly on the frontend.


programmes-developers:

src/DsShared/Helpers/GuidanceWarningHelper.php
```diff
@@ -0,0 +1,119 @@
...
+     * Given a list of guidance warning codes separated by comas it returns the guidance warning text
+     *
+     * @param string $guidanceWarningCodes
+     * @param string $textLength
+     * @return string
+     */
+    public function getText(string $guidanceWarningCodes, $textLength = GuidanceWarningEnum::TEXT_LONG): string
+    {
+        $codes = array_map('trim', explode(',', $guidanceWarningCodes));
+        $codes = $this->orderCodes($codes);
```
votemike: No need to order the codes until you know there are more than 0.

src/DsShared/Helpers/Enumeration/GuidanceWarningEnum.php
```diff
@@ -0,0 +1,128 @@
...
+
+    public const TEXT_SHORT = 'SHORT';
+    public const TEXT_LONG = 'LONG';
+
+    /**
+     * Guidance warning short text
+     */
+
+    // Language
+    public const L1_SHORT = 'some strong language';
```
votemike: Do these not need to be translated?

src/DsShared/Helpers/GuidanceWarningHelper.php
```diff
@@ -0,0 +1,119 @@
...
+        $position = 1;
+        // start building the text
+        $text = 'Contains ';
+        $and = ' and ';
+        // Handle a different behaviour for and edge case
+        if ($textLength === GuidanceWarningEnum::TEXT_LONG && in_array(GuidanceWarningEnum::W2, $codes)) {
+            $text = 'Deals ';
+            $and = ' and contains ';
+        }
+        foreach ($codes as $code) {
```
votemike: Instead of this what about something like:
```
$newCodes = [];
foreach($codes as $code) {
    $newCodes = $this->getTextForSingleCode($code, $textLength);
}
$last = array_pop($newCodes);
$text .= implode(',', $newCodes) . $and . $last . '.';
```

src/DsShared/Helpers/Enumeration/GuidanceWarningEnum.php
Tasiobg: No, there is no translations support for it

src/DsShared/Helpers/GuidanceWarningHelper.php
Tasiobg: Despite only valid codes are expected if for any reason an invalid one is present the orderCodes() won't add it to the new array so is safer to use the counter after ordering the array.

src/DsShared/Helpers/GuidanceWarningHelper.php
Tasiobg: This one works fine
```
        $newCodes = [];
        foreach ($codes as $code) {
            $newCodes[] = $this->getTextForSingleCode($code, $textLength);
        }
        $last = array_pop($newCodes);
        $and = ($numberOfCodes > 1) ? $and : '';
        $text .= implode(', ', $newCodes) . $and . $last . '.';
```
@BjornTwachtmann @benolot any prefences?

src/DsShared/Helpers/GuidanceWarningHelper.php
```diff
@@ -0,0 +1,120 @@
...
+     * @param array $codes
+     * @return array
+     */
+    private function orderCodes(array $codes): array
+    {
+        $orderedCodes = [];
+        foreach ($codes as $code) {
+            // first "watershed"
+            if (in_array($code, [
+                GuidanceWarningEnum::WV,
```
BjornTwachtmann: Is it worth defining these lists as constant arrays in the Enum? Or is this more readable? Just a thought.

tests/DsShared/Helpers/GuidanceWarningHelper/GuidanceWarningHelperTest.php
```diff
@@ -0,0 +1,38 @@
...
+    /** @var GuidanceWarningHelper */
+    private $helper;
+
+    public function setUp()
+    {
+        $this->helper = new GuidanceWarningHelper();
+    }
+
+    /** @dataProvider getAllCombinationsProvider */
+    public function testAllCombinations($adult, $language, $sex, $violence, $disturbing, $rfi, $watershed, $shortText, $longText)
```
BjornTwachtmann: Did you figure out why this test was slow? How long is it taking now?

src/DsShared/Helpers/GuidanceWarningHelper.php
BjornTwachtmann: Probably this version is slightly easer to read. I don't have a strong preference though

src/DsShared/Helpers/GuidanceWarningHelper.php
votemike: Hmmm.... OK

tests/DsShared/Helpers/GuidanceWarningHelper/GuidanceWarningHelperTest.php
Tasiobg: Doesn't seems to affect the total time, I thought unitest was faster but turns out that even removing this test takes around 22s in my laptop

src/DsShared/Helpers/GuidanceWarningHelper.php
```diff
@@ -0,0 +1,101 @@
...
+        if ($textLength === GuidanceWarningEnum::TEXT_LONG && in_array(GuidanceWarningEnum::W2, $codes)) {
+            $text = 'Deals ';
+            $and = ' and contains ';
+        }
+        $newCodes = [];
+        foreach ($codes as $code) {
+            $newCodes[] = $this->getTextForSingleCode($code, $textLength);
+        }
+        $last = array_pop($newCodes);
+        $and = ($numberOfCodes > 1) ? $and : '';
```
votemike: You could even do:
```
if (empty($newCodes)) {
   return $last . '.';
}
```
It might be a bit more obvious what's happening.

PR #653, opened 2018-07-13
PROGRAMMES-6477 Profile Page Scaffolding

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6477


programmes-developers:

templates/profiles/show.html.twig
```diff
@@ -0,0 +1,13 @@
+{% extends 'base_ds2013.html.twig' %}
+
+{% block page_classes %}text-base{% endblock %}
+
+{% block body %}
+    <div class="prog-layout">
```
votemike: These classes seem to be elsewhere on different pages. https://github.com/bbc/programmes-frontend/blob/develop/templates/profiles/index.html.twig#L4-L7
Does anyone know why?

src/Controller/Helpers/IsiteKeyHelper.php
```diff
@@ -0,0 +1,58 @@
...
+        return substr($characters, 0, 8) . '-' .
+            substr($characters, 8, 4) . '-' .
+            substr($characters, 12, 4) . '-' .
+            substr($characters, 16, 4) . '-' .
+            substr($characters, 20, 12);
+    }
+
+    private function convertBase(string $characters, string $fromString, string $toString): string
+    {
+        $iFromBase = \strlen($fromString);
```
Tasiobg: Why is `\strlen` prefixed with the backslash? I find it clear to use php core functions without namespace like you already do with substr

votemike: For performance. It stops PHP looking in the current namespace for a method with that name.

src/Controller/Profiles/ShowController.php
```diff
@@ -0,0 +1,26 @@
...
+
+class ShowController extends BaseController
+{
+    public function __invoke(string $key, string $slug, Request $request, IsiteKeyHelper $isiteKeyHelper)
+    {
+        if (strpos($key, '-') !== false) { // if contains dashes it is a guid
+            $key = $isiteKeyHelper->convertGuidToKey($key);
+
+            $params = ['key' => $key, 'slug' => $slug];
+            if ($request->query->has('preview') && $request->query->get('preview')) {
```
Tasiobg: Why do you use the second condition in the if? is because if "preview" is empty this condition would be evaluated to false?

votemike: Yeah, preview needs to have the correct value as well as being present.

templates/profiles/show.html.twig
Tasiobg: Seems to be used acros the site to control the layouts, there are some comments in _prog-layout.scss

src/Controller/Helpers/IsiteKeyHelper.php
Tasiobg: Why is added only to `strlen` and not to `strpos` or other functions?
Does it make a difference in performance? if so we should consider to use namespaces always like this but I found strange that this could be noticed in the performance

templates/profiles/show.html.twig
votemike: I saw those, I didn't quite understand the comments though.
How do you think these classes should be used? As they are here? Or as they are in the linked file?

src/Controller/Helpers/IsiteKeyHelper.php
votemike: See https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/ and https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3048#issue-256982855
I assumed everyone was using these. Does it not highlight for you in Intellij?

templates/profiles/show.html.twig
Tasiobg: I think this one is right and looks like v2

src/Controller/Helpers/IsiteKeyHelper.php
Tasiobg: Doesn't seems to be highlight in my Intellij.
I'm suprise abot this PHP behaviour, thanks

templates/profiles/show.html.twig
votemike: @BjornTwachtmann any views on this?

BjornTwachtmann: These were created before my time here. It would _seem_ to make more sense to add them to the page_classes, as elsewhere, don't know if there's a reason this is different or not tbh. Try it, we can always change it back.

votemike: Moving this one seems fine...

PR #654, opened 2018-07-13
Fix docker build blowing up when dev dependencies not present in prod…

BjornTwachtmann:
…uction

Sam fixed this in blogs already, just applying to programmes. It prevents our build breaking when dev dependencies are not present in production.


programmes-developers:

PR #655, opened 2018-07-16

Tasiobg:
Update all HTTP APIs to use the Multi client and retire the single URL HttpApiClient

This is only for Electron, I'll open more PR for the remaining APIs and for removing the single HttpApiClient

- Updated Electron to use Multi client
- Fix missing ParseException in HttpApiMultiClient.php
- Fix MultiParseException, $responseKey can be integer or string
- Fix wrong use of MultiParseException in TupacService.php


programmes-developers:

src/ExternalApi/Exception/MultiParseException.php
```diff
@@ -10,7 +10,7 @@ class MultiParseException extends \RuntimeException
     /** @var string */
     private $responseKey;
 
-    public function __construct(string $responseKey, $message = "", $code = 0, Exception $previous = null)
```
votemike: Why was this removed?

Tasiobg: We found an issue because `$responseKey` can be integer or string

votemike: Brilliant.... OK.

votemike: Should you update the docblock then?

src/ExternalApi/Tupac/Service/TupacService.php
```diff
@@ -63,13 +63,13 @@ private function generateRequestUrl(array $recordIds, bool $isUk = false): strin
     private function parseResponse(array $responses): array
     {
         $results = [];
-        foreach ($responses as $response) {
+        foreach ($responses as $key => $response) {
             if (!$response instanceof Response) {
                 throw new MultiParseException("TUPAC callback received non-response object!");
```
BjornTwachtmann: You missed the response key here, no?

PR #656, opened 2018-07-16
PROGRAMMES-6480 Profile Listing iSite Queries

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6480


programmes-developers:

src/ExternalApi/Isite/Domain/Profile.php
```diff
@@ -0,0 +1,45 @@
+<?php
+declare(strict_types = 1);
+
+namespace App\ExternalApi\Isite\Domain;
+
+class Profile
+{
+    //Image
```
votemike: I think we'll end up needing these, but I haven't added them yet as I don't want to add redundant fields. I reckon they can be added as part of https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6500, https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6478 or https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6479

src/ExternalApi/Isite/IsiteFeedResponseHandler.php
```diff
@@ -0,0 +1,99 @@
...
+
+class IsiteFeedResponseHandler
+{
+    /** @var ProfileMapper */
+    private $mapper;
+
+    /** @var XmlParser */
+    private $xmlParser;
+
+    public function __construct(ProfileMapper $mapper, XmlParser $xmlParser)
```
votemike: We only need the `ProfileMapper` at the moment. When it comes time to do the Articles, we'll want this to accept a generic mapper, but I didn't want to pre-emptivley convolute this code.

src/ExternalApi/Isite/IsiteFeedResponseHandler.php
```diff
@@ -0,0 +1,99 @@
...
+            throw new IsiteResultException('Invalid Isite response body.', 0, $e);
+        }
+
+        $envelopeName = $decodedResponseBody->getName();
+
+        try {
+            if ($envelopeName === 'search') {
+                return $this->parseSearchEnvelope($decodedResponseBody);
+            }
+            if ($envelopeName === 'result') {
```
votemike: Not used yet, but will be soon.

src/ExternalApi/Isite/Domain/Profile.php
Tasiobg: This looks good but I suggest to check v2 for edge case like if we have to deal with other writing system, also I noticed that for "O’Donnell" in v2 the slug is "o-donnell" while here is "odonnell" might be right but just mentioning it in case you were not aware

src/ExternalApi/Isite/Domain/Profile.php
```diff
@@ -0,0 +1,56 @@
...
+        $this->key = $key;
+    }
+
+    public function getKey()
+    {
+        return $this->key;
+    }
+
+    public function getSlug()
+    {
```
Tasiobg: This looks good but I suggest to check v2 for edge case like if we have to deal with other writing system, also I noticed that for "O’Donnell" in v2 the slug is "o-donnell" while here is "odonnell" might be right but just mentioning it in case you were not aware

src/ExternalApi/Isite/Service/IsiteService.php
```diff
@@ -0,0 +1,100 @@
...
+                ],
+            ])
+            ->setPage($page)
+            ->setPageSize($limit);
+
+        $url = $this->baseUrl . $query->getPath();
+
+        $cacheKey = $this->clientFactory->keyHelper(__CLASS__, __FUNCTION__, $programme->getPid(), $page, $limit);
+
+        return $this->clientFactory->getHttpApiClient(
```
Tasiobg: Please use `getHttpApiMultiClient`, I'm removing `getHttpApiClient` in my current branch

Tasiobg:
Not sure if you are planing to do it as part of this ticket but new APIs should be added to the circuit breakers and set some alars.
I think you only need to update this 3 classes for the circuit breakers:
```
CircuitBreakerFactory.php
ApiTypeEnum.php
UriToApiTypeMapper.php
```
Also you should do something like this for this API https://github.com/bbc/programmes-infrastructure/commit/2e8136ef3e3b12070f3366be556c28aa4a461b79

src/ExternalApi/Isite/Domain/Profile.php
votemike: Does it? Do you have example URL's where you're seeing this?

Tasiobg: http://www.bbc.co.uk/programmes/profiles/3HXvg4YTQL7VFZxC7ZyB79Q/o-donnell

votemike: It comes out as "o-donnell" still in V3 for me....

votemike:
Ticket for alarms is here: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6521

src/ExternalApi/Isite/SearchQuery.php
```diff
@@ -0,0 +1,85 @@
...
+        return $this;
+    }
+
+    public function setPageSize(int $resultsPerPage): self
+    {
+        if ($resultsPerPage > self::MAX_PAGE_SIZE || $resultsPerPage < 0) {
+            throw new InvalidArgumentException('$resultsPerPage must be between 0 and ' . self::MAX_PAGE_SIZE);
+        }
+
+        $this->q->pageSize = (string) $resultsPerPage;
```
tatitati: Is interesting that we delegate to $this->q the whole state of this SearchQuery class. From my point of view is easier to understand this class if you open it and see that the state of this object (a search query) is specified by properties like a pagesize, a query, namespace, a depth, project and a sort (sometimes even protocol, headers, method....). Those parameters conform what you expects to see about a query.

src/Controller/Profiles/IndexController.php
```diff
@@ -4,20 +4,30 @@
...
+
         $parameters = ['coreEntity' => $coreEntity];
         if ($coreEntity instanceof Programme) {
             $parameters['programme'] = $coreEntity; //so the the base 2013 template sets the footer
+
+            $profilesPromise = $isiteService->getProfilesByProgramme($coreEntity, $this->getPage());
         }
 
-        return $this->renderWithChrome('profiles/index.html.twig', $parameters);
+        $resolvedPromises = $this->resolvePromises([
```
BjornTwachtmann: This needs to return a 404 status code if there are no profiles I thing, that's what live does at least

tests/ExternalApi/Isite/IsiteResultTest.php
```diff
@@ -0,0 +1,66 @@
...
+            $this->createMock(Profile::class),
+            null,
+        ];
+
+        $result = new IsiteResult(1, 1, 1, $mockDomainModels);
+
+        $this->assertInternalType('array', $result->getDomainModels());
+        $this->assertContainsOnlyInstancesOf(Profile::class, $result->getDomainModels());
+        $this->assertCount(3, $result->getDomainModels());
+    }
```
tatitati: Not sure if this might improve the readability:

```
        $mockDomainModels = [
            null,
            $prof1 = $this->createMock(Profile::class),
            $prof2 = $this->createMock(Profile::class),
            null,
            $prof3 = $this->createMock(Profile::class),
            null,
        ];


        $result = new IsiteResult(1, 1, 1, $mockDomainModels);

        $this->assertEquals([$prof1, $prof2, $prof3], $result);
```

src/ExternalApi/Isite/SearchQuery.php
votemike: It was mostly to make the `getPath()` method clean.
Otherwise that method would be something like:
```
public function getPath(): string
{
    $q = new stdClass();
    if (!empty($this->page)) {
        $q->page = $this->page;
    }
    ............................
    return '/search?' . http_build_query(['q' => json_encode($q)]);
}
```
I guess if you did it with separate class variables, you could json_encode the class. But I don't like the sound of that.

BjornTwachtmann:
It might also be useful to add some tests for IsiteService, a lot of the other stuff uses that for end-to-end tests, e.g. https://github.com/bbc/programmes-frontend/blob/develop/tests/ExternalApi/Electron/Service/ElectronServiceTest.php . Not essential though.

src/ExternalApi/Isite/SearchQuery.php
Tasiobg: May be is a good idea to create a domain class for `$this->q`

votemike: SearchQueryQuery?

src/ExternalApi/Isite/Service/IsiteService.php
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * @param Response[] $responses
+     * @return IsiteResult
+     */
+    public function parseResponse(array $responses): IsiteResult
+    {
+        return $this->responseHandler->getIsiteResult($responses[0]);
+    }
+
+    private function makeClient(Programme $programme, int $page, int $limit, string $ttl, string $nullTtl): HttpApiClient
```
Tasiobg: returns HttpApiMultiClient

src/ExternalApi/Isite/SearchQuery.php
Tasiobg: Not sure what could be best name, but is just a suggestion, not a required change, up to you.
Another option could be to set the propertis in SearchQuery.php and use this code.
```return '/search?' . http_build_query(['q' => json_encode(get_object_vars($this))]);```
 not sure if works, need testing*

File: tests/ExternalApi/Isite/IsiteResultTest.php (hunk 8 above)
Comment (votemike) [reply tests/ExternalApi/Isite/IsiteResultTest.php]: That doesn't seem to work as it also checks the array keys.

**Comment by votemike** [comment]
@BjornTwachtmann so you'd like me to create some dummy XML files?

File: src/ExternalApi/Isite/SearchQuery.php (hunk 6 above)
Comment (votemike) [reply src/ExternalApi/Isite/SearchQuery.php]: I think I'm happiest with how it is now. Although happy to change if someone comes up with a better way or wants to over-rule me.

File: src/ExternalApi/Isite/Service/IsiteService.php [hunk 10]
```diff
@@ -0,0 +1,162 @@
...
+                        'direction' => 'asc',
+                    ],
+                ]);
+//            ->setPage($page); //@TODO there's no limit, do I need the page?
+
+            $urls[] = $this->baseUrl . $query->getPath();
+            $cacheKeys[] = $profile->getFileId();
+        }
+
+        $cacheKey = $this->clientFactory->keyHelper(__CLASS__, __FUNCTION__, implode(',', $cacheKeys)/*, $page, $limit*/);
```
votemike: Do we want to paginate this query? Or just get everything?
@BjornTwachtmann  ?

src/ExternalApi/Isite/Service/IsiteService.php
```diff
@@ -0,0 +1,162 @@
...
+use App\ExternalApi\Isite\IsiteFeedResponseHandler;
+use App\ExternalApi\Isite\IsiteResult;
+use App\ExternalApi\Isite\SearchQuery;
+use BBC\ProgrammesCachingLibrary\CacheInterface;
+use BBC\ProgrammesPagesService\Domain\Entity\Programme;
+use Closure;
+use GuzzleHttp\Promise\PromiseInterface;
+use GuzzleHttp\Psr7\Response;
+
+class IsiteService
```
votemike: Do people think I should change this to `ProfileService`?

BjornTwachtmann: Depends if we want separate queries for different services. I suspect many of the queries for profiles/articles will be very similar, so maybe not yet?

src/ExternalApi/Isite/Service/IsiteService.php
BjornTwachtmann: Allowing data to be retrieved without limit sounds a bit dangerous, I'd have pagination and a page/limit myself. We can always set it high.

src/ExternalApi/Isite/SearchQuery.php
BjornTwachtmann: Don't really have a strong preference myself. Isite queries are weird and magical, not sure there's a way round it

src/ExternalApi/Isite/Service/IsiteService.php
```diff
@@ -0,0 +1,169 @@
...
+        );
+
+        return $client->makeCachedPromise();
+    }
+
+    public function getProfilesByProgramme(
+        Programme $programme,
+        int $page = 1,
+        int $limit = 48,
+        $ttl = CacheInterface::NORMAL,
```
votemike: Do we want to pass $ttl and $nullTtl through each time or just hardcode it in the call to `getHttpApiMultiClient`?

src/ExternalApi/Isite/Service/IsiteService.php
```diff
@@ -0,0 +1,169 @@
...
+                'and' => [
+                    ['profile:parent_pid', '=', (string) $programme->getPid()],
+                    [
+                        'not' => [
+                            ['profile:parent', 'contains', 'urn:isite'],
+                        ],
+                    ],
+                ],
+            ])
+            ->setSort([
```
BjornTwachtmann: Do you think it's worth abstracting the duplicate parts of this query, or is that just muddying the waters?

votemike: Just in the middle of it.... :-D

src/ExternalApi/Isite/Mapper/ProfileMapper.php
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    public function getDomainModel(SimpleXMLElement $isiteObject): Profile
+    {
+        $formMetaData = $this->getFormMetaData($isiteObject);
+
+        $key = $this->isiteKeyHelper->convertGuidToKey($this->getString($this->getMetaData($isiteObject)->guid));
+        $title = $this->getString($formMetaData->title);
+        $type = $this->getString($formMetaData->type);
+        $fileId = $this->getString($this->getMetaData($isiteObject)->fileId); // NOTE: This is the metadata fileId, not the form data file_id
```
votemike: In blogs-frontend this is a valueobject so that it's value can be validated. https://github.com/bbc/blogs-frontend/blob/master/src/BlogsService/Domain/ValueObject/FileID.php do we want the same thing here?

votemike:
I have added the tests for IsiteService asked for by @BjornTwachtmann

src/ExternalApi/Isite/Service/IsiteService.php
benolot: I don't have any strong feelings either way, but personally I think it's better to pass it through each time.

src/ExternalApi/Isite/Mapper/ProfileMapper.php
benolot: This is generated on iSite and returned to us via iSite so I'm not sure if theres much point in us checking whether it's valid or not? As mentioned we don't validate type, so why just file ID?

Tasiobg: Not familiar with iSite API but I suppose that only valid `fileId` should be returned so maybe is not required to create a valuableobject

src/ExternalApi/Isite/Service/IsiteService.php
Tasiobg: In the other APIs the function calling getHttpApiMultiClient is not passing the $ttl so I think it is more consisten to hardcode it directly in getHttpApiMultiClient. Also it would be strange to be able to se the same redis key with different TTL's, generally is a good practive to use same TTL for same key.

BjornTwachtmann: The question for me would be "Will we need this parameter". I think probably not in this case. We have this as a parameter in pages-service as it's a library which may have different use cases. This function probably only has one use case. I don't think it's a problem to have it either though.

src/ExternalApi/Isite/Domain/Profile.php
```diff
@@ -0,0 +1,93 @@
...
+        $this->fileId = $fileId;
+        $this->type = $type;
+    }
+
+    /**
+     * @return Profile[]
+     */
+    public function getChildren(): array
+    {
+        return $this->children;
```
BjornTwachtmann: This is potentially misleading, I think. The children of a profile aren't always fetched, and aren't required by the constructor, so somebody could read this property without querying the children and think there simply were none. Pages-service handles this by throwing a DataNotFetchedException when the setter hasn't been called, you think that would make sense here?

votemike: I would hope the devs are sensible, but we could throw a NotFetchedException if you like. How do you imagine it working?
Set the property as null, then when it's fetched it might end up as an empty array?

BjornTwachtmann: yeah, default it to null, throw an exception if it's null, setChildren can set an empty array if there are no children

programmes-developers:

votemike:
Test this please

PR #657, opened 2018-07-16
[PROGRAMMES-6348][new] - Now smp player reproduce playlist

tatitati:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6348

Specs of smp player:
- [x] It can provide url to embed the video
- [x] It remembers the last moment was played in order to resume reproduction
- [x] It remembers volume level
- [x] It can display in full-screen
- [x] It start from start when the video finishes
- [x] it can start from the desired moment in the url. Example: **/programmes/p06dtqzw#t=00h00m30s**
 

Url to test:
http://sandbox.bbc.co.uk/programmes/p06dtqzw?__flush_cache=1


programmes-developers:

assets/js/smp/playerRepository.js
```diff
@@ -0,0 +1,95 @@
...
+                    if (cache[i].pid === programmePid) {
+                        return cache[i].time;
+                    }
+                }
+            }
+
+            return 0;
+        };
+
+        this.addVolume = function (volumeValue) {
```
votemike: Does this only addVolume, or does it subtract volume? Should the method be something like changeVolume?

assets/js/smp/smp.js
```diff
@@ -0,0 +1,186 @@
...
+        var self  = this;
+
+        var restoreUserPrefs = function() {
+            self.options.smpSettings.volume = self.playerRepostiroy.findByVolume();
+            self.options.smpSettings.muted = self.options.smpSettings.volume === 0;
+        };
+
+        var resumeReproductionTime = function() {
+            var timeInUrl = getTimeFromUrl();
+            if (timeInUrl != 0) {
```
votemike: What if they want it to start from 0?

assets/js/smp/smp.js
```diff
@@ -0,0 +1,186 @@
...
+
+        var listenVolumeEvents = function() {
+            self.player.bind('volumechange', function (event) {
+                self.playerRepostiroy.addVolume(event.volume);
+            });
+        };
+
+        var listenProgressTimeEvents = function ()
+        {
+            var saved_time = 0;
```
votemike: Should vars be snake_case or camelCase?

assets/js/smp/smpFactory.js
```diff
@@ -0,0 +1,65 @@
...
+        var defaultOptions = {
+            'uasConfig': {}
+        };
+
+        this.runSmp = function (factoryOptions, smpOptions) {
+            var requiredItems = ['smp/smp'];
+            factoryOptions = mergeOptions(factoryOptions);
+
+            if (factoryOptions.uasConfig) {
+                requiredItems.push(null); // @todo
```
votemike: Do you need to remove this todo?

templates/base_ds2013.html.twig
```diff
@@ -43,7 +43,10 @@
                     'picturefill': "{{ asset_js('js/picturefill.js') }}",
                     'popup': "{{ asset_js('js/popup.js') }}",
                     'snippets': "{{ asset_js("js/snippets.js") }}",
-                    'bump-3': "//emp.bbci.co.uk/emp/bump-3/bump-3"
+                    'bump-3': "//emp.bbci.co.uk/emp/bump-3/bump-3",
+                    'smp/smpFactory': "{{ asset_js("js/smp/smpFactory.js") }}",
```
votemike: Will this load all of these on every page?

assets/js/smp/cookieStore.js
```diff
@@ -0,0 +1,67 @@
...
+                var expires = new Date(new Date().getTime() + COOKIE_LIFETIME);
+
+                window.bbccookies.set(document.cookie = [
+                    encodeURIComponent(name) + '=' + window.escape(value),
+                    'path=/programmes',
+                    expires ? 'expires=' + expires.toUTCString() : '',
+                    'domain=.bbc.co.uk'
+                ].join('; '));
+            }
+            return false;
```
votemike: Is this return value ever used?

assets/js/smp/cookieStore.js
```diff
@@ -0,0 +1,67 @@
...
+                return {};
+            }
+            var i = 0,
+                cookies = window.bbccookies.get().split(';'),
+                cookieValue,
+                currentCookie,
+                cookieName,
+                parsedCookies = {};
+
+            for (; i < cookies.length; i++) {
```
votemike: Should there be something before the first semi-colon?

assets/js/smp/playerRepository.js
```diff
@@ -0,0 +1,95 @@
...
+         *
+         * [
+         *      {pid: ..., time: ...},
+         *      {pid: ..., time: ...},
+         *      ....
+         * ]
+         */
+        var cache = [];
+
+        this.findByVolume = function () {
```
votemike: `findByVolume`? or `findVolume`

assets/js/smp/playerRepository.js
```diff
@@ -0,0 +1,95 @@
...
+                }
+            }
+
+            return 0;
+        };
+
+        this.addVolume = function (volumeValue) {
+            dataStore.persist(ID_USER_PREFS, {
+                'volume': volumeValue,
+                'muted': volumeValue === 0 ? true : false,
```
votemike: `'muted': volumeValue === 0,`?

assets/js/smp/playerRepository.js
```diff
@@ -0,0 +1,95 @@
...
+                'subtitles': subtitles
+            });
+        };
+
+        this.updateTimeResume = function (timestamp) {
+            timestamp = parseInt(timestamp, 10);
+            if (timestamp === 0) {
```
votemike: Why should 0 not be stored?

assets/js/smp/playerRepository.js
```diff
@@ -0,0 +1,95 @@
...
+
+
+        // Privates
+
+        var replaceTimeResumeForProgramme = function (newTimeResume) {
+            // remove all timeResume from pid
+            for (var i = 0, len = cache.length; i < len; i++) {
+                if (cache[i].pid === programmePid) {
+                    cache.splice(i, 1);
+                    break;
```
votemike: I don't like `break;`s, can this be refactored?

templates/base_ds2013.html.twig
tatitati: This is party of the requirejs.CONFIG. Is not requiring any module actually. 
You can see that all this modules configured in require are later required only on clip page:
https://github.com/bbc/programmes-frontend/pull/657/files#diff-18287153ed931668c784bada6c65961aR16

programmes-developers:

src/DsShared/Helpers/SmpPlaylistHelper.php
```diff
@@ -66,6 +68,61 @@ public function getPlaylist(
...
 
+    /**
+     * This method returns the SMP config we actually use
+     *
+     * @param ProgrammeItem $programmeItem
+     * @param Version $version
+     * @return array
+     */
+    public function getSmpPlaylist(
+        ProgrammeItem $programmeItem,
```
votemike: Do you need to pass the `ProgrammeItem` through? Can't you just pass the version and then call `->getProgrammeItem`?

tatitati: No, but you are right, indeed is something we might consider to improve.

It happens that the relations of Version->ProgrammesItems is unidirectional, that is fine and we have our version referencing a programmeItem. However we are not fetching the parent of that programmeITem contained in this version, so when getting the list playlist it explodes because the parent is necessary to build some programme Titles. 

https://github.com/bbc/programmes-frontend/blob/develop/src/DsShared/Helpers/SmpPlaylistHelper.php#L79

Instead we have this parent in the programmeItem we are passing apart, cause of that we use it. The question in here is if we are not fetching this parent programme for performance reasons or is just something that should be changed.

votemike: Surely if you've already got the info, the query has been cached?

BjornTwachtmann: It would be better to rely on version->getProgrammeItem. Not all the queries for versions hydrate all the data on the parent programme, so this is easier. I can't be arsed to fix it here, but it might be worth a ticket to tidy up

assets/js/smp/cookieStore.js
```diff
@@ -0,0 +1,65 @@
...
+    var CookieStore = function() {
+
+        var COOKIE_LIFETIME = 1000 * 60 * 60 * 24 * 90; // 90 days
+
+        this.findByName = function (name) {
+            var all = findAll();
+            var cookie = all[name];
+
+            if (cookie) {
+                return JSON.parse(cookie)
```
BjornTwachtmann: Why JSON.parse?

assets/js/smp/cookieStore.js
```diff
@@ -0,0 +1,65 @@
...
+                if (typeof value === 'object') {
+                    value = JSON.stringify(value);
+                }
+
+                var expires = new Date(new Date().getTime() + COOKIE_LIFETIME);
+
+                window.bbccookies.set(document.cookie = [
+                    encodeURIComponent(name) + '=' + window.escape(value),
+                    'path=/programmes',
+                    expires ? 'expires=' + expires.toUTCString() : '',
```
BjornTwachtmann: Why hardcode the expiry time?

assets/js/smp/playerRepository.js
```diff
@@ -0,0 +1,88 @@
+define(['jquery-1.9','smp/cookieStore'], function ($, DataStore) {
+
+    var PlayerRepository = function (programmePid) {
+        var dataStore = new DataStore();
+        var programmePid = programmePid;
+
+        var KEY_USER_PREFS = 'ckps_progs_playbackCookie';
```
BjornTwachtmann: COOKIE_USER_PREFS or something? Key is ambiguous

assets/js/smp/playerRepository.js
```diff
@@ -0,0 +1,88 @@
+define(['jquery-1.9','smp/cookieStore'], function ($, DataStore) {
+
+    var PlayerRepository = function (programmePid) {
```
BjornTwachtmann: SmpStatePersistance, PlayerStateStorage, something like that? Repository is one of those programming words that can mean literally anything.

assets/js/smp/cookieStore.js
```diff
@@ -0,0 +1,65 @@
...
+            }
+
+            return null;
+        };
+
+        this.persist = function (name, value) {
+            if (window.bbccookies !== undefined && window.bbccookies.cookiesEnabled() && bbccookies.isAllowed(name)) {
+                if (typeof value === 'object') {
+                    value = JSON.stringify(value);
+                }
```
BjornTwachtmann: If we're only accepting objects, document that, and make the caller aware. Otherwise this code will fail on on JS object input.

assets/js/smp/playerRepository.js
```diff
@@ -0,0 +1,88 @@
...
+        var programmePid = programmePid;
+
+        var KEY_USER_PREFS = 'ckps_progs_playbackCookie';
+        var KEY_RESUME = 'ckps_progs_player_resume';
+
+        var volume = 0.6;
+        var subtitles = false;
+
+        /*
+         * @example:
```
BjornTwachtmann: Example of what

assets/js/smp/playerRepository.js
```diff
@@ -0,0 +1,88 @@
...
+
+
+        // Privates
+
+        var replaceTimeResumeForProgramme = function (newTimeResume) {
+            removeProgrammeFromCache();
+
+            // add a new time resume for the pid
+            cache.push({pid: programmePid, time: newTimeResume});
+            cache = cache.slice(-10);
```
BjornTwachtmann: A comment would be good

templates/base_ds2013.html.twig
BjornTwachtmann: Might be worth investigating how many HTTP requests are actually made for this stuff.

tatitati: 3 async JS requests:
http://sandbox.bbc.co.uk/assets/js/smp/smpFactory.js
http://sandbox.bbc.co.uk/assets/js/smp/smp.js
http://sandbox.bbc.co.uk/assets/js/smp/recommendationsBump.js

assets/js/smp/smp.js
```diff
@@ -0,0 +1,180 @@
+/**
+ * SMP module is responsible to start the smp player. Starting this player means:
+ *      - Load user-player-preferences from cookie
+ *      - resumeHandler to an specific moment in the video if specified in the URL or in a cooke
+ *      - Display button to allow (or not) to embed the video
+ *      - Avoid autoplay if specified
+ */
+define(['jquery-1.9', 'smp/smpStatePersistance', 'bump-3'], function ($, PlayerRepository, Bump) {
```
BjornTwachtmann: Can you rename PlayerRepository here too please :)

PR #658, opened 2018-07-17

Tasiobg:
- Add DocBlock with @deprecated comment to getHttpApiClient
- Updated FavouritesButton API to use Multi client

This PR is only for FavouritesButton, next work will be done in a separate PR


src/ExternalApi/Client/HttpApiClientFactory.php
```diff
@@ -29,6 +29,20 @@ public function __construct(
         $this->logger = $logger;
     }
 
+    /**
+     * @deprecated
+     * To be replaces by getHttpApiMultiClient
```
votemike: replaces => replaced

programmes-developers:

PR #661, opened 2018-07-18
[PROGRAMMES-6517] - fixed difference V2/V3 of titles-sizes in footer

tatitati:
## What
You can see that titles in footer are displaying different in V2/V3 for episode page.

example episode URL:
V3: http://sandbox.bbc.co.uk/programmes/b0bb3tmv
V2: http://pal.sandbox.dev.bbc.co.uk/programmes/b0bb3tmv

## Why
With the introduction of accesisiblity in V3 we started to be more strict with the order of headers H1, H2, H3, etc. This produced some differences between V2 and V3. In summary we have:

V3-footer has a hidden title for accessibility. It informs to the user about a section with related content:
```
<div class="footer">
       <h2 class="visually-hidden">Related Content</h2>
       <div class="content-footer">  
            <h3>You may also like</h3>
            <h3>Similar programmes</h3>
             ....
</div>
```

However in V2 we don't have accessibility so that hidden title doesn't exist, instead we have:
```
<div class="footer">
       <div class="content-footer">  
            <h2>You may also like</h2>
            <h2>Similar programmes</h2>
             ....
</div>
```


programmes-developers:

PR #662, opened 2018-07-18

Tasiobg:
- Updated Recipes to use Multi client
- Removed "use" for a non existing "HttpApiService" class

This is the last API to update, in the next PR I'll remove HttpApiClient


programmes-developers:

PR #664, opened 2018-07-20
[PROGRAMMES-6478][new] - Basic information of profiles in profile page

tatitati:
We want to consume XML profile information from isite and display this info into our profile pages.

This PR also include a fix for a bug in our system not falling back to a default branding_id. This fix feature(fix) was required because in case that our xml-profile doesn't include any branding_id (is an optional field) we want to use this default branding_id.

### URL example: Group profile:
Form on Isite: https://production.live.bbc.co.uk/isite2-xforms/fr/progs-radio4and4extra/programmes-profile/edit/a8ed3fbc-db98-3f8e-b974-3a6348083c0f

On API Isite (content consumed by V3): https://api.live.bbc.co.uk/isite2-content-reader/content?contentId=a8ed3fbc-db98-3f8e-b974-3a6348083c0f&depth=1

Page on V3: http://sandbox.bbc.co.uk/programmes/profiles/3YDBGmJwZTYtGTk2PCCbsXw/womans-hour-presenters?__flush_cache=1

### URL example: Individual profile:
Form on iSite: https://production.live.bbc.co.uk/isite2-xforms/fr/progs-radio4and4extra/programmes-profile/edit/3db57c6a-44f0-3153-ba52-2c1ea9ed9416

On API Isite (content consumed by V3): https://api.live.bbc.co.uk/isite2-content-reader/content?contentId=3db57c6a-44f0-3153-ba52-2c1ea9ed9416&depth=1

Page on V3: http://sandbox.bbc.co.uk/programmes/profiles/1rtmf0y9DWkzm3Pwb0cHP2g/jane-garvey?__flush_cache=1

## Ref: construction of the url in V2:
https://github.com/bbc/programmes/blob/develop/webapp/php/sharedlib/BBC/Programmes/IsiteContent.php#L160


programmes-developers:

config/services_test.yaml
```diff
@@ -11,6 +11,7 @@ parameters:
     app.recipe.service.class: App\ExternalApi\Recipes\Service\RecipesStubService
     app.favourites.service.class: App\ExternalApi\FavouritesButton\Service\FavouritesButtonStubService
     app.tupac.service.class: App\ExternalApi\Tupac\Service\TupacStubService
+    app.isite.service.class: App\ExternalApi\Isite\Service\IsiteServiceFake
```
votemike: The rest are called XXXXStubService. Is there a reason this is XXXXServiceFake?

src/Controller/Profiles/ShowController.php
```diff
@@ -5,22 +5,67 @@
...
             return $this->cachedRedirectToRoute('programme_profile', $params, 301);
         }
 
-        return $this->renderWithChrome('profiles/show.html.twig', []);
+        $guid = $isiteKeyHelper->convertKeyToGuid($key);
+
+        /** @var IsiteResult $isiteResult */
+        $isiteResult = $isiteService->findByContentId($guid)->wait(true);
+        if (!$isiteResult) {
+            throw $this->createNotFoundException('None results found for guid');
```
votemike: When does this case happen?

src/Controller/Profiles/ShowController.php
```diff
@@ -5,22 +5,67 @@
...
+        }
+
+        /** @var Profile $profile */
+        $profiles = $isiteResult->getDomainModels();
+        if (count($profiles) == 0) {
+            throw $this->createNotFoundException('Non profiles found');
```
votemike: I think you probably need a different view for this case: E.G. http://www.bbc.co.uk/programmes/profiles/1rtmf0y9DWkzm3Pwb0cHP2e

src/Controller/Profiles/ShowController.php
```diff
@@ -5,22 +5,67 @@
...
+
+        $this->setContextAndPreloadBranding($programme);
+
+        if ($profile->getType() === Profile::INDIVIDUAL) {
+            return $this->renderWithChrome('profiles/individual.html.twig', ['profile' => $profile]);
+        }
+
+        /** @var string|null $projectSpaceOption */
+        $projectSpaceOption = $programme->getOption('project_space');
+        if (null === $projectSpaceOption) {
```
votemike: Does this ever happen?

src/ExternalApi/Isite/GuidQuery.php
```diff
@@ -0,0 +1,43 @@
+<?php
+namespace App\ExternalApi\Isite;
+
+class GuidQuery implements QueryInterface
+{
+    /** @var (int|string)[] */
+    private $parameters = [];
```
votemike: Is there a reason that this is different to https://github.com/bbc/programmes-frontend/blob/develop/src/ExternalApi/Isite/SearchQuery.php ? If you want to use $parameter, is it possible to change SearchQuery to also use $parameters?

src/ExternalApi/Isite/GuidQuery.php
```diff
@@ -0,0 +1,43 @@
...
+
+    public function setProject(string $project): self
+    {
+        $this->parameters['project'] = 'blogs-' . $project;
```
votemike: blogs doesn't sound correct....

src/ExternalApi/Isite/Service/IsiteService.php
```diff
@@ -78,6 +80,37 @@ public function getChildrenOfProfiles(
         return $client->makeCachedPromise();
     }
 
+    public function findByContentId(string $guid, bool $preview = false): PromiseInterface
```
votemike: Other methods start with "get", should this method also start with "get"?

src/ExternalApi/Isite/Service/IsiteService.php
```diff
@@ -117,6 +150,11 @@ public function getProfilesByProgramme(
         return $client->makeCachedPromise();
     }
 
+    public function parseOneProfile(array $responses): IsiteResult
```
votemike: This is the same as `parseProfilesByProgrammeResponses()` could they both be refactored in to one method?

templates/profiles/group.html.twig
```diff
@@ -0,0 +1,27 @@
+{% extends 'base_ds2013.html.twig' %}
+
+{% block title %}{{ meta_context.titlePrefix() }} - {{ tr('profiles') }}{% endblock %}
+{% block page_classes %}text-base prog-layout{% endblock %}
+
+{% block body %}
+    <div class="programmes-page profile--group">
+        <div class="grid-wrapper">
+            <div class="grid">
+                    <div class="br-box-page">
```
votemike: This line is indented too far

tests/ExternalApi/Isite/Domain/ProfileTest.php
```diff
@@ -3,26 +3,28 @@
...
-
-        $profile = new Profile('Title-of an    article!@£$%^&*()', 'anything', 'anything', 'anything');
-        $this->assertSame('title-of-an-article', $profile->getSlug());
-
-        $profile = new Profile('A title~with "quotes" that should/ strip', 'anything', 'anything', 'anything');
-        $this->assertSame('a-title-with-quotes-that-should-strip', $profile->getSlug());
-
-        $profile = new Profile('A title~with apostrophe\'s that should/ strip', 'anything', 'anything', 'anything');
-        $this->assertSame('a-title-with-apostrophes-that-should-strip', $profile->getSlug());
+        $profile = ProfileBuilder::any()->with(['title' => $withTitle])->build();
```
votemike: Does a Profile need a builder?

tests/ExternalApi/Isite/Domain/ProfileTest.php
```diff
@@ -3,26 +3,28 @@
...
+        $this->assertSame($expectSlug, $profile->getSlug());
+    }
 
-        $profile = new Profile('A cööl titlé wîth accènts', 'anything', 'anything', 'anything');
-        $this->assertSame('a-cool-title-with-accents', $profile->getSlug());
+    public function titlesAndSlugsProvider()
+    {
+        return [
+            ['title', 'title'],
```
votemike: Should these array's have keys?

src/Builders/IsiteResultBuilder.php
```diff
@@ -0,0 +1,32 @@
...
+
+class IsiteResultBuilder extends AbstractBuilder
+{
+    protected function __construct()
+    {
+        $faker = Factory::create();
+        $this->classTarget = IsiteResult::class;
+
+        $this->blueprintConstructorTarget = [
+            'page' => $faker->numberBetween(0, 5),
```
votemike: Should all the parameter's here make sense? For example, you can't have page 0 with 200 results and 3 domain models...

config/services_test.yaml
tatitati: In testing there are 4 naming conventions:

**stub**: hardcoded results with no logic at all
**fake**: It has logic, it can replace the original implementation, however is not suitable for production:
**mock**: just used for expectations
**dummy**: just to pass through variables

src/Controller/Profiles/ShowController.php
votemike: Ahh, I assumed you'd just get an empty array of domain models. Cool. In that case, you probably want a new view to say the profile wasn't found. E.G. http://www.bbc.co.uk/programmes/profiles/3YDBGmJwrerTYtGTk2PCCbsXw/womans-hour-presenters?__flush_cache=1

tatitati: We don't need to create a new view for each 404 we throw. That is not practical at all. Symsony allow to create error pages, and we are already using them. As an example of this working you can compare these two urls:

Brand page in Sandbox: http://sandbox.bbc.co.uk/programmes/oiuoiuh
Same Brand page in Live: http://www.bbc.co.uk/programmes/oiuoiuh

votemike: I should learn our system better. I assumed it would be a view saying "Profile not found"

src/Controller/Profiles/ShowController.php
tatitati: Yes. Keep in mind that project_space is inherited from the programmes hierarchy. So taking this programme as example: http://bbc.co.uk/programmes/p0332xt3

You can see in the source code that this programme belongs to the project space:
```<!-- iSite2 Project (Tenant space): progs-radio4and4extra -->```

However, this programme has not a project_space option in our database.

As a summary we have:

| amount of programmes without project_space option | type |
| --- | --- |
| 2046 | brand |
| 74 | clip |
| 1306 | collection |
| 1574 | episode |
| 10 | franchise |
| 23 | gallery |
| 44 | season |
| 1257 | series |

Comment (votemike) [reply src/Controller/Profiles/ShowController.php]: So this will never have profiles?

File: tests/ExternalApi/Isite/Domain/ProfileTest.php [hunk 13]
```diff
@@ -3,26 +3,28 @@
...
+        $profile = ProfileBuilder::any()->with(['title' => $withTitle])->build();
+        $this->assertSame($expectSlug, $profile->getSlug());
+    }
 
-        $profile = new Profile('A cööl titlé wîth accènts', 'anything', 'anything', 'anything');
-        $this->assertSame('a-cool-title-with-accents', $profile->getSlug());
+    public function titlesAndSlugsProvider()
+    {
+        return [
+            'Case1:' => ['title', 'title'],
```
votemike: 'Case1', 'Case2' etc... aren't very descriptive...
How about:
'Case1' => 'alpha'
'Case2' => 'special-chars'
'Case3' => 'quotes'
'Case4' => 'apostrophes'
'Case5' => 'accents'
The you'll get the more descriptive message:
`Tests\App\ExternalApi\Isite\Domain\ProfileTest::testSlugify with data set "accents:"` rather than:
`Tests\App\ExternalApi\Isite\Domain\ProfileTest::testSlugify with data set "Case5:"`

src/ExternalApi/Isite/GuidQuery.php
tatitati: Both request need a different query format. 

For profiles we have this kind of simple-flat url:
```...../content?contentId=AAAAAAAAAAAAA&allowNonLive=true&preview=true&depth=1```

However, the url built by SearchQuery is more complex as it uses a series of nested json items, all of them nested in a ```q``` parameter. Also the basepath is different ("/search"):

```
/search?q={
  "depth": 0,
  "namespaces": {
    "profile": "https://production.bbc.co.uk/isite2/project/progs-eastenders/programmes-profile"
  },
  "page": "1",
  "pageSize": "48",
  "project": "progs-eastenders",
  "query": {
    "and": [
      [
        "profile:parent_pid",
        "=",
....
```

I reject the idea of modifying SearchQuery to use $parameters or GuidQuery to use $q, as they are simply doing different things, is not needed to merge internal functionality of these two Query builders, they simply build different Queries.

Comment (votemike) [reply src/ExternalApi/Isite/GuidQuery.php]: Ahh, fair enough then.

File: src/Builders/IsiteResultBuilder.php [hunk 14]
```diff
@@ -0,0 +1,36 @@
...
+            'pageSize' => 1,
+            'total' => count($profiles),
+            'domainModels' => $profiles,
+        ];
+    }
+
+    public static function anyWithProfiles(array $profiles)
+    {
+        return self::any()->with([
+            'domainModels' => $profiles,
```
votemike: What about the correct pageSize and totals?
I'm not sure this method is needed, you could probably just construct this in the test or styleguide when you need it.

src/Builders/IsiteResultBuilder.php
```diff
@@ -0,0 +1,36 @@
...
+
+class IsiteResultBuilder extends AbstractBuilder
+{
+    protected function __construct()
+    {
+        $faker = Factory::create();
+        $this->classTarget = IsiteResult::class;
+
+        $profiles = $faker->randomElements([
+            ProfileBuilder::any()->build(),
```
votemike: This will make 5 Profiles even if you only make 1 won't it? That seems like a bit of a waste.

votemike: And if you supply your own profiles, this will make 5 Profiles of which none will be used.

config/services_test.yaml
votemike: `IsiteFakeService` then maybe?

tests/Controller/Profiles/ShowControllerTest.php
```diff
@@ -0,0 +1,84 @@
...
+    public function testTemplateIndividualIsDisplayed()
+    {
+        $this->userRequest('/programmes/profiles/' . self::KEY_INDIVIDUAL . '/' . self::SLUG);
+        $this->thenDisplaySingleProfileTemplate();
+    }
+
+    /**
+     * Helpers
+     */
+    private function userRequest($url)
```
votemike: What's `userRequest`?

tests/Controller/Profiles/ShowControllerTest.php
```diff
@@ -0,0 +1,84 @@
...
+class ShowControllerTest extends BaseWebTestCase
+{
+    const KEY_NOT_FOUND = '5YDBGmJwZTYtGTk2PCCbsXw';
+    const KEY_NO_PROFILES = '6YDBGmJwZTYtGTk2PCCbsXw';
+    const KEY_GROUP = '3YDBGmJwZTYtGTk2PCCbsXw';
+    const KEY_INDIVIDUAL = '4YDBGmJwZTYtGTk2PCCbsXw';
+
+    const SLUG = 'men-minutes-presenters';
+
+    private $crawler;
```
votemike: Why is this a property?

src/Builders/IsiteResultBuilder.php
```diff
@@ -0,0 +1,36 @@
...
+            ProfileBuilder::any()->build(),
+            ProfileBuilder::any()->build(),
+        ], $count = $faker->numberBetween(0, 5));
+
+        $this->blueprintConstructorTarget = [
+            'page' => 1,
+            'pageSize' => 1,
+            'total' => count($profiles),
+            'domainModels' => $profiles,
```
votemike: Do you think this would be better if by default it returned 0 results? Then you wouldn't have to do weird things like `IsiteResultBuilder::anyWithProfiles([])->build();`.

tests/Controller/Profiles/ShowControllerTest.php
tatitati: Pick one:
```$this->userRequests('www.google.com');```
```$this->userNavigatesTo('www.google.com');```
```$this->userLoads('www.google.com');```
```$this->userRequires('www.google.com');```
```$this->userAskFor('www.google.com');```

File: tests/Controller/Profiles/ShowControllerTest.php (hunk 17 above)
Comment (tatitati) [reply tests/Controller/Profiles/ShowControllerTest.php]: to make more readable tests. 

I found that I like more to do
```php
$this->userNavigateTo('www.google.com');

$this->thenUserSeeASearchBar()
$this->thenUserSeeALinkToGmail()
```

Than
```php
$html = $this->userNavigateTo('www.google.com');

$this->thenUserSeeASearchBar($html);
$this->thenUserSeeALinkToGmail($html);
```

src/Builders/IsiteResultBuilder.php
tatitati: Yeah, I agree

tests/Controller/Profiles/ShowControllerTest.php
votemike: It was more of a question of what is the use of it?
The method is called "userRequest", but it also checks that the status code is a 200. But it's perfectly possible for a user to request a page that 404s or 500s. So I thought the name was a bit odd.
If you then remove that line, you're left with a single line which seems quite self explanatory by itself. I was just wondering what the benefit was by putting this in a method.

Tasiobg: Those helpers functions are too short to bring a big value, some of them seems to be used only once but I don't really mind, doesn't look bad either

src/Controller/Profiles/ShowController.php
```diff
@@ -5,22 +5,64 @@
...
 use App\Controller\BaseController;
 use App\Controller\Helpers\IsiteKeyHelper;
+use App\ExternalApi\Isite\Domain\Profile;
+use App\ExternalApi\Isite\IsiteResult;
+use App\ExternalApi\Isite\Service\IsiteService;
+use BBC\ProgrammesPagesService\Domain\ValueObject\Pid;
+use BBC\ProgrammesPagesService\Service\CoreEntitiesService;
 use Symfony\Component\HttpFoundation\Request;
 
 class ShowController extends BaseController
```
BjornTwachtmann: ProfileController?

src/Controller/Profiles/ShowController.php
```diff
@@ -5,22 +5,64 @@
...
 {
-    public function __invoke(string $key, string $slug, Request $request, IsiteKeyHelper $isiteKeyHelper)
+    public function __invoke(string $key, string $slug, Request $request, IsiteService $isiteService, IsiteKeyHelper $isiteKeyHelper, CoreEntitiesService $coreEntitiesService)
     {
-        if (strpos($key, '-') !== false) { // if contains dashes it is a guid
-            $key = $isiteKeyHelper->convertGuidToKey($key);
+        if ($this->isKeyAGuid($key)) {
```
BjornTwachtmann: Add this functionality to iSiteKeyHelper and use it in there too

src/Controller/Profiles/ShowController.php
```diff
@@ -5,22 +5,64 @@
...
             }
+
             return $this->cachedRedirectToRoute('programme_profile', $params, 301);
         }
 
-        return $this->renderWithChrome('profiles/show.html.twig', []);
+        $guid = $isiteKeyHelper->convertKeyToGuid($key);
+
+        /** @var IsiteResult $isiteResult */
+        $isiteResult = $isiteService->getByContentId($guid)->wait(true);
```
BjornTwachtmann: Add the preview parameter from the query string here.

src/Controller/Profiles/ShowController.php
```diff
@@ -5,22 +5,64 @@
...
+        }
+
+        /** @var Profile $profile */
+        $profiles = $isiteResult->getDomainModels();
+        if (count($profiles) == 0) {
+            throw $this->createNotFoundException('Non profiles found');
+        }
+
+        $profile = $profiles[0];
+        $programme = $coreEntitiesService->findByPidFull(new Pid($profile->getParentPid()));
```
BjornTwachtmann: Check, but I don't think parent pid is actually required.

src/Controller/Profiles/ShowController.php
```diff
@@ -5,22 +5,64 @@
...
+
+        $this->setContextAndPreloadBranding($programme);
+
+        if ($profile->getType() === Profile::INDIVIDUAL) {
```
BjornTwachtmann: Maybe have an isIndividual method?

src/Controller/Profiles/ShowController.php
```diff
@@ -5,22 +5,57 @@
...
+        /** @var IsiteResult $isiteResult */
+        $isiteResult = $isiteService->getByContentId($guid)->wait(true);
+        /** @var Profile $profile */
+        $profiles = $isiteResult->getDomainModels();
+        if (!$profiles) {
+            throw $this->createNotFoundException('No profiles found for guid');
+        }
+
+        $profile = reset($profiles);
+        $programme = $coreEntitiesService->findByPidFull(new Pid($profile->getParentPid()));
```
BjornTwachtmann: Parent pid is optional. If it is specified, do this query. Also check the project space from the programme's options against the project space from the profile, and throw a 404 if they do not match

src/Controller/Profiles/ShowController.php
```diff
@@ -5,22 +5,57 @@
...
+
+        $this->setContextAndPreloadBranding($programme);
+
+        if ($profile->getType() === Profile::INDIVIDUAL) {
+            return $this->renderWithChrome('profiles/individual.html.twig', ['profile' => $profile]);
+        }
+
+        $childPromise = $isiteService->getChildrenOfProfiles($profile->getProjectSpace(), [$profile]);
```
BjornTwachtmann: Are you specifying the right page size and limit?

src/Controller/Profiles/ShowController.php
```diff
@@ -5,22 +5,57 @@
...
+
+        /** @var IsiteResult $isiteResult */
+        $isiteResult = $isiteService->getByContentId($guid)->wait(true);
+        /** @var Profile $profile */
+        $profiles = $isiteResult->getDomainModels();
+        if (!$profiles) {
+            throw $this->createNotFoundException('No profiles found for guid');
+        }
+
+        $profile = reset($profiles);
```
BjornTwachtmann: After this line, before doing anything with the profile, you need to check the "slug"
```if ($slug != $this->context->profile->url_slug) {
            $this->getResponse()->setHeader('cache-control', 'public, max-age=3600');
            return $this->_helper->redirector->setCode(301)->gotoUrlAndExit(Zend_View_Helper_Url::url(array(
                'key' => $this->context->profile->key,
                'slug' => $this->context->profile->url_slug
                ),
                'programme_profile_article'
            ) . (($preview) ? '?preview=true' : ''));
        }```

File: src/Controller/Profiles/ShowController.php (hunk 19 above)
Comment (votemike) [reply src/Controller/Profiles/ShowController.php]: There is another controller under the Profile directory for Listing Profiles. Would it not be confusing if one was the ProfileController and one was the ListController?

File: bin/.6.5.md5 [hunk 27]
```diff
@@ -0,0 +1,2 @@
+91a9a156ac0314b66fea26b5918f757c
```
votemike: I don't think we need this file....

src/Controller/Profiles/ShowController.php
tatitati: Reference:
https://github.com/bbc/programmes/blob/develop/webapp/conf/programmes.ini#L49

src/Controller/Profiles/ShowController.php
```diff
@@ -5,22 +5,79 @@
...
+        if ('' !== $profile->getParentPid()) {
+            $programmeOwner = $coreEntitiesService->findByPidFull(new Pid($profile->getParentPid()));
+
+            if ($profile->getProjectSpace() !== $programmeOwner->getOption('project_space')) {
+                throw $this->createNotFoundException('Project space Profile-Programme not matching');
+            }
+
+            $this->setContextAndPreloadBranding($programmeOwner);
+        } else {
+            // @TODO: Set a generic branding
```
tatitati: @BjornTwachtmann how can I set a generic branding?, I tried to find something in the code, but all of them seems to put an specific branding-programme

src/Controller/Helpers/IsiteKeyHelper.php
```diff
@@ -26,6 +26,11 @@ public function convertKeyToGuid(string $key): string
             substr($characters, 20, 12);
     }
 
+    public function isKeyAGuid($key): bool
```
votemike: string typehint?

src/Controller/Profiles/ShowController.php
```diff
@@ -5,22 +5,79 @@
...
+            $params = [
+                'key' => $profile->getKey(),
+                'slug' => $profile->getSlug(),
+            ];
+
+            if ($preview) {
                 $params['preview'] = 'true';
             }
+
             return $this->cachedRedirectToRoute('programme_profile', $params, 301);
```
votemike: This is very similar to the redirect above. Is it possible to combine or refactor them?

src/Controller/Profiles/ShowController.php
votemike: Is there a ticket to do this? Or are you going to do it as part of this PR?

src/Controller/Profiles/IndexController.php
```diff
@@ -36,7 +36,7 @@ public function __invoke(CoreEntity $coreEntity, IsiteService $isiteService)
                     $groupProfiles[] = $profile;
                 }
             }
-            $childPromise = $isiteService->getChildrenOfProfiles($project, $groupProfiles);
+            $childPromise = $isiteService->setChildProfilesOn($project, $groupProfiles);
```
votemike: Tiny thing, it might be nicer to have the profiles as the first param so that it reads `setChildProfilesOn($groupProfiles, ........`

src/Controller/Profiles/ShowController.php
```diff
@@ -5,22 +5,79 @@
...
+        if ('' !== $profile->getParentPid()) {
+            $programmeOwner = $coreEntitiesService->findByPidFull(new Pid($profile->getParentPid()));
+
+            if ($profile->getProjectSpace() !== $programmeOwner->getOption('project_space')) {
+                throw $this->createNotFoundException('Project space Profile-Programme not matching');
+            }
+
+            $context = $programmeOwner;
+        }
+        $this->setContext($context);
```
votemike: You you just set this in the `if` statement above?

src/ExternalApi/Isite/Service/IsiteService.php
```diff
@@ -88,7 +125,7 @@ public function getProfilesByProgramme(
         $query = $this->getBaseQuery($project, $page, $limit);
         $query->setQuery([
             'and' => [
-                ['profile:parent_pid', '=', (string) $programme->getPid()],
+                ['profile:parent_pid', '=', (string)$programme->getPid()],
```
votemike: Why no space?

src/ExternalApi/Isite/Service/IsiteService.php
```diff
@@ -121,19 +158,20 @@ public function getProfilesByProgramme(
      * @param Response[] $responses
      * @param Profile[] $profiles
      */
-    public function parseChildrenOfProfilesResponses(array $responses, array $profiles): void
+    public function parseChildrenOfProfilesResponses(array $responses, array $profiles): array
```
votemike: Add the return type to the docblocks?

src/ExternalApi/Isite/Service/IsiteService.php
```diff
@@ -165,4 +203,33 @@ private function getDefaultSort(): array
             ],
         ];
     }
+
+
+    /**
+     * Given a promise from
```
votemike: This seems incomplete...

src/ExternalApi/Isite/Service/IsiteService.php
```diff
@@ -165,4 +203,33 @@ private function getDefaultSort(): array
...
+     * @param array $profiles
```
votemike: `@param Profile[] $profiles`

src/Controller/Profiles/ShowController.php
```diff
@@ -5,22 +5,77 @@
...
+            return $this->renderWithChrome('profiles/individual.html.twig', ['profile' => $profile]);
+        }
+
+        $childPromise = $isiteService->setChildProfilesOn([$profile], $profile->getProjectSpace());
+        $this->resolvePromises([$childPromise]);
+
+        return $this->renderWithChrome('profiles/group.html.twig', ['profile' => $profile]);
+    }
+
+    private function redirectWith($key, $slug, $preview)
```
votemike: Typehints?

src/Builders/ProfileBuilder.php
```diff
@@ -0,0 +1,40 @@
...
+            'fileId' => $faker->word,
+            'type' => $faker->randomElement(['group', 'individual']),
+            'projectSpace' => $faker->word,
+            'parentPid' => new Pid($faker->regexify('[0-9b-df-hj-np-tv-z]{8,15}')),
+            'longSynopsis' => $faker->sentence(5),
+            'brandingId' => $faker->word,
+        ];
+    }
+
+    public static function anyGroup()
```
votemike: I'm not convinced we need methods like this. I think these `with`s could just be added in the tests.

src/Controller/BaseController.php
```diff
@@ -178,8 +178,10 @@ protected function resolvePromises(array $promises): array
             $promises['branding'] = $this->brandingPromise;
         }
         $unwrapped = \GuzzleHttp\Promise\unwrap($promises);
-        $this->branding = $unwrapped['branding'];
-        unset($unwrapped['branding']);
+        if (isset($unwrapped['branding'])) {
+            $this->branding = $unwrapped['branding'];
```
votemike: Does this mean branding may not be set? Is that OK?

src/ExternalApi/Isite/Service/IsiteService.php
```diff
@@ -165,4 +204,31 @@ private function getDefaultSort(): array
...
+     * @param Profile[] $profiles
+     * @param PromiseInterface $promise
+     * @return PromiseInterface
+     */
+    private function chainHydrationPromise(array $profiles, PromiseInterface $promise): PromiseInterface
+    {
+        return $promise->then(
+            function ($responses) use ($profiles) {
+                // Success callback
+                return $this->hydrateProfiles($profiles, $responses);
```
votemike: Do you need to return this? I thought you just need to run `hydrateProfiles()`?

votemike: This will just return void won't it?

src/Controller/BaseController.php
tatitati: Yeah. Is an optional field in the profile. When branding id is not set our system fallback to a default one:
https://github.com/bbc/programmes-frontend/blob/develop/src/Controller/BaseController.php#L33

PR #666, opened 2018-07-20
PROGRAMMES-6465 Update all HTTP APIs 

Tasiobg:
Update all HTTP APIs to use the Multi client and retire the single URL HttpApiClient

This is the last PR for this branch, all APIs has been migrated to HttpApiMultiClient and HttpApiClient has been removed. Understanding this PR could be a bit challenging so feel free to ask me and I can walk you through

Changes:
- Updated Ada API to use HttpApiMultiClient
- Fix some return type declarations
- Removed HttpApiClient so now we only have one client for HTTP request
- Updated guzzle to enable verifications, this fix an issue we have with TUPAC and provides more security like preventing preventing man in the middle attacks
- Added more unitest to HttpApiMultiClientTest


src/ExternalApi/Client/HttpApiMultiClient.php
```diff
@@ -100,14 +144,46 @@ private function handleResponses(array $responses)
             array_unshift($args, $responses);
             $result = call_user_func_array($this->parseResponseCallable, $args);
         } catch (MultiParseException $e) {
-            $url = $this->requestUrls[$e->getResponseKey()] ?? $this->requestUrl;
+            $url = $this->requestUrls[$e->getResponseKey()] ?? reset($this->requestUrls);
```
votemike: What if the error wasn't with the first URL?

src/ExternalApi/Client/HttpApiMultiClient.php
```diff
@@ -100,14 +144,46 @@ private function handleResponses(array $responses)
...
         } catch (ParseException $e) {
-            $this->logger->error('Error parsing feed for "{0}". Error was: {1}', [$this->requestUrl, $e->getMessage()]);
+            $this->logger->error('Error parsing feed for one of this URLs: "{0}". Error was: {1}', [implode(',', $this->requestUrls), $e->getMessage()]);
             return $this->nullResult;
         }
         $this->cache->setItem($this->cacheItem, $result, $this->standardTTL);
         return $result;
     }
+
+    private function handleGuzzleException(GuzzleException $e)
```
votemike: Return typehints?

src/ExternalApi/Client/HttpApiMultiClient.php
```diff
@@ -100,14 +144,46 @@ private function handleResponses(array $responses)
...
+            }
+            $this->logger->error("HTTP Error status $responseCode for one of this URLs $urls : " . $e->getMessage());
+        } elseif ($this->notFoundTTL !== CacheInterface::NONE) {
+            // 404s get cached for a shorter time
+            $this->cache->setItem($this->cacheItem, $this->nullResult, $this->notFoundTTL);
+        }
+        return $this->nullResult;
+    }
+
+    private function handleAsyncError($reason)
```
votemike: Typehints?

src/ExternalApi/Client/HttpApiMultiClient.php
Tasiobg: `$e->getResponseKey()` should always return the URL but if for any reason it doesn't (bad implementation for example) at least we would be able to log something.
Example of a rare case when this could happen: our code throws MultiParseException with the $key value '0' but the real key is an interger, in this case we are able to log something. We could even log different messages if the array key doesn't exist but the chances of this happenign are low enough to be worried for a more advance solution

src/ExternalApi/Client/HttpApiMultiClient.php
Tasiobg: This should return `$this->nullResult` which is defined when instanciating HttpApiMultiClient so it vary depending on the API

src/ExternalApi/Client/HttpApiMultiClient.php
Tasiobg: Same reason as handleGuzzleException

PR #667, opened 2018-07-24
PROGRAMMES-6524 Updating versions considered alternate

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6524


programmes-developers:

PR #669, opened 2018-07-24
PROGRAMMES-6520 changes to Account Wrapper integration

Tasiobg:
Fix the language string


programmes-developers:

PR #674, opened 2018-07-27
[Extends: PROGRAMMES-6509][Fix] - Added break lines for very long workds on titles

tatitati:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6509

## What: 
Some titles don't fit nicely when the screen is resized to mobile size. Is not a problem of "how much text we display" as we already use truncation for titles for both render systems as you can see here:

Truncation In Ds2013: https://github.com/bbc/programmes-frontend/blob/develop/src/Ds2013/Presenters/Domain/CoreEntity/SharedSubPresenters/core_entity_title.html.twig#L23

Truncation in DsAmen: https://github.com/bbc/programmes-frontend/blob/develop/src/DsAmen/Presenters/Domain/CoreEntity/Base/SubPresenter/BaseTitlePresenter.php#L89

Instead this problem arise when the system is not able to split a simple word when the space is too small into a new line.

## URL examples to test:
DsAmen Brand Page: http://sandbox.bbc.co.uk/programmes/b006qj8q?__flush_cache=1
Ds2013 Episode Page: http://sandbox.bbc.co.uk/programmes/b09kkt1m?__flush_cache=1

## About the fix:
Support of "hyphens" in browsers: https://caniuse.com/#search=hyphens

Not necessary to add prefixes for "hyphens" as we add them when building assets:
https://github.com/bbc/programmes-frontend/blob/develop/gulpfile.js#L99


programmes-developers:

src/Ds2013/Presenters/Domain/CoreEntity/Programme/_programme.scss
```diff
@@ -60,6 +60,8 @@ $programme-icon-bpw-box-size: gel-typography('bpw', 'delta', 'line-height') + 18
 .programme__title {
     display: block;
     margin-bottom: 0;
+    hyphens: auto;
```
benolot: Do we want to apply this to every programme title?

tatitati: Good question. I decided to put it for every programme title as I tried different ones (typing in the html dom very long words) and it was happening the same problem in all of them. 

But is just my idea.....any other idea "touching less" programmes is welcome

PR #676, opened 2018-07-31
[PROGRAMMES-6351] - Added UAS module to send SMP player activities

tatitati:
## UAS module Specs:
- [x] UAS service only send data to Activity server when the user is logged in (by default), but it includes some parameters to send data when the user is also logged out (configurable in code).
These requests look like this:

| Code response | Method | Type | Protocol | Url | Call |
|---|---|---|---|---|---|
| 202 | POST | plays | https | activity.test.api.bbc.co.uk | xhr |
| 202 | POST | heartbeat | https | heartbeat.activity.test.api.bbc.co.uk | xhr |
| 202 | POST | reads | https | activity.test.api.bbc.co.uk | xhr |

- [x] UAS send different activities: **started, paused, ended, reads, heartbeat**. Example:
```json
{
"action" : "started",
"actionContext" : "urn:bbc:tv:version_offset:p06f8f0f#22",
"resourceDomain" : "tv",
"resourceId" : "p06f8f0b",
"resourceType" : "clip",
} 
```
- [x] "Heartbeat" event is sent every 30 seconds
- [x] "Reads" event is sent when the user visit the page through a recommendation and is loggedin
- [x] UAS only runs in clip page.
- [x] UAS only send data for int/test/stage/live environments.
- [x] Uas stop sending data to Activity server for a few minutes if it receives error responses (A.K.A circuitbreaker in js)

## To test it:
1. Go to an INT v3 clip page like: https://programmes-frontend.int.api.bbc.co.uk/programmes/p06f8f0b?__flush_cache=1
2. Login
3. refresh the page after login. You should see a the next requests similar to the table above.

## Question:
Uas module use an ApiKey when sending activities to the server:
https://github.com/bbc/programmes-frontend/blob/uas-module/src/Ds2013/Presenters/Utilities/SMP/SmpPresenter.php#L132
Not sure if is useful to store it in any secure place as is gonna be dumped later in clear text in the javascript DOM. In V2 this is the case.

<details>
  <summary>Expand information about the implementation of UAS module:</summary>

I tried to find an existing UAS client that we could use. However I was no lucky. As a prove of this you can see some interesting options I found, but all of them were not reliable as they are done in scala, node, etc:
* https://github.com/bbc/rms-uas-client
* https://github.com/bbc/ibl-user-activity
* https://github.com/bbc/ibl-user-activity-client
* https://github.com/bbc/node-uas-client

From my conversations with some teams I got:
- IBL team doesn't use any UAS client, instead they send data to IBL, and IBL transform it into UAS format. However /programmes wouldn't be able to use IBL as it transform into an IPlayer specific way.
- Iplayer team use a backend UAS client only usable by Java or scala systems.

UAS client Confluence page says that we can just require it on the fly, this might be a reliable solution), however I didn't do it as I assume there is a reason to not be required in V2:
https://confluence.dev.bbc.co.uk/display/activity/User+Activity+Client+Library+User+Guide#UserActivityClientLibraryUserGuide-Gettingstarted

I finished choosing the existing V2:  https://github.com/bbc/uas-client-pal
</details>

<details>
  <summary>How I review this?</summary>
  In order to make a review you must know that the only file I could refactor (clean and improve properly) was UasService.js
This file acts as a façade to UAS module and as you can imagine this UasService relies in the UasClient to send activity user.
</details>


programmes-developers:

assets/js/smp/smp.js
```diff
@@ -145,9 +145,28 @@ define(['jquery-1.9', 'smp/smpStatePersistance', 'bump-3'], function ($, SmpStat
         var listenProgressTimeEvents = function ()
         {
             var savedTime = 0;
+            var time = 0;
```
votemike: What's time? The current time? The time the user started listening? The time the media was first broadcast?

assets/js/uas/corsRequest.js
```diff
@@ -0,0 +1,304 @@
...
+                env = '';
+            }
+
+            if (optionsForRequest.userIsNotSignedIn) {
+                baseUrl = 'https://nsi.activity.';
+            } else {
+                baseUrl = 'https://activity.';
+            }
+
+            baseUrl = baseUrl + env + 'api.bbc.' + this.config.domainSuffix + '/my/' + optionsForRequest.activityType;
```
votemike: `baseUrl += env + 'api.bbc.' + this.config.domainSuffix + '/my/' + optionsForRequest.activityType;`?

assets/js/uas/corsRequest.js
```diff
@@ -0,0 +1,304 @@
...
+
+            if (this.config.baseUri) {
+                baseUrl = this.config.baseUri + '/my/' + optionsForRequest.activityType;
```
votemike: Line 54 to 60 were just spent constructing the baseUrl, now you're overwriting it. Is there a way to simplify or refactor this logic?

assets/js/uas/uasService.js
```diff
@@ -0,0 +1,213 @@
...
+                    uasData.metaData = JSON.stringify(self.recommendationData);
+                }
+
+                sendToUas(uasData);
+            }
+        };
+
+        // private
+
+        var self = this;
```
votemike: I like that sometimes we use `self` and sometimes we use `_this`

src/Controller/FindByPid/ClipController.php
```diff
@@ -40,7 +41,8 @@ public function __invoke(
         RelatedLinksService $relatedLinksService,
         SegmentEventsService $segmentEventsService,
         StructuredDataHelper $structuredDataHelper,
-        VersionsService $versionsService
+        VersionsService $versionsService,
+        CosmosInfo $cosmosInfo
```
votemike: All of the other params coming though seem alphabetised, shall we keep it that way?

src/Ds2013/Presenters/Utilities/SMP/SmpPresenter.php
```diff
@@ -128,9 +134,10 @@ public function getFactoryOptions(): array
      */
     private function getUasConfig(): array
     {
+
         return [
-            'apiKey' => '', // @todo
-            'env' => '', // @todo
+            'apiKey' => 'bapd63mcqopnp', // @todo
```
votemike: This isn't a LIVE key is it? As now it's in a public repo.

Tasiobg: Francisco mention there was a key in the javascript PAL library so he decided make it public here as well, in other words, for v2 this key can be seen in the javascript.

programmes-developers:

assets/js/smp/smp.js
```diff
@@ -145,33 +145,56 @@ define(['jquery-1.9', 'smp/smpStatePersistance', 'bump-3'], function ($, SmpStat
         var listenProgressTimeEvents = function ()
         {
             var savedTime = 0;
+            var currentInstantInReproduction = 0;
```
BjornTwachtmann: currentTime?

src/Controller/FindByPid/ClipController.php
```diff
@@ -119,6 +121,7 @@ public function __invoke(
             'downloadableVersion' => $linkedVersions['downloadableVersion'],
             'streamableVersion' => $linkedVersions['streamableVersion'],
             'segmentEvents' => $segmentEvents,
+            'cosmosInfo' => $cosmosInfo,
```
BjornTwachtmann: Avoid passing this into the view. Pass it into the constructor for the presenter factory as it's known at compile time.

templates/base_ds2013.html.twig
```diff
@@ -46,7 +46,12 @@
                     'bump-3': "//emp.bbci.co.uk/emp/bump-3/bump-3",
                     'smp/smpFactory': "{{ asset_js("js/smp/smpFactory.js") }}",
                     'smp/smp': "{{ asset_js("js/smp/smp.js") }}",
-                    'smp/recommendationsBump': "{{ asset_js("js/smp/recommendationsBump.js") }}"
+                    'smp/recommendationsBump': "{{ asset_js("js/smp/recommendationsBump.js") }}",
+                    'uas/uasService': "{{ asset_js("js/uas/uasService.js") }}",
+                    'uasclient': "https://static.bbc.co.uk/uasclient/0.1.42/modules/uasclient",
+                    'idcta/idcta-1': "https://mybbc.files.bbci.co.uk/s/id/account-idcta/1.8.4/modules/idcta-v2/idcta-1",
```
BjornTwachtmann: You shouldn't need to add this.

templates/base_ds2013.html.twig
```diff
@@ -46,7 +46,12 @@
                     'bump-3': "//emp.bbci.co.uk/emp/bump-3/bump-3",
                     'smp/smpFactory': "{{ asset_js("js/smp/smpFactory.js") }}",
                     'smp/smp': "{{ asset_js("js/smp/smp.js") }}",
-                    'smp/recommendationsBump': "{{ asset_js("js/smp/recommendationsBump.js") }}"
+                    'smp/recommendationsBump': "{{ asset_js("js/smp/recommendationsBump.js") }}",
+                    'uas/uasService': "{{ asset_js("js/uas/uasService.js") }}",
```
BjornTwachtmann: change the path for this, make it live under SMP I think

assets/js/uas/uasService.js
```diff
@@ -0,0 +1,213 @@
...
+                600000  // 10 minutes
+            ]
+        }, uasConfig);
+
+        this.init = function() {
+            if (this.options.apiKey && userIsLoggedInAndHasTrackingEnabled()) {
+                var self = this;
+                istats.isReady(function() {
+                    self.recommendationData = getRecommendationDataFromIstats();
+                    require (['uasclient'], function(UasClient) {
```
BjornTwachtmann: Maybe move the require for uasclient to the initial define() block at the top?

assets/js/smp/smp.js
```diff
@@ -41,7 +41,7 @@ define(['jquery-1.9', 'smp/smpStatePersistance', 'bump-3'], function ($, SmpStat
                     volume: null,
                     autoplay: false
                 },
-                rememberResume: true,
+                rememberResume: false,
```
tatitati: This was wrong. By default is false in V2:
https://github.com/bbc/programmes/blob/develop/webapp/static-versioned/js-modules/smp/init.js#L61

assets/js/smp/smpStatePersistance.js
```diff
@@ -54,25 +54,25 @@ define(['jquery-1.9','smp/cookieStore'], function ($, DataStore) {
...
         this.updateTimeResume = function (timestamp) {
-            dataStore.persist(COOKIE_RESUME, replaceTimeResumeForProgramme(timestamp));
+            replaceTimeResumeForProgrammeInCache(timestamp);
+            dataStore.persist(COOKIE_RESUME, cache);
         };
 
         this.removeTimeResume = function () {
-            dataStore.persist(COOKIE_RESUME, removeProgrammeFromCache());
+            removeProgrammeFromCache();
+            dataStore.persist(COOKIE_RESUME, cache);
```
tatitati: Fixed BUG. removeProgrammeFromCache() return null. When the reproduction was ended we were storing a null into the cookie

programmes-developers:

PR #677, opened 2018-08-01
[PROGRAMMES-6500] Profile Index page

benolot:
Ticket: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6500

Okay so I have a couple of questions about how I've done stuff, to keep in mind when reviewing.

The work done in the domain object, Is this okay? Is this too much logic and needs moving elsewhere? I can't imagine where else it should go, hence why it is in there.

With regards to Image height in the domain object (get image URL), is it okay to be decided how it is, or should it be a parameter you pass in?

If no profiles are found for that pid, do we want to return a 404 (as it currently does)?

I have removed the block link from the recipe, into the dedicated _block-link.scss that already existed, and updated it to apply to all block links, as it does in V2. Is this okay with everyone? Can anyone think of cases this might break stuff in? Especially looking for feedback about this point from @tatitati and @votemike.

Example of normal page: https://sandbox.bbc.co.uk/programmes/b006m86d/profiles
Paginated page: https://sandbox.bbc.co.uk/programmes/b006q2x0/profiles?page=2
Page with no profiles: https://sandbox.bbc.co.uk/programmes/b04ww8fq/profiles


programmes-developers:

src/Controller/Profiles/IndexController.php
```diff
@@ -17,12 +18,17 @@ public function __invoke(CoreEntity $coreEntity, IsiteService $isiteService)
 
         /** @var Profile[] $profiles */
         $profiles = [];
+        $paginator = null;
         $parameters = ['coreEntity' => $coreEntity, 'profiles' => $profiles];
```
votemike: $parameters = ['coreEntity' => $coreEntity, 'profiles' => $profiles, 'paginatorPresenter' => null];

src/Controller/Profiles/IndexController.php
```diff
@@ -17,12 +18,17 @@ public function __invoke(CoreEntity $coreEntity, IsiteService $isiteService)
...
         if ($coreEntity instanceof Programme) {
             $parameters['programme'] = $coreEntity; //so the the base 2013 template sets the footer
 
             $profilesResult = $isiteService->getProfilesByProgramme($coreEntity, $this->getPage())->wait();
             $profiles = $profilesResult->getDomainModels();
+
+            if ($profilesResult->getTotal() > 48) {
+                $paginator = new PaginatorPresenter($this->getPage(), 48, $profilesResult->getTotal());
```
votemike: `$parameters['paginatorPresenter'] = new PaginatorPresenter($this->getPage(), 48, $profilesResult->getTotal());`

src/Ds2013/Presenters/Domain/Recipe/_recipe.scss
```diff
@@ -17,29 +17,10 @@
     }
 }
 
-.recipe .block-link__overlay-link {
```
votemike: Is this no longer needed then?

src/ExternalApi/Isite/Domain/Profile.php
```diff
@@ -115,6 +125,11 @@ public function getSlug()
         );
     }
 
+    public function getShortSynopsis()
```
votemike: return typehint?

src/ExternalApi/Isite/Domain/Profile.php
```diff
@@ -135,6 +150,41 @@ public function getBrandingId()
...
+    public function getImage(): string
+    {
+        return $this->image;
+    }
+
+    /**
+     * @param string $pid
+     * @return string
+     */
+    public function getImageSrcsets(string $pid): string
```
votemike: This feels more like it belongs in a presenter.

templates/profiles/index.html.twig
```diff
@@ -6,13 +6,79 @@
...
+                                                            </div>
+                                                        </div>
+                                                        <div class="programme__body">
+                                                            <h3 class="programme__titles delta">
+                                                        <span class="programme__title">
+                                                            <a href="{{ path('programme_profile', {'key': child.getKey(), 'slug': child.getSlug()}) }}" class="br-blocklink__link">{{ child.getTitle() }}</a>
+                                                        </span>
+                                                            </h3>
+                                                            <p class="programme__synopsis text--subtle">
+                                                                <span>{{ child.getShortSynopsis() }}</span>
```
votemike: This is VERY indented..... It feels like it could do with a refactor. Probably into a presenter.

templates/profiles/index.html.twig
```diff
@@ -6,13 +6,79 @@
...
         <h1><span class="context">{{ ds_shared('entityContext', coreEntity) }} </span>{{ tr('profiles') }}</h1>
-        {% for profile in profiles %}
-            <a href="{{ path('programme_profile', {'key': profile.getKey(), 'slug': profile.getSlug()}) }}">{{ profile.getTitle() }}</a>
-            {% if profile.getChildren() %}
-                {% for child in profile.getChildren() %}
-                    <a href="{{ path('programme_profile', {'key': child.getKey(), 'slug': child.getSlug()}) }}">{{ child.getTitle() }}</a>
+        <div class="component__body br-box-page">
+            <ul class="grid-wrapper highlight-box-wrapper highlight-box-wrapper--grid">
+                {%  if profiles|length <= 0 %}
+                    <div class="grid">
```
votemike: Does the grid class help for a piece of text like this?

templates/profiles/index.html.twig
```diff
@@ -6,13 +6,79 @@
...
+                        <p class="note">Sorry, there are no profiles for <a href="{{ path('find_by_pid', {'pid': programme.pid}) }}" class="br-blocklink__link">{{ programme.getTitle() }}</a></p>
```
votemike: Should this be translated?

templates/profiles/index.html.twig
```diff
@@ -6,13 +6,79 @@
...
+                    </div>
+                {% endif %}
+                {% for profile in profiles %}
+                    {% if not profile.getChildren() %}
```
votemike: Why not just do `{% if profile.getChildren() %}` and then an `else`?

templates/profiles/index.html.twig
benolot: Yeah, the grid-wrapper gives it negative letter spacing, the letter-spacing is fixed by adding a grid class (As it says to do in the comments of _grid.scss)

src/Ds2013/Presenters/Domain/Recipe/_recipe.scss
benolot: Moved it into the _block-link.scss file and applied it to all A tags rather than just on recipe's as I mentioned in PR notes. Would like Francisco to confirm it doesn't break his stuff before I merge it in.

votemike: block-link__overlay-link is only ever a class on anchor tags isn't it? So it will be applied to every instance. Won't it?

votemike:
There is an H1 that goes straight to a bunch of H3's on both https://sandbox.bbc.co.uk/programmes/b006m86d/profiles and https://sandbox.bbc.co.uk/programmes/b006q2x0/profiles?page=2

templates/profiles/index.html.twig
votemike: Also, why is a p tag in a ul? And why does the ul have no lis?

benolot:
Test this please

src/ExternalApi/Isite/Mapper/ProfileMapper.php
```diff
@@ -58,14 +60,16 @@ public function getDomainModel(SimpleXMLElement $isiteObject): Profile
...
             }
         }
-        // @codingStandardsIgnoreEnd
 
         $onwardJourneyBlock = null;
         if (!empty($form->profile->onward_journeys)) {
-            $onwardJourneyBlock = $this->mapperFactory->createContentBlockMapper()->getDomainModel(
-                $form->profile->onward_journeys->result
-            );
+            if (!empty($block->content_block->result->metadata->guid)) { // Must be published
```
votemike: My bad, I think `$block->content_block->result->metadata->guid` should be something like `$form->profile->onward_journeys->result->metadata->guid` or `$form->profile->onward_journeys->metadata->guid` something like that

templates/profiles/index.html.twig
```diff
@@ -6,13 +6,49 @@
...
+                                <div class="component component--box component--box-flushbody-vertical component--box--primary">
+                                    <div class="component__header br-box-page">
+                                        <h2><a href="{{ path('programme_profile', {'key': profile.getKey(), 'slug': profile.getSlug()}) }}">{{ profile.getTitle() }}</a></h2>
+                                    </div>
+                                    <div class="component__body br-box-page">
+                                        <ul class="highlight-box-wrapper grid-wrapper">
+                                            {% for child in profile.getChildren() %}
+                                                <li class="grid 1/2@bpb1 1/3@bpb2 1/3@bpw 1/4@bpw2 1/4@bpe">
+                                                    {{ ds2013('Profile', child, {
+                                                        'heading_level': 'h2',
```
votemike: Should this not be h3?

templates/profiles/index.html.twig
```diff
@@ -6,13 +6,45 @@
...
-            <a href="{{ path('programme_profile', {'key': profile.getKey(), 'slug': profile.getSlug()}) }}">{{ profile.getTitle() }}</a>
-            {% if profile.getChildren() %}
-                {% for child in profile.getChildren() %}
-                    <a href="{{ path('programme_profile', {'key': child.getKey(), 'slug': child.getSlug()}) }}">{{ child.getTitle() }}</a>
+        <div class="component__body br-box-page">
+            <ul class="grid-wrapper highlight-box-wrapper highlight-box-wrapper--grid">
+                {%  if profiles|length <= 0 %}
+                    <li>
+                        <div class="grid">
+                            <p class="note">{{ tr('profiles_none', {'%1': "<a href='" ~ path('find_by_pid', {'pid': programme.pid}) ~ "' class='br-blocklink__link'>" ~ programme.getTitle() ~ "</a>"})|raw }}</p>
```
votemike: I still don't think this message should be in a list.

src/ExternalApi/Isite/Mapper/ProfileMapper.php
```diff
@@ -12,22 +12,24 @@ class ProfileMapper extends Mapper
...
+        $formMetaData = $this->getFormMetaData($isiteObject);
         // @codingStandardsIgnoreStart
         // Ignored PHPCS cause of snake variable fields included in the xml
         $shortSynopsis = $this->getString($formMetaData->short_synopsis);
         $longSynopsis = $this->getString($formMetaData->long_synopsis);
         $parentPid = $this->getString($formMetaData->parent_pid);
         $brandingId = $this->getString($formMetaData->branding_id);
         $imagePortrait = $this->getString($form->profile->image_portrait);
-
+        // @codingStandardsIgnoreEnd
```
votemike: Any reason you've moved this so far up from where it was?

benolot: I figured theres no point ignoring the below code, as it's not snake cased, so might as well check it for coding standards.

votemike: This is snake cased: https://github.com/bbc/programmes-frontend/pull/677/files#diff-69f83bb43a02f3d252b5eeab94e4b13aR33

votemike: Because you moved the other one up.... a second one was needed.

votemike: You can see it was covered by the original one.

benolot: Yes but what I'm saying is, we're ignoring code that doesn't need to be ignored by applying one big one, as opposed to two smaller ones.

benolot: Lines 24-30 were being ignored, when theres no good reason to ignore them. The linter is there for the purpose of checking the code quality, so imo theres no point ignoring these lines if we don't need to.

src/Controller/Profiles/IndexController.php
```diff
@@ -6,6 +6,7 @@
 use App\Controller\BaseController;
 use App\ExternalApi\Isite\Domain\Profile;
 use App\ExternalApi\Isite\Service\IsiteService;
+use App\Ds2013\Presenters\Utilities\Paginator\PaginatorPresenter;
```
votemike: This should be alphabetical. If in doubt you can run Optimize Imports in Intellij

votemike:
This page throws an Exception. https://sandbox.bbc.co.uk/programmes/b09kkt1m/profiles

src/ExternalApi/Isite/Mapper/ProfileMapper.php
votemike: Only because you've moved what is currently line 24-30 inside the ignore statement... Move it back to before when the ignore case is started and all will be fine with the world.

votemike:
On http://sandbox.bbc.co.uk/programmes/b006qpgr/profiles Oliver Sterlings image seems smaller than LIVE.

src/ExternalApi/Isite/Mapper/ProfileMapper.php
```diff
@@ -58,14 +60,16 @@ public function getDomainModel(SimpleXMLElement $isiteObject): Profile
...
             }
         }
-        // @codingStandardsIgnoreEnd
 
         $onwardJourneyBlock = null;
         if (!empty($form->profile->onward_journeys)) {
-            $onwardJourneyBlock = $this->mapperFactory->createContentBlockMapper()->getDomainModel(
-                $form->profile->onward_journeys->result
-            );
+            if (!empty($form->profile->onward_journeys->result->metadata->guid)) { // Must be published
```
votemike: This line now almost duplicates https://github.com/bbc/programmes-frontend/pull/677/files#diff-69f83bb43a02f3d252b5eeab94e4b13aR55 would it be worth refactoring in to a private method along the lines of `if ($this->contentBlockIsPublished($form->profile->onward_journeys)))`

src/ExternalApi/Isite/Mapper/ProfileMapper.php
```diff
@@ -12,7 +12,7 @@ class ProfileMapper extends Mapper
     public function getDomainModel(SimpleXMLElement $isiteObject): Profile
     {
         $form = $this->getForm($isiteObject);
-        $formMetaData = $form->metadata;
+        $formMetaData = $this->getFormMetaData($isiteObject);
```
votemike: Is this change beneficial?

src/ExternalApi/Isite/Mapper/ProfileMapper.php
```diff
@@ -26,8 +26,8 @@ public function getDomainModel(SimpleXMLElement $isiteObject): Profile
         $parentPid = $this->getString($formMetaData->parent_pid);
         $brandingId = $this->getString($formMetaData->branding_id);
         $imagePortrait = $this->getString($form->profile->image_portrait);
-
```
votemike: Was this removed for a reason?

src/ExternalApi/Isite/Mapper/ProfileMapper.php
```diff
@@ -26,8 +26,8 @@ public function getDomainModel(SimpleXMLElement $isiteObject): Profile
...
         $keyFacts = [];
+        // Ignored PHPCS cause of snake variable fields included in the xml
```
votemike: Is this the appropriate place for this comment?

benolot:
@votemike Theres a ticket for the exceptions, https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6553.
Also yeah the oliver sterling image is out by 3px (183px in Sandbox, 186px in live), I'll look into it, but I assume it's the image presenters ratio.

src/ExternalApi/Isite/Mapper/ProfileMapper.php
benolot: It's how it was working originally before the rebase, and it just stuck like that in my code after rebase, I don't think it really matters either way as there appears to be no data differences.

templates/profiles/index.html.twig
```diff
@@ -6,13 +6,42 @@
...
-                    <a href="{{ path('programme_profile', {'key': child.getKey(), 'slug': child.getSlug()}) }}">{{ child.getTitle() }}</a>
-                {% endfor %}
+        <div class="component__body br-box-page">
+            {%  if profiles|length <= 0 %}
+                <p class="note">{{ tr('profiles_none', {'%1': "<a href='" ~ path('find_by_pid', {'pid': programme.pid}) ~ "' class='br-blocklink__link'>" ~ programme.getTitle() ~ "</a>"})|raw }}</p>
+            {% else %}
+                <ul class="grid-wrapper highlight-box-wrapper highlight-box-wrapper--grid">
+                    {% for profile in profiles %}
+                        {% if profile.getChildren() %}
+                            <div class="grid-wrapper">
```
votemike: https://www.w3.org/TR/html50/grouping-content.html#the-ul-element
See "Content model:". Only li's are allowed as direct children of a ul.

votemike:
On http://sandbox.bbc.co.uk/programmes/b006qpgr/profiles , surely as Sterlings is an H2, then Oliver Sterling should be an H3.

templates/profiles/index.html.twig
```diff
@@ -6,13 +6,42 @@
 {% block body %}
     <div class="programmes-page prog-layout__primary br-box-page">
         <h1><span class="context">{{ ds_shared('entityContext', coreEntity) }} </span>{{ tr('profiles') }}</h1>
-        {% for profile in profiles %}
-            <a href="{{ path('programme_profile', {'key': profile.getKey(), 'slug': profile.getSlug()}) }}">{{ profile.getTitle() }}</a>
-            {% if profile.getChildren() %}
-                {% for child in profile.getChildren() %}
-                    <a href="{{ path('programme_profile', {'key': child.getKey(), 'slug': child.getSlug()}) }}">{{ child.getTitle() }}</a>
-                {% endfor %}
+        <div class="component__body br-box-page">
```
votemike: A component__body should be inside a compoent

benolot:
Same markup as is currently done in Live, is it worth creating a new ticket for headings refactor? No doubt there will be many more cases of this across all articles/profiles pages.

src/ExternalApi/Isite/Mapper/ProfileMapper.php
benolot: How do you imagine that working? (Like the check if it's published function), just doing if isset($form->profile->onward_journeys->result->metadata->guid) then return true?

votemike: Yeah. `return isset($block->result->metadata->guid)`

templates/profiles/index.html.twig
```diff
@@ -6,13 +6,42 @@
...
+                                <div class="grid">
+                                    <div class="component component--box component--box-flushbody-vertical component--box--primary">
+                                        <div class="component__header br-box-page">
+                                            <h2><a href="{{ path('programme_profile', {'key': profile.getKey(), 'slug': profile.getSlug()}) }}">{{ profile.getTitle() }}</a></h2>
+                                        </div>
+                                        <div class="component__body br-box-page">
+                                            <ul class="highlight-box-wrapper grid-wrapper">
+                                                {% for child in profile.getChildren() %}
+                                                    <li class="grid 1/2@bpb1 1/3@bpb2 1/3@bpw 1/4@bpw2 1/4@bpe">
+                                                        {{ ds2013('Profile', child, {'heading_level': 'h3'}) }}
```
votemike: Our standard is to use lowercase for presenters could you change that both here and below.

templates/profiles/index.html.twig
```diff
@@ -6,13 +6,42 @@
...
-                    <a href="{{ path('programme_profile', {'key': child.getKey(), 'slug': child.getSlug()}) }}">{{ child.getTitle() }}</a>
-                {% endfor %}
+        <div class="br-box-page">
+            {%  if profiles|length <= 0 %}
+                <p class="note">{{ tr('profiles_none', {'%1': "<a href='" ~ path('find_by_pid', {'pid': programme.pid}) ~ "' class='br-blocklink__link'>" ~ programme.getTitle() ~ "</a>"})|raw }}</p>
+            {% else %}
+                <ul class="grid-wrapper highlight-box-wrapper highlight-box-wrapper--grid">
+                    {% for profile in profiles %}
+                        {% if profile.getChildren() %}
+                            <li class="grid-wrapper">
```
votemike: This li has a class of grid-wrapper as does it's immediate parent. That doesn't sound right.

templates/profiles/index.html.twig
```diff
@@ -6,13 +6,42 @@
...
-                {% for child in profile.getChildren() %}
-                    <a href="{{ path('programme_profile', {'key': child.getKey(), 'slug': child.getSlug()}) }}">{{ child.getTitle() }}</a>
-                {% endfor %}
+        <div class="br-box-page">
+            {%  if profiles|length <= 0 %}
+                <p class="note">{{ tr('profiles_none', {'%1': "<a href='" ~ path('find_by_pid', {'pid': programme.pid}) ~ "' class='br-blocklink__link'>" ~ programme.getTitle() ~ "</a>"})|raw }}</p>
+            {% else %}
+                <ul class="grid-wrapper highlight-box-wrapper highlight-box-wrapper--grid">
+                    {% for profile in profiles %}
+                        {% if profile.getChildren() %}
```
votemike: Are you using this to check if it's a group? What if it is a group, but has no children. Does the correct thing happen in this case?

templates/profiles/index.html.twig
```diff
@@ -6,13 +6,42 @@
...
+                        {% else %}
+                            <li class="grid 1/2@bpb1 1/3@bpb2 1/3@bpw 1/4@bpw2 1/4@bpe">
+                                {{ ds2013('Profile', profile) }}
+                            </li>
+                        {% endif %}
+                    {% endfor %}
+                </ul>
             {% endif %}
-        {% endfor %}
+            {%- if paginatorPresenter -%}
```
votemike: There's not going to be a paginator if the length of profiles is lessthan or equal to zero.

benolot:
@BjornTwachtmann @Tasiobg - As myself and @votemike just discussed offline, the markup is the same between key pages and pid pages (https://sandbox.bbc.co.uk/programmes/profiles/2nv0rtJZlqCpyLc2qW3w6F7/characters and https://www.bbc.co.uk/programmes/b006m86d/profiles) - should we split this into another partial and then use that in both page types?

templates/profiles/index.html.twig
votemike: ¯\_(ツ)_/¯ I would query iSite, but I'm not sure how to do that query. Maybe just click about until you find one? Or do an `if (profile.isGroup() and !profile.hasChildren())` and then dump and die.

felipeparaujo: ¯\\_(ツ)_/¯\\ There you go, I think you dropped this

programmes-developers:

templates/profiles/index.html.twig
```diff
@@ -7,41 +7,20 @@
...
+                'paginatorPresenter': paginatorPresenter
+            } only %}
         </div>
     </div>
 {% endblock %}
+
+
+
+
+{#{% include 'profiles/partials/profile_list.html.twig' with {#}
```
votemike: What's this? Who put that there?

templates/profiles/partials/profile_list.html.twig
```diff
@@ -0,0 +1,35 @@
...
+                                    <li class="grid 1/2@bpb1 1/3@bpb2 1/3@bpw 1/4@bpw2 1/4@bpe">
+                                        {{ ds2013('profile', child, {'heading_level': 'h3'}) }}
+                                    </li>
+                                {%- endfor -%}
+                            </ul>
+                        </div>
+                    </div>
+                </li>
+            {% else %}
+                <li class="grid 1/2@bpb1 1/3@bpb2 1/3@bpw 1/4@bpw2 1/4@bpe">
```
votemike: Is this li supposed to have the same classes as the one above?

templates/profiles/index.html.twig
```diff
@@ -6,13 +6,12 @@
...
         <h1><span class="context">{{ ds_shared('entityContext', coreEntity) }} </span>{{ tr('profiles') }}</h1>
-        {% for profile in profiles %}
-            <a href="{{ path('programme_profile', {'key': profile.getKey(), 'slug': profile.getSlug()}) }}">{{ profile.getTitle() }}</a>
-            {% if profile.getChildren() %}
-                {% for child in profile.getChildren() %}
-                    <a href="{{ path('programme_profile', {'key': child.getKey(), 'slug': child.getSlug()}) }}">{{ child.getTitle() }}</a>
-                {% endfor %}
-            {% endif %}
-        {% endfor %}
+        <div class="br-box-page">
```
votemike: templates/profiles/group.html.twig has a load of `component--` classes. Does this page need them? It would be nice to make the pages as similar as possible. If not, nevermind.

PR #679, opened 2018-08-01
PROGRAMMES-6479 Bulk of Profile Page

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6479


programmes-developers:

src/Controller/Profiles/ShowController.php
```diff
@@ -56,13 +58,26 @@ public function __invoke(string $key, string $slug, Request $request, IsiteServi
         }
 
         if ($profile->isIndividual()) {
-            return $this->renderWithChrome('profiles/individual.html.twig', ['profile' => $profile]);
+            $extraProfiles = $profile->getParents();
+            $siblingPromise = $isiteService->setChildProfilesOn($extraProfiles, $profile->getProjectSpace());
+            $this->resolvePromises([$siblingPromise]);
```
votemike: Should we resolve the promises in an array like this? Or seeing as we're doing 1 at a time, shall we just put a `->wait()` on the line above.

Tasiobg: Please check if `resolvePromises` is resolving something else like branding

src/Controller/Profiles/ShowController.php
```diff
@@ -56,13 +58,26 @@ public function __invoke(string $key, string $slug, Request $request, IsiteServi
...
+            return $this->renderWithChrome('profiles/individual.html.twig', ['profile' => $profile, 'programme' => $context]);
         }
 
-        $childPromise = $isiteService->setChildProfilesOn([$profile], $profile->getProjectSpace());
-        $this->resolvePromises([$childPromise]);
+        $childPromise = $isiteService->setChildProfilesOn([$profile], $profile->getProjectSpace(), $this->getPage());
+        $responses = $this->resolvePromises([$childPromise]);
+        $extraProfiles = array_merge($profile->getChildren(), $profile->getParents());
```
Tasiobg: If we have to look into that in the future could be a bit complicated to understand what any of those variables are, do you think adding some comments like this could help?

```
// $extraProfiles are profiles at a higher lever and somehow related to this profile page
// example if this is "The Beales" family profile page, the extra profiles can be a list of the others families and groups
```

src/Ds2013/Presenters/Domain/Profile/profile.html.twig
```diff
@@ -0,0 +1,28 @@
+<div
+    class="br-box-subtle programme block-link highlight-box--grid programme--grid br-keyline br-blocklink-page br-page-linkhover-onbg015--hover">
+    <div class="programme__img 1/1 programme__img--hasimage">
+        <div class="programme__img-box">
+            <img class="image lazyload rsp-img"
```
Tasiobg: I think you are moving this to the new image presenter otherwhise you probably want to remove `rsp-img`

src/ExternalApi/Isite/Mapper/Mapper.php
```diff
@@ -0,0 +1,67 @@
...
+     */
+    protected function getMetaData(SimpleXMLElement $isiteObject): SimpleXMLElement
+    {
+        return $isiteObject->metadata;
+    }
+
+    /**
+     * Method to get the metadata form isite document
+     * @param  SimpleXMLElement $isiteObject
+     * @return SimpleXMLElement
```
Tasiobg: this can return null as well

src/Controller/Profiles/ShowController.php
votemike: Rejigged, should be sorted now.

src/ExternalApi/Isite/Mapper/Mapper.php
votemike: Doesn't it match? When does it return null?

PR #680, opened 2018-08-01
PROGRAMMES-6261-segments-fixes

Tasiobg:
- Break text with hyphens (fix small screen issues)
- Display Song title as "Untitled" if there is no title (same as v2)


programmes-developers:

PR #681, opened 2018-08-02
Image presenters

votemike:
Making a presenter that doesn't need an Image Entity


programmes-developers:

composer.lock
```diff
@@ -3495,16 +3495,16 @@
         },
         {
             "name": "symfony/http-foundation",
-            "version": "v3.4.12",
+            "version": "v3.4.14",
```
Tasiobg: Is this updated on purpose? just to make sure is not acidentally

votemike: Yes. Jenkins complains as 3.4.12 has been found to have a security hole. So the build fails.

PR #683, opened 2018-08-06
PROGRAMMES-6526 Changing audio clip playout to Sounds

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6526

I assume the method of deciding if the Clip is audio is fine.
I also assume that the data-linktrack is acceptable.


programmes-developers:

PR #684, opened 2018-08-06
[PROGRAMMES-6544] Add an offset class

benolot:
Unable to recreate this issue outside of the page to report to mozzila, but have fixed it in the page.

Ticket: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6544


programmes-developers:

src/Ds2013/Css/Core/Base/_main.scss
```diff
@@ -56,7 +56,11 @@ abbr[title] {
 
 .programmes-page--inset,
 .inset {
-    padding-left: 8px;
+    padding-right: 8px;
```
votemike: Why is this 8px, but if the page is in the bpw range, it's 0px?

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/details.html.twig
```diff
@@ -7,8 +7,13 @@
         {% else %}
             <h1 class="no-margin">{{ details.getEpisode().getTitle() }}</h1>
             <div class="gamma">
-                {{ ds_shared('entityContext', details.getEpisode(), {'include_self': false}) }}
-
+                {% if details.getEpisode().getParent().isTleo() %}
```
votemike: What was the reason for this `if`?

src/Ds2013/Css/Core/Base/_main.scss
```diff
@@ -56,7 +56,11 @@ abbr[title] {
...
+}
+
+.offset {
```
votemike: Is offset an appropriate name for this class? Is there a better and less generic name?

benolot: Can't think of a better one, feel free to suggest

src/Ds2013/Css/Core/Base/_main.scss
benolot: Fixed it for the BPW range, it should still apply.

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/details.html.twig
benolot: Don't apply a right padding to the div if we don't want it to be padded (See the if on the spans where it sets inset)

programmes-developers:

PR #685, opened 2018-08-07
PROGRAMMES-6536 Correcting classes on segments

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6536


programmes-developers:

PR #686, opened 2018-08-08
add stage account wrapper

Tasiobg:
Use stage account wrapper on stage environment


programmes-developers:

PR #688, opened 2018-08-09
Change to check if should stream via playspace

votemike:
Bjorn would like this change.


programmes-developers:

PR #689, opened 2018-08-09
[No Ticket] - Fix randomly broken test related with profiles controller

tatitati:
## What:
We have a test failing randomly.

## Why:
Random parameters in our fixture was creating the next situation:
We created a Group Profile with children. However, this children is ```any()```. This means that our Group-profile can have Individual or group children. https://github.com/bbc/programmes-frontend/blob/develop/src/ExternalApi/Isite/Service/IsiteFakeService.php#L22

These two options set the scenario to make it fails randomly:
- **It works when:** children profile is a INDIVIDUAL-PROFILE, as they have by default ```$children = []``` (meaning: "not necessary to fetch children").  https://github.com/bbc/programmes-frontend/blob/develop/src/ExternalApi/Isite/Domain/Profile.php#L85

- **It fails when:** children profile is a GROUP-PROFILE having by default ```$children = null``` (meaning: necessary to fetch children if you want to use ```$profile->getChildren()```). 

## The fix:
My initial idea when creating profile tests for the controller was not about testing the whole hierarchy with children and parents, but more about testing that the right layout was being used, that code responses are correct, and redirects work as expected. So The fix consist in setting an empty array of children for every group profile we create as this removes the random part of the test and this is also a valid value.

## URLs to tests

Individual profile:
http://sandbox.bbc.co.uk/programmes/profiles/1rtmf0y9DWkzm3Pwb0cHP2g/jane-garvey?__flush_cache=1

Group profile:
http://sandbox.bbc.co.uk/programmes/profiles/3YDBGmJwZTYtGTk2PCCbsXw/womans-hour-presenters?__flush_cache=1


programmes-developers:

src/ExternalApi/Isite/Service/IsiteFakeService.php
```diff
@@ -73,6 +60,26 @@ public function setChildProfilesOn(
...
+                1,
+                0,
+                [$this->buildGroupProfile(new Pid('p3000000'))]
+            )
+        ]);
+    }
+
+    private function buildGroupProfile(Pid $pid): Profile
+    {
+        $profileB = ProfileBuilder::anyGroup()->with([
```
votemike: Just call it `$profile`

src/ExternalApi/Isite/Service/IsiteFakeService.php
```diff
@@ -3,39 +3,26 @@
 
 use App\Builders\IsiteResultBuilder;
 use App\Builders\ProfileBuilder;
+use App\ExternalApi\Isite\Domain\Profile;
 use App\ExternalApi\Isite\IsiteResult;
 use BBC\ProgrammesPagesService\Domain\ValueObject\Pid;
 use GuzzleHttp\Promise\FulfilledPromise;
+use GuzzleHttp\Promise\Promise;
```
votemike: Where is this used?

src/ExternalApi/Isite/Service/IsiteFakeService.php
```diff
@@ -73,6 +59,21 @@ public function setChildProfilesOn(
         int $page = 1,
         int $limit = 48
     ): PromiseInterface {
-        return new FulfilledPromise([new IsiteResult(1, 1, 0, [])]);
+        return new FulfilledPromise([
+            new IsiteResult($page, $limit, 100, $profiles),
```
votemike: This does not mimic the actual functionality of setChildProfilesOn() though. Does it?

programmes-developers:

src/ExternalApi/Isite/Service/IsiteFakeService.php
```diff
@@ -75,4 +61,17 @@ public function setChildProfilesOn(
...
+
+    private function buildGroupProfile(Pid $pid): Profile
+    {
+        $profile = ProfileBuilder::anyGroup()->with([
+            'parentPid' => $pid,
+            'projectSpace' => 'progs-radio4and4extra',
+            'title' => 'men-minutes-presenters',
+            'key' => '6YDBGmJwZTYtGTk2PCCbsXw',
+        ])->build();
+        $profile->setChildren([]);
```
votemike: If this is a Group profile, surely by definition it HAS to have children?

src/ExternalApi/Isite/Service/IsiteService.php
```diff
@@ -50,10 +50,6 @@ public function setChildProfilesOn(
         int $page = 1,
         int $limit = 48
     ): PromiseInterface {
-        if (empty($profiles)) {
```
votemike: Why are you removing this?

tatitati: In the controller we throw an exception if profiles is empty. That is before using this "setChildProfilesOn" function:
https://github.com/bbc/programmes-frontend/blob/develop/src/Controller/Profiles/ShowController.php#L36

Basically we never hit this function if we don't have profiles.

src/ExternalApi/Isite/Service/IsiteFakeService.php
tatitati: I had the same question.....in the end I did it in this way because children are set in a posterior step. Are you happier setting children=[] by default in the builder for groups?, I will be happy to do it

src/ExternalApi/Isite/Service/IsiteService.php
votemike: Just because we check in that controller doesn't mean we'll check everywhere. I think it's best to leave it in.

src/ExternalApi/Isite/Service/IsiteFakeService.php
votemike: In that case, we shouldn't be setting the children here.... should we?

src/ExternalApi/Isite/Service/IsiteService.php
```diff
@@ -51,7 +51,7 @@ public function setChildProfilesOn(
         int $limit = 48
     ): PromiseInterface {
         if (empty($profiles)) {
-            return new FulfilledPromise(null);
+            return new FulfilledPromise([]);
```
votemike: Has this been changed from null for a reason?

src/ExternalApi/Isite/Service/IsiteFakeService.php
tatitati: Exactly. I will move it then to the builder

src/ExternalApi/Isite/Service/IsiteService.php
tatitati: Yes. For consistency. We were returning an array always:
https://github.com/bbc/programmes-frontend/blob/develop/src/ExternalApi/Isite/Service/IsiteService.php#L74

votemike: That's on error.

src/Builders/ProfileBuilder.php
```diff
@@ -44,4 +44,14 @@ public static function anyIndividual()
             'type' => 'individual',
         ]);
     }
+
+    public function build() {
+        // We cannot pass children into the constructor of a group profile directly
+        $profile = parent::build();
+        if ($profile->isGroup()) {
+            $profile->setChildren([]);
```
votemike: What? A group would have children!

src/Controller/Profiles/ShowController.php
```diff
@@ -64,12 +64,15 @@ public function __invoke(string $key, string $slug, Request $request, IsiteServi
             return $this->renderWithChrome('profiles/individual.html.twig', ['profile' => $profile, 'programme' => $context]);
         }
 
+        /**
+         * @var IsiteResult[] $response
```
votemike: Could probably go on 1 line

PR #690, opened 2018-08-10
PROGRAMMES-6487 replace iPlayer radio icon

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6487

This PR makes me feel sick. The CSS logic for our CTA's is grim! Once Sounds is 100% done, we should probably go through and refactor all this a bit, as well as clear out any references to iPlayer Radio and also choose a consistent term for Sounds/Playspace/CAP.

This PR is only for the CTA colour as the icon was done as a previous PR.


programmes-developers:

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/_playout.scss
```diff
@@ -74,11 +74,21 @@ $size: 64px;
...
-        color: #ffffff !important;
     }
 }
 
+.iplayer {
+    .playout__blocklink:not([disabled]):hover,
+    .playout__blocklink:not([disabled]):active,
+    .playout__blocklink:not([disabled]):focus {
+        .playout__message {
+            background: $iplayer-pink !important; //pink is the brand colour
```
tatitati: Maybe we can reduce some duplication here doing:

```css
.iplayer {
    .playout__blocklink:not([disabled]) {
       &:hover, &:active, &:focus {
            .playout__message {
                   background: ....

File: src/Ds2013/Css/Components/_sounds-icon.scss [hunk 2]
```diff
@@ -0,0 +1,11 @@
+.sounds-icon {
+    color: #ffffff !important;
```
BjornTwachtmann: !imporant !!!!

BjornTwachtmann:
Also, have a look at the listen live icon on a radio brand page where the episode image isn't being displayed. It's a big box now, where before it was a small box with some text next to it.

programmes-developers:

Tasiobg:
Looks good but I suggest to avoid `!important` if possible, I tried locally and I coudn't find a case where the `!imporant` was required so I can't see how this breaks to suggest a better solution

src/Ds2013/Css/Components/_sounds-icon.scss
votemike: If I change the selector to `a div.programme__icon.sounds-icon {` I can get rid of `!important`

BjornTwachtmann: yeah, definitely

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/_playout.scss
```diff
@@ -74,11 +74,21 @@ $size: 64px;
...
-        color: #ffffff !important;
     }
 }
 
+.iplayer .playout__blocklink:not([disabled]) {
+    &:hover,
+    &:active,
+    &:focus {
+        .playout__message {
+            background: $iplayer-pink !important; //pink is the brand colour
```
BjornTwachtmann: Does this still need the !important?

votemike: The previous one did, so I left it on. I was thinking, if it causes issues in the future, it can be investigated then. I imagine it originally had !important for a reason.

tatitati:
Just checked the code, not tested in my local

Tasiobg:
test this please

votemike:
Test this please

PR #691, opened 2018-08-13
PROGRAMMES-6464 Sounds Nav Bar

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6464


programmes-developers:

src/Sounds/SoundsNavClient.php
```diff
@@ -0,0 +1,170 @@
...
+    public function getContentAsync()
+    {
+        $url = $this->getUrl();
+        $cacheItem = $this->fetchCacheItem($url);
+        if ($cacheItem->isHit()) {
+            $brandingObject = $this->mapResultToObject($cacheItem->get());
+            return new FulfilledPromise($brandingObject);
+        }
+
+        //@TODO Do we want to catch Exceptions and do anything with them?
```
votemike: Do we want to catch if the call to the SoundsNav fails and fail more gracefully?

src/Sounds/SoundsNavClient.php
```diff
@@ -0,0 +1,170 @@
...
+        $requestPromise = $this->client->requestAsync('GET', $url, [
```
votemike: Copy/Pasted from Branding/Orbit is this correct?

src/Sounds/SoundsNavClient.php
```diff
@@ -0,0 +1,170 @@
...
+        ]);
+        $promise = $requestPromise->then(
+        // Success callback
+            function ($response) use ($cacheItem) {
+                $result = $this->parseAndCacheResponse($response, $cacheItem);
+                return $this->mapResultToObject($result);
+            },
+            // Error callback
+            function ($reason) {
+                throw $reason;
```
votemike: Do we want to do something cleverer here?

src/Sounds/SoundsNavClient.php
```diff
@@ -0,0 +1,170 @@
...
+            $result['body'],
+            $result['foot']
+        );
+    }
+
+    private function parseAndCacheResponse(ResponseInterface $response, CacheItemInterface $cacheItem)
+    {
+        $result = json_decode($response->getBody()->getContents(), true);
+        if (!$result || !isset($result['head'])) {
+            throw new Exception('Invalid Sounds Nav Response. Response JSON object was invalid or malformed');
```
votemike: Are we happy with using the generic Exception?

src/Sounds/SoundsNavClient.php
```diff
@@ -0,0 +1,170 @@
...
+    public function __construct(
+        Client $client,
+        CacheItemPoolInterface $cache,
+        array $options = []
+    ) {
+        $this->client = $client;
+        $this->cache = $cache;
+
+        if (array_key_exists('env', $options) && !in_array($options['env'], self::SUPPORTED_ENVIRONMENTS)) {
+            throw new Exception(sprintf(
```
votemike: Are we happy with the generic Exception?

src/Sounds/SoundsNavClient.php
```diff
@@ -0,0 +1,170 @@
...
+                'Invalid environment supplied, expected one of "%s" but got "%s"',
+                implode(', ', self::SUPPORTED_ENVIRONMENTS),
+                $options['env']
+            ));
+        }
+
+        if (array_key_exists('cacheTime', $options) && !(is_int($options['cacheTime']) && $options['cacheTime'] >= 0)) {
+            throw new Exception(sprintf(
```
votemike: Are we happy with the generic Exception?

src/Cosmos/Dials.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    private $dials;
+
+    public function __construct(string $path)
+    {
+        // Default object to empty
+        $this->dials = (object) [];
+
+        // Try read the file in
+        if (file_exists($path)) {
```
BjornTwachtmann: Can you make this lazy so it's only fopen()'d when it's needed in get()

src/Sounds/SoundsNavClient.php
benolot: Personally I would say yes, as the page can function perfectly correctly without the nav (apart from looking a little weird) as far I understood it?

programmes-developers:

src/Sounds/SoundsNavClient.php
```diff
@@ -0,0 +1,176 @@
...
+use DateTime;
+use Exception;
+use GuzzleHttp\Client;
+use GuzzleHttp\Promise\FulfilledPromise;
+use function GuzzleHttp\Psr7\parse_header;
+use Psr\Cache\CacheItemInterface;
+use Psr\Cache\CacheItemPoolInterface;
+use Psr\Http\Message\ResponseInterface;
+
+class SoundsNavClient
```
BjornTwachtmann: Any reason this isn't using the standard API client patterns and living in ExternalApi, it seems pretty standard apart from the variable cache times. Do we need those? Was that something sounds were keen on?

src/Sounds/SoundsNavClient.php
BjornTwachtmann: InvalidArgumentException?

src/Sounds/SoundsNavClient.php
BjornTwachtmann: I don't think this API failing should take down the page

src/Sounds/SoundsNavClient.php
BjornTwachtmann: Take a look at https://github.com/bbc/programmes-frontend/blob/develop/src/ExternalApi/FavouritesButton/Service/FavouritesButtonService.php . The branding client isn't really a good pattern to follow.

src/Controller/Schedules/SchedulesBaseController.php
```diff
@@ -0,0 +1,42 @@
...
+        if ($this->shouldShowSoundsNav()) {
+            $this->branding->overrideOption('orb_header', 'white');
+        }
+    }
+
+    protected function renderWithChrome(string $view, array $parameters = [])
+    {
+        if ($this->shouldShowSoundsNav()) {
+            $soundsNavPromise = $this->container->get(SoundsNavClient::class)->getContentAsync();
+            $resolvedPromises = $this->resolvePromises(['soundsNav' => $soundsNavPromise]);
```
BjornTwachtmann: Doing it this way, our HTTP requests are going to end up being made in sequence (first everything else, then this one). I think we want to find a way to do them all in parallel if we can

src/EventSubscriber/CacheFlushSubscriber.php
```diff
@@ -53,6 +54,9 @@ public function setupCacheFlush(GetResponseEvent $event)
 
             $morphCache = $this->container->get(MorphClient::class);
             $morphCache->setFlushCacheItems(true);
+
+            $soundsNavCache = $this->container->get(SoundsNavClient::class);
+            $soundsNavCache->setFlushCacheItems(true);
```
BjornTwachtmann: The API Client should be using our Cache class, and this is already handled in there. The only reason the branding stuff is in here is because that's a shared project that doesn't use our internal bits and pieces

src/Controller/Schedules/SchedulesBaseController.php
```diff
@@ -0,0 +1,42 @@
...
+{
+    public static function getSubscribedServices()
+    {
+        return array_merge(parent::getSubscribedServices(), [
+            Dials::class,
+            SoundsNavClient::class,
+        ]);
+    }
+
+    protected function preRender()
```
BjornTwachtmann: I'm not sure I like overriding this and making the branding accessible everywhere. Would be be better to have some  kind of an overrideBrandingOption($key, $value) method on the baseController that sets a variable fulfilledBrandingPromise() can read from?

src/Sounds/SoundsNavClient.php
votemike: I was following the branding client, I'll get rid of the cache times.

programmes-developers:

src/Controller/BaseController.php
```diff
@@ -42,9 +44,7 @@
     private $brandingPromise;
 
     /** @var Branding */
-    private $branding;
-
-    private $context;
+    protected $branding;
```
BjornTwachtmann: This can be private if you're not accessing it outside of the baseController, right?

config/services_prod_stage.yaml
```diff
@@ -3,3 +3,4 @@ parameters:
     app.feed_env.branding: 'live'
     app.isite.baseurl: 'https://api.test.bbc.co.uk/isite2-content-reader'
     app.vote.accountwrapperurl: 'https://account-wrapper.stage.files.bbci.co.uk/vote-wrapper.min.js'
+    app.sounds_nav.baseurl: 'https://sounds-nav.test.api.bbci.co.uk/json'
```
BjornTwachtmann: Can stage use the live nav? Seems like we should try and keep things talking to live where we can on stage

src/ExternalApi/SoundsNav/Service/SoundsNavService.php
```diff
@@ -0,0 +1,69 @@
...
+    {
+        $cacheKey = $this->clientFactory->keyHelper(__CLASS__, __FUNCTION__);
+
+        return $this->clientFactory->getHttpApiMultiClient(
+            $cacheKey,
+            [$this->url],
+            Closure::fromCallable([$this, 'parseResponse']),
+            [],
+            null,
+            CacheInterface::MEDIUM,
```
BjornTwachtmann: We can probably cache this for quite a long time. CacheInterface::LONG is 2 hours, which is probably about right.

src/Cosmos/Dials.php
```diff
@@ -0,0 +1,43 @@
...
+        return null;
+    }
+
+    private function loadDials(): void
+    {
+        // Default object to empty
+        $this->dials = (object) [];
+
+        // Try read the file in
+        if (file_exists($this->path)) {
```
Tasiobg: I would like to know what the others thinks about performance here. As far as I know opening files has a negative performnace impact, in this particular case and due to AWS having fast IO may not be apreciable but is diffcult to know without more testing. Opening PHP script files with PHP functions like `include` shouldn't have this impact because it gets compiles and caches in memory thanks to opcache, this is transparent for us and we don't need to do anything but opening non PHP files is different, it always generates some IO which is nice to avoid if possible.
 
The only way I can think of improving it is by using APCU but not sure if is worth the hassle.

I have quickly tested to call this `if` in a loop 100 times in my local sandbox (probably slow IO) and it takes around 35ms while if I hardcode the file content inside `json_decode` it takes less than 1ms.

What the other think? is worth chaning anything or investigating it more?

BjornTwachtmann: It's not ideal, but running this on an actual c5 AWS instance,...

```
<?php
$iters = 100000;
$t0 = microtime(true);
for ($i = 0; $i < $iters; $i++) {
        testIt();
}
$t1 = microtime(true);
echo "Total Time taken =" . (($t1 - $t0)) . "s for $iters iterations\n";
echo "Time taken =" . (($t1 - $t0) / $iters) . "s per fopen()\n";

function testIt() {
$path = '/home/bjorn_twachtmann/dials.json';
if (file_exists($path)) {
            $file = json_decode(file_get_contents($path));
            if ($file !== null) {
                $stuff = $file;
            }
    }
}```
Gives the following output:
```Total Time taken =0.78897595405579s for 100000 iterations
Time taken =7.8897595405579E-6s per fopen()```

I think it should be fine, especially as it only happens when people actually request a dial

votemike:
@Tasiobg the favourite button doesn't seem to have alarms or run book entries.

Tasiobg:
@votemike probably someone forgot to do it, what do you think @BjornTwachtmann ?

PR #692, opened 2018-08-13
PROGRAMMES-6509: Stop programme titles being unnecessarily hyphenated

BjornTwachtmann:
The previous PR for this introduced new behaviour in firefox/safari where programme titles would get hyphenated even when there was plenty of space to put them on a new line. This PR reverts that behaviour, stuff will now break onto a new line when the word is too long to fit anywhere. This does the same for segment labels too.


programmes-developers:

votemike:
Test this please

PR #693, opened 2018-08-14
PROGRAMMES-6545 Redirect credits route to FBP page

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6545

I'm redirecting no matter what. If there are no credits, the page just won't scroll.


programmes-developers:

BjornTwachtmann:
Hmm, I hadn't considered that case. Do iPlayer link to /credits even when there are none? Is that going to cause problems?

votemike:
No, they don't. I don't think it should cause any problems.

PR #695, opened 2018-08-14
PROGRAMMES-6560 Directing all non-TV audio to playout in Sounds

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6560

Seemed suspiciously easy. I must be missing something.


programmes-developers:

benolot:
Is that the correct URL? I thought sounds/playspace was `radio/play/live:<SID>`. I might be misunderstanding things though, so feel free to ignore me.

src/DsShared/Helpers/StreamableHelper.php
```diff
@@ -46,7 +44,7 @@ public function shouldStreamViaIplayer(ProgrammeItem $programmeItem): bool
     public function shouldStreamViaPlayspace(ProgrammeItem $programmeItem): bool
     {
         $network = $programmeItem->getNetwork();
-        if ($network && in_array($network->getNid(), $this->playspaceNetworksList) && $this->shouldTreatProgrammeItemAsAudio($programmeItem)) {
+        if ($network && !$network->isTv() && $this->shouldTreatProgrammeItemAsAudio($programmeItem)) {
```
BjornTwachtmann: I think this is probably correct for now. Thinking about it, I wonder what is supposed to happen in the case that a thing doesn't have a network, there's quite a lot of clips like this, and at least 1 episode.

BjornTwachtmann: I've raised it with product. For now this is correct.

PR #696, opened 2018-08-14
PROGRAMMES-6542 Fixing available soon for international networks

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6542


src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/PlayoutPresenter.php
```diff
@@ -74,11 +74,6 @@ public function doesntHaveOverlay(): bool
             return false;
         }
 
-        // International networks don't get overlays
-        if ($this->episode->getNetwork() && $this->episode->getNetwork()->isInternational()) {
```
BjornTwachtmann: What's the desired behaviour here for world news, which this will also modify?

BjornTwachtmann: Also, world service languages?

votemike: Was the previous logic ever a requirement. After looking at V2, I think this logic may have been put in after misunderstanding the V2 code.

BjornTwachtmann: So, the difference here is in the overlay on unavailable episode pages, see /programmes/n3ct4f71 , your code now displays the "Sorry, this episode is not currently available" message.

It's not the end of the world, but probably rather than removing this check, it should be changed to be just for world news, or possibly world news + world service languages. Ask cathal maybe

programmes-developers:

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/PlayoutPresenter.php
BjornTwachtmann: @claudioneves , there are special cases, there are always special cases hidden under about 15 layers of crap. In this case, they live here 
https://github.com/bbc/programmes/blob/develop/webapp/php/lib/BBC/Programmes/View/Episode/Playout.php#L97
https://github.com/bbc/programmes/blob/develop/webapp/php/lib/BBC/Programmes/View/Episode/Playout.php#L115

BjornTwachtmann: I'm not saying they should be implemented, just pointing them out. I would prefer they were killed.

programmes-developers:

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/PlayoutPresenter.php
claudioneves: @BjornTwachtmann yes, sorry. I meant that I couldn't find any special cases for this 'Coming soon' label specifically, after looking in the V2 Controller, but also in the View Helper class and the view partials.

claudioneves: @BjornTwachtmann I was looking at the wrong label, thought this was about the 'Coming soon' label in the On demand column of brand pages...

votemike:
Test this please

PR #700, opened 2018-08-15
PROGRAMMES-6259 Fix favourite button

Tasiobg:
Favourites button for segments wasn't working fine, this is a list of changes to fix it
- fix title
- moved "product" parameter to fav.render because is a constant value
- make profile an optional parameter

### DIFF
--- a/src/Ds2013/Presenters/Section/Clip/Details/clip_details.html.twig
+++ b/src/Ds2013/Presenters/Section/Clip/Details/clip_details.html.twig
@@ -45,7 +45,8 @@
                             'clip-map-favourites-button',
                             clip_details.getClip().getPid(),
                             'programme',
-                            clip_details.getClip().getPid()
+                            clip_details.getClip().getPid(),
+                            clip_details.getClip().getTitle()
                     ) }}
                     <div id="clip-map-favourites-button" class="favourites-module-wrapper map__buttons favourites-button buttons--favourites-small"></div>
                 {%- endif -%}
--- a/src/Ds2013/Presenters/Section/Segments/SegmentItem/AbstractMusicSegmentItemPresenter.php
+++ b/src/Ds2013/Presenters/Section/Segments/SegmentItem/AbstractMusicSegmentItemPresenter.php
@@ -160,5 +160,62 @@ abstract class AbstractMusicSegmentItemPresenter extends AbstractSegmentItemPres
         return $segmentEndTime->diffForHumans();
     }
 
+    /**
+     * This function generates a track title as in https://confluence.dev.bbc.co.uk/display/MAT/Item+Types
+     * this follows same logic we have in the twig template but without any HTML.
+     * This values is send to Favourites API and is not visible by the users
+     *
+     * @return string
+     */
+    public function getFavouriteButtonTitle(): string
+    {
+        return $this->getArtistWithContributions() . ' || ' . $this->getTrackTitleWithContributions();
+    }
+
     abstract protected function setupContributions(): void;
+
+    private function getArtistWithContributions(): string
+    {
+        $artistWithContributions = '';
+        if (!$this->getVersusContributions() && !$this->getPrimaryContributions()) {
+            return $artistWithContributions;
+        }
+
+        if ($this->getPrimaryContributions()) {
+            $artistWithContributions = $this->concatContributorNames(' & ', $this->getPrimaryContributions());
+        }
+
+        $artistWithContributions .= ($this->getVersusContributions() && $this->getPrimaryContributions()) ? ' vs ' : '';
+
+        if ($this->getVersusContributions()) {
+            $artistWithContributions .= $this->concatContributorNames(' & ', $this->getVersusContributions());
+        }
+
+        return $artistWithContributions;
+    }
+
+    private function getTrackTitleWithContributions(): string
+    {
+        if (!$this->segment->getTitle()) {
+            return 'Untitled';
+        }
+        if (!$this->getFeaturedContributions()) {
+            return $this->segment->getTitle();
+        }
+        return $this->segment->getTitle() . ' (feat. ' . $this->concatContributorNames(' & ', $this->getFeaturedContributions()) . ')';
+    }
+
+    /**
+     * @param string $glue
+     * @param Contribution[] $contributions
+     * @return string
+     */
+    private function concatContributorNames(string $glue, array $contributions): string
+    {
+        $names = [];
+        foreach ($contributions as $contribution) {
+            $names[] = $contribution->getContributor()->getName();
+        }
+        return implode($glue, $names);
+    }
 }
--- a/src/Twig/FavouriteButtonExtension.php
+++ b/src/Twig/FavouriteButtonExtension.php
@@ -21,19 +21,17 @@ class FavouriteButtonExtension extends Twig_Extension
         ];
     }
 
-    public function addButton(string $elementId, string $id, string $type, string $contextId, ?string $profile = null, ?string $product = null)
+    public function addButton(string $elementId, string $id, string $type, string $contextId, string $title, ?string $profile = null)
     {
         $button = [
             'element_id' => $elementId,
             'id' => $id,
             'type' => $type,
             'context_id' => $contextId,
+            'title' => $title,
         ];
         if ($profile) {
             $button['profile'] = $profile;
-        }
-        if ($product) {
-            $button['product'] = $product;
         }
         $this->buttons[] = $button;
     }
--- a/templates/find_by_pid/episode.html.twig
+++ b/templates/find_by_pid/episode.html.twig
@@ -345,11 +345,13 @@
                     'appid': 'radio',
                     'type': '{{ button['type'] }}',
                     'id': '{{ button['id'] }}',
-                    'title': '{{ programme.getTitle() }}',
+                    'title': '{{ button['title']|e('js') }}',
                     'variant': 'small',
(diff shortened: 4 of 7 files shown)


programmes-developers:

src/Ds2013/Presenters/Section/Segments/SegmentItem/AbstractMusicSegmentItemPresenter.php
```diff
@@ -160,5 +160,58 @@ public function getTiming(): ?string
...
+                foreach ($contributions as $contribution) {
+                    $names[] = $contribution->getContributor()->getName();
+                }
+                $artistWithContributions .= implode(' & ', $names);
+            }
+        }
+        return $artistWithContributions;
+    }
+
+    private function getTrackTitleWithContributions(): string
```
votemike: Could this be simplified to something like:
```
private function getTrackTitleWithContributions(): string
{
    if (!$this->segment->getTitle()) {
        return 'Untitled';
    }
    if (!$this->getFeaturedContributions()) {
        return $this->segment->getTitle();
    }
    return $this->segment->getTitle() . ' (feat. ' . $this->concatContributorNames(' & ', $this->getFeaturedContributions()) . ')';
}
```

src/Ds2013/Presenters/Section/Segments/SegmentItem/AbstractMusicSegmentItemPresenter.php
```diff
@@ -160,5 +160,58 @@ public function getTiming(): ?string
...
+     */
+    public function getFavouriteButtonTitle(): string
+    {
+        return $this->getArtistWithContributions() . ' || ' . $this->getTrackTitleWithContributions();
+    }
+
+    private function getArtistWithContributions(): string
+    {
+        $artistWithContributions = '';
+        if ($this->getVersusContributions() || $this->getPrimaryContributions()) {
```
votemike: You could reduce the indentation by reversing this check.

templates/find_by_pid/episode.html.twig
```diff
@@ -345,11 +345,13 @@
                     'appid': 'radio',
                     'type': '{{ button['type'] }}',
                     'id': '{{ button['id'] }}',
-                    'title': '{{ programme.getTitle() }}',
+                    'title': {% if button['title'] is defined %}'{{ button['title'] }}'{% else %}'{{ programme.getTitle() }}'{% endif %},
```
votemike: Would it be better to enforce a title?

Tasiobg: Not sure what do you mean.
`button['title']` is always sets for tracks, the value is comming from `getFavouriteButtonTitle()`. If is not set is because the favourite button is not for a track.

src/Ds2013/Presenters/Section/Clip/Details/clip_details.html.twig
```diff
@@ -45,7 +45,8 @@
                             'clip-map-favourites-button',
                             clip_details.getClip().getPid(),
                             'programme',
-                            clip_details.getClip().getPid()
+                            clip_details.getClip().getPid(),
+                            clip_details.getClip().getTitle()|e('js')
```
votemike: What does `|e('js')` do? Would it be better to put that as part of `'title': '{{ button['title'] }}',`? e.g. `'title': '{{ button['title']|e('js') }}',`

src/Twig/FavouriteButtonExtension.php
```diff
@@ -32,8 +32,8 @@ public function addButton(string $elementId, string $id, string $type, string $c
         if ($profile) {
             $button['profile'] = $profile;
         }
-        if ($product) {
-            $button['product'] = $product;
+        if ($title) {
```
votemike: This isn't optional. So you don't need the `if`. Right?

src/Ds2013/Presenters/Section/Segments/SegmentItem/AbstractMusicSegmentItemPresenter.php
```diff
@@ -160,5 +160,59 @@ public function getTiming(): ?string
...
     abstract protected function setupContributions(): void;
+
+    private function getArtistWithContributions(): string
+    {
+        $artistWithContributions = '';
+        if ($this->getVersusContributions() || $this->getPrimaryContributions()) {
+            if ($contributions = $this->getPrimaryContributions()) {
+                $artistWithContributions = $this->concatContributorNames(' & ', $this->getPrimaryContributions());
+            }
+
```
votemike: If there are no `$this->getVersusContributions()`, then you could return early here.

src/Ds2013/Presenters/Section/Segments/SegmentItem/AbstractMusicSegmentItemPresenter.php
```diff
@@ -160,5 +160,62 @@ public function getTiming(): ?string
...
+    /**
+     * This function generates a track title as in https://confluence.dev.bbc.co.uk/display/MAT/Item+Types
+     * this follows same logic we have in the twig template but without any HTML.
+     * This values is send to Favourites API and is not visible by the users
+     *
+     * @return string
+     */
+    public function getFavouriteButtonTitle(): string
+    {
+        return $this->getArtistWithContributions() . ' || ' . $this->getTrackTitleWithContributions();
```
votemike: So.... this could end up being " || blah" or "blah || "? Is that OK?

src/Ds2013/Presenters/Section/Segments/SegmentItem/AbstractMusicSegmentItemPresenter.php
```diff
@@ -160,5 +160,62 @@ public function getTiming(): ?string
...
     abstract protected function setupContributions(): void;
+
+    private function getArtistWithContributions(): string
+    {
+        $artistWithContributions = '';
+        if (!$this->getVersusContributions() && !$this->getPrimaryContributions()) {
+            return $artistWithContributions;
+        }
+
+        if ($contributions = $this->getPrimaryContributions()) {
```
votemike: You're not using `$contributions`.

src/Ds2013/Presenters/Section/Segments/SegmentItem/AbstractMusicSegmentItemPresenter.php
```diff
@@ -160,5 +160,62 @@ public function getTiming(): ?string
...
+     * @return string
+     */
+    public function getFavouriteButtonTitle(): string
+    {
+        return $this->getArtistWithContributions() . ' || ' . $this->getTrackTitleWithContributions();
+    }
+
     abstract protected function setupContributions(): void;
+
+    private function getArtistWithContributions(): string
```
votemike: Dunno if it's any cleaner, but maybe this?
```
private function getArtistWithContributions(): string
{
    if (!$this->getVersusContributions() && !$this->getPrimaryContributions()) {
        return '';
    }

    $artistWithContributions = '';

    if ($contributions = $this->getPrimaryContributions()) {
        $artistWithContributions = $this->concatContributorNames(' & ', $this->getPrimaryContributions());
    }

    if (!$this->getVersusContributions()) {
        return $artistWithContributions;
    }

    if ($contributions = $this->getVersusContributions()) {
        if ($artistWithContributions) {
            $artistWithContributions .= ' vs ';
        }
        $artistWithContributions .= $this->concatContributorNames(' & ', $this->getVersusContributions());
    }

    return $artistWithContributions;
}
```
Or maybe keep what you have. I dunno which I prefer.

src/Ds2013/Presenters/Section/Segments/SegmentItem/AbstractMusicSegmentItemPresenter.php
Tasiobg: example for `/programmes/b0b913fj`: Tiësto &amp; Dzeko || Jackie Chan (feat. Preme &amp; Post Malone)

tests/Ds2013/Presenters/Section/Segments/SegmentItem/AbstractMusicSegmentItemPresenterTest.php
```diff
@@ -132,6 +134,48 @@ public function testDefaultImage()
         $this->assertSame('https://ichef.bbci.co.uk/images/ic/96x96/p01c9cjb.png', $stub->getImageUrl());
     }
 
+    public function testGetFavouriteButtonTitle()
```
Tasiobg: On my point of view unitest heres is not required because this function doesn't have an strong impact in the favourites API but having something could be helpful so I decided to implement only a test for one of the complex cases.

src/Ds2013/Presenters/Section/Segments/SegmentItem/AbstractMusicSegmentItemPresenter.php
votemike: I mean if the are no artists with contributions or no track title with contributions?

Tasiobg: Yes it could build something like ` || Untitled`.  The only information I found about it is that the right format is `'title' => "{{Artist Name}} || {Track Title}}"` having a look at v2 this is not well implemented, I think this value is used for something minor, maybe tracking on Favourites side so I think the real value here is not that importat for edge cases

programmes-developers:

PR #703, opened 2018-08-16
PROGRAMMES-6487 Fixing CTA bug

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6487?focusedCommentId=3989142&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-3989142

I can't find an example of this locally, but I think this is the fix.


programmes-developers:

PR #704, opened 2018-08-17
PROGRAMMES-6502 Adding simple profile blocks

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6502

Hopefully the blocks are generic enough to be expanded upon.


src/Controller/Profiles/ShowController.php
```diff
@@ -96,4 +108,23 @@ private function getPaginator(IsiteResult $iSiteResult): ?PaginatorPresenter
 
         return new PaginatorPresenter($this->getPage(), 48, $iSiteResult->getTotal());
     }
+
+    /**
+     * @param Profile $profile
+     * @return Presenter[]
+     */
+    private function buildContentBlockPresenters(Profile $profile): array
```
votemike: Do we want an Exception or anything here if the block type is unknown?

src/ExternalApi/Isite/Domain/ContentBlock/AbstractContentBlock.php
```diff
@@ -0,0 +1,9 @@
+<?php
+declare(strict_types = 1);
+
+namespace App\ExternalApi\Isite\Domain\ContentBlock;
+
+abstract class AbstractContentBlock
```
votemike: Should we had a title property and getter to this class?

programmes-developers:

src/Controller/Profiles/ShowController.php
BjornTwachtmann: I'm never sure with this kind of thing. If you do throw an exception it can end up torpedoing pages on live. If you don't you can miss bugs quite easily. I'd probably lean towards yes. 

In terms of where this logic lives, would it be better to have a contentBlockPresenter method in the PresenterFactory that does this? Bit more re-usable between articles/profiles/lists etc that way. It doesn't feel like the controller should need to know this.

src/ExternalApi/Isite/Domain/ContentBlock/Links.php
```diff
@@ -0,0 +1,29 @@
+<?php
+declare(strict_types=1);
+
+namespace App\ExternalApi\Isite\Domain\ContentBlock;
+
+class Links extends AbstractContentBlock
+{
+    /** @var array */
```
tatitati: if we know, can we add the type of the items that contains this array? (string, int..)

src/Ds2013/Presenters/Domain/Profile/Image/image.html.twig
```diff
@@ -0,0 +1,9 @@
+<div class="component component--box component--box-flushbody-vertical component--box--primary">
+    <div class="component__header br-box-page">
+        <h2>{{ image.getImage().getTitle() }}</h2>
+    </div>
+    <div class="component__body br-box-page">
+        {{ ds_shared('image', image.getImage().getImage(), 256, '', {'image_classes': 'rsp-img'}) }}
```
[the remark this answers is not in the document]
votemike: The first getImage is the image object. The second is the image itself.

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
```diff
@@ -3,15 +3,67 @@
...
+                    $links[$this->getString($link->link_title)] = $this->getString($link->url);
+                    // @codingStandardsIgnoreEnd
+                }
+                $contentBlock = new Links(
+                    $this->getString($contentBlockData->title),
+                    $links
+                );
+                break;
+            default:
+//                throw new Exception('Invalid content block type. Found ' . $type);
```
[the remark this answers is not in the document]
votemike: Yep. Because not all the block types can be mapped yet, which would make testing impossible.

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
```diff
@@ -3,15 +3,67 @@
 
 namespace App\ExternalApi\Isite\Mapper;
 
-use App\ExternalApi\Isite\Domain\ContentBlock;
+use App\ExternalApi\Isite\Domain\ContentBlock\AbstractContentBlock;
+use App\ExternalApi\Isite\Domain\ContentBlock\Image;
+use App\ExternalApi\Isite\Domain\ContentBlock\Links;
+use Exception;
```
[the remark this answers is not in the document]
votemike: Not sure

src/Ds2013/Presenters/Domain/Profile/Image/image.html.twig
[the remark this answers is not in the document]
benolot: That seems very confusing... any way we can clarify that? Maybe change the methods?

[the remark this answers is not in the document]
benolot: getImageObject or getImagePid?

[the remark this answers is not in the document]
votemike: `image.getImageObject().getImagePid()`?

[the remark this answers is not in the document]
benolot: Seems more reasonable than `image.getImage().getImage()`

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
```diff
@@ -3,15 +3,67 @@
...
 use SimpleXMLElement;
 
 class ContentBlockMapper extends Mapper
 {
+    /**
+     * public function getDomainModel(SimpleXMLElement $isiteObject): AbstractContentBlock
```
tatitati: is it residual this comment?

votemike: Once all block cases are accounted for, we should be able to uncomment this line.

src/Ds2013/Presenters/Domain/ContentBlock/Image/image.html.twig
```diff
@@ -0,0 +1,9 @@
+<div class="component component--box component--box-flushbody-vertical component--box--primary">
+    <div class="component__header br-box-page">
+        <h2>{{ image.getImage().getTitle() }}</h2>
+    </div>
+    <div class="component__body br-box-page">
+        {{ ds_shared('image', image.getImage().getImage(), 256, '', {'image_classes': 'rsp-img'}) }}
```
BjornTwachtmann: Can you get the list of sizes from V2, I don't think hardcoding this to 256 is right

src/Ds2013/Presenters/Domain/ContentBlock/Links/links.html.twig
```diff
@@ -0,0 +1,14 @@
+<div class="component component--box component--box-flushbody-vertical component--box--primary">
+    <div class="component__header br-box-page">
+        <h2>{{ links.getLinks().getTitle() }}</h2>
+    </div>
+    <div class="component__body br-box-page">
+        <ul class="no-margin-vertical list-spaced">
+            {% for title, url in links.getLinks().getLinks() %}
+                <li class="related-link">
+                    <a href="{{ url }}">{{ title }}</a>
```
BjornTwachtmann: In V2, external URLs get the domain in brackets after the link title

src/Ds2013/Presenters/Domain/ContentBlock/Image/image.html.twig
```diff
@@ -0,0 +1,9 @@
+<div class="component component--box component--box-flushbody-vertical component--box--primary">
```
BjornTwachtmann: Not quite sure where to put this comment, so I might as well put it here. In article pages, these content blocks can live in one of two columns, and the component classes change accordingly. It's not an issue for profile pages, but this stuff is going to have to change in some way. Probably we'll want something in the base content block presenter class to work out what these classes should be or something. It doesn't need to be in this PR, but it might be worth having a look at that whole mess at some point.

src/Ds2013/Presenters/Domain/ContentBlock/Image/image.html.twig
votemike: 256 is the default size.

src/Ds2013/Presenters/Domain/ContentBlock/Image/image.html.twig
votemike: Would it be an idea to get these 2 block types working in Article pages too before we continue with more block types?

src/Ds2013/Presenters/Domain/ContentBlock/Image/image.html.twig
BjornTwachtmann: Presumably it's not fixed sized? Shouldn't we have a sizes array here?

src/Ds2013/Presenters/Domain/ContentBlock/Image/image.html.twig
BjornTwachtmann: probably not a bad idea

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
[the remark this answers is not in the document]
BjornTwachtmann: Do we intend to catch these exceptions? If not they can be generic

src/Ds2013/Presenters/Domain/ContentBlock/Image/image.html.twig
votemike: I don't think sizes are used if it's lazy loaded.

BjornTwachtmann: it will auto size lazy-loaded, yeah, didn't actually realise that was the default. 

I don't think this image should be lazy loaded, if editorial are plonking an "image" block into the page, I'd consider that core content and always load it.

Also, rsp-img is not a thing in V3, it's a V2 thing, don't think it will do anything here.

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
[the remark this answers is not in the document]
votemike: Do we want it to be fatal if an unknown block type exists? There shouldn't be random block types that we don't know about.

src/Ds2013/Presenters/Domain/ContentBlock/Links/links.html.twig
votemike: Done

src/Ds2013/Presenters/Domain/ContentBlock/Image/image.html.twig
votemike: Done. I think there are other rsp-img classes in V3. Should we remove them?

src/Ds2013/Presenters/Domain/ContentBlock/Image/image.html.twig
```diff
@@ -0,0 +1,9 @@
...
+    <div class="component__header br-box-page">
+        <h2>{{ image.getImage().getTitle() }}</h2>
+    </div>
+    <div class="component__body br-box-page">
+        {{ ds_shared('image', image.getImage().getImage(), 256, {0: 1, 770: 1/2, 1008: '490px'}, {'is_lazy_loaded': false}) }}
```
Tasiobg: Would it be more clear to rename `ImagePresenter` to `ImageBlockPresenter` and `getImage()` to `getImageBlock()` or something else? Not sure if this is a good idea but calling `image.getImage().getImage()` seems a bit weird

src/Ds2013/Presenters/Domain/ContentBlock/Links/links.html.twig
```diff
@@ -0,0 +1,16 @@
...
+                    {% set externalUrl = links.extractExternalHost(url) %}
+                    {% if externalUrl %} <span class="milli">({{ externalUrl }})</span>{% endif %}
```
Tasiobg: I think the round brackets are also added in `extractExternalHost()`

src/ExternalApi/Isite/Domain/ContentBlock/AbstractContentBlock.php
Tasiobg: Sounds good to me, title can be removed from the other classes

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
```diff
@@ -3,15 +3,67 @@
...
+                    $this->getString($contentBlockData->image_caption)
+                    // @codingStandardsIgnoreEnd
+                );
+                break;
+            case 'links':
+                $contentBlockData = $form->content;
+                $links = [];
+                foreach ($contentBlockData->links as $link) {
+                    // @codingStandardsIgnoreStart
+                    $links[$this->getString($link->link_title)] = $this->getString($link->url);
```
Tasiobg: I'm assuming there won't be different URLs with the same link_title

src/Ds2013/Presenters/Domain/ContentBlock/Links/links.html.twig
votemike: What do you mean "also added"?

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
votemike: I assume so too.
Do you think it should be done in some different way?

Tasiobg: looks good, just to make sure you are aware

src/Ds2013/Presenters/Domain/ContentBlock/Links/links.html.twig
Tasiobg: In LinksPresenter.php L35 it wraps the host name with round brackets while in this line it wrap the same text with round brackets again, this create something like:
`<span class="milli">((domain.com))</span>`
I think adding only once the brackets to the twig file is enough

src/Ds2013/Presenters/Domain/ContentBlock/Links/links.html.twig
```diff
@@ -0,0 +1,16 @@
+<div class="component component--box component--box-flushbody-vertical component--box--primary">
+    <div class="component__header br-box-page">
+        <h2>{{ links.getLinks().getTitle() }}</h2>
+    </div>
+    <div class="component__body br-box-page">
+        <ul class="no-margin-vertical list-spaced">
+            {% for title, url in links.getLinks().getLinks() %}
```
Tasiobg: Would it make sense to also rename `getLinks()` similar to what you did with `getImages()`? Not necessary but the change you did for images looks nice

src/Ds2013/Presenters/Domain/ContentBlock/Links/links.html.twig
votemike: Oops

PR #706, opened 2018-08-20
[PROGRAMMES-6307] - Fix css in episode map playout section. 

tatitati:
## What:
Playout panel in episode map looks shorter than Tx panel for some sizes.
We "fixed" this already, however in all the examples it was tested (table below) the tx column present a short text. So the problem was not visible until we found a case with a longer text such as:
http://sandbox.bbc.co.uk/programmes/m00006jr?__flush_cache=1

## Why:
The "magic" css we added:
```css
margin-bottom: -50em;
padding-bottom: 50em;
```
was being added for sizes bigger (or equal) than bpw2 (this is, bpw2, and bpe), however we forgot to add it for bpe. Then at this point (bpe) we had again:
```css
margin-bottom: 0;
padding-bottom: 0;
```
As a reminder:
```
 *                            0           320    480   600        770      1008
 * [unprefixed]               |----------------------------------------------->
 * bpb1                                   |------------|
 * bpb2                                          |-----|
 * bpw                                                 |---------------------->
 * bpw1                                                |-------------------|
 * bpw2                                                           |--------|
 * bpe                                                                     |-->
```

More examples (same as before) to check that still the previous examples keep working properly:

| Tx | More | Empty | URL example:
|--- | --- | --- |--- |
| 0 | 0 | 1 | http://sandbox.bbc.co.uk/programmes/b0074ggc | 
| 0 | 1 | 0 | http://sandbox.bbc.co.uk/programmes/p06467mn | 
| 1 | 0 | 1 | http://sandbox.bbc.co.uk/programmes/b00744w1|
| 1 | 1 | 0 | http://sandbox.bbc.co.uk/programmes/b0b6v9qt|


programmes-developers:

BjornTwachtmann:
You say that this is missing at a specific breakpoint, but your CSS has nothing to do with that breakpoint, it adds specificity. So I'm confused, why have you done this?

PR #707, opened 2018-08-21
PROGRAMMES-6510 Using Reith font

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6510


programmes-developers:

templates/bundles/TwigBundle/Exception/base_no_orb.html.twig
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="{{ tr('language_code') }}">
+<html class="b-reith-sans-font" lang="{{ tr('language_code') }}">
```
benolot: Does this work with no orb?

PR #708, opened 2018-08-21
PROGRAMMES-6483 Adding Article skeletons

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6483

Just basic skeletons.


programmes-developers:

src/Controller/Articles/IndexController.php
```diff
@@ -0,0 +1,29 @@
...
+class IndexController extends BaseController
+{
+    public function __invoke(CoreEntity $coreEntity, IsiteService $isiteService)
+    {
+        $this->setContextAndPreloadBranding($coreEntity);
+
+        $articles = [];
+        $parameters = ['coreEntity' => $coreEntity, 'articles' => $articles, 'paginatorPresenter' => null];
+        if ($coreEntity instanceof Programme) {
+            $parameters['programme'] = $coreEntity; //so the the base 2013 template sets the footer
```
tatitati: just wondering if this might be wrong or maybe not. But if the check is true we would have:
```php
$parameters[
    'coreEntity' => $coreEntity,
    'programme' => $coreEntity,
    .....
```

templates/articles/partials/article_list.html.twig
```diff
@@ -0,0 +1 @@
+The list of articles/features
```
benolot: I assume this is being done later?

templates/articles/show.html.twig
```diff
@@ -0,0 +1,10 @@
+{% extends 'base_ds2013.html.twig' %}
+
+{% block title %}{{ meta_context.titlePrefix() }} - {{ tr('features') }}{% endblock %}
+{% block page_classes %}text-base{% endblock %}
+
+{% block body %}
+    <div class="prog-layout">
+        The article, or list of them
```
benolot: Content block to be done later?

src/Controller/Articles/IndexController.php
```diff
@@ -0,0 +1,29 @@
...
+use BBC\ProgrammesPagesService\Domain\Entity\Programme;
+
+class IndexController extends BaseController
+{
+    public function __invoke(CoreEntity $coreEntity, IsiteService $isiteService)
+    {
+        $this->setContextAndPreloadBranding($coreEntity);
+
+        $articles = [];
+        $parameters = ['coreEntity' => $coreEntity, 'articles' => $articles, 'paginatorPresenter' => null];
```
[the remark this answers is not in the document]
votemike: Nah, it's just the skeletons for these 2 pages. There is no article data in the page yet, so nothing to paginate.

templates/articles/partials/article_list.html.twig
votemike: It will be. Yes.

templates/articles/show.html.twig
votemike: It will be. Yes.

PR #710, opened 2018-08-21
PROGRAMMES-6487 Fixing Clip CTA bug

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6487?focusedCommentId=3989142&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-3989142

Might be better to look at using https://github.com/bbc/programmes-frontend/pull/710/files?w=1


programmes-developers:

src/Ds2013/Presenters/Domain/CoreEntity/Programme/SubPresenters/programme_overlay.html.twig
```diff
@@ -28,37 +28,28 @@
...
+                            <a href="{{ programme_overlay.getPlaybackUrl() }}"
+                               class="programme__availability programme__player{{ programme_overlay.programmeIsAudio() ? '' : ' iplayer-icon--container' }}"
+                               title="{{ programme_overlay.getAvailabilityInWords() }}"
+                               data-linktrack="{{ programme_overlay.getOption('cta_link_location_track') }}"
+                               tabindex="-1"
+                            >
+                                <div class="{{ programme_icon_classes }}">
+                                    {{ gelicon('audio-visual', programme_overlay.getMediaIconName(), 'gelicon--half-container') }}
+                                </div>
+                                <meta property="description" content="{{ programme_overlay.getProgramme().getShortSynopsis() }}" />
```
votemike: Do we need these meta tags any more?

src/Ds2013/Presenters/Domain/CoreEntity/Programme/SubPresenters/programme_overlay.html.twig
```diff
@@ -28,37 +28,28 @@
                             'sounds-icon': programme_overlay.programmeIsAudio(),
                             'programme__icon--orphaned': programme_overlay.getOption('show_standalone_cta'),
                         }) %}
-                        {% if programme_overlay.isEpisode() %}
```
votemike: Is it OK to remove this? Will it knacker video clips? Should I instead change the check to be if episode OR audio?

src/Ds2013/Presenters/Domain/CoreEntity/Programme/SubPresenters/programme_overlay.html.twig
BjornTwachtmann: I created another PR to kill them everywhere. I think they're just dead code at this point

src/Ds2013/Presenters/Domain/CoreEntity/Programme/SubPresenters/programme_overlay.html.twig
BjornTwachtmann: I don't think this is the right fix for this really. The issue in the ticket seems more like a CSS bug. Ultimately, the isEpisode() check isn't 100% correct anymore _either_ though. What this is trying to do is avoid outputting a superflous link when both the main link and the CTA link go to the same place. Clips always used to go to /programmes, so hence that check.

We probably should have need a new check called hasSeparatePlayoutLink or something that returns true on iPlayer and Playspace ProgrammeItems. 

Ultimately, once you've fixed that, the CSS bug will still be there, I imagine something is targeting the <a> incorrectly, so that will need to be fixed in either case.

votemike: The ticket is _sort_ of a CSS bug. But part of the problem is the overly complex logic of the markup. This *does* solve the problem.
If what you say is the case, I think that should be solved and then the CSS adjusted to not rely on a link being there (or whatever elements and classes should be there once the markup is un-screwed).

BjornTwachtmann: This does solve the problem, it's just that it inserts a duplicate link to the same destination when it does it. It's not the end of the world, just doesn't seem ideal.

votemike: Can we re-jig the HTML to be "better" and _then_ fix the CSS?

votemike: Do you prefer https://github.com/bbc/programmes-frontend/pull/713 ?

PR #711, opened 2018-08-21
Remove remaining schema.org markup from presenters

BjornTwachtmann:
I don't think any of this markup serves any purpose any more? 

It looks like the old schema.org stuff. Anyone think this is still doing anything?


programmes-developers:

src/Ds2013/Presenters/Domain/Broadcast/broadcast.html.twig
```diff
@@ -28,10 +28,6 @@
             </h3>
 
             {#end time and zone#}
-            <meta property="endDate" datatype="xsd:dateTime" content="{{ broadcast.getEndAt()|date('c')}}" />
-            <span>
-                <meta property="name" content="{{ broadcast.getServiceName() }}" />
```
votemike: Is getServiceName() used any more?

src/Ds2013/Presenters/Domain/CoreEntity/Programme/SubPresenters/programme_overlay.html.twig
```diff
@@ -9,7 +9,6 @@
 
     <div class="{{ outer_div_classes }}">
         {% if programme_overlay.getOption('show_image') %}
-            <meta property="image" content="{{ programme_overlay.getImageUrl(480,270)}}" />
```
votemike: Is getImageUrl() used any more?

src/Ds2013/Presenters/Domain/Broadcast/broadcast.html.twig
BjornTwachtmann: nope,removed

src/Ds2013/Presenters/Domain/CoreEntity/Programme/SubPresenters/programme_overlay.html.twig
BjornTwachtmann: ditto

PR #714, opened 2018-08-28
Decrease default API ttl

Tasiobg:
Default API TTL is too high, would be safer 5 minutes for standard TTL and 1 minute for not found TTL.

@BjornTwachtmann Do you know if I should also update any of the APIs that are not using this default values?


programmes-developers:

BjornTwachtmann:
@Tasiobg . If you're updating the defaults, which I think is sensible, you may want to update some of the things using them to use the _old_ defaults, because a lot of these things can be cached for quite a while:
ElectronService can use the old defaults
RecEngService can use the old defaults
TupacService (IMO, but you would know better than me) can cache for longer as new recordIds means a new query and its data will rarely change.

I would reduce Cache lifetimes to _new_ default values for:
AdaClassService
AdaProgrammesService

let me know if that makes sense

Tasiobg:
@BjornTwachtmann updated.
For what I can see the second TTL (notFoundTTL) is used when the API respond a 404, this shoudn't happen frequently for most API but if if does, I think a shorter TTL would make sense here, what do you think? should I update all API to use `CacheInterface::SHORT` or do you think is fine now?

Tasiobg:
I have updated pages-serivice as well in this PR

Tasiobg:
I see, well in this case I have changed it only for TUPAC which shoudn't return 404s

Tasiobg:
@BjornTwachtmann I only runned ```composer update "bbc/programmes-pages-service"``` and I found this a bit strange but I thought maybe it was required as well.
I reverted and run the same command again and is only updating pages service so I don't really know how this happen before

PR #715, opened 2018-08-28
PROGRAMMES-6513

Tasiobg:
Add required parameter otherwise it fails because missing arguments.
Parameter type is `?CollapsedBroadcast $firstBroadcast` but Clips don't have CollapsedBroadcast so I set it to null


programmes-developers:

templates/find_by_pid/clip.html.twig
```diff
@@ -72,7 +72,7 @@
             {% endif %}
 
             {% if segmentEvents %}
-                {{ ds2013('segmentsList', programme, segmentEvents) }}
+                {{ ds2013('segmentsList', programme, segmentEvents, null) }}
```
BjornTwachtmann: Would it be better to make the argument optional in the factory?

Tasiobg: yep, passing null looks pointless

PR #716, opened 2018-08-28
PROGRAMMES-6591 Article Page

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6591

Updating iSite calls to be able to deal with both Profiles AND Articles.


config/services.yaml
```diff
@@ -289,10 +288,15 @@ services:
             arguments:
                 $url: '%app.favourites.baseurl%'
 
-    App\ExternalApi\Isite\Service\IsiteService:
-        class: '%app.isite.service.class%'
+    App\ExternalApi\Isite\Service\ArticleService:
+        factory: ['@App\ExternalApi\Isite\Service\ServiceFactory', 'getArticleService']
```
BjornTwachtmann: I would try and handle this differently. I _think_ you shouldn't need a factory. 

You need to specify the 2 different IsiteFeedResponseHandler as named services (i.e. 
```yaml
isite_feed_response_handler_article:
    ....
isite_feed_response_handler_profile:
   ....
```
The specify for articles/profiles service in here that it should use one of those named services in the constructor. I think that should work.

src/Controller/Articles/ShowController.php
```diff
@@ -22,7 +25,42 @@ public function __invoke(string $key, string $slug, Request $request, IsiteServi
...
+
+        /** @var Article $article */
+        $article = reset($articles);
+
+        if ($slug !== $article->getSlug()) {
+            return $this->redirectWith($article->getKey(), $article->getSlug(), $preview);
+        }
+
+        $context = null;
+        if ('' !== $article->getParentPid()) {
```
BjornTwachtmann: !empty ?

src/ExternalApi/Isite/Domain/Article.php
```diff
@@ -0,0 +1,155 @@
+<?php
+declare(strict_types = 1);
+
+namespace App\ExternalApi\Isite\Domain;
+
+use App\ExternalApi\Isite\DataNotFetchedException;
+
+class Article
```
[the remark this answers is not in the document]
BjornTwachtmann: If there's enough common code to make it worth it. Otherwise I doubt they'll be used interchangably.

src/ExternalApi/Isite/Domain/Article.php
```diff
@@ -0,0 +1,155 @@
...
+    private $brandingId;
+
+    /** @var Article[] */
+    private $parents;
+
+    /** @var string */
+    private $image;
+
+    /** @var array */
+    private $rows;
```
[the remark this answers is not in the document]
BjornTwachtmann: I think an object with primary and secondary fields probably makes sense

src/Controller/Articles/ShowController.php
```diff
@@ -22,7 +25,42 @@ public function __invoke(string $key, string $slug, Request $request, IsiteServi
...
+        /** @var Article $article */
+        $article = reset($articles);
+
+        if ($slug !== $article->getSlug()) {
+            return $this->redirectWith($article->getKey(), $article->getSlug(), $preview);
+        }
+
+        $context = null;
+        if (!empty($article->getParentPid())) {
+            $context = $coreEntitiesService->findByPidFull(new Pid($article->getParentPid()));
```
tatitati: should getParentPid return a Pid?, I feel might be error prone if you call getParentPid() and you get an string.

src/ExternalApi/Isite/Domain/Article.php
```diff
@@ -0,0 +1,155 @@
...
+    {
+        return $this->title;
+    }
+
+    public function getProjectSpace(): string
+    {
+        return $this->projectSpace;
+    }
+
+    public function getBrandingId()
```
benolot: Should this be typehinted? Won't a branding ID always be a string?

src/ExternalApi/Isite/Mapper/RowMapper.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        //check if module is in the data
+        if (empty($isiteObject->{$type})) {
+            return [];
+        }
+
+        $blocks = $isiteObject->{$type};
+        $name = $type . '-blocks';
+        if (empty($blocks[0]->{$name}->result)) {
+            return null; // Content blocks have not been fetched
```
benolot: Do we want to throw an error rather than return null?

votemike: @BjornTwachtmann any view?

src/ExternalApi/Isite/Domain/Row.php
```diff
@@ -0,0 +1,31 @@
...
+
+class Row
+{
+    /** @var null|AbstractContentBlock[] */
+    private $primary;
+
+    /** @var null|AbstractContentBlock[] */
+    private $secondary;
+
+    public function __construct(?array $primary, ?array $secondary)
```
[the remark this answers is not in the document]
votemike: We could do. I don't think there would be any difference. Shall I make that change?

src/Controller/Articles/ShowController.php
```diff
@@ -5,13 +5,16 @@
 
 use App\Controller\BaseController;
 use App\Controller\Helpers\IsiteKeyHelper;
-use App\ExternalApi\Isite\Service\IsiteService;
+use App\ExternalApi\Isite\Domain\Article;
+use App\ExternalApi\Isite\IsiteResult;
+use App\ExternalApi\Isite\Service\ArticleService;
+use BBC\ProgrammesPagesService\Domain\ValueObject\Pid;
```
Tasiobg: I think you don't need this line in showController

src/ExternalApi/Isite/Mapper/ArticleMapper.php
```diff
@@ -0,0 +1,46 @@
+<?php
+declare(strict_types = 1);
+
+namespace App\ExternalApi\Isite\Mapper;
+
+use App\ExternalApi\Exception\ParseException;
```
Tasiobg: ParseException is not used

src/ExternalApi/Isite/Mapper/ProfileMapper.php
```diff
@@ -73,21 +73,4 @@ public function getDomainModel(SimpleXMLElement $isiteObject): Profile
...
     }
-
-    protected function getProjectSpace(SimpleXMLElement $form): string
-    {
-        $namespaces = $form->getNamespaces();
-        $namespace = reset($namespaces);
-        $matches = [];
-        preg_match('{https://production(?:\.int|\.test|\.stage|\.live)?\.bbc\.co\.uk/isite2/project/([^/]+)/}', $namespace, $matches);
-        if (empty($matches[1])) {
-            throw new ParseException('iSite XML does not specify project space and is therefore invalid');
```
Tasiobg: now you can remove 
```use App\ExternalApi\Exception\ParseException;```
from this class

File: templates/articles/show.html.twig [hunk 12]
```diff
@@ -5,6 +5,59 @@
...
+                                {% if row.getPrimaryBlocks() %}
+                                    {# If we're the first row then show the article title#}
+                                    {% if loop.first %}
+                                        <div class="prog-layout__primary island--vertical">
+                                            <h1>{{ article.getTitle() }}</h1>
+                                        </div>
+                                    {% endif %}
+
+                                    {% for block in row.getPrimaryBlocks() %}
+                                        {% if block is not null %} {# Remove this line once all content block types are popualted #}
```
Tasiobg: I think there is a typo in the word "popualted", same in the next comment

Tasiobg:
Not sure if is an issue but when going to `/programmes/profiles/49sPLJnsQBd0DCh0Vc11cGy/families-and-groups` now I can see all the "Past characters" listed in this page while in live it is only displaying a link to the profile page
screen shot 2018-08-29 at 16 27 20
Live
screen shot 2018-08-29 at 16 27 30

src/ExternalApi/Isite/Service/ProfileService.php
```diff
@@ -0,0 +1,133 @@
...
+        $cacheKey = $this->clientFactory->keyHelper(__CLASS__, __FUNCTION__, implode(',', $cacheKeys), $page, $limit);
+
+        $client = $this->clientFactory->getHttpApiMultiClient(
+            $cacheKey,
+            $urls,
+            Closure::fromCallable([$this, 'parseResponses']),
+            [$profiles],
+            [],
+            CacheInterface::NORMAL,
+            CacheInterface::NONE,
```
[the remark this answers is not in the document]
votemike: The iSite stuff loads the majority of the page content. I think we should keep retrying until iSite responds, otherwise we risk taking down the page for a whole 60 seconds. Whereas if this were just a small portion, it'd be fine.
@BjornTwachtmann any views?

src/ExternalApi/Isite/Mapper/ArticleMapper.php
```diff
@@ -0,0 +1,46 @@
...
+
+class ArticleMapper extends Mapper
+{
+    public function getDomainModel(SimpleXMLElement $isiteObject): Article
+    {
+        $form = $this->getForm($isiteObject);
+        $formMetaData = $this->getFormMetaData($isiteObject);
+        $projectSpace = $this->getProjectSpace($formMetaData);
+        $key = $this->isiteKeyHelper->convertGuidToKey($this->getString($this->getMetaData($isiteObject)->guid));
+        $title = $this->getString($formMetaData->title);
```
Tasiobg: I checked this article page
```/programmes/articles/5H6SWP4lplW1hyrHrWy4THJ/2015-live-week-crew-list```
It seems that we display the same title but nothing else inside this content block, I checked the API response and the content is inside `<prose>` so I think we should be mapping this too but not too familiar with iSite to know which fields should be mapped

File: templates/articles/show.html.twig [hunk 15]
```diff
@@ -5,6 +5,59 @@
...
+                                        {% endif %}
+                                    {% endfor %}
+                                {% endif %}
+                            </div>
+                        </div>
+                        {% if row.getSecondaryBlocks() %}
+                            <div class="grid 5/12@bpw2 5/12@bpe">
+                                {% for block in row.getSecondaryBlocks() %}
+                                    {% if block is not null %} {# Remove this line once all content block types are popualted #}
+                                        {{ ds2013('contentBlock', block) }}
```
Tasiobg: This content block seems quite different from what is generated on v2 example: ```/programmes/articles/5H6SWP4lplW1hyrHrWy4THJ/2015-live-week-crew-list```

src/ExternalApi/Isite/Domain/Row.php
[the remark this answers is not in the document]
BjornTwachtmann: might as well, seems simpler

src/ExternalApi/Isite/Mapper/RowMapper.php
BjornTwachtmann: Probably make it throw if this shouldn't happen

src/ExternalApi/Isite/Service/ProfileService.php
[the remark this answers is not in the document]
BjornTwachtmann: I'd check if having no child profiles means 404, or if it means 200 with no content or something. If a 404 legitimately indicates no child profiles, that seems worth caching for at least 1 minute or something

votemike:
@Tasiobg the families-and-groups problem seems to be on develop too. Maybe raise a ticket to have a look in to it if it hasn't been tested already? Or dig out the ticket and put a comment on it.

src/ExternalApi/Isite/Mapper/ProfileMapper.php
votemike: Can I? It looks used to me

src/ExternalApi/Isite/Mapper/ArticleMapper.php
votemike: We're not mapping all the content block types yet. There are tickets in this sprint to do that.

src/ExternalApi/Isite/Mapper/ProfileMapper.php
votemike: I think I was looking in the wrong file. I've removed it now.

src/ExternalApi/Isite/Mapper/ArticleMapper.php
votemike: It looks like it's just because those blocks haven't been created yet. There are tickets in this sprint to do that.

src/ExternalApi/Isite/Service/ProfileService.php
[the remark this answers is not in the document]
BjornTwachtmann: fair enough then

PR #717, opened 2018-08-30
PROGRAMMES-6598 CTA Presenter

votemike:
This PR creates a DS2013 Presenter for the CTA.
In the future, hopefully it will encompass the Amen CTAs and the MAP Playout CTA, but for now, it only covers the little square CTAs (and the Watch/Listen Live with those 3 pulsing bars.)

The design is white on black (black on white on hover) for everything except iPlayer. So there's no longer any colours from branding in these CTAs, nor are there any transparent backgrounds or anything. These are my choices, not UX or products. Do you agree that these all look OK? Once the team is happy, I'll ask UX. Then we'll hope product doesn't complain.

Please review the CTAs and site in general, but also the Styleguide page and make sure it contains all the information you'd like it to contain.

In order to test, checkout the branch, run `yarn run gulp` then visit http://sandbox.bbc.co.uk/programmes/styleguide/ds2013/utilities/cta or any of the 2013 pages.

I don't dare attempt to replicate these CTAs in V2. Is that OK?

Code TODOs
- [x] Ensure all old CSS is removed
- [x] Ensure all old PHP is removed
- [x] Decide if the template for the Live CTA should extend the CTA template.
- [x] Once the above is done, write some tests

Signoff TODOs
- [x] Get the team to accept the design
- [x] Put branch on Stage
- [x] Get Cathal to accept the design
- [x] Get UX to accept the design
- [ ] Get testers to test
- [ ] Merge to develop


programmes-developers:

src/Builders/ClipBuilder.php
```diff
@@ -53,7 +53,7 @@ protected function __construct()
     public static function anyRadioClip()
     {
         return self::any()->with([
-            'masterBrand' => MasterBrandBuilder::anyRadioMasterBrand()->build(),
+            'masterBrand' => MasterBrandBuilder::anyRadioMasterBrand()->build(), //should this set media type?
```
votemike: If we're building a Radio Clip, should we set the media type as audio?

src/Builders/EpisodeBuilder.php
```diff
@@ -57,7 +57,7 @@ public static function anyRadioEpisode()
         $self = new self();
 
         return $self->with([
-            'masterBrand' => MasterBrandBuilder::anyRadioMasterBrand()->build(),
+            'masterBrand' => MasterBrandBuilder::anyRadioMasterBrand()->build(), // should this set media type and network too?
```
votemike: If we're building a Radio Episode, should we set the media type as audio?

src/Controller/Styleguide/Ds2013/Utilities/CtaController.php
```diff
@@ -0,0 +1,75 @@
...
+class CtaController extends BaseController
+{
+    public function __invoke(Request $request, CoreEntitiesService $coreEntitiesService)
+    {
+        if ($request->query->has('branding_context')) {
+            $coreEntity = $coreEntitiesService->findByPidFull(new Pid($request->query->get('branding_context')));
+            $this->setContextAndPreloadBranding($coreEntity);
+        }
+        // @TODO
+        // Variation of CTAs. E.g. no mediaType, no Network
```
votemike: Do we also want examples of CTAs where the CoreEntity is missing media types and Networks?

src/Ds2013/Presenters/Utilities/Cta/CtaPresenter.php
```diff
@@ -0,0 +1,152 @@
...
+        return 'media';
+    }
+
+    public function getUrl()
+    {
+        if (!$this->getOption('include_link')) {
+            throw new Exception('You said you didn\'t want a link, why are you calling this method?');
+        }
+        if (!$this->coreEntity instanceof ProgrammeItem) {
+            return 'TODO'; // Is this even needed? Should I throw an exception?
```
votemike: Should I throw an Exception here?

src/Ds2013/Presenters/Utilities/Cta/CtaPresenter.php
```diff
@@ -0,0 +1,152 @@
...
+
+        return $this->router->generate($routeName, $routeArguments, UrlGeneratorInterface::ABSOLUTE_URL);
+    }
+
+    public function programmeIsAudio(): bool
+    {
+        return $this->streamUrlHelper->shouldTreatProgrammeItemAsAudio($this->coreEntity);
+    }
+
+    protected function validateOptions(array $options): void
```
votemike: Do I need any extra validation other than what's currently here?

src/Ds2013/Presenters/Utilities/Cta/live_cta.html.twig
```diff
@@ -0,0 +1,24 @@
+{% import _self as self %}
```
votemike: Should this template extend the normal CTA template?

templates/styleguide/ds2013/Utilities/cta.html.twig
```diff
@@ -0,0 +1,267 @@
...
+            {{ ds2013('broadcastProgramme', radioCollapsedBroadcast, null, { 'body_options': {'show_overlay': true}}) }}
+        </div>
+    </div>
+    <h3>Programmes</h3>
+    <h4>Link</h4>
+    <div class="grid-wrapper">
+        {% for blurb, entity in programmes %}
+            <div class="grid 1/2 1/4@bpw">
+                <h5>{{ blurb }}</h5>
+                <div style="overflow: hidden;"> {# TODO what shall we do about this? #}
```
votemike: It seems CTA's rely on their containing element having this. Are we OK just having this style outside the CTA concept?

src/Ds2013/Presenters/Utilities/Cta/live_cta.html.twig
```diff
@@ -0,0 +1,24 @@
...
+    <a href="{{ live_cta.getUrl() }}"
+       class=""
+       title="{{ live_cta.getAvailabilityInWords() }}"
+       data-linktrack=""
+       tabindex="-1"
+    >
+        <div class="cta__icon {{ live_cta.getMysteryClasses() }} cta__with_link">
+            {{ gelicon(live_cta.getWhateverThisIs(), live_cta.getMediaIconName(), 'gelicon--half-container') }}
+        </div>
+        <div class="programme__overlay-data br-box-page br-page-bg-onbg080">
```
votemike: Do we need all these classes?

src/Builders/ClipBuilder.php
votemike: Although it is possible to have radio clips without media types

src/Ds2013/Presenters/Utilities/Cta/cta.html.twig
```diff
@@ -0,0 +1,22 @@
+{% import _self as self %}
+
+<div class="cta{{ cta.getOption('is_overlay') ? ' cta__overlay' : ' cta__standalone' }}">
+    {% if cta.getOption('include_link') %}
+        <a href="{{ cta.getUrl() }}"
+           class=""
+           title="{{ cta.getAvailabilityInWords() }}"
+           data-linktrack=""
```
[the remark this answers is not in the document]
Tasiobg: This is for iStats tracking, probably can be removed becasue there is a new traking system comming soon and I think is not compatible at all with the previous one. Feel free to ask Claudio for mode details

Tasiobg:
Looks like a good idea and code looks fine but I haven't tested or in detail understand how everythings works, I would need to spend a fair amount of time testing and learning more in detail if required.

src/Builders/EpisodeBuilder.php
tatitati: Is something I thought when designing this builder. In the end I came with the idea of being strict to the definition of a radio Entity, which doesn't have in account (for my surprise) anything about the episode, but instead only check the network/masterbrand. 
https://github.com/bbc/programmes-pages-service/blob/master/src/Domain/Entity/CoreEntity.php#L211-L231

src/Builders/ClipBuilder.php
BjornTwachtmann: It is possible, but it seems sensible to test against the usual media type

src/Controller/Styleguide/Ds2013/Utilities/CtaController.php
BjornTwachtmann: Would probably prevent a few edge case related bugs

src/Controller/Styleguide/Ds2013/Utilities/CtaController.php
```diff
@@ -0,0 +1,75 @@
...
+        $groups = [];
+        $programmes = [];
+        $programmes['TV Clip'] = ClipBuilder::anyTvClip()->with(['mediaType' => MediaTypeEnum::VIDEO])->build();
+        $programmes['Radio Clip'] = ClipBuilder::anyRadioClip()->with(['mediaType' => MediaTypeEnum::AUDIO])->build();
+        $programmes['TV Episode'] = EpisodeBuilder::anyTVEpisode()->with(['mediaType' => MediaTypeEnum::VIDEO])->build();
+        $programmes['Radio Episode'] = EpisodeBuilder::anyRadioEpisode()->with(['mediaType' => MediaTypeEnum::AUDIO])->build();
+        $groups['Gallery'] = GalleryBuilder::any()->build();
+        $groups['Collection'] = CollectionBuilder::any()->build();
+
+        $tvClip = $coreEntitiesService->findByPidFull(new Pid('p057d91w'));
```
BjornTwachtmann: What happens when these programme items change/become unavailable in the database. Clips and Radio things have pretty long availabilities, but the TV stuff changes pretty quicly.

src/Ds2013/Presenters/Utilities/Cta/_cta.scss
```diff
@@ -0,0 +1,104 @@
...
+
+            &.sounds-icon {
+                color: #ff4900;
+                background: #ffffff;
+            }
+        }
+    }
+}
+
+@include mq-range('bpb2') {
```
BjornTwachtmann: This size applies from 480 -> 769, is it worth combining with the block below?

src/Ds2013/Presenters/Utilities/Cta/live_cta.html.twig
BjornTwachtmann: That's what the Amen CTA presenters do, which are structured quite similarly to this, although the Amen stuff has a separate CTA presenter for Media Icons (there are 3 Amen CTA presenters, Streamable, Live and MediaIcon). Dunno if you think that structure would work here or simplify anything, just pointing out that it might be worth looking at.

src/Ds2013/PresenterFactory.php
```diff
@@ -116,6 +119,22 @@ public function calendarPresenter(
         );
     }
 
+    public function liveCtaPresenter(
```
BjornTwachtmann: So, for the sub-presenters generally, we don't make them available through the factory as they don't make sense without their context. 

I would say these should be instantiated in the presenters making them, and not available generically through the factory, but I dunno, what do you think?

src/Ds2013/Presenters/Domain/CoreEntity/Programme/_programme.scss
```diff
@@ -274,23 +206,13 @@ $programme-icon-bpw-box-size: gel-typography('bpw', 'delta', 'line-height') + 18
         visibility: visible;
     }
 
-    .programme__icon,
     .programme__img--standalone-cta:not(.programme__img--available) {
```
BjornTwachtmann: This line looks CTA-ish and could probably be killed

BjornTwachtmann: Or moved, or whatever.

src/Ds2013/Presenters/Utilities/Cta/CtaPresenter.php
```diff
@@ -0,0 +1,152 @@
...
+        // Incase we want Sounds branding e.g. sounds-icon in the future
+        if ($this->programmeIsAudio()) {
+            return '';
+//            return 'sounds-icon';
+        }
+
+        return $this->coreEntity instanceof Episode ? 'iplayer-icon' : '';
+    }
+
+    public function getWhateverThisIs()
```
BjornTwachtmann: Wasn't the aim of this refactor to get rid of the mystery crap? Any idea what this does?

src/Controller/Styleguide/Ds2013/Utilities/CtaController.php
votemike: I was wondering that. Should we query for these things instead?

src/Ds2013/Presenters/Utilities/Cta/CtaPresenter.php
votemike: I thought I had changed the name of them.

src/Builders/ClipBuilder.php
votemike: OK, I'll just leave it as it is then.

src/Builders/EpisodeBuilder.php
votemike: Maybe it's best to leave it as it is?

src/Controller/Styleguide/Ds2013/Utilities/CtaController.php
votemike: I thought I had written a comment about querying them. Should we instead query for these cases?

src/Ds2013/Presenters/Utilities/Cta/_cta.scss
votemike: In what way?

src/Controller/Styleguide/Ds2013/Utilities/CtaController.php
votemike: Done

src/Builders/EpisodeBuilder.php
tatitati: For me at least is fine as it is, but I don't have an strong opinion because if you introduce that change, still you will get a radio episode, which is fine, is not gonna change.

src/Controller/Styleguide/Ds2013/Utilities/CtaInSituController.php
```diff
@@ -0,0 +1,62 @@
...
+
+class CtaInSituController extends BaseController
+{
+    public function __invoke(Request $request, CoreEntitiesService $coreEntitiesService)
+    {
+        if ($request->query->has('branding_context')) {
+            $coreEntity = $coreEntitiesService->findByPidFull(new Pid($request->query->get('branding_context')));
+            $this->setContextAndPreloadBranding($coreEntity);
+        }
+        // @TODO Should we instead load examples from queries here instead of hardcoding PIDs?
```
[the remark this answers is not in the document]
votemike: The hardcoded PIDs may end their availability.

src/Ds2013/PresenterFactory.php
votemike: If I get rid of this, I'll have to pass the helperfactory and router in to all the Presenters or Controllers that need to make a LiveCtaPresenter. Is that preferred?

BjornTwachtmann: That kind of thing is already done extensively. It's not ideal, but I really don't like the idea of cluttering up this class with presenters you can't actually use.

src/Ds2013/Presenters/Domain/CoreEntity/Programme/CollapsedBroadcastSubPresenters/collapsed_broadcast_programme_overlay.html.twig
```diff
@@ -2,24 +2,7 @@
...
-                    <div class="timebar">
-                        <span class="timebar-barbox timebar-barbox--live"></span>
-                        <span class="timebar-barbox timebar-barbox--live"></span>
-                        <span class="timebar-barbox timebar-barbox--live"></span>
-                    </div>
-                </div>
-            </a>
-        </div>
-        {% endspaceless %}
+        {{ ds2013('LiveCta', programme_overlay.getCollapsedBroadcast(), programme_overlay.getOption('context_service')) }}
```
BjornTwachtmann: Rather than passing all this stuff into the view and back out again, the DsAmen stuff just instantiates the subpresenters directly in the main presenter and would then use ds2013_presenter(...) here, e.g. https://github.com/bbc/programmes-frontend/blob/develop/src/DsAmen/Presenters/Domain/CoreEntity/Programme/programme.html.twig#L14

tatitati:
I tested it and work. It would be good to have someone else testing it, as it is a big PR, but for seems ok. I don't approve it for now as he specified so.

votemike:
Test this please

PR #718, opened 2018-08-31
PROGRAMMES-6505 Table Content Block

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6505


programmes-developers:

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
```diff
@@ -49,6 +50,57 @@ public function getDomainModel(SimpleXMLElement $isiteObject)
                     $links
                 );
                 break;
+            case 'table':
```
votemike: What would be the best way to test this?

src/Ds2013/Presenters/Domain/ContentBlock/Table/table.html.twig
```diff
@@ -0,0 +1,25 @@
+<div class="component component--box{{ table.isInPrimaryColumn() ? ' component--box-flushbody-vertical component--box--primary' : ' component--box--striped component--box--secondary' }}">
```
BjornTwachtmann: How common is this logic across the different content block types. Is it, theoretically, worth creating a BaseContentBlockPresenter that does this logic everywhere, and just has getXClasses() methods that its children can call?

votemike: Not sure, in V2 it's all hidden in that component renderer. I have a feeling that the 2 blocks that have already been done for the profile page don't take in to account the secondary column. Maybe the person who picks up the next ticket can look in to genericising it.

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
```diff
@@ -6,6 +6,7 @@
 use App\ExternalApi\Isite\Domain\ContentBlock\AbstractContentBlock;
```
Tasiobg: I think you can remove
use App\ExternalApi\Isite\Domain\ContentBlock\AbstractContentBlock;
use Exception;

votemike: Both of those classes will be used once all blocks have been implemented.

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
votemike: Done

PR #724, opened 2018-09-10
[Programmes-6600][fix] - Improved how we handle cookie information when parsing

tatitati:


programmes-developers:

tatitati:
I tried deleting the cookie with:
```
window.bbccookies.delete(cookieName);
window.bbccookies.del(cookieName);
window.bbccookies.remove(cookieName);
window.bbccookies.removeData(cookieName);
```
None of them deleted it. So I decided just reset the content to empty (is the same value that we set when the video is finished). SMP player is designed to understand this value as a mean to start the reproduction from the start.

assets/js/smp/cookieStore.js
```diff
@@ -10,12 +10,12 @@ define(function () {
         this.findByName = function (name) {
             var all = findAll();
             var cookie = all[name];
-
-            if (cookie !== undefined) {
```
BjornTwachtmann: Why remove this check?

PR #725, opened 2018-09-10
[PROGRAMMES-6572] - Improved accessibility-audio-messages for download button

tatitati:
# What
The message "High/low quiality 128kbps" was little descriptive for the user listening.

## Example urls:
Episode page with multiple downloads: 
http://sandbox.bbc.co.uk/programmes/b007dpz8?__flush_cache=1

Episode page with one download:
http://sandbox.bbc.co.uk/programmes/b007519z?__flush_cache=1

----
Same fix for V2 here: https://github.com/bbc/programmes/pull/890


src/Ds2013/Presenters/Utilities/Download/download.html.twig
```diff
@@ -33,7 +33,7 @@
                         {% for translationKey,url in download.getPodcastUrls() %}
                             <li>
                                 <a class="link-complex popup__list__item island--squashed br-secondary-bg-ontext br-secondary-bg-onbg--hover br-secondary-link-ontext--hover" href="{{ url }}"
-                                   download="{{ download.getPodcastFileName() }}">
+                                   download="{{ download.getPodcastFileName() }}" aria-label="{{tr('download') }} {{ tr(translationKey)}} ">
```
BjornTwachtmann: There is no translation for "download" in the file. There's "downloads", which you'd have to call with a number. But no "download". The default behaviour of the translation layer is to spit the supplied string back at you if it doesn't recognise it, hence what you see on the page. 

Change
`{{tr('download') }}` to `{{ tr('downloads', 1) }}` as is used elsewhere in this very file.

PR #726, opened 2018-09-11
[PROGRAMMES-6615][fix] - Fixed teXt in related programmes  

tatitati:
## What
When showing the list of related programmes, we were saying always "This episode is related to <topic>". However, the programme displayed in the list sometimes is not an episode and can be a clip as well. 

## The fix
Currently we have only one translation in our *.po files. This is for ```episode_is_related_to```, so we don't have anything else like 'programme_is_related_to', or 'clip_is_related_to'. Nothing similar. 
As discussed with @claudioneves we will try to add translations in the future for "clip_is_related_to", but for now we are not too worried about them. So only for "clip is related to" we won't have any translation (always in english)

## Example URLs
Episode page with related programmes: http://sandbox.bbc.co.uk/programmes/b03kpdhg?__flush_cache=1

Clip page with related programmes: http://sandbox.bbc.co.uk/programmes/p01djbmh?__flush_cache=1


programmes-developers:

BjornTwachtmann:
@tatitati Text like this belongs in the translation file, not hardcoded into the template. Add a new translation for clip_is_related_to to the translation files, populate it for English, and let Product know that it will need to be populated for other languages.

templates/find_by_pid/clip.html.twig
```diff
@@ -106,7 +106,7 @@
                                     {{ ds2013('programme', programme.getProgramme(), {
                                         'title_options': {'title_format': 'item::ancestry', 'link_location_track': 'clip_suggested_item_' ~ loop.index },
                                         'body_options': {
-                                            'body_suffix': '<p class="programme__child-availability">' ~ tr('episode_is_related_to', {'%1': '<a class="block-link__link" href="' ~ path('topics_topic', {topic: programme.getRelatedByClasses()[0].getId()}) ~'" data-linktrack="clip_suggested_relationship">' ~ programme.getRelatedByClasses()[0].getTitle() ~ '</a></p>'}),
+                                            'body_suffix': '<p class="programme__child-availability">' ~ tr( programme.getProgramme().getType() ~ '_is_related_to', {'%1': '<a class="block-link__link" href="' ~ path('topics_topic', {topic: programme.getRelatedByClasses()[0].getId()}) ~'" data-linktrack="clip_suggested_relationship">' ~ programme.getRelatedByClasses()[0].getTitle() ~ '</a></p>'}),
```
Tasiobg: I think the type is not dynamic, I suggest to use a constant string 'clip_is_related_to'
Similar change can be done for episode.html.twig

PR #728, opened 2018-09-11
[PROGRAMMES-5892] Added break-word to titles and subtitles

benolot:
Ticket link: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5892

Not sure if this will break in other places, but it seems reasonable to apply the same overflow fix to those places.


programmes-developers:

src/Ds2013/Presenters/Domain/CoreEntity/Programme/_programme.scss
```diff
@@ -61,6 +61,7 @@ $programme-icon-bpw-box-size: gel-typography('bpw', 'delta', 'line-height') + 18
     display: block;
     margin-bottom: 0;
     word-wrap: break-word;
+    hyphens: auto;
```
tatitati: as far as I remember we decided to have broken words without hyphens because visually is more pleasant. The second reason for this is that the frecuency of "broken words" when adding hyphen increase in browsers, so is easier to see hyphens every where.

benolot:
Too many issues - need to rethink this approach

PR #729, opened 2018-09-11
PROGRAMMES-6508 Populate Promotion content block

Tasiobg:


programmes-developers:

src/Ds2013/Presenters/Domain/ContentBlock/Promotions/promotions.html.twig
```diff
@@ -0,0 +1,19 @@
+{% extends "@Ds2013/Presenters/Domain/ContentBlock/content_block.html.twig" %}
+
+{% block title %}
+    {% if content_block.getBlock().getTitle() %}
```
BjornTwachtmann: Do we need this block? Looks like it's just duplicating what's in the base template.

src/Ds2013/Presenters/Domain/ContentBlock/Promotions/promotionsgrid.html.twig
```diff
@@ -0,0 +1,22 @@
+<ul class="grid-wrapper highlight-box-wrapper--grid@bpb1 highlight-box-wrapper--grid@bpw">
+    {% for promotion in content_block.getBlock().getPromotions() %}
+    <li class="grid 1/2@bpb1 1/2@bpw 1/3@bpe">
```
BjornTwachtmann: This may be a stupid question. But I'm going to plow ahead regardless. Why not use the promotionPresenter?

BjornTwachtmann: Anwsering my own question: This is not a PIPs promotion. Because of course it's not.

src/Ds2013/Presenters/Domain/ContentBlock/Promotions/promotionsgrid.html.twig
```diff
@@ -0,0 +1,22 @@
...
+        <div class="grid__inner">
+            <div class="promotion block-link highlight-box--list-borderless promotion--grid@bpb1 highlight-box--grid@bpb1 promotion--grid@bpw highlight-box--grid@bpw br-keyline br-blocklink-page br-page-linkhover-onbg015--hover">
+                <div class="promotion__img one-whole">
+                    {{ ds_shared('image', promotion.promotedItemId, 163, '', {'image_classes': 'rsp-img'}) }}
```
BjornTwachtmann: rsp-img is not a thing in V2, it's from V2 and should be removed. 

Also, 163 is not an image size that exists in any of our recipes (https://github.com/bbc/programme-images/tree/master/webapp/ichef/recipes ), and will probably return a 404. Where is that from?

src/Ds2013/Presenters/Domain/ContentBlock/Promotions/promotions.html.twig
Tasiobg: I may have missed, thanks

src/Ds2013/Presenters/Domain/ContentBlock/Promotions/promotionsgrid.html.twig
Tasiobg: I forgot to ask about this, I check with Francisco but we where not sure what `rsp-img` does, looking at all the other usages of `ds_shared('image'` I found all using this class so I adde it as well.
On the other hand as per the image size I wasn't sure what values are valid, so I tried to use same size as the image in test was rendering and seems working fine but I don't know where I can find the right value and I forgot to ask before opening the PR.

BjornTwachtmann: Please remove that class anywhere anywhere you see it, it's dead. 

Valid widths are in the file I linked. In this instance the image is lazy-loaded, and so the default width is not used in most browsers, lazySizes will choose the closest size from the default list of srcsets attached here https://github.com/bbc/programmes-frontend/blob/develop/src/DsShared/Utilities/Image/ImagePresenter.php#L21 . That list isn't necessarily optimal here, on the other hand cluttering up the whole codebase with a load of srcsets is not that fun either.

src/Ds2013/Presenters/Domain/ContentBlock/Promotions/promotionslist.html.twig
```diff
@@ -0,0 +1,24 @@
+<ul class="list-unstyled">
+    {% for promotion in content_block.getBlock().getPromotions() %}
+    <li>
+        <div class="promotion block-link highlight-box--list-borderless br-keyline br-blocklink-page br-page-linkhover-onbg015--hover">
+            <div class="promotion__img 1/4@bpb1  1/3@bpb2 1/3@bpw">
+                {{ ds_shared('image', promotion.promotedItemId, 192, '', {'image_classes': 'rsp-img'}) }}
```
BjornTwachtmann: rsp-img

BjornTwachtmann:
Question: Have you tried putting this in the right column of an article and seeing if it renders the same as V2?

src/ExternalApi/Isite/Domain/ContentBlock/Promotions.php
```diff
@@ -0,0 +1,34 @@
+<?php
+declare(strict_types = 1);
+
+namespace App\ExternalApi\Isite\Domain\ContentBlock;
+
+class Promotions extends AbstractContentBlock
+{
+    /**
+     * @var array
```
tatitati: can we specify the type of variables contained in this array?

tests/ExternalApi/Isite/Mapper/promotions.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tatitati: Not sure if I might be wrong, but I think that might be useful to add as an XML comment in these kind of fixtures the url used to obtain it. It provide useful context about the response and is useful for debugging porpuses

PR #584, opened 2018-06-11
PROGRAMMES-6383 Adding clip group attribution

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6326

PR #730, opened 2018-09-12
[PROGRAMMES-6624][change] - Added image to display in featured in for clips

tatitati:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6624

"Featured in" for clips was added previously in this PR: https://github.com/bbc/programmes-frontend/pull/584

However the jira ticket for that ticket (https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6383) didn't specify to show the image, so this is to fix that problem.

The default value is show_image=true as you can see in the presenter:
https://github.com/bbc/programmes-frontend/blob/develop/src/Ds2013/Presenters/Domain/CoreEntity/Group/GroupPresenter.php#L18

## Examples urls:
Clip pages:
Collection with image: http://sandbox.bbc.co.uk/programmes/p03l1bf7?__flush_cache=1
Collection with no image: http://sandbox.bbc.co.uk/programmes/p016w9yl?__flush_cache=1

Episode page (this PR doesn't modify anything in episode page, but is useful to compare the behaviour with clip pages):
http://sandbox.bbc.co.uk/programmes/p022hmc8?__flush_cache=1


programmes-developers:

PR #732, opened 2018-09-14
Make sure that speech and chapter segments get different identifying …

BjornTwachtmann:
Make sure that speech and chapter segments get different identifying classes for testers to use.

https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6536


programmes-developers:

PR #735, opened 2018-09-18
UI | Added ids in html dom to help QA to test frontend

tatitati:
For testing:

related_links:
sandbox.bbc.co.uk/programmes/p0033n7z 

related_topics:
sandbox.bbc.co.uk/programmes/p01djbmh

related_programmes
sandbox.bbc.co.uk/programmes/p01djbmh


programmes-developers:

templates/find_by_pid/clip.html.twig
```diff
@@ -48,7 +48,7 @@
     <div class="prog-layout programmes-page">
         <div class="br-box-page prog-box">
             {% if relatedTopics %}
-                <div class="component component--box--primary">
+                <div id="related_topics" class="component component--box--primary">
```
votemike: On the Episode page, these ids are on a different div. https://github.com/bbc/programmes-frontend/blob/develop/templates/find_by_pid/episode.html.twig#L23
Would it be an idea to keep it consistent?

tatitati: It makes sense, but I just checked and the html layout is lightly different, as you can see 
```<div id="related_topics" class="grid grid--bounded grid--unbounded@bpw2 grid--unbounded@bpe">``` appear in episode page, however, we don't have any div with that class in clip.

Episode.html.twig
```twig
{% if relatedTopics %}
    <div id="related_topics" class="grid grid--bounded grid--unbounded@bpw2 grid--unbounded@bpe">
        <div class="br-box-page prog-box">
            <div class="component component--box--primary">
               <div class="component__header br-box-page">
                                <h2>{{ tr('related_topics') }}</h2>
               </div>
           <div class="component__body br-box-page">
               {{ ds2013('relatedTopics', relatedTopics, 'episode_topic') }}
           </div>
     </div>
 </div>
</div>
{% endif %}
```
Clip.html.twig
```twig
            {% if relatedTopics %}
                <div class="component component--box--primary">
                    <div class="component__header br-box-page">
                        <h2>{{ tr('related_topics') }}</h2>
                    </div>
                    <div class="component__body br-box-page">
                        {{ ds2013('relatedTopics', relatedTopics, 'clip_topic') }}
                    </div>
                </div>
{% endif %}
```

votemike: I see. It seems odd that it's so messy like that. Obviously it would be better if there was parity between the two.
What do others think about making the markup more similar. Otherwise the code is going to end up being messy.
I hate to suggest it, but is it worth creating a presenter for "component"s?

BjornTwachtmann: Is the display actually different

BjornTwachtmann: https://github.com/bbc/programmes/blob/develop/webapp/php/sharedmodules/programmes/views/partials/generic/component.spv

I will just leave that to speak for itself.

tatitati: OK, our V3 html match exactly the V2 html. However, related topics html is different for clips and episode.
To avoid to block QA of testing what they need I decided to create a ticket in order to fix this subtle issue and make them to have the same HTML in both programme items. 

The ticket can be found here: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6635

PR #737, opened 2018-09-18
[PROGRAMMES-5837] Is today method refactor

tomwinnington:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5837

When iterating 1000 times, I found it saved 2 milliseconds.


programmes-developers:

src/Ds2013/Presenters/Utilities/DateList/DayDateListItemPresenter.php
```diff
@@ -32,6 +33,11 @@ public function isGmt(): bool
         return $this->options['user_timezone'] == 'GMT';
     }
 
+    public function isToday(): bool
+    {
+        return ApplicationTime::getTime()->format('Y-m-d') == $this->getDateTime()->toDateString();
```
votemike: Should be able to still use `->toDateString()` on the ApplicationTime

benolot: There's no method like that in the Application Time class, and as far as I understood, format was the best way to format a `DateTimeImmutable` object?

src/Ds2013/Presenters/Utilities/DateList/date_list.html.twig
```diff
@@ -1,5 +1,7 @@
 <ol class="nav nav--banner date-list {{ date_list.getOption('css_classes') }}">
+
```
BjornTwachtmann: ARGH WHITESPACE ARGH ARGHGHGHGH!!!!!111111ELEVEN

src/Ds2013/Presenters/Utilities/DateList/DayDateListItemPresenter.php
BjornTwachtmann: In Practice, ApplicationTime will return Chronos, not DateTimeImmutable, so what mike's saying is right, if counterintuitve

src/Ds2013/Presenters/Utilities/DateList/date_list.html.twig
votemike: ARGHGHGHGH!!!!!sin(90)cos(0)

src/Ds2013/Presenters/Utilities/DateList/DayDateListItemPresenter.php
```diff
@@ -32,6 +33,11 @@ public function isGmt(): bool
         return $this->options['user_timezone'] == 'GMT';
     }
 
+    public function isToday(): bool
```
votemike: I think a small doc block explaining why we aren't just using `$this->getDateTime()->isToday()` would be useful.

PR #738, opened 2018-09-19
Christian would like classes on music segment items to reflect DB typ…

BjornTwachtmann:
…es exactly


programmes-developers:

PR #740, opened 2018-09-19
PROGRAMMES-6610 Standalone playout page

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6610


programmes-developers:

templates/ancillary_pages/player.html.twig
```diff
@@ -0,0 +1,234 @@
...
+            background: rgba(0, 0, 0, 0.7);
+            letter-spacing: 0;
+            font-weight: normal;
+            margin: 0;
+            padding: 8px;
+            display: table;
+            vertical-align: middle;
+        }
+
+        .smp__messagetext {
```
votemike: This is copy pasted from V2, but I can't see it used anywhere. Shall I get rid of it?

config/routes/3rd_party.yaml
```diff
@@ -101,12 +101,6 @@ programme_topics_topic:
     requirements: { pid: '[0-9b-df-hj-np-tv-z]{8,15}' }
     schemes: [http]
 
-programme_player:
-    path: /programmes/{pid}/player
-    defaults: { _controller: FrameworkBundle:Redirect:redirect, route: '' }
-    requirements: { pid: '[0-9b-df-hj-np-tv-z]{8,15}' }
-    schemes: [https]
```
votemike: Do I need to keep the https thing? Aren't all our pages https?

src/Controller/AncillaryPages/PlayerController.php
```diff
@@ -0,0 +1,75 @@
...
+        foreach ($clip->getAncestry() as $ancestor) {
+            $longerTitleParts[] = $ancestor->getTitle();
+        }
+
+        $twitterTitle = implode(', ', $longerTitleParts) . $suffix;
+
+        $hasChrome = null === $request->get('chromeless');
+        $subtitle = implode(', ', array_map(function (Programme $t) {
+            return $t->getTitle();
+        }, array_reverse(\array_slice($clip->getAncestry(), 1))));
```
tatitati: I really like it!

src/Controller/AncillaryPages/PlayerController.php
```diff
@@ -0,0 +1,75 @@
...
+    public function __invoke(
+        Request $request,
+        UrlGeneratorInterface $router,
+        SegmentEventsService $segmentEventsService,
+        VersionsService $versionsService,
+        Clip $clip
+    ) {
+        $suffix = ' - BBC';
+        if ($clip->getTleo()->getNetwork()) {
+            $suffix = ' - ' . $clip->getTleo()->getNetwork()->getName();
```
cisco87: Shouldn't it be `$suffix .=`?

src/Controller/AncillaryPages/PlayerController.php
```diff
@@ -0,0 +1,75 @@
...
+
+        $longerTitleParts = [];
+        foreach ($clip->getAncestry() as $ancestor) {
+            $longerTitleParts[] = $ancestor->getTitle();
+        }
+
+        $twitterTitle = implode(', ', $longerTitleParts) . $suffix;
+
+        $hasChrome = null === $request->get('chromeless');
+        $subtitle = implode(', ', array_map(function (Programme $t) {
```
cisco87: Don't you think that it might require some more brain work in order to tokenize for a human the various part of it, what do you think of changing it to like:

```
$subtitle = implode(
            ', ',
            array_map(
                function (Programme $t) {
                    return $t->getTitle();
                },
                array_reverse(\array_slice($clip->getAncestry(), 1))
            )
        );
```
would it be worth it?

config/routes/3rd_party.yaml
BjornTwachtmann: Scheme isn't needed for pages in V3, should be fine

src/Controller/AncillaryPages/PlayerController.php
```diff
@@ -0,0 +1,75 @@
...
+        if ($clip->getTleo()->getNetwork()) {
+            $suffix = ' - ' . $clip->getTleo()->getNetwork()->getName();
+        }
+
+        $longerTitleParts = [];
+        foreach ($clip->getAncestry() as $ancestor) {
+            $longerTitleParts[] = $ancestor->getTitle();
+        }
+
+        $twitterTitle = implode(', ', $longerTitleParts) . $suffix;
```
BjornTwachtmann: Would the TItleLogicHelper be useful for the title/subtitle here? This seems like the sort of thing the ProgrammePresenter normally does.

votemike: It's the opposite way round to the normal title.

src/Controller/AncillaryPages/PlayerController.php
votemike: I think it's clear enough as-is. Does anyone else have any opinions?

src/Controller/AncillaryPages/PlayerController.php
BjornTwachtmann: That thing does have an option for doing things the other way round, but it might not be right, honestly not sure, just wondering.

src/Controller/AncillaryPages/PlayerController.php
```diff
@@ -0,0 +1,75 @@
...
+
+        $hasChrome = null === $request->get('chromeless');
+        $subtitle = implode(', ', array_map(function (Programme $t) {
+            return $t->getTitle();
+        }, array_reverse(\array_slice($clip->getAncestry(), 1))));
+
+        $context = new MetaContext($clip, $router->generate('programme_player', ['pid' => $clip->getPid()]));
+
+        $available = $clip->isExternallyEmbeddable();
```
BjornTwachtmann: don't you also need to be checking streamable here?

templates/ancillary_pages/player.html.twig
```diff
@@ -0,0 +1,234 @@
...
+                max-width: 128px;
+            }
+        }
+    </style>
+{% endblock %}
+
+{% block body %}
+    <div class="episode-playout">
+        {% if available %}
+            {{ ds2013('smp', clip, streamable_version, segment_events, '', {'app_name': 'programmes', 'prod_name': 'programmes'}, {'uas': false, 'sizes': {1: 1/1}}) }}
```
BjornTwachtmann: Can you pass autoplay: false in here. Tasio's adding that option, and these shouldn't autoplay

src/Controller/AncillaryPages/PlayerController.php
votemike: Doesn't seem to do what we want for the Twitter title. But did use it for the subtitle.

src/Controller/AncillaryPages/PlayerController.php
benolot: I think Mikes is fine, it also matches PSR2 and PSR2 says you MAY split it across multiple lines, not you have to: `Argument lists MAY be split across multiple lines, where each subsequent line
is indented once. When doing so, the first item in the list MUST be on the
next line, and there MUST be only one argument per line.`

templates/ancillary_pages/player.html.twig
votemike: @Tasiobg I don't think this is related, is it?

templates/ancillary_pages/player.html.twig
```diff
@@ -0,0 +1,190 @@
...
+    <script type="text/javascript" src="https://nav.files.bbci.co.uk/orbit/2.0.0-296.9bee3fd/js/require.min.js"></script>
+    <script>
+        requirejs.config({
+            "paths": {
+                'jquery-1.9': "https://static.bbci.co.uk/frameworks/jquery/0.4.1/sharedmodules/jquery-1.9.1",
+                'bump-3': "//emp.bbci.co.uk/emp/bump-3/bump-3",
+                'smp/smpFactory': "{{ asset_js("js/smp/smpFactory.js") }}",
+                'smp/smp': "{{ asset_js("js/smp/smp.js") }}",
+                'smp/recommendationsBump': "{{ asset_js("js/smp/recommendationsBump.js") }}",
+                'smp/uasService': "{{ asset_js("js/smp/uasService.js") }}"
```
BjornTwachtmann: You shouldn't need this, I think, it's not really usable without ORB/IDCTA etc.

templates/ancillary_pages/player.html.twig
```diff
@@ -0,0 +1,190 @@
...
+
+            <p class="info__more">
+                <a href="{{ path('find_by_pid', { 'pid': clip.getTleo.getPid() }) }}" target="_blank">
+                    {{ clip.getTleo.getTitle() }}
+                </a>
+            </p>
+        </div>
+        <div class="footer">
+            <div class="footer__logo">
+                <img src="https://static.bbc.co.uk/programmes/2.288.1.328/img/bbc-blocks-light.png" alt="">
```
BjornTwachtmann: It doesn't seem very smart to link to the V2 image here. Wouldn't it be better to bring in the PNG to this project and link to it with the asset() thing?

PR #742, opened 2018-09-21
Programmes 6630 improve programme filter logic

cisco87:
Making sure AdaProgrammeService::parseAggregateResponses is able to recognize and delete duplicate records from lists, and unit tests are considering it


programmes-developers:

.gitignore
```diff
@@ -17,3 +17,8 @@ config/parameters.yaml
 .phpunit
 /phpunit.xml
 ###< symfony/phpunit-bridge ###
+
+###> IntelliJ
+.idea/
```
votemike: We tend to use a global gitignore for this. Otherwise this file will grow for each IDE people choose to use.

src/ExternalApi/Ada/Service/AdaProgrammeService.php
```diff
@@ -79,7 +94,29 @@ private function parseAggregateResponses(array $responses, int $limit): array
...
+            return [];
+        }
+
+        $uniqueProgrammes = array_reduce(
+            $programmes,
+            function ($carry, $item) {
+                if (!is_array($carry)) {
+                    $carry = [];
+                }
+                if (is_array($item) && isset($item['pid'])) {
```
BjornTwachtmann: is_array is redundant here I think, isset should cover it

BjornTwachtmann:
A test for the case "PIDs are the same, but the arrays returned by the ADA service differ" would be useful, as Mike said.

PR #744, opened 2018-09-21
[PROGRAMMES-6506] - Display in Profile/Articles pages single clips and clip streams. 

tatitati:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6506

## URL Examples to test (don't forget to run ```yarn run gulp```):
Article page with clips
http://sandbox.bbc.co.uk/programmes/articles/3mh1WBFYSqTSVN2znrmr8gM/testing-embeds-galore?__flush_cache=1

Profile page with clips:
http://sandbox.bbc.co.uk/programmes/profiles/z5gWDtzYM5nmgZD7zfgDFy/test-profile?__flush_cache=1

Light profile:
http://sandbox.bbc.co.uk/programmes/profiles/2grMBcgWb9lHNR46f9yJGzy/any-title?__flush_cache=1


programmes-developers:

src/Ds2013/Presenters/Domain/ContentBlock/Clip/ClipStandAlone/ClipStandalonePresenter.php
```diff
@@ -0,0 +1,30 @@
+<?php
+namespace App\Ds2013\Presenters\Domain\ContentBlock\Clip\ClipStandAlone;
```
votemike: Alone is capitalised in the namespace but not the path. This is the same in a few places in this PR. Could you please make them all the same.

src/Ds2013/Presenters/Utilities/SMP/SmpPresenter.php
```diff
@@ -105,8 +105,8 @@ public function getSmpConfig(): array
                 'playlistObject' => $smpPlaylist,
                 'statsObject' => [
                     'siteId' => $this->analyticsLabels['bbc_site'] ?? '',
-                    'product' => $this->analyticsLabels['prod_name'],
-                    'appName' => $this->analyticsLabels['app_name'],
+                    'product' => $this->analyticsLabels['prod_name'] ?? '',
```
votemike: Are these optional?

src/ExternalApi/Isite/Domain/ContentBlock/ClipBlock/ClipStream.php
```diff
@@ -0,0 +1,26 @@
...
+    public function __construct(string $title, array $streamItems = [])
+    {
+        parent::__construct($title);
+        $this->streamItems = $streamItems;
+    }
+
+    /**
+     * @return StreamItem[]
+     */
+    public function getStreamItems()
```
votemike: Typehint `: array`?

src/Ds2013/Presenters/Domain/CoreEntity/SharedSubPresenters/core_entity_title.html.twig
```diff
@@ -19,7 +19,13 @@
 {% endspaceless %}
 
 {% macro render_title(programme, presenter) %}
-{% spaceless %}
-        <span>{{ truncate(programme.getTitle(), presenter.getOption('truncation_length'))}}</span>
-{% endspaceless %}
+    {% if presenter.getOption('override_title') is not null %}
+        {% set titleProgramme = presenter.getOption('override_title') %}
```
votemike: Would it not be better to have this logic in the presenter rather than the template?

src/Ds2013/PresenterFactory.php
```diff
@@ -4,6 +4,9 @@
 
 use App\Ds2013\Presenters\Domain\Broadcast\BroadcastPresenter;
 use App\Ds2013\Presenters\Domain\BroadcastEvent\BroadcastEventPresenter;
+use App\Ds2013\Presenters\Domain\ContentBlock\Clip\ClipPresenter;
```
votemike: Is this used?

src/Ds2013/Presenters/Domain/ContentBlock/Clip/ClipStandalone/clip_standalone__list.html.twig
```diff
@@ -0,0 +1,27 @@
...
+<div id="clips-p{{ id }}-items">
+    <ul class="list-unstyled article-clips grid-wrapper stream__list">
+        <li class="stream__item grid 1">
+            <div class="grid__inner stream__item-inner br-page-bg-onborder">
+                <div class="programme programme--clip block-link programme--grid br-keyline br-blocklink-subtle br-subtle-linkhover-onbg015--hover">
+                    <div class="programme__img">
+                        {{ ds2013('smp', content_block.getClip(), content_block.getStreamableVersion(), [], '', []) }}
+                    </div>
+
+                    <div class="programme__body" style="height: 16px;">
```
votemike: We should use inline styles.

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
```diff
@@ -93,6 +99,41 @@ public function getDomainModel(SimpleXMLElement $isiteObject)
...
+                $contentBlockData = $form->content;
+
+                $captions = [];
+                foreach ($contentBlockData->clips as $isiteClip) {
+                    $captions[$this->getString($isiteClip->pid)] = $this->getString($isiteClip->caption);
+                }
+
+                // fill programme key
+                $clipsProgs = $this->coreEntitiesService->findByPids($this->stringToPids(array_keys($captions)));
+                if (count($clipsProgs) > 1) {
```
votemike: What if there are 0 clips?

tatitati: In isite is required  (and validated) to introduce a PID clip for "Content - Clips Box". The caption however is optional and can be empty.

src/Ds2013/Presenters/Domain/ContentBlock/Clip/ClipStandalone/clip_standalone.html.twig
```diff
@@ -0,0 +1,18 @@
...
+{% block extraClasses %}
+    {% if content_block.isInPrimaryColumn() == false %}
+        component--box-flushbody
+    {% endif %}
+{% endblock %}
+
+{% set id = random() %}
+
+{% block blockContent %}
+        {% if content_block.isInPrimaryColumn() %}
```
BjornTwachtmann: Please make this the same for 1 and 2 column layouts. It's a fucking box.

src/Ds2013/Presenters/Domain/ContentBlock/Clip/ClipStandalone/clip_standalone.html.twig
```diff
@@ -0,0 +1,40 @@
...
+            {% if content_block.isInPrimaryColumn() %}
+                <h3 class="no-margin">
+                    <a href="{{ path('find_by_pid', {'pid': content_block.getClip().getPid()}) }}">
+                        {{ content_block.getClip().getTitle() }}
+                    </a>
+                </h3>
+
+                <p>{{ content_block.getClip().getShortSynopsis() }}</p>
+            {% else %}
+                <h4 class="programme__titles text-base">
```
votemike: I think these heading levels probably need to be configurable like in other presenters

src/Ds2013/Presenters/Domain/ContentBlock/Clip/ClipStream/clip_stream.html.twig
```diff
@@ -0,0 +1,63 @@
...
+            </li>
+            {%- endfor -%}
+
+        </ul>
+    </div>
+
+    <script>
+        {% set br = content_block.isInPrimaryColumn() ? 'page' : 'highlight' %}
+
+        require(['jquery-1.9','stream'], function($, Stream) {
```
votemike: Do we want a require for each clip block there is?

assets/js/stream.js
```diff
@@ -0,0 +1,490 @@
...
+            }
+            this.updateClasses();
+
+
+            setTimeout(function () {
+                //_this.fetchImages();
+                _this.setButtons();
+            },100);
+        },
+        createIndicators : function () {
```
votemike: Are we keeping the indicators?

src/Ds2013/Css/PageSpecfic/Profiles/_clipstream.scss
```diff
@@ -0,0 +1,110 @@
...
+
+.stream__window {
+    display: block;
+    overflow-x: scroll;
+    overflow-y: hidden;
+    touch-action: pan-x;
+}
+
+// indicator buttons
+.component__header .stream__indicator-wrap {
```
votemike: Are we keeping the indicators?

src/Ds2013/Presenters/Domain/ContentBlock/Clip/ClipStream/clip_stream.html.twig
tatitati: yes. Actually this was improved because in v2 it was required even for not stream blocks. So, for this example:
https://www.test.bbc.co.uk/programmes/articles/3mh1WBFYSqTSVN2znrmr8gM/testing-embeds-galore?__flush_cache=1

You will see that there are two streams in the page, and one stand alone (right column). 

if you search in source-code: 'indicator_button' (is part of the code of stream.js)  this code appear three times. The extra one is because of an standalone clip, which doesn't need it.

src/Ds2013/Presenters/Domain/ContentBlock/Clip/ClipStream/clip_stream.html.twig
```diff
@@ -0,0 +1,56 @@
+{% extends "@Ds2013/Presenters/Domain/ContentBlock/content_block.html.twig" %}
+
+{% set id = random() %}
+
+{% block blockContent %}
+    {# Carousel stream #}
+    <div id="clips-p{{ id }}-items">
+        <ul class="list-unstyled article-clips grid-wrapper stream__list">
+
+            {%- for index, carouselItem in content_block.getStreamItems() -%}
```
Tasiobg: Is index used?

src/Ds2013/Presenters/Domain/ContentBlock/Clip/ClipStream/clip_stream.html.twig
```diff
@@ -0,0 +1,56 @@
...
+    </div>
+
+    <script>
+        require(['jquery-1.9','stream'], function($, Stream) {
+            var params = {
+                'panel_class': '',
+                'next_button_text': '{{ gelicon('basics', 'next', 'gelicon--half-container') }}' ,
+                'previous_button_text': '{{ gelicon('basics', 'previous', 'gelicon--half-container') }}',
+                'panel_class': 'br-box-subtle',
+                'set_to_same': '.programme__body' ,
```
Tasiobg: Does `.programme__body` and `.programme__img` requires a dot? Looks a bit odd but could be right.

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
```diff
@@ -15,18 +17,22 @@
 use App\ExternalApi\Isite\Domain\ContentBlock\ThirdParty;
 use BBC\ProgrammesPagesService\Domain\ValueObject\Pid;
 use BBC\ProgrammesPagesService\Service\CoreEntitiesService;
-use Exception;
+use BBC\ProgrammesPagesService\Service\VersionsService;
```
Tasiobg: I think you don't need to remobe Exception, it would be used when uncomented 
```
throw new Exception('Invalid content block type. Found ' . $type);
```
I think this should be uncommented when the last content block is added

src/Ds2013/Presenters/Domain/ContentBlock/Clip/ClipStream/clip_stream.html.twig
tatitati: Is used in that way in V2:
https://github.com/bbc/programmes/blob/develop/webapp/php/sharedmodules/programmes/views/partials/content-block/clips.spv#L89

src/Ds2013/Presenters/Domain/ContentBlock/Clip/ClipStandalone/clip_standalone.html.twig
```diff
@@ -0,0 +1,40 @@
...
+
+{% block blockContent %}
+    <div id="clips-p{{ id }}-player" class="smp">
+        <div class="stream__item-inner br-page-bg-onborder">
+            {{ ds2013('smp', content_block.getClip(), content_block.getStreamableVersion(), [], '', []) }}
+        </div>
+    </div>
+
+    <div id="clips-p{{ id }}-detail">
+        <div class="island br-box-subtle">
```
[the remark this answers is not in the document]
tatitati: Yeah, I tried with br-box-subtle, but the resulting style was not the desired. Also I compared with another panels, I think is fine

Tasiobg:
Please rebase so we can review after merge conflicts are resolved

src/Ds2013/Presenters/Domain/ContentBlock/Clip/ClipStandalone/clip_standalone.html.twig
```diff
@@ -0,0 +1,40 @@
...
+            {% if content_block.isInPrimaryColumn() %}
+                <h3 class="no-margin">
```
votemike: Have you checked that the headers are semantically correct?

src/Ds2013/Presenters/Domain/ContentBlock/Clip/ClipStream/clip_stream.html.twig
votemike: Would we not prefer something similar to the favourite buttons?

src/Ds2013/Presenters/Domain/ContentBlock/Clip/ClipStandalone/clip_standalone.html.twig
tatitati: I just double checked. I confirm that is in the correct order.
```
h1
  |- h2
        |-h3
```

tatitati:
> 
>@Tasiobg 
>Please rebase so we can review after merge conflicts are resolved

Done

src/Ds2013/Presenters/Domain/ContentBlock/Clip/ClipStream/clip_stream.html.twig
tatitati: I think this improvement might be done as part of the "improvements/simplifications" ticket:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6605

BjornTwachtmann: @tatitati , that ticket was for investigating how to get rid of some of the V2 clip switching madness. It's not for refactoring this code

config/services.yaml
```diff
@@ -27,7 +27,7 @@ parameters:
     app.electron.service.class: App\ExternalApi\Electron\Service\ElectronService
     app.favourites.baseurl: 'https://www.bbc.co.uk/modules/plugin/musicfavourite/init'
     app.favourites.service.class: App\ExternalApi\FavouritesButton\Service\FavouritesButtonService
-    app.isite.baseurl: 'https://api.live.bbc.co.uk/isite2-content-reader'
+    app.isite.baseurl: 'https://api.test.bbc.co.uk/isite2-content-reader'
```
votemike: NO!
Please don't commit this change.

src/Ds2013/Presenters/Domain/ContentBlock/content_block.html.twig
```diff
@@ -15,3 +15,31 @@
         {% block blockContent %}{% endblock %}
     </div>
 </div>
+
+{% block javascripts %}
```
votemike: This will render for each content block won't it? Would it be possible to just call `get_streams()` in the page template?

config/services.yaml
tatitati: arg!, damn it, indeed a mistake

templates/base_ds2013.html.twig
```diff
@@ -114,7 +115,6 @@
             {% block body %}{% endblock %}
         </div>
         {% if programme is defined %}{{ ds2013('footer', programme) }}{% endif %}
-        {{ gelicons_source() }}
```
BjornTwachtmann: why move this?

src/Twig/StreamClipsExtension.php
```diff
@@ -0,0 +1,36 @@
+<?php
+declare(strict_types = 1);
+
+namespace App\Twig;
+
+use Twig_Extension;
+use Twig_Function;
+
+class StreamClipsExtension extends Twig_Extension
```
BjornTwachtmann: Why StreamClips instead of ClipStream?

src/Ds2013/Presenters/Domain/ContentBlock/Clip/ClipStandalone/ClipStandalonePresenter.php
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    public function getCaption(): string
+    {
+        return $this->block->getCaption();
+    }
+
+    public function getStreamableVersion(): Version
+    {
+        return $this->block->getStreamableVersion();
```
BjornTwachtmann: So, here's a fun question that you will enjoy and in no way want to hurt me because of...

What happens if somebody adds a clip to an article, and later that clip stops being available/streamable. The clip will still be returned by the query, but it won't have a streamableVersion, and this will blow up. 

Should we deal with that? It's an edge case, admittedly.

templates/base_ds2013.html.twig
tatitati: In our javascript block in twig we have something like:
```
{% block javascripts %}
              ....
              var params = {
                    ...
                    'next_button_text': '{{ gelicon('basics', 'next', 'gelicon--half-container') }}' ,
                    'previous_button_text': '{{ gelicon('basics', 'previous', 'gelicon--half-container') }}',
                    ...
{% endblock %}
```
The problem with this is that the javascript-block is printed in the resulting HTML after the call to {{gelicon_source}}. So if we need gelicons in our javascript and they were not added before to the list of icons, then they won't have any definition to use (the icon won't appear)

I thought that this might solve that issue (and possibly more in the future)

src/Ds2013/Presenters/Domain/ContentBlock/Clip/ClipStandalone/ClipStandalonePresenter.php
tatitati: I didn't predict that scenario, but Is worth to. I will try to cover it

src/Twig/StreamClipsExtension.php
tatitati: I can change it, for me StreamClips was more like saying "Stream Of Clips". But there is no science behind of that decisision.

templates/base_ds2013.html.twig
BjornTwachtmann: yeah, that makes sense actually, that stuff is way too magic...

templates/shared/stream_clips.html.twig
```diff
@@ -0,0 +1,26 @@
...
+        require(['jquery-1.9','stream'], function($, Stream) {
+            {% for stream_clip_id in streams_clips_ids %}
+                var params = {
+                    'panel_class': '',
+                    'next_button_text': '{{ gelicon('basics', 'next', 'gelicon--half-container') }}' ,
+                    'previous_button_text': '{{ gelicon('basics', 'previous', 'gelicon--half-container') }}',
+                    'panel_class': 'br-box-subtle',
+                    'set_to_same': '.programme__body' ,
+                    'match_button_height': '.programme__img',
+                    'items_in_window': [
```
BjornTwachtmann: Weren't we going to adjust this so that the main (left) column also only had 1 item. It honestly looks terrible with 4 next to each other and no main item.

tatitati: Is not a mistake from my side, @claudioneves and @votemike decided to remove the main item. Do we want to discuss it again guys?

votemike: Happy to discus again.

BjornTwachtmann: I'm not saying don't remove the same item.

I'm saying that displaying 4 clips next to each other in a narrow bar looks shit. We should change that to a carousel where only 1 clip is displayed at a time (by changing this code) and it will look less shit.

src/Ds2013/Presenters/Domain/ContentBlock/Clip/ClipStandalone/ClipStandalonePresenter.php
tatitati: Studied and covered the edge case
See:
https://github.com/bbc/programmes-frontend/blob/c9d2557ab365a814400bb01556ca4578cd60a657/tests/ExternalApi/Isite/Mapper/Clip/MultipleStreamableClipsContentBlockMapperTest.php

BjornTwachtmann:
This is looking pretty good now. The clips carousel looks much nicer with just one in it. There are still some bugs in the mobile functionality for the clips carousel I think. Try it in mobile mode and compare to test.

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
```diff
@@ -190,27 +190,44 @@ public function getDomainModel(SimpleXMLElement $isiteObject)
...
+                            $this->getString($contentBlockData->title),
+                            $carouselItems[0]->getTitle(),
+                            $carouselItems[0]->getClip(),
+                            $this->streamableVersions[(string) $carouselItems[0]->getClip()->getPid()]
+                        );
 
-                    break;
+                        break;
+                    } else {
+                        break;
```
votemike: Shouldn't this throw an exception?

Tasiobg:
When I open this page programmes/profiles/z5gWDtzYM5nmgZD7zfgDFy/test-profile?__flush_cache=1
some clips auto

src/Ds2013/Presenters/Domain/ContentBlock/Clip/ClipStandalone/clip_standalone.html.twig
```diff
@@ -0,0 +1,40 @@
...
+        component--box-flushbody
+    {% endif %}
+{% endblock %}
+
+{% set id = random() %}
+
+{% block blockContent %}
+    <div id="clips-p{{ id }}-player" class="smp">
+        <div class="stream__item-inner br-page-bg-onborder">
+            {{ ds2013('smp', content_block.getClip(), content_block.getStreamableVersion(), [], '', []) }}
```
Tasiobg: Autoplay is enable by defaul so when I go a an article with some clips all of them start to play when loaded,
I belive you only need to change this to 
```
{{ ds2013('smp', content_block.getBlock().getClip(), content_block.getBlock().getStreamableVersion(), [], null, null, {'autoplay': 'false'} ) }}
```
I tested it and seems working fine

src/ExternalApi/Isite/Domain/ContentBlock/ClipBlock/ClipStandAlone.php
```diff
@@ -0,0 +1,41 @@
+<?php
```
Tasiobg: Is `declare(strict_types = 1);` missing on purpose? Is also not present in other classes

PR #745, opened 2018-09-24
[PROGRAMMES-6552] Bug fix- Navigation links being cut off.

tomwinnington:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6552

Changed the breakpoints to resolve issue.

Link:
http://sandbox.bbc.co.uk/programmes/b0bcmgp1?__flush_cache=1

Before:
before

After:
after


programmes-developers:

BjornTwachtmann:
Hmm, so this is shrinking the space for genres/formats in order to make more space for the service boxes? 

That sort of works I suppose, but it would still break in the cases of longer words. It also still cuts things off at smaller breakpoints, have a look at your example at 755px screen width. I'm also not sure if reducing the space for genres/formats could end up cutting things off _there_ on some programmes. 

I'd probably first look to slap `word-wrap: break-word` and `hyphens: auto` on `footer__service__nav`. Probably also reduce the right padding on `footer__service__nav`. It may still be worth it to create more space for this, but we should run that change past product.

Tasiobg:
Looks nice but I can't remember if we should be using `word-wrap: break-word` or `hyphens: auto`

PR #746, opened 2018-09-24
Programmes 6592 article page social embeds

cisco87:


programmes-developers:

assets/js/utils/cookie.js
```diff
@@ -0,0 +1,74 @@
+define('utils/cookie', function () {
```
tatitati: nice work!, only one thing. We have a file already that makes exactly the same than this cookie.js (or at least in a first glance seems quite duplicated). You can find it in here:
https://github.com/bbc/programmes-frontend/blob/develop/assets/js/smp/cookieStore.js

If it sounds like a duplicated, we should need to consider to put that cookieStore.js in a place where is easy to reuse (probably outside the folder smp might be enough).

src/Ds2013/Presenters/Domain/ContentBlock/ThirdParty/third_party.html.twig
```diff
@@ -0,0 +1,30 @@
+{% extends "@Ds2013/Presenters/Domain/ContentBlock/content_block.html.twig" %}
+
+{% block blockContent %}
+    {% set content_block_box_id = 'third-party-' ~ content_block.getIncrementalContainerId() %}
```
tatitati: for id generations we are using the twig function "random()". Have a look in here, maybe it is useful for this as well:
https://github.com/bbc/programmes-frontend/blob/develop/src/Ds2013/Presenters/Domain/ContentBlock/Faq/faq.html.twig#L8

src/ExternalApi/Isite/Domain/ContentBlock/ThirdParty.php
```diff
@@ -0,0 +1,85 @@
...
+    ) {
+        parent::__construct($title);
+
+        $this->name = $name;
+        $this->url = $url;
+        $this->altText = $altText;
+        $this->type = $this->computeType($this->url);
+    }
+
+    public function getURL()
```
tatitati: Typehints for every getter, example: ```public function getURL(): string```

assets/js/utils/cookie.js
cisco87: I didn't know about it, but I just brought over a piece of code (third-party.js) that has a dependency on this file. I guess it also depends on the method signature of the way this has been designed, so I guess deleting it would mean also changing the way third-party.js interact with the cookie class. I think it's worth it but I need to ask @BjornTwachtmann to see if I can have time to do it

BjornTwachtmann: Yeah, don't duplicate code, use the cookieStore.js, I'm assuming it's suitable for that purpose @tatitati  ?

tatitati: Yeah, if you see both files they are doing exactly the same.

assets/js/third-party.js
```diff
@@ -0,0 +1,195 @@
...
+
+            url += "format=" + provider.format + "&url=" + escape(externalUrl) + qs;
+            if(provider.dataType!='json') url += "&" + provider.callbackparameter + "=?";
+
+            return url;
+        },
+        generateCode : function (embedProvider, externalUrl) {
+            var policy = cookie.readPolicy('personalisation'); //functional cookie
+
+            if (!policy || embedProvider.templateRegex) {
```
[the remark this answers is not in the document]
cisco87: I think maybe at a certain point in time someone had the idea to define templates to display some third party embeds in a personalised way, defining a template and saving the setting in a cookie, I think now that you make me notice, there is nothing like that existing right now, so I guess since it is the only dependency on the cookie class it can all be removed, what do you think?

assets/js/third-party.js
```diff
@@ -0,0 +1,196 @@
...
+            new Provider("soundcloud", "rich", ["soundcloud.com/.+","snd.sc/.+"], "https://soundcloud.com/oembed?maxheight=166",{format:'js'}),
+            new Provider("twitter", "rich", ["twitter.com/.+"], "https://api.twitter.com/1/statuses/oembed.json"),
+            new Provider("vimeo", "video", ["www\.vimeo\.com\/groups\/.*\/videos\/.*", "www\.vimeo\.com\/.*", "vimeo\.com\/groups\/.*\/videos\/.*", "vimeo\.com\/.*"], "//vimeo.com/api/oembed.json"),
+            new Provider("youtube", "video", ["youtube\\.com/watch.+v=[\\w-]+&?", "youtu\\.be/[\\w-]+","youtube.com/embed"], 'https://www.youtube.com/embed/$1?wmode=transparent&rel=0&iv_load_policy=3', {
+                templateRegex: /.*(?:v\=|be\/|embed\/)([\w\-]+)&?.*/,
+                embedtag: {
+                    tag: 'iframe'
+                }
+            }),
+            new Provider("vine", "video", ["vine\\.co\/v\/.+"], 'https://vine.co/v/$1/embed/simple', {
```
votemike: Vine doesn't exist anymore. Do we still need it?

assets/sass/components/_third-party.scss
```diff
@@ -0,0 +1,57 @@
...
+    iframe {
+        width: 100%; // be responsive
+    }
+}
+// make responsive
+.third-party--youtube,
+.third-party--instagram,
+.third-party--touchcast,
+.third-party--vimeo,
+.third-party--vine {
```
votemike: Vine doesn't exist anymore, do we still need this?

composer.json
```diff
@@ -59,6 +59,7 @@
     "require-dev": {
         "doctrine/doctrine-fixtures-bundle": "^3.0",
         "escapestudios/symfony2-coding-standard": "^3.0.0",
+        "fzaninotto/faker": "^1.7",
```
votemike: Faker is already in the main require block. Is there a reason it's been added here?

assets/sass/components/_third-party.scss
```diff
@@ -0,0 +1,57 @@
+.third-party {
```
votemike: Is this the correct path for this file? Should it not live in the same directory and the presenter?

assets/js/third-party.js
BjornTwachtmann: It still works, seems pointless to kill it

assets/sass/components/_third-party.scss
cisco87: Yup sorry I didn't see it, Bjorn just showed me where they are put, thank you!

assets/sass/components/_third-party.scss
cisco87: It still works so I guess since it requires no maintainance we can just keep it?

src/ExternalApi/Isite/Domain/ContentBlock/ThirdParty.php
```diff
@@ -0,0 +1,85 @@
+<?php
+declare(strict_types = 1);
+
+namespace App\ExternalApi\Isite\Domain\ContentBlock;
+
+class ThirdParty extends AbstractContentBlock
+{
+    private const SUPPORTED_THIRDPARTY_ENDPOINT = [
```
BjornTwachtmann: I think this doesn't cover all the cases

templates/base_ds2013.html.twig
```diff
@@ -51,9 +51,10 @@
                     'smp/smp': "{{ asset_js("js/smp/smp.js") }}",
                     'smp/recommendationsBump': "{{ asset_js("js/smp/recommendationsBump.js") }}",
                     'smp/uasService': "{{ asset_js("js/smp/uasService.js") }}",
+                    'smp/cookieStore': "{{ asset_js("js/smp/cookieStore.js") }}",
```
BjornTwachtmann: Is it worth moving this cookieStore utility if it's going to be used elsewhere. Also, shouldn't it be concatinated into the JS files that require it, so we don't need this?

src/Ds2013/Presenters/Domain/ContentBlock/ThirdParty/third_party.html.twig
```diff
@@ -0,0 +1,30 @@
...
+    {% set content_block_box_id = 'third-party-' ~ random() %}
+    {% set third_party_block = content_block.getBlock() %}
+
+    <div class="third-party" id="{{ content_block_box_id }}">
+        <div class="icon-box br-box-page">
+            <a
+                class="icon-box__link br-box-page__link br-page-link-onbg015 br-page-linkhover-onbg015--hover"
+                href="{{ third_party_block.getUrl() }}"
+            >
+                {{ gelicon('basics', 'recommend', 'icon-box__icon br-pseudolink') }}
```
BjornTwachtmann: Change to `{{ gelicon('core', 'external-link', 'icon-box__icon br-pseudolink') }}`

assets/js/third-party.js
votemike: Just seems like extra bytes.

src/Ds2013/Presenters/Domain/ContentBlock/ThirdParty/third_party.html.twig
```diff
@@ -0,0 +1,36 @@
+{% extends "@Ds2013/Presenters/Domain/ContentBlock/content_block.html.twig" %}
+
+
```
votemike: Probably don't need 2 blank lines.

src/Ds2013/Presenters/Domain/ContentBlock/ThirdParty/third_party.html.twig
```diff
@@ -0,0 +1,36 @@
...
+{% block extraClasses %}
+    {% if not content_block.isInPrimaryColumn() %}
+        component--box-flushbody
+    {% endif %}
+{% endblock %}
+
+{% block blockContent %}
+    {% set content_block_box_id = 'third-party-' ~ random() %}
```
votemike: Should this not be set in the presenter?

src/Ds2013/Presenters/Domain/ContentBlock/ThirdParty/third_party.html.twig
```diff
@@ -0,0 +1,36 @@
...
+                {{ gelicon('core', 'external-link', 'icon-box__icon br-pseudolink') }}
+                <div class="icon-box__hgroup">
+                    <h3 class="icon-box__title gamma">{{ third_party_block.getName() }}</h3>
+                </div>
+            </a>
+        </div>
+    </div>
+
+    <script>
+        require(['jquery-1.9', 'third-party'], function($, ThirdParty) {
```
votemike: Should this not all be built up and then output in one go in a similar way to Favourite Button?

src/Ds2013/Presenters/Domain/ContentBlock/content_block.html.twig
```diff
@@ -1,4 +1,11 @@
-<div class="component component--box {{ content_block.isInPrimaryColumn() ? 'component--box-flushbody-vertical component--box--primary' : 'component--box--striped component--box--secondary' }} {% block extraClasses %}{% endblock %}">
+{% if content_block.isInPrimaryColumn() %}
```
votemike: Would it be better if this logic was in the presenter?

src/Ds2013/Presenters/Domain/ContentBlock/ThirdParty/third_party.html.twig
cisco87: I'm not sure, it seems a template concerns?

src/Ds2013/Presenters/Domain/ContentBlock/content_block.html.twig
cisco87: If it a boolean check that looks like a template corcern for me, but maybe we can involve another dev to see how the voting balances? :D

src/Ds2013/Presenters/Domain/ContentBlock/ThirdParty/third_party.html.twig
cisco87: I will have a look about the new way to include this kind of file, do you know where I can see it?

src/Ds2013/Presenters/Domain/ContentBlock/content_block.html.twig
cisco87: I mean, my idea is that this code is setting a container css class, now this css class under the circumstanses of this template/layout/style might be needed, but might not be needed if you want to change the template/layout/style. If you have it here, you can move on and use a different css class in another template/layout/style, right? But if we put it in the Presenter, which might be shared across different t/l/s, then you would use code that there is not shared or in the worst case scenario would force you to do ifs/else based on the layout you want to use.

This way, this css class only affects this template and makes it changeable/not affecting others in the future, IMHO

src/Ds2013/Presenters/Domain/ContentBlock/ThirdParty/third_party.html.twig
votemike: Look at:
- https://github.com/bbc/programmes-frontend/blob/develop/src/Twig/FavouriteButtonExtension.php
- https://github.com/bbc/programmes-frontend/blob/develop/templates/find_by_pid/clip.html.twig#L161
- https://github.com/bbc/programmes-frontend/blob/develop/src/Ds2013/Presenters/Section/Clip/Details/clip_details.html.twig#L44

cisco87: Thank you!

src/Ds2013/Presenters/Domain/ContentBlock/content_block.html.twig
BjornTwachtmann: There are a few other places in the codebase where the logic around CSS classes was complex enough that I moved it into a Presenter. Not really out of any point of principle but because it involved complex logic and became completely unreadable in the template. 

This is simple enough. You can do either. It might look less messy to use the abastraction we have for this...
https://github.com/bbc/programmes-frontend/blob/develop/src/Ds2013/Presenters/Domain/Broadcast/broadcast.html.twig#L1-L9

cisco87:
> I seem to get a load of console errors on http://sandbox.bbc.co.uk/programmes/articles/16DQcvm1MDdfsLB2tVjsTPw/third-party-embeds

I am not sure what is going on, there is something that I am missing, I tried locally to run phpcs and phpunit before commit, yarn gulp, tested the stuff but apparently you get js errors and jenkins fails to build it, I am going to re-check all the stuff and in case will ask for help if someone can figure out what is happening

src/Ds2013/PresenterFactory.php
```diff
@@ -260,6 +265,9 @@ public function contentBlockPresenter(AbstractContentBlock $contentBlock, bool $
         if ($contentBlock instanceof Table) {
             return new TablePresenter($contentBlock, $inPrimaryColumn, $options);
         }
+        if ($contentBlock instanceof ThirdParty) {
```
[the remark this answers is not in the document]
cisco87: yeah, let's write a tech debt ticket

src/Ds2013/Presenters/Domain/ContentBlock/ThirdParty/_third-party.scss
```diff
@@ -0,0 +1,69 @@
...
+        height: 100%;
+    }
+}
+
+// set aspect ratios
+.third-party--instagram {
+    max-width: 618px; // instagram doesn't support wider
+
+    .third-party__embed {
+        padding-bottom: 122%;
```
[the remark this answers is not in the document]
cisco87: It's sass taken from V2, I guess somehow at a point in time there was the need for it

assets/js/cookieStore.js
```diff
@@ -61,6 +65,15 @@ define(function () {
 
             return parsedCookies;
         };
+
+        this.readPolicy = function(policy) {
+            if (window.bbccookies !== undefined && window.bbccookies.cookiesEnabled()) {
```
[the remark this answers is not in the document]
cisco87: I was told to don't worry about it by @BjornTwachtmann because it would be injected by another component in the page window.bbccookies

assets/js/cookieStore.js
```diff
@@ -7,6 +7,10 @@ define(function () {
 
         var COOKIE_LIFETIME = 1000 * 60 * 60 * 24 * 90; // 90 days
 
+        if (window.bbccookies === undefined || !window.bbccookies.cookiesEnabled()) {
```
BjornTwachtmann: I think this could break things.

src/ExternalApi/Isite/Domain/ContentBlock/ThirdParty.php
```diff
@@ -0,0 +1,44 @@
...
+    private $altText;
+
+    /** @var string */
+    private $name;
+
+    public function __construct(
+        ?string $title,
+        string $url,
+        string $altText,
+        ?string $name
```
BjornTwachtmann: Either this is not nullable, or getName() must also be nullable return type

PR #747, opened 2018-09-25
Programmes 6537 optimise episode page queries

BjornTwachtmann:


programmes-developers:

composer.json
```diff
@@ -32,7 +32,7 @@
         "bbc/branding-client": "^v2.6.2",
         "bbc/gel-iconography-assets": "^2.0",
         "bbc/programmes-morph-library": "^1.0",
-        "bbc/programmes-pages-service": "dev-master@dev",
+        "bbc/programmes-pages-service": "dev-PROGRAMMES-6537-optimise-episode-page-queries@dev",
```
votemike: This needs changing back before merging.

BjornTwachtmann: Indeed

composer.lock
```diff
@@ -3640,6 +3640,11 @@
         {
             "name": "symfony/lts",
             "version": "v3",
+            "source": {
```
benolot: So we don't want to get it from packagist?

BjornTwachtmann: Why the fuck did it change that. DAMN YOU COMPOSER

PR #750, opened 2018-09-25
[PROGRAMMES-6638] - Added redirect controller with cache

tatitati:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6638

The default Symfony redirect doesn't add any caching, so I created a custom one. Not sure if is the best approach, but feedbacks are welcome.

URL examples:

| URL | Redirect (302) to: |
|--- |--- |
| https://sandbox.bbc.co.uk/programmes/b0bjyw29/credits | https://sandbox.bbc.co.uk/programmes/b0bjyw29#credits | 
| https://sandbox.bbc.co.uk/programmes/b09kkt1m/broadcasts | https://sandbox.bbc.co.uk/programmes/b09kkt1m#broadcasts | 

This is parto of another ticket: https://github.com/bbc/programmes-findbypid-proxy/pull/43


programmes-developers:

src/Controller/RedirectAndCacheController.php
```diff
@@ -0,0 +1,23 @@
+<?php
+namespace App\Controller;
+
+use Symfony\Component\HttpFoundation\Request;
+
+class RedirectAndCacheController extends BaseController
+{
+    public function __invoke(Request $request)
```
BjornTwachtmann: It seems like this would be a lot more flexible and require fewer changes to the routing files if you just copied the code from Symfony's RedirectController::redirectAction and modified it to include cache lifetime headers.

120s is also not generally the amount of time we want to cache these for.

BjornTwachtmann: See https://github.com/symfony/symfony/blob/4.0/src/Symfony/Bundle/FrameworkBundle/Controller/RedirectController.php#L56

config/routes.yaml
```diff
@@ -25,10 +25,14 @@ programme_article_listings:
...
+programme_broadcasts:
+    path: /programmes/{pid}/broadcasts
+    defaults: { _controller: App\Controller\RedirectAndCacheController }
+    requirements: { pid: '[0-9b-df-hj-np-tv-z]{8,15}' }
+
 programme_credits:
     path: /programmes/{pid}/credits
-    controller: Symfony\Bundle\FrameworkBundle\Controller\RedirectController::redirectAction
-    defaults: { route: find_by_pid, _fragment: 'credits' }
+    defaults: { _controller: App\Controller\RedirectAndCacheController }
```
votemike: Any reason you've created a controller for this instead of how it previously was?

tatitati: Michael the reason is that the previous Symfony controller don't use caching when redirecting.

src/Controller/RedirectAndCacheController.php
```diff
@@ -0,0 +1,16 @@
+<?php
+namespace App\Controller;
+
+use Symfony\Bundle\FrameworkBundle\Controller\RedirectController;
+use Symfony\Component\HttpFoundation\Request;
+
+class RedirectAndCacheController extends RedirectController
```
tatitati: NOTE:
Symfony\Bundle\FrameworkBundle\Controller\RedirectController is marked as @final, so from that point of view I shouldn't override any function that exists in there. However, I think that because I'm just extending something really small (one line to add cache) It might be worth.  But still, if you have a different opinion just let me know. I thought it might be worth this solution as it doesn't duplicate any logic. But....

src/Controller/RedirectAndCacheController.php
```diff
@@ -0,0 +1,31 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Exception\HttpException;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+
+class RedirectAndCacheController extends BaseController
+{
+    public function redirectAction(Request $request, $route, UrlGeneratorInterface $router, $permanent = false, $ignoreAttributes = false)
+    {
+        if ('' == $route) {
+            throw new HttpException($permanent ? 410 : 404);
```
cisco87: You could use 
Symfony\Component\HttpFoundation\Response::HTTP_GONE
Symfony\Component\HttpFoundation\Response::HTTP_NOT_FOUND
Symfony\Component\HttpFoundation\Response::HTTP_MOVED_PERMANENTLY
Symfony\Component\HttpFoundation\Response::HTTP_FOUND

What do you think?

tatitati: Actually good point. It happens that I'm basically duplicating a code that already exist in Symfony framework: https://github.com/symfony/symfony/blob/4.0/src/Symfony/Bundle/FrameworkBundle/Controller/RedirectController.php#L56

The reason to duplicate this code is not trivial. The Symfony redirect doesn't add cache, and I cannot extend that Symfony controller as it is marked as "@final". I might, but even in that case PHPStan shouts quite a lot. So in the end I opted for duplicating that code. Is for this reason that I would like to keep it as much as possible the same than the original

PR #751, opened 2018-09-26
PROGRAMMES-6593 Article List Page

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6593


programmes-developers:

templates/articles/partials/article_list.html.twig
```diff
@@ -1 +1,16 @@
-The list of articles/features
+{%  if articles|length <= 0 %}
```
votemike: This is similar to templates/profiles/partials/profile_list.html.twig is it worth extracting them in to a common template?

cisco87: I agree with the idea

src/Controller/Articles/IndexController.php
```diff
@@ -4,24 +4,34 @@
...
 
 class IndexController extends BaseController
 {
-    public function __invoke(CoreEntity $coreEntity, IsiteService $isiteService)
+    public function __invoke(CoreEntity $coreEntity, ArticleService $isiteService)
     {
         $this->setContextAndPreloadBranding($coreEntity);
 
         $articles = [];
         $parameters = ['coreEntity' => $coreEntity, 'articles' => $articles, 'paginatorPresenter' => null];
```
cisco87: Sorry, where is this used? Only in the template or somewhere else in PHP? Because maybe it could be worth using constants?

src/Controller/Articles/IndexController.php
```diff
@@ -4,24 +4,34 @@
...
         if ($coreEntity instanceof Programme) {
             $parameters['programme'] = $coreEntity; //so the the base 2013 template sets the footer
+
+            $articlesResult = $isiteService->getArticlesByProgramme($coreEntity, $this->getPage())->wait();
+            $articles = $articlesResult->getDomainModels();
+
+            if ($articlesResult->getTotal() > 48) {
+                $parameters['paginatorPresenter'] = new PaginatorPresenter($this->getPage(), 48, $articlesResult->getTotal());
```
cisco87: I think as a documenting way it could be nice to use a constant for the value `48` to describe what is its meaning, I guess it would be hard for people in the future to figure why has it been used and what it represents? What do you think?

src/Controller/Articles/IndexController.php
```diff
@@ -4,24 +4,34 @@
...
+            }
         }
 
         if (empty($articles)) {
             $this->response()->setStatusCode(404);
+        } else {
+            $parameters['articles'] = $articles;
```
cisco87: I think here I found the place where the above array indexes are being used? So the 'articles' bit is being repeated and I am fairly pro using a constant in order to refer to it

src/ExternalApi/Isite/Service/ArticleService.php
```diff
@@ -4,14 +4,55 @@
...
 use Closure;
 use GuzzleHttp\Promise\FulfilledPromise;
 use GuzzleHttp\Promise\PromiseInterface;
 
 class ArticleService extends IsiteService
 {
+    public function getArticlesByProgramme(
+        Programme $programme,
+        int $page = 1,
+        int $limit = 48
```
cisco87: I definitely think that this 48 should be hidden between a constant label :D

src/Controller/Articles/IndexController.php
benolot: Do you mean where is `$parameters` used? They are passed into the template as seen on line 37.

benolot: I strongly disagree that constants are the way to go here, we pass in an array of parameters to the template throughout the app, it would also be difficult to implement constants with how we add lots of parameters as we go through the controller. If I recall correctly, you cannot add array elements to a constant once set, so we'd have to do all the logic in an array just to set it as a constant, not sure if that's actually going to improve anything

cisco87: Yeah I didn't see it, but what I was more concerned about is the use of the 'articles' index as a magic value without a constant, on live 34, imho should be a constant, what do you think?

cisco87: No what I didn't mean to use a constant array what I meant was

const ARTICLES_KEY /*or something more meaningfully self-documenting  */ = 'articles';
$parameters = [self::ARTICLES_KEY => $articles];
$parameters[self::ARTICLES_KEY] = blabla;

as described here: https://en.wikipedia.org/wiki/Magic_number_(programming)

BjornTwachtmann: @cisco87 you mean use constants for the keys? Then I have to do something like `constant('Namespace\\Classname::CONSTANT_NAME')` every time I want to access a template variable. I don't see a huge benefit to that. These variable names aren't going to change and they're not shared across templates.

cisco87: Well this index is being used 3 times, and most of the problems always come after the thought "it's never going to change" :D

cisco87: It was just a suggestion anyway right, I will just solve the conversation if I am the only one thinking it, sorry

src/Controller/Articles/IndexController.php
votemike: How about a constant on the controller? Or would a parameter do?

src/Controller/Articles/IndexController.php
votemike: I'm not sure what you mean.

src/ExternalApi/Isite/Service/ArticleService.php
votemike: This one will stay as an int

templates/articles/partials/article_list.html.twig
benolot: Didn't we extract a bunch of other shared code between the two? Would make sense to move this too then

src/Controller/Articles/IndexController.php
cisco87: If it was me I would change it to:

class Controller {
   const MAXIMUM_DISPLAYABLE_RESULTS = 48;

  if ($articlesResult->getTotal() > self::MAXIMUM_DISPLAYABLE_RESULTS) {
                 $parameters['paginatorPresenter'] = new PaginatorPresenter($this->getPage(), self::MAXIMUM_DISPLAYABLE_RESULTS, $articlesResult->getTotal()); 
  }
}

and also below

$limit = self::MAXIMUM_DISPLAYABLE_RESULTS,

I think this way you can achieve easier changeability of the limit and documentation for the value itself

No?

src/ExternalApi/Isite/Service/ArticleService.php
cisco87: I replied above about it, but since it is in another layer I wouldn't be too much worried about leaving it this way, also because I mean making the Service layer dependant on a constant in the controller layer would break the Dependency inversion principle and create spaghetti situation, and placing in the Service a constant which represents the number of items which can be displayed on the template would be bad too, so I think I agree with you in leaving this one the way it is

cisco87: Sorry for commenting then, I didn't think enough about the possible outcomes in the first place

src/Controller/Articles/IndexController.php
cisco87: This one was related to the suggestion of using constants labels for this array keys, but apparently I am the only one who thinks it would be an improvement to use them in place of magic values, so just ignore it if you don't agree either

templates/articles/partials/article_list.html.twig
votemike: Just looked in to this. It doesn't actually look that nice, so I'm going to leave it as-is.

votemike:
Test this please

PR #752, opened 2018-09-27
[PROGRAMMES-6603]  articles page

tomwinnington:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6603


programmes-developers:

src/Controller/Articles/ShowController.php
```diff
@@ -56,6 +56,11 @@ public function __invoke(string $key, string $slug, Request $request, ArticleSer
             $this->setBrandingId($article->getBrandingId());
         }
 
+        $extraProfiles = $article->getParents();
+        $siblingPromise = $isiteService->setChildProfilesOn($extraProfiles, $article->getProjectSpace());
+        $this->resolvePromises([$siblingPromise]);
+
+
```
votemike: TOO MANY BLANK LINES!!!!!!! ARGH!!!!!!

src/Controller/Articles/ShowController.php
```diff
@@ -56,6 +56,11 @@ public function __invoke(string $key, string $slug, Request $request, ArticleSer
             $this->setBrandingId($article->getBrandingId());
         }
 
+        $extraProfiles = $article->getParents();
```
votemike: `$extraProfiles` was a bad name before. These could be called `$parents`

src/Controller/Articles/ShowController.php
```diff
@@ -56,6 +56,11 @@ public function __invoke(string $key, string $slug, Request $request, ArticleSer
...
+        $siblingPromise = $isiteService->setChildProfilesOn($extraProfiles, $article->getProjectSpace());
+        $this->resolvePromises([$siblingPromise]);
```
votemike: This could be combined with the resolvePromise line below. Ask me if you need help.

templates/articles/show.html.twig
```diff
@@ -45,8 +45,20 @@
...
             {% if article.getChildren() %}
                 <div class="br-box-page prog-box">
-                    Children go here
-                    <div class="component">Pagination</div>
+                    <div class="component component--box component--box--primary">
+                        <div class="component__body br-box-page">
+                            <ul class="grid-wrapper highlight-box-wrapper highlight-box-wrapper--grid">
+                                {%- for child in article.getChildren() -%}
+                                    <li class="grid 1/2@bpb1 1/3@bpb2 1/3@bpw 1/4@bpw2 1/4@bpe">
+                                        {{ ds2013('article', child, {'show_synopsis': true, 'heading_level': 'h2'}) }}
```
votemike: show_synopsis is true by default and the heading_level is h2 by default, so you don't need to pass that option in.

templates/articles/show.html.twig
```diff
@@ -55,7 +67,27 @@
...
+                                    <a href="{{ path('programme_article', {'key': parent.getKey(), 'slug': parent.getSlug()}) }}">{{ parent.getTitle() }}</a>
+                                </h2>
+                            </div>
+                            <div class="component__body">
+                                {% if parent.getChildren() %}
+                                    <ul class="grid-wrapper highlight-box-wrapper highlight-box-wrapper--grid">
+                                        {%- for child in parent.getChildren() -%}
+                                            <li class="grid 1/2@bpb1 1/3@bpb2 1/3@bpw 1/4@bpw2 1/5@bpe">
+                                                {{ ds2013('article', child, {'show_synopsis': false, 'heading_level': 'h3'}) }}
+
```
votemike: ANOTHER SPARE LINE!!!!!!!

templates/articles/show.html.twig
```diff
@@ -45,8 +45,20 @@
             {% endif %}
             {% if article.getChildren() %}
                 <div class="br-box-page prog-box">
-                    Children go here
-                    <div class="component">Pagination</div>
+                    <div class="component component--box component--box--primary">
+                        <div class="component__body br-box-page">
+                            <ul class="grid-wrapper highlight-box-wrapper highlight-box-wrapper--grid">
+                                {%- for child in article.getChildren() -%}
```
BjornTwachtmann: So, having a quick read of the V2 code for this. There is actually pagination on this bit when there are more than 48 child articles

src/Controller/Articles/ShowController.php
```diff
@@ -56,8 +56,10 @@ public function __invoke(string $key, string $slug, Request $request, ArticleSer
             $this->setBrandingId($article->getBrandingId());
         }
 
+        $parents = $article->getParents();
+        $siblingPromise = $isiteService->setChildProfilesOn($parents, $article->getProjectSpace());
```
BjornTwachtmann: I think you will need to implement pagination on this one unfortunately. It's something that will come up. Not massively easy to test, I created a bunch of child articles of "Testing Embeds Galore" and reduced the page size on my sandbox to 4.

Basically, where there are more than 48 children, you need to create a paginationPresenter. You should be able to get totals from the result object (which should be returned by the setChildProfileOn promise, I think).

Can chat about this on Monday if it's unclear

templates/articles/show.html.twig
votemike: Done as you can see below.

src/Controller/Articles/ShowController.php
```diff
@@ -56,10 +57,13 @@ public function __invoke(string $key, string $slug, Request $request, ArticleSer
             $this->setBrandingId($article->getBrandingId());
         }
 
-        $childPromise = $isiteService->setChildProfilesOn([$article], $article->getProjectSpace());
-        $this->resolvePromises([$childPromise]);
+        $parents = $article->getParents();
+        $siblingPromise = $isiteService->setChildProfilesOn($parents, $article->getProjectSpace()); //if more than 48, extras are removed
```
BjornTwachtmann: Is this getting children for _all_ the parents? Wouldn't we only need the first one? Or am I just confused.

votemike: Yes, all parents. Why would it be only the first one?

src/ExternalApi/Isite/Service/ArticleService.php
```diff
@@ -48,7 +48,7 @@ public function setChildProfilesOn(
             $cacheKeys[] = $article->getFileId();
         }
 
-        $cacheKey = $this->clientFactory->keyHelper(__CLASS__, __FUNCTION__, implode(',', $cacheKeys), $page, $limit);
+        $cacheKey = $this->clientFactory->keyHelper(__CLASS__, __FUNCTION__, implode(',', $cacheKeys), $page, $limit) ;
```
Tasiobg: Looks like an extra space has been added here, not breaking anything but will look more consistent without the extra space.

Tasiobg:
If I go to this page on the sandbox http://www.bbc.co.uk/programmes/articles/RHLqYBSkQdMNgQzvQvTs2k/top-50-stories-2017-age-5-to-9

I can't see this line
"Just click the links below to listen to and read all the stories..."

Is this suppose to be implement in this ticket or somewhere else?

src/Controller/Articles/ShowController.php
BjornTwachtmann: Sweet holy jebus, we have multiple inheritance.

PR #754, opened 2018-09-27
Programmes 6647 implement telescope vote content block

cisco87:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6647

I need to fix the styling/markup and I need some clarification from @claudioneves but I think the code would be the one you see so I guess it's good to see the PR


programmes-developers:

src/Ds2013/Presenters/Domain/ContentBlock/Telescope/telescope.html.twig
```diff
@@ -0,0 +1,20 @@
+{#{% extends "@Ds2013/Presenters/Domain/ContentBlock/content_block.html.twig" %}#}
```
votemike: Should this be commented out?

src/ExternalApi/Isite/Domain/ContentBlock/Telescope.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace App\ExternalApi\Isite\Domain\ContentBlock;
+
+class Telescope extends AbstractContentBlock
+{
+    /**
+     * @var string
```
votemike: Could we have single-line docblock for class properties please.

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
```diff
@@ -164,6 +165,17 @@ public function getDomainModel(SimpleXMLElement $isiteObject)
                     $rows
                 );
                 break;
+            case 'programmes-telescope-vote':
```
votemike: Isn't "programmes" stripped from the front of the type?

src/Ds2013/Presenters/Domain/ContentBlock/Telescope/telescope.html.twig
```diff
@@ -0,0 +1,20 @@
...
+
+{% block blockContent %}
+
+    {% set container_html_id = 'vote-telescope-' ~ random() %}
+
+    <div class="component component--box component--box--striped component--box-flushbody component--box--secondary">
+        <div class="component component--box component--box--primary">
+            {% include 'shared/telescope/target_element.html.twig' with {
```
votemike: Would this be better as a presenter or something? We don't really use `include` much.

src/Ds2013/Presenters/Domain/ContentBlock/Telescope/telescope.html.twig
cisco87: No I've seen the current one from V2 and apparently it doesn't use the common style as others blocks (i.e. doesn't have the title header), but is presented as-is without any surrounding/background, so I guess that was the easiest to just display it without extending the parent block

src/Ds2013/Presenters/Domain/ContentBlock/Telescope/telescope.html.twig
cisco87: Well it's a shared block of code, you don't use include that much, but I guess if you have a shared partial that only contains a template with 2 variables and using include make it shared between 2 components with 5 lines of markup, there isn't much sense to call the cavalry?

PR #755, opened 2018-09-28
changed the description tag

tomwinnington:
For schedules the meta_context.description()  returned  "" and a search engine would just show the page content.


programmes-developers:

votemike:
I see what you've gone for here. I'm not sure that's the best way to go about it.
I think that the you should leave the MetaContext spitting out a blank description if it's blank.
But for the schedule pages we should somehow override the Meta Description. You can see in BaseController that the method $this->getMetaNoIndex() is called and passed in to the MetaContext as the 3rd param. I think it might be an idea to add a 4th param for an overridden description. Then set that in the Day and Week Controllers.

src/ValueObject/MetaContext.php
```diff
@@ -38,12 +38,11 @@ class MetaContext
     /** @var  bool */
     private $metaNoIndex;
 
-    public function __construct($context = null, string $canonicalUrl = '', bool $metaNoIndex = false)
+    public function __construct($context = null, string $canonicalUrl = '', bool $metaNoIndex = false, $optionalDescription = null)
```
votemike: I'd call this overriddenDescription instead.

src/Ds2013/Presenters/Pages/Schedules/ByWeekPage/schedules_by_week_page.html.twig
```diff
@@ -1,3 +1,4 @@
+<meta name="description" context="This is the broadcast schedule for {{ schedules_by_week_page.service().name() }}">
```
votemike: This meta tag no longer needs to be here.

src/Ds2013/Presenters/Pages/Schedules/ByDayPage/schedules_by_day_page.html.twig
```diff
@@ -3,6 +3,7 @@
     '1/2@bpw2 1/2@bpe': schedules_by_day_page.hasSiblingServiceLink(),
 }) -%}
 
+<meta name="description" context="This is the broadcast schedule for {{ schedules_by_day_page.service().name() }}">
```
votemike: This meta tag no longer needs to be here.

src/Controller/BaseController.php
```diff
@@ -224,7 +224,7 @@ protected function renderWithChrome(string $view, array $parameters = [])
 
         $parameters = array_merge([
             'orb' => $orb,
-            'meta_context' => new MetaContext($this->context, $this->getCanonicalUrl(), $this->getMetaNoIndex()),
+            'meta_context' => new MetaContext($this->context, $this->getCanonicalUrl(), $this->getMetaNoIndex(), $optionalDescription),
```
votemike: Instead of passing the description though like this, try and make it work more like `$this->getMetaNoIndex()`

src/Controller/Schedules/ByWeekController.php
```diff
@@ -102,7 +102,7 @@ public function __invoke(
         ];
 
         $serviceIsActiveInThisPeriod = $this->serviceIsActiveDuringWeek($service, $broadcastWeek);
-        $this->overridenDescription = "This is the broadcast schedule for " . $service->getName();
+        $this->overridenDescription = "This is the Weekly broadcast schedule for " . $service->getName();
```
votemike: lower case "W" please

src/Controller/BaseController.php
```diff
@@ -62,6 +62,8 @@
     /** @var bool */
     protected $metaNoIndex;
 
+    protected $overridenDescription;
```
cisco87: No PHPDoc?  Is it mandatory or non-mandatory?( I forgot the word:D)

PR #756, opened 2018-09-28
[PROGRAMMES-6638][fix] - Change from 302 to 301 redirect

tatitati:
@jatinkaswa pointed me that I was redirecting with the wrong code. 

Original PR here:
https://github.com/bbc/programmes-frontend/pull/750


programmes-developers:

PR #757, opened 2018-09-28
[No Ticket][change] - Keeping consistency in booleans for SMP player

tatitati:
Cleaning some code. Seems that SMP player check for text and booleans when checking some stuff. So if we pass an string or a boolean the SMP player still work as expected


programmes-developers:

PR #761, opened 2018-10-03
Programmes 6654 implement quiz content block

cisco87:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6654


src/Ds2013/PresenterFactory.php
```diff
@@ -273,6 +275,9 @@ public function contentBlockPresenter(
         if ($contentBlock instanceof ThirdParty) {
             return new ThirdPartyPresenter($contentBlock, $inPrimaryColumn, $options);
         }
+        if ($contentBlock instanceof Quiz) {
```
votemike: Is it worth making these alphabetical?

src/ExternalApi/IdtQuiz/IdtQuizService.php
```diff
@@ -0,0 +1,76 @@
+<?php
+/**
```
votemike: Probably don't need this docblock

src/ExternalApi/IdtQuiz/IdtQuizService.php
```diff
@@ -0,0 +1,76 @@
...
+
+use App\ExternalApi\Client\HttpApiClientFactory;
+use GuzzleHttp\Promise\PromiseInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Closure;
+use BBC\ProgrammesCachingLibrary\CacheInterface;
+
+class IdtQuizService
+{
+    /** @var Request */
```
votemike: Is it a Request?

src/ExternalApi/Isite/Domain/ContentBlock/Quiz.php
```diff
@@ -0,0 +1,47 @@
...
+class Quiz extends AbstractContentBlock
+{
+    /** @var string */
+    private $name;
+
+    /** @var string */
+    private $quizId;
+
+    /** @var PromiseInterface */
+    private $htmlContentQueryPromise;
```
votemike: This doesn't seem used.

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
```diff
@@ -165,6 +175,20 @@ public function getDomainModel(SimpleXMLElement $isiteObject)
                     $headings,
                     $rows
                 );
+                break;
+            case 'idt-quiz':
+                $quizId = $this->getString($form->content->idt_id);
+                $htmlContent = $this->idtQuizService->getQuizContentPromise($quizId)->wait();
```
votemike: Are we happy `wait`ing here? I can't think of a better way of the top of my head, but may be worth investigating.

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
```diff
@@ -165,6 +175,20 @@ public function getDomainModel(SimpleXMLElement $isiteObject)
...
+
+                // @codingStandardsIgnoreStart
```
votemike: Do you need these here?

tests/ExternalApi/Isite/Mapper/ContentBlockMapperTest.php
```diff
@@ -20,9 +26,68 @@ class ContentBlockMapperTest extends TestCase
...
+                $idtQuizService
+            ),
+            $keyHelper,
+            $ces,
+            $idtQuizService
+        );
+    }
+
+    /**
+     * @dataProvider getTestMappingQuizObjectData
```
votemike: Do you need a provider for just one test?

tests/ExternalApi/Isite/Mapper/ContentBlockMapperTest.php
```diff
@@ -20,9 +26,68 @@ class ContentBlockMapperTest extends TestCase
...
+            ->setMethods(['getQuizContentPromise'])
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $idtQuizService
+            ->expects($this->once())
+            ->method('getQuizContentPromise')
+            ->with($this->equalTo($quizId))
+            ->willReturn($promiseMock);
+        ReflectionHelper::setObjectPropertyValue($this->mapper, 'idtQuizService', $idtQuizService);
```
votemike: Is there no other way to do this?

cisco87: I usually had always some sort of reflector in Tests to do this kind of things, also because instantiating the needed idtQuizService mock for all the tests, even if not needed seems much worse, but I think it's quite a spread convention?

tests/ExternalApi/Isite/Mapper/ContentBlockMapperTest.php
cisco87: I think I am used to label/have test data coming from a dataProvider to have the test code just caring about mocks/asserts, but if you want I can move it to the test code, it just seems cleaner and easier to add new test data if needed to extend, no?

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
cisco87: Apparently no, but I usually always had a problem with fields having _ in that method, so just did it without paying too much attention to the fact that they were now absent, sorry

src/ExternalApi/IdtQuiz/IdtQuizService.php
cisco87: No sorry

src/Ds2013/PresenterFactory.php
cisco87: Probably yes while we don't refactor this

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
cisco87: Well maybe @BjornTwachtmann has an idea?

tests/ExternalApi/Isite/Mapper/ContentBlockMapperTest.php
cisco87: But this was meant as a unit test, for the logic we have control on, not as a integration test, so I am not sure it is not ok?

src/ExternalApi/IdtQuiz/IdtQuizService.php
```diff
@@ -0,0 +1,70 @@
...
+    private function parseResponse(array $responses): string
+    {
+        if (count($responses) <= 0) {
+            return '';
+        }
+
+        /** @var Response $response */
+        $response = $responses[0];
+
+        if ($response->getStatusCode() == \Symfony\Component\HttpFoundation\Response::HTTP_OK) {
```
BjornTwachtmann: You don't need this. Any non-200 has already ended up somewhere else

src/ExternalApi/IdtQuiz/IdtQuizService.php
```diff
@@ -0,0 +1,70 @@
...
+            return $response->getBody()->getContents();
+        }
+
+        return '';
+    }
+
+    private function getQuizUrl(string $quizId): string
+    {
+        return sprintf(
```
BjornTwachtmann: urlencode()?

src/ExternalApi/IdtQuiz/IdtQuizService.php
cisco87: Oh ok got it

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
BjornTwachtmann: Tasio has a method for grouping together all the SQL queries in an article page. I think we should do the same for this, but I'm happy to merge this as is for now...

tests/ExternalApi/Isite/Mapper/ContentBlockMapperTest.php
tatitati: OK,  forget my comment, is just that I have a different opinion, because I see guzzle being mocked, and reflection helpers,  and mocks returning mocks in the same test, but is just my opinion. I don't want to be over-rule.

BjornTwachtmann: Why is it necessary to use reflection here at all? I don't understand, can you pop over and explain to me when you have a sec?

src/ExternalApi/IdtQuiz/IdtQuizService.php
cisco87: I'm fairly sure that it would damage the functionality ;o

config/services_test.yaml
```diff
@@ -13,6 +13,7 @@ parameters:
     app.tupac.service.class: App\ExternalApi\Tupac\Service\TupacStubService
     app.isite.article_service.class: App\ExternalApi\Isite\Service\ArticleFakeService
     app.isite.profile_service.class: App\ExternalApi\Isite\Service\ProfileFakeService
+    app.services.smallprox.endpoint: 'https://smallprox.vj.tools.bbc.co.uk'
```
BjornTwachtmann: This file is for unit tests. We want a stub to prevent this from making HTTP requests in controller tests, not an endpoint.

config/services_prod_test.yaml
```diff
@@ -9,5 +9,5 @@ parameters:
     app.recipes.baseurl: 'https://api.test.bbc.co.uk/food/recipes'
     app.morph.baseurl: 'https://morph.test.api.bbci.co.uk'
     app.sounds_nav.baseurl: 'https://sounds-nav.test.api.bbci.co.uk/json'
-
     app.vote.accountwrapperurl: 'https://account-wrapper.test.files.bbci.co.uk/vote-wrapper.min.js'
+    app.services.smallprox.endpoint: 'https://smallprox.vj.tools.bbc.co.uk'
```
BjornTwachtmann: You don't need to add this if it's the same as services.yaml. Only to override. Same for stage/dev.

config/parameters.yaml.dist
```diff
@@ -24,3 +24,4 @@ parameters:
     fixture_secure_key: developer
     receng_video_key: 'videokey'
     receng_audio_key: 'audiokey'
+    app.services.smallprox.endpoint: 'https://smallprox.vj.tools.bbc.co.uk'
```
BjornTwachtmann: You don't need this.

config/services_dev.yaml
```diff
@@ -1,6 +1,5 @@
 parameters:
     app.metric_backend.class: App\Metrics\Backend\LogMetricBackend
-
     app.vote.accountwrapperurl: 'https://account-wrapper.test.files.bbci.co.uk/vote-wrapper.min.js'
-
     app.cosmos.dials.path: '../dials.json'
+    app.services.smallprox.endpoint: 'https://smallprox.vj.tools.bbc.co.uk'
```
votemike: Do we need this if this value is already in config/parameters.yaml.dist

PR #764, opened 2018-10-03
[PROGRAMMES-6666][fix] - Link location parameter for stats was not passing proper value

tatitati:
We had all the values in place for link_location parameter (stats), however seems that we forgot to pass them through the different layers of presenters

## URLs to test:
tv prog: http://sandbox.bbc.co.uk/programmes/p04yly99?__flush_cache=1
radio prog: http://sandbox.bbc.co.uk/programmes/b01slvgp?__flush_cache=1


programmes-developers:

PR #765, opened 2018-10-04
Programmes 6656 implement interactive game content box

cisco87:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6656


programmes-developers:

src/Ds2013/Presenters/Domain/ContentBlock/Game/game.html.twig
```diff
@@ -0,0 +1,13 @@
...
+{% set game_params = {
+    'width': content_block.getBlock().getWidth(),
+    'height': content_block.getBlock().getHeight(),
+    'path': content_block.getBBCDownloadsUrl() ~ content_block.getBlock().getPath()
+} %}
+
+<div id="og-frame-holder"></div>
+<script>
+    require(["kite/embed"], function(embed) {
+        embed({{ loader_url|json_encode|raw }}, {{ game_params|json_encode|raw }});
```
tatitati: do we need to json_encode a url? (I'm just ignorant)

BjornTwachtmann: This is the right way to make sure the parameter is correctly formatted and quoted for including in JS

cisco87: I saw how embed() is called on V2, yeah it gets the json_encode(d) version of the arguments ;o

src/ExternalApi/Isite/Mapper/Mapper.php
```diff
@@ -77,6 +77,11 @@ protected function getString(?SimpleXMLElement $val): ?string
         return trim($val);
     }
 
+    protected function getInt(?SimpleXMLElement $val): ?int
```
benolot: This will never return null, if you pass a null into intval it just returns with 0

src/Ds2013/Presenters/Domain/ContentBlock/Game/GamePresenter.php
```diff
@@ -0,0 +1,20 @@
...
+use App\ExternalApi\Isite\Domain\ContentBlock\Game;
+
+class GamePresenter extends ContentBlockPresenter
+{
+    public function __construct(Game $block, bool $inPrimaryColumn, array $options = [])
+    {
+        parent::__construct($block, $inPrimaryColumn, $options);
+    }
+
+    public function getBBCDownloadsUrl(): string
```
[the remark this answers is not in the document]
cisco87: It is used twice in the template, so I was thinking to don't have the magic string duplicated, a constant would have been better, but the syntax for getting constants values in the template is not nice, so I created a method and I can use it both places, I could also have used a twig variable, but I think the way the Presenter pattern is conceived is for passing data to the template, so I thought it was ok

templates/base_ds2013.html.twig
```diff
@@ -54,7 +54,9 @@
                     'cookieStore': "{{ asset_js("js/cookieStore.js") }}",
                     'uasclient': "https://static.bbc.co.uk/uasclient/0.1.42/modules/uasclient",
                     'relay-1': "https://static.bbc.co.uk/frameworks/relay/0.2.6/sharedmodules/relay-1",
-                    'third-party': "{{ asset_js("js/third-party.js") }}"
+                    'third-party': "{{ asset_js("js/third-party.js") }}",
+                    'embed': 'http://play.bbc.co.uk/play/sdk/embed/v1/latest',
```
benolot: 1.) HTTPS! 🔒 
2.) I get a 404 on this address, is it correct?

src/Ds2013/Presenters/Domain/ContentBlock/Game/GamePresenter.php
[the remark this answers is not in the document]
benolot: Ah fair enough - I didn't see it used twice only once! My bad!

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
```diff
@@ -200,6 +201,18 @@ public function getDomainModel(SimpleXMLElement $isiteObject)
...
                 // @codingStandardsIgnoreEnd
                 break;
+            case 'kitegame':
+                // @codingStandardsIgnoreStart
+                $contentBlock = new Game(
+                    $this->getString($form->content->title),
+                    $this->getString($form->metadata->name),
+                    $this->getString($form->content->game_loader_url),
+                    $this->getString($form->content->path),
+                    $this->getInt($form->content->width),
```
Tasiobg: I think "with" and "height" are string, this is an example of v2 paramters passed to the JS
```{"height":"577px","width":"944px","path":"\/\/downloads.bbc.co.uk\/radio\/games\/h2g2\/"}```
Please could you double "height" and "width" are right and works fine?

**Comment by Tasiobg** [comment]
I think calling it "Game" could be too generic, we have different games systems and we could even have more in a future, what do you think @BjornTwachtmann should we use something more specific? I can only think of calling it "InteractiveActivity" which doesn't help much but at least match the naming used by iSite.

**Comment by cisco87** [comment]
> I think calling it "Game" could be too generic, we have different games systems and we could even have more in a future, what do you think @BjornTwachtmann should we use something more specific? I can only think of calling it "InteractiveActivity" which doesn't help much but at least match the naming used by iSite.

Maybe KiteGame?

**Comment by benolot** [comment]
> > I think calling it "Game" could be too generic, we have different games systems and we could even have more in a future, what do you think @BjornTwachtmann should we use something more specific? I can only think of calling it "InteractiveActivity" which doesn't help much but at least match the naming used by iSite.
> 
> Maybe KiteGame?

Kite no longer make the games, they just had the contract to port them over

**Comment by cisco87** [comment]
> > > I think calling it "Game" could be too generic, we have different games systems and we could even have more in a future, what do you think @BjornTwachtmann should we use something more specific? I can only think of calling it "InteractiveActivity" which doesn't help much but at least match the naming used by iSite.
> > 
> > 
> > Maybe KiteGame?
> 
> Kite no longer make the games, they just had the contract to port them over



> > > I think calling it "Game" could be too generic, we have different games systems and we could even have more in a future, what do you think @BjornTwachtmann should we use something more specific? I can only think of calling it "InteractiveActivity" which doesn't help much but at least match the naming used by iSite.
> > 
> > 
> > Maybe KiteGame?
> 
> Kite no longer make the games, they just had the contract to port them over

KitePortedGame?

**Comment by Tasiobg** [comment]
@cisco87 I think using iSite name "InteractiveActivity" will keep it more consisten otherwise could be and issue to related the class "Game" in our code with the isite component "InteractiveActivity"

**Comment by cisco87** [comment]
> @cisco87 I think using iSite name "InteractiveActivity" will keep it more consisten otherwise could be and issue to related the class "Game" in our code with the isite component "InteractiveActivity"

Gotcha! Done

File: src/Ds2013/Presenters/Domain/ContentBlock/InteractiveActivity/InteractiveActivityPresenter.php [hunk 6]
```diff
@@ -0,0 +1,20 @@
...
+class InteractiveActivityPresenter extends ContentBlockPresenter
+{
+    public function __construct(InteractiveActivity $block, bool $inPrimaryColumn, array $options = [])
+    {
+        parent::__construct($block, $inPrimaryColumn, $options);
+    }
+
+    public function getBBCDownloadsUrl(): string
+    {
+        return '//downloads.bbc.co.uk/';
```
votemike: Is this useful? Would it be better to just hardcode it in the template?

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
```diff
@@ -222,6 +223,18 @@ public function getDomainModel(SimpleXMLElement $isiteObject)
                 );
                 // @codingStandardsIgnoreEnd
                 break;
+            case 'kitegame':
+                // @codingStandardsIgnoreStart
```
votemike: Is this needed? Which line is the CS checker complaining about?

src/Ds2013/Presenters/Domain/ContentBlock/InteractiveActivity/InteractiveActivityPresenter.php
Tasiobg: It seems to be used twice in the template so probably is not a bad idea, I think this is more of a personal preference or at least I don't really know which approach is really best.

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
cisco87: game_loader_url we need camelCase but the xml has underscored params

votemike: I see. OK then.

src/Ds2013/Presenters/Domain/ContentBlock/InteractiveActivity/InteractiveActivityPresenter.php
votemike: I would just hardcode it I think. Don't know what others think.

Tasiobg: I found any of this options good enought:
- Use the existing getBBCDownloadsUrl()
- Set a variable inside interactive_activity.html.twig 
- Hardcode the URL into interactive_activity.html.twig twice

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
Tasiobg: This change seems generating exactly same js parameters as before. I think you also need to change `interactive_activity.html.twig`.
Please also test if the fix you do works fine by editing the values in test iSite and checking if this affects the size of the game

src/Ds2013/Presenters/Domain/ContentBlock/InteractiveActivity/interactive_activity.html.twig
```diff
@@ -0,0 +1,13 @@
+{% set loader_url = content_block.getBBCDownloadsUrl() ~ content_block.getBlock().getLoaderUrl() %}
+{% set game_params = {
+    'width': content_block.getBlock().getWidth(),
+    'height': content_block.getBlock().getHeight(),
+    'path': content_block.getBBCDownloadsUrl() ~ content_block.getBlock().getPath()
```
BjornTwachtmann: path is optional in V2, as in, if it's not present, the variable won't be added to the game parameters. I think this needs to do the same

cisco87: Got it!

config/services.yaml
```diff
@@ -27,7 +27,7 @@ parameters:
     app.electron.service.class: App\ExternalApi\Electron\Service\ElectronService
     app.favourites.baseurl: 'https://www.bbc.co.uk/modules/plugin/musicfavourite/init'
     app.favourites.service.class: App\ExternalApi\FavouritesButton\Service\FavouritesButtonService
-    app.isite.baseurl: 'https://api.live.bbc.co.uk/isite2-content-reader'
+    app.isite.baseurl: 'https://api.test.bbc.co.uk/isite2-content-reader'
```
BjornTwachtmann: Please don't commit this :)

BjornTwachtmann:
Other than changing the iSite URL, looks good.

src/Ds2013/Presenters/Domain/ContentBlock/InteractiveActivity/InteractiveActivityPresenter.php
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    public function getParams()
```
votemike: `public function getParams(): array`?

PR #767, opened 2018-10-05
IdtQuizStubService

cisco87:
bjorn [12:29 PM]
@Francesco (The Italian one) the stub is for controller tests where the entire rendering stack is initialised, you can't really use a simple mock in that case, but you still don't want things running off and making real HTTP requests from the unit test layer.
It's a pretty simple thing, see


programmes-developers:

config/services.yaml
```diff
@@ -25,9 +25,10 @@ parameters:
     app.ada.programme_service.class: App\ExternalApi\Ada\Service\AdaProgrammeService
     app.electron.baseurl: 'https://api.live.bbc.co.uk/electron'
     app.electron.service.class: App\ExternalApi\Electron\Service\ElectronService
+    app.idtquiz.service.class: App\ExternalApi\IdtQuiz\IdtQuizService
     app.favourites.baseurl: 'https://www.bbc.co.uk/modules/plugin/musicfavourite/init'
     app.favourites.service.class: App\ExternalApi\FavouritesButton\Service\FavouritesButtonService
-    app.isite.baseurl: 'https://api.live.bbc.co.uk/isite2-content-reader'
+    app.isite.baseurl: 'https://api.test.bbc.co.uk/isite2-content-reader'
```
Tasiobg: I think this change will make the LIVE page to use TEST iSite2

src/ExternalApi/IdtQuiz/IdtQuizStubService.php
```diff
@@ -0,0 +1,14 @@
...
+namespace App\ExternalApi\IdtQuiz;
+
+use GuzzleHttp\Promise\FulfilledPromise;
+use GuzzleHttp\Promise\PromiseInterface;
+
+class IdtQuizStubService extends IdtQuizService
+{
+    public function getQuizContentPromise(string $quizId): PromiseInterface
+    {
+        return new FulfilledPromise([]);
```
BjornTwachtmann: Shouldn't this be returning an empty string? The IDT quiz class doesn't return arrays does it?

cisco87: Oh yeah ok I thought this had to be considered list of results

PR #769, opened 2018-10-08
PROGRAMMES-6674 Fixing CTA padding

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6674

I _think_ this will fix it, I have tested with various screen widths and it all seems to be OK.


src/Ds2013/Presenters/Domain/CoreEntity/Programme/_programme.scss
```diff
@@ -38,7 +38,7 @@ $programme-icon-bpw-box-size: gel-typography('bpw', 'delta', 'line-height') + 18
     &.programme__img--standalone-cta {
         float: right;
         margin-right: 0;
-        margin-left: 0;
+        margin-left: 8px;
```
cisco87: You can't achieve the same result by using 0.8rem?

votemike: We don't use rem for margins

BjornTwachtmann: Can you use the variable here though, $basic-spacing, rather than hardcoding. Other than that, seems perfect.

PR #771, opened 2018-10-09
PROGRAMMES-6648 Making pages 500 if iSite fails

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6648


src/Controller/Articles/IndexController.php
```diff
@@ -23,6 +23,10 @@ public function __invoke(CoreEntity $coreEntity, ArticleService $isiteService)
             $parameters['programme'] = $coreEntity; //so the the base 2013 template sets the footer
 
             $articlesResult = $isiteService->getByProgramme($coreEntity, $this->getPage())->wait();
+            if ($articlesResult === null) {
+                $this->response()->setStatusCode(500);
+                return $this->renderWithChrome('articles/index.html.twig', $parameters); //@TODO do we want to show the page? Or fire bubbles?
```
votemike: Do we still want to render the page if the articles aren't retrieveable from iSite? It will display the message "There are no features for X" if we do.

programmes-developers:

BjornTwachtmann:
Won't this also 500 if iSite throws a 404? Or am I missing something obvious?

votemike:
It should only equal null if there is an error. If there are no results, it should have an empty IsiteResult, which will then make the page 404.

votemike:
@BjornTwachtmann I have updated with your suggestions

PR #772, opened 2018-10-09
PROGRAMMES-6613 Cleaning up meta descriptions

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6613


programmes-developers:

src/Controller/Articles/IndexController.php
```diff
@@ -36,6 +36,8 @@ public function __invoke(CoreEntity $coreEntity, ArticleService $isiteService)
             $parameters['articles'] = $articles;
         }
 
+        $this->overridenDescription = 'Articles about ' . $coreEntity->getSearchTitle();
```
BjornTwachtmann: Why are you using searchTitle?

votemike: Because this string will appear when searching Google. It sounded like the right thing to use. Is it not? Should I just use the plain title?

src/Controller/Podcast/EpisodesDownloadsController.php
```diff
@@ -25,6 +25,8 @@ public function __invoke(string $pid, ProgrammesService $programmesService, Pres
 
         $this->setContextAndPreloadBranding($programme);
 
+        $this->overridenDescription = 'Downloads for ' . $programme->getSearchTitle();
```
BjornTwachtmann: Might be good to mention podcasts in this one. 

"Podcast downloads for" or something maybe? It's not live yet anyway, but just a thought.

src/Controller/Articles/IndexController.php
BjornTwachtmann: Search Title, looking at it, is actually something I created on the backend for use in the MySQL search. It's got punctuation and a few other things stripped out.

I have literally no idea why Ben added it to the domain models, it is completely useless to the frontend. Title is much more correct.

src/Controller/ProgrammeEpisodes/GuideController.php
```diff
@@ -68,6 +68,8 @@ public function __invoke(
 
         $schema = $this->getSchema($structuredDataHelper, $programme, $children, $upcomingBroadcasts);
 
+        $this->overridenDescription = 'All episodes of ' . $programme->getSearchTitle();
```
BjornTwachtmann: How do we deal with the case of an episode guide for a series? Normally there's some logic that constructs a title from the heirarchy, but I'm really not sure that belongs in the controller either.

votemike: I guess it would say something like "All episodes of Series 11". Would you like these to all have the full title name?

PR #776, opened 2018-10-15
Programmes 6610 fix twitter embeddability

BjornTwachtmann:


programmes-developers:

src/Controller/AncillaryPages/PlayerController.php
```diff
@@ -23,6 +23,8 @@ public function __invoke(
         VersionsService $versionsService,
         Clip $clip
     ) {
+        // This frame needs to display off the BBC site
+        $this->response()->headers->remove('X-Frame-Options');
```
Tasiobg: What do you think about only removing this header when `$clip->isExternallyEmbeddable()` is true?

BjornTwachtmann: This page is only shown when isExternallyEmbeddable is true

BjornTwachtmann: it's 404 otherwise

PR #777, opened 2018-10-16
fix without chrome branding

Tasiobg:
fix and update alarms as well


programmes-developers:

BjornTwachtmann:
Make sure to merge the PR into master

PR #778, opened 2018-10-17
Enable radionav in the sandbox

Tasiobg:
This will enable the radio nav in the sandbox and also make the default value true.


programmes-developers:

PR #779, opened 2018-10-17
[PROGRAMMES-6693] Fix profile pages 500 when PID is unavailable

benolot:
Ticket link: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6693


programmes-developers:

src/Controller/Profiles/ShowController.php
```diff
@@ -46,7 +46,7 @@ public function __invoke(string $key, string $slug, Request $request, ProfileSer
         if ('' !== $profile->getParentPid()) {
             $context = $coreEntitiesService->findByPidFull(new Pid($profile->getParentPid()));
 
-            if ($profile->getProjectSpace() !== $context->getOption('project_space')) {
+            if ($context && $profile->getProjectSpace() !== $context->getOption('project_space')) {
```
cisco87: Maybe you should do $context instanceof WhatItHasToBe?

benolot: Not sure it's too easy to do that, because a PID could be a brand, could be a Series, could be an episode. Unless maybe we could check just if it's a core entity?

benolot: @BjornTwachtmann what are your thoughts on this?

cisco87: Then you could check it towards an interface?

BjornTwachtmann: See the typehint on the findByPidFull method. It can be null or a Programme, so checking for a truthy value is adequate IMO.

PR #782, opened 2018-10-18
Clips favourite button colour

BjornTwachtmann:
* Fix colour and title of favourites button on clip page
* Abstract favourites JS to partial so changes apply everywhere


programmes-developers:

PR #789, opened 2018-10-23
[PROGRAMMES-6663] If theres a contact form, redirect them away!

benolot:
Ticket: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6663


config/routes/3rd_party.yaml
```diff
@@ -83,6 +83,12 @@ programme_clips:
     requirements: { pid: '[0-9b-df-hj-np-tv-z]{8,15}' }
     schemes: [https]
 
+article_with_contact_form:
+    path: /programmes/articles/{key}/{slug}/contact
+    defaults: { _controller: FrameworkBundle:Redirect:redirect, route: '' }
```
BjornTwachtmann: We have our own controller for issuing cached redirects which works the same way. See https://github.com/bbc/programmes-frontend/blob/develop/config/routes.yaml#L28

BjornTwachtmann: These redirect won't be cacheable otherwise

src/Controller/Articles/ShowController.php
```diff
@@ -27,8 +28,13 @@ public function __invoke(string $key, string $slug, Request $request, ArticleSer
...
         $guid = $isiteKeyHelper->convertKeyToGuid($key);
 
-        /** @var IsiteResult $isiteResult */
-        $isiteResult = $isiteService->getByContentId($guid, $preview)->wait(true);
+        $isiteResult = null;
+        try {
+            /** @var IsiteResult $isiteResult */
+            $isiteResult = $isiteService->getByContentId($guid, $preview)->wait(true);
+        } catch (HasContactFormException $e) {
+            return $this->cachedRedirectToRoute('article_with_contact_form', ['key' => $key, 'slug' => $slug], 307);
```
BjornTwachtmann: Why a 307 rather than a 302 here. I'm always slightly wary of how browsers might react to less frequently used status codes.

src/Controller/BaseController.php
```diff
@@ -283,15 +283,16 @@ protected function cachedRedirect($url, $status = 302, int $cacheLifetime = null
...
+     * @param int $status The status code to use for the Response
+     * @param int|null $lifetime The varnish cache lifetime
      *
      * @return RedirectResponse
      */
-    protected function cachedRedirectToRoute($route, array $parameters = array(), $status = 302): RedirectResponse
+    protected function cachedRedirectToRoute($route, array $parameters = array(), $status = 302, int $lifetime = null): RedirectResponse
     {
-        return $this->cachedRedirect($this->generateUrl($route, $parameters), $status);
+        return $this->cachedRedirect($this->generateUrl($route, $parameters), $status, $lifetime);
```
BjornTwachtmann: Is this going to break horribly if $lifetime is null, i.e. do you need to change the typehint on cachedRedirect to ?int, or does this work fine? Honestly asking, I cannot remember.

PR #791, opened 2018-10-24
[PROGRAMMES-6693] If someone inputs an incorrect pid, don’t 500…

benolot:
Ticket: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6693

Very simple fix for some "great" twig behavior. If something is null it will be true for "defined", as we use null as our fallback value, we need to add a specific check for that.


programmes-developers:

templates/base_ds2013.html.twig
```diff
@@ -117,7 +117,7 @@
             <div id="programmes-content" class="visually-hidden" tabindex="-1">Main content</div>
             {% block body %}{% endblock %}
         </div>
-        {% if programme is defined %}{{ ds2013('footer', programme) }}{% endif %}
+        {% if programme is defined and programme is not null %}{{ ds2013('footer', programme) }}{% endif %}
```
BjornTwachtmann: Could this not just if `if programme`

benolot: No because if it's not defined that will error

BjornTwachtmann: ffs twig

PR #795, opened 2018-10-25
Programmes 6641 move 301 redirect for pid podcasts partials

cisco87:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6641


programmes-developers:

BjornTwachtmann:
Behaviour for this in V2 is that the redirect is only issued when the PID is a brand or a series. I'm pondering if that matters or if this is better...

PR #800, opened 2018-10-29
PROGRAMMES-6717: Fix text size icon alignment

BjornTwachtmann:
This fixes icon alignment in various places and removes some overrides. I don't _think_ it should break anything but it will need a bit of testing.


Tasiobg:
I notice the issue when having a look at a different PR, I did some testing and looks fine

PR #803, opened 2018-10-30
Programmes 6652 recommendations controller get episode could return error

cisco87:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6652


programmes-developers:

PR #804, opened 2018-10-31
Programmes 6543 super promo first element

cisco87:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6543


programmes-developers:

src/Controller/FindByPid/TlecController.php
```diff
@@ -143,7 +143,7 @@ public function __invoke(
 
         $priorityPromotion = null;
         if ($programme->getOption('brand_layout') === 'promo' && !empty($promotions) && $programme->isTlec() && !$showMiniMap) {
-            $priorityPromotion = $this->extractPriorityPromotionAndUpdateList($promotions);
+            $priorityPromotion = array_shift($promotions);
```
cisco87: in case of empty / non-array `$priorityPromotion` will be null which is compatible with the previous behaviour of the removed unneeded method

PR #805, opened 2018-10-31
[PROGRAMMES-6718] Parent links

tomwinnington:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6718

Based it off of the v2 code. It works, but I'm sure it's not the most elegant way.


programmes-developers:

src/DsShared/Helpers/ParentLinksHelper.php
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+
+    public function parentLinks($programme, $include_self = false, $item_class = 'programme')
+    {
+        $strings = array();
+        if ($programme->getParent()) {
+            $parent = $programme;
+            while (!is_null($parent) && !$parent->isTleo()) {
+                $full_url = 'https://www.bbc.co.uk'. $this->router->generate('find_by_pid', array('pid' => $programme->getPid()));
```
cisco87: I think this might not be correct, we usually always try to go to the hostname we're currently on, so you can use generate with ABSOLUTE_URL const, or am  I wrong?

src/DsShared/Helpers/ParentLinksHelper.php
```diff
@@ -0,0 +1,62 @@
...
+                $schema = '';
+                if ($programme->getPid() != $parent->getPid()) { // don't say partOf when talking about current programme
+                    $schema = ' property="isPartOf" typeof="' . $this->typeToSchema($parent->getType()) . '" resource="' . $full_url . '"';
+                }
+                $strings[] = '<span' . $schema . '><a property="url" href="'. $this->router->generate('find_by_pid', array('pid' => $parent->getPid())) . '" content="' . $full_url . '"><span property="name">' . htmlspecialchars($parent->gettitle()) . '</span></a></span>';
+                $parent = $parent->getParent();
+            }
+            $strings = array_reverse($strings);
```
cisco87: Can you put these templating related stuff into a twig file and get the output of rendering through $twigEnvironment->render()? you can still save it in a string and don't have to use much string interpolation/concatenation, you can pass to the template rendering function a boolean representing the result of logical operations, to keep the template logic-frei, I think it might be better to find a compromise to have the PHP not knowking about the template and the template not knowing about the logical stuff

src/DsShared/Helpers/ParentLinksHelper.php
```diff
@@ -0,0 +1,62 @@
...
+        if (!$include_self) {
+            array_pop($strings);
+        }
+        if (empty($strings)) return null;
+        return '<span class="' . $item_class . '">' . implode(', </span><span class="' . $item_class . '">', $strings) . '</span>';
+    }
+
+    private function typeToSchema($type)
+    {
+        switch ($type) {
```
cisco87: Would you like to create a enum-like entity to use it like 

switch (.//) {
   SomethingTypeEnum::BRAND_TYPE /..etc../?
}

benolot:
Test this please

src/DsShared/Helpers/ParentLinksHelper.php
```diff
@@ -0,0 +1,62 @@
+<?php
+/**
+ * Created by IntelliJ IDEA.
```
benolot: Cut out this comment that IntelliJ loves to stick in everywhere 😄

src/Controller/Profiles/ShowController.php
```diff
@@ -96,4 +96,6 @@ private function getPaginator(IsiteResult $iSiteResult): ?PaginatorPresenter
 
         return new PaginatorPresenter($this->getPage(), 48, $iSiteResult->getTotal());
     }
+
```
benolot: Random white space?

src/Controller/Profiles/ShowController.php
```diff
@@ -15,8 +16,9 @@
 
 class ShowController extends BaseController
 {
-    public function __invoke(string $key, string $slug, Request $request, ProfileService $isiteService, IsiteKeyHelper $isiteKeyHelper, CoreEntitiesService $coreEntitiesService)
+    public function __invoke(string $key, string $slug, Request $request, ProfileService $isiteService, IsiteKeyHelper $isiteKeyHelper, CoreEntitiesService $coreEntitiesService, ParentLinksHelper $parentLinksHelper)
     {
+
```
benolot: Don't add white space to the start of a function - See PSR/2 - I can dig up the relevant section if needed

src/DsShared/Helpers/ParentLinksHelper.php
tomwinnington: Seeing as this is just a helper class I don't think this would be necessary, and would just make it more complicated. Unless anyone disagrees?

BjornTwachtmann:
Hmm. Couple of points here. 
a) This should really be a presenter rather than a helper if it's going to be used in this way (directly from the view). That way it can have a view file too and you don't need to do as much string concatenation.
b) I think it's doing something very similar to the entityContextPresenter, could you see if that could be re-used in some way here?

templates/profiles/group.html.twig
```diff
@@ -6,7 +6,10 @@
             <div class="grid">
                 <div class="br-box-page">
                     <div class="prog-layout__primary island--vertical">
-                        <h1>{{ profile.getTitle() }}</h1>
+                        <h1><span class="context">
+                            {{ ds_shared('entityContext', programme) }}
```
BjornTwachtmann: We need an if block here to check if the context programme exists, as it's nullable. See /programmes/profiles/GnmKv7Rmjt0NxvZdZhbxw6/stupid-test-profile

templates/profiles/group.html.twig
```diff
@@ -6,7 +6,14 @@
...
                     <div class="prog-layout__primary island--vertical">
-                        <h1>{{ profile.getTitle() }}</h1>
+                        
+                        <h1>
+                        {% if programme %}
+                            <span class="context">
+                                {{ ds_shared('entityContext', programme) }}
+                            </span>
+                        {% endif %}
+                        {{ profile.getTitle() }}</h1>
```
cisco87: /h1 on newline and indentation fix?

PR #813, opened 2018-11-07
Programmes 6642 max age

cisco87:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6642


programmes-developers:

benolot:
Does this need to be done? I thought it wasn't an issue that it was being cached when discussed on slack?

PR #816, opened 2018-11-13
[PROGRAMMES-6679] removed dial for sounds

tomwinnington:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6679


programmes-developers:

Tasiobg:
Not sure what is required as part of this task but it we only want to remove the dials for sounds, can we empty the file instead of removing it?, this way it should be easy to add new dials in the future.
What do you think @BjornTwachtmann? or is this task to completly remove the dials integration?

Tasiobg:
I have investigating this with @tomwinnington and we found that if dial agent is present cosmos deployment will fail unless there is at least one valid entry in dials.json

With this PR changes It may not be easy to remember how to add dials again in the future but I can't think of better way.

PR #820, opened 2018-11-15
PROGRAMMES-6735 Content block mapper fix

Tasiobg:


programmes-developers:

Tasiobg:
The whole process still confusing, I'll try to do more refactor and then create a new PR which will incluede this changes

PR #821, opened 2018-11-15
PROGRAMMES-6302 - added ARIA labels for schedule navs

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6302

The nav error is occurring due to `<nav>` being a section element. I don't think the navigation needs a heading as it is fairly obvious what it does. I have added ARIA labels to make the functionality clearer for users of assistive technology.

Any thoughts? Do you think the labels are clear enough?


programmes-developers:

Bobkidbob:
Tool issue

PR #827, opened 2018-11-21
noticed and removed issue from layout page

tomwinnington:
Layout page of the style guide had a random 'a' at the start of the {% extends..... %} tag.


programmes-developers:

PR #828, opened 2018-11-21
[PROGRAMMES-6742] Ati stats stuff

benolot:
Ticket: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6742

To test: you need to enable the ati-stats dial in dials.json (change it from false to true). Happy to explain anything people might have questions over. 

Please read the comments around line 228 and let me know your thoughts.

### DIFF
--- a/dials.json
+++ b/dials.json
@@ -1,4 +1,4 @@
 {
     "//": "Keep this file updated with the defaults from cosmos/dials.json",
-    "ati-stats": "false"
+    "ati-stats": "true"
 }
--- a/dials.json
+++ b/dials.json
@@ -1,4 +1,4 @@
 {
     "//": "Keep this file updated with the defaults from cosmos/dials.json",
-    "ati-stats": "true"
+    "ati-stats": "false"
 }
--- a/dials.json
+++ b/dials.json
@@ -1,0 +1,4 @@
+{
+    "//": "Keep this file updated with the defaults from cosmos/dials.json",
+    "ati-stats": "false"
+}
--- a/src/Controller/BaseController.php
+++ b/src/Controller/BaseController.php
@@ -216,30 +216,20 @@ abstract class BaseController extends AbstractController
         $this->preRender();
 
         $cosmosInfo = $this->container->get(CosmosInfo::class);
-        $analyticsCounterName = null;
-        $istatsAnalyticsLabels = null;
-        $atiAnalyticsLabels = null;
-        $istatsAnalyticsLabelsValues = null;
-
+        $atiAnalyticsLabelsValues = null;
         if ($this->container->get(Dials::class)->get('ati-stats') === 'true') {
-            $atiAnalyticsLabels = new AtiAnalyticsLabels($this->context, $this->istatsProgsPageType);
-            $atiAnalyticsLabels = $atiAnalyticsLabels->orbLabels();
-        } else {
-            $analyticsCounterName = (string) new AnalyticsCounterName($this->context, $this->request()->getPathInfo());
-            $istatsAnalyticsLabels = new IstatsAnalyticsLabels($this->context, $this->istatsProgsPageType, $cosmosInfo->getAppVersion(), $this->istatsExtraLabels);
-            $istatsAnalyticsLabelsValues = $istatsAnalyticsLabels->orbLabels();
-            // Why rename it? Are you mad? That looks awful!
-            // I mean, I'm not denying the madness, but sadly it must be done this way
-            // See line 253, where it creates an instance of ComscoreAnalyticsLabels. This needs an
-            // instance of IstatsAnalyticsLabels not the array of orb labels.
-            // if anyone can think of a cleaner way to handle this, let me know.
+            $atiAnalyticsLabelsValues = new AtiAnalyticsLabels($this->context, $this->istatsProgsPageType);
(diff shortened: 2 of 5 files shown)


src/Controller/BaseController.php
```diff
@@ -215,17 +216,34 @@ protected function renderWithChrome(string $view, array $parameters = [])
...
+            $atiAnalyticsLabels = $atiAnalyticsLabels->orbLabels();
+        } else {
+            $analyticsCounterName = (string) new AnalyticsCounterName($this->context, $this->request()->getPathInfo());
+            $istatsAnalyticsLabels = new IstatsAnalyticsLabels($this->context, $this->istatsProgsPageType, $cosmosInfo->getAppVersion(), $this->istatsExtraLabels);
+            $istatsAnalyticsLabelsValues = $istatsAnalyticsLabels->orbLabels();
+            // Why rename it? Are you mad? That looks awful!
+            // I mean, I'm not denying the madness, but sadly it must be done this way
+            // See line 253, where it creates an instance of ComscoreAnalyticsLabels. This needs an
+            // instance of IstatsAnalyticsLabels not the array of orb labels.
+            // if anyone can think of a cleaner way to handle this, let me know.
```
benolot: Thoughts?

programmes-developers:

src/ValueObject/AtiAnalyticsLabels.php
```diff
@@ -0,0 +1,48 @@
+<?php
```
cisco87: const for the strings?

benolot: What do you mean?

Bobkidbob: The strings will be changed in later tickets, so probably best keep them as variables.

dials.json
```diff
@@ -1,4 +1,4 @@
 {
     "//": "Keep this file updated with the defaults from cosmos/dials.json",
-    "ati-stats": "false"
+    "ati-stats": "true"
```
Bobkidbob: Should this have changed to `true` in the fix commit?

benolot: No, good spot.

src/Controller/BaseController.php
```diff
@@ -215,28 +216,52 @@ protected function renderWithChrome(string $view, array $parameters = [])
...
 
         $cosmosInfo = $this->container->get(CosmosInfo::class);
-        $istatsAnalyticsLabels = new IstatsAnalyticsLabels($this->context, $this->istatsProgsPageType, $cosmosInfo->getAppVersion(), $this->istatsExtraLabels);
-        $analyticsCounterName = (string) new AnalyticsCounterName($this->context, $this->request()->getPathInfo());
+        $analyticsCounterName = null;
+        $istatsAnalyticsLabels = null;
+        $atiAnalyticsLabels = null;
+        $istatsAnalyticsLabelsValues = null;
+
+        if ($this->container->get(Dials::class)->get('ati-stats') === 'true') {
```
Bobkidbob: We should look into what @cisco87 was talking about at some point. Obviously it isn't ideal using `'true'` instead of `true`. This could perhaps be solved by creating a `Dial` class with `isTrue()` or `isOn()` and returning an instance of `Dial` from `Dials->get()` instead of the raw value.

src/ValueObject/AtiAnalyticsLabels.php
cisco87: No I was just saying that we could use a const to mimick a label in place of the string itself, then save it to the variable, which also works as a self-documentation bit

i.e.

const ALLOWED_SERVICE_NID = 'bbc_world_news';

if([...]$this->context->getNetwork()->getNid() === self::ALLOWED_SERVICE_NID[...]){ 


Or
const MASTERBRAND_ID = '...';
        // TODO: This needs to be set based on the masterbrand - seperate ticket incoming
        $producer = self::MASTERBRAND_ID;

src/Cosmos/dials.json
```diff
@@ -0,0 +1,17 @@
+[
+    {
+        "name": "ati-stats",
+        "default-value": "false",
+        "values": [
+            {
+                "value": "true",
+                "description": "Use ATI stats rather than iStats"
```
Tasiobg: As far as I know when this is set to true iStats won't stop working

src/ValueObject/AtiAnalyticsLabels.php
cisco87: https://en.wikipedia.org/wiki/Magic_number_(programming)

The term magic number or magic constant refers to the anti-pattern of using numbers directly in source code. This has been referred to as breaking one of the oldest rules of programming, dating back to the COBOL, FORTRAN and PL/1 manuals of the 1960s.[1] The use of unnamed magic numbers in code obscures the developers' intent in choosing that number,[2] increases opportunities for subtle errors (e.g. is every digit correct in 3.14159265358979323846 and is this equal to 3.14159?) and makes it more difficult for the program to be adapted and extended in the future.[3] Replacing all significant magic numbers with named constants makes programs easier to read, understand and maintain.[4] 

Now use the same approach to strings, which is up to date with our languages, according to Clean Code principles from Uncle Bob ^^

This is what I meant

src/Cosmos/dials.json
benolot: Updated the description to be clearer now the dial changes have happened 😄

PR #834, opened 2018-11-23
[No ticket] Quick fix for broken stats

benolot:
This is why you should assign things slugs...


programmes-developers:

PR #835, opened 2018-11-23
PROGRAMMES-6735 Content block mapping refactor

Tasiobg:
- Preload data (run grouped queries for performance improvements)
- Added galleries to the preload data
- Handle case when the content block is not valid


src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
```diff
@@ -76,21 +83,23 @@ public function __construct(
...
      * @param array $contentBlocksList
-     * @return AbstractContentBlock[]
      */
-    public function getDomainModels(array $contentBlocksList): array
+    public function preloadData(array $contentBlocksList): void
     {
-        $clipsPid = [];
-        // get clips queries
+        $clipsPids = [];
+        $galleriesPids = [];
```
Tasiobg: Are `$clipsPids` and `$galleriesPids` good names? I not sure if the plurar is well used

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
```diff
@@ -376,7 +400,7 @@ public function getDomainModel(SimpleXMLElement $isiteObject)
             case 'contactform':
                 throw new HasContactFormException('Contact form found');
             default:
-//                throw new Exception('Invalid content block type. Found ' . $type);
+                $this->logger->error('Invalid content block type. Found ' . $type);
```
Tasiobg: In case a block type is not an expected case, log the error so we can investigate but don't stop the page from rendering

src/ExternalApi/Isite/Domain/Article.php
```diff
@@ -38,7 +38,7 @@ class Article
     /** @var string */
     private $image;
 
-    /** @var Row[] */
+    /** @var RowGroup[] */
     private $rows;
```
BjornTwachtmann: Is it worth changing this to rowGroups, along with the associated getter? Or is that too much work for marginal gain in readability?

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
BjornTwachtmann: I would say clipPids and galleryPids is more correct pluralisation, but both are understandable.

src/ExternalApi/Isite/Mapper/MapperFactory.php
```diff
@@ -78,9 +78,9 @@ public function createProfileMapper(): ProfileMapper
         return $this->findMapper(ProfileMapper::class);
     }
 
-    public function createRowMapper(): RowMapper
+    public function createRowMapper(): RowGroupMapper
```
BjornTwachtmann: createRowGroupMapper?

BjornTwachtmann:
Looks good to me other than some very minor formatting points.

PR #836, opened 2018-11-23
Programmes 6707 clips page tests

cisco87:


programmes-developers:

tests/Controller/Clips/ClipsListControllerTest.php
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    public function testClipsListControllerCanLoadNoClipsPage()
+    {
+        $crawler = $this->client->request('GET', '/programmes/b006pnjk/clips');
+
+        $this->assertResponseStatusCode($this->client, 200);
+        $this->assertCount(
+            1,
+            $crawler->filter('p.no_clips_page')
```
Bobkidbob: There must be a better way than adding a class that isn't used to test this.

PR #839, opened 2018-11-27
Article controller only needs to query for children of profile _groups_

BjornTwachtmann:


programmes-developers:

PR #844, opened 2018-11-28
PROGRAMMES-6753 - added default titles to clip blocks

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6753


programmes-developers:

Bobkidbob:
Should probably try to get rid of the repetition tomorrow. The same code is also used in the galleries block.

src/Ds2013/Presenters/Domain/ContentBlock/Clip/ClipStandalone/clip_standalone.html.twig
```diff
@@ -6,6 +6,18 @@
     {% endif %}
 {% endblock %}
 
+{% block title %}
```
cisco87: I think you could put them in a shared/partial and pass as argument 1 or 2 like {% include 'shared_name.html.twig' with {'number_of_clips': 1 or 2} %} and read it on the tr('clips', xyz), at least I guess we don't have a duplicate template? what do you think? they look quite identical to me

Bobkidbob:
Pretty sure that code review was before the refactor, so waiting for another.

PR #850, opened 2018-12-03
Fix 500 error

cisco87:
https://sandbox.bbc.co.uk/programmes/articles/5ZyPNMLqNyqtddDxjVnYXt3/episode-transcript-episode-90-jade-bi

http://sandbox.bbc.co.uk/programmes/profiles/2FgQZ2jkQy8510skZTsRhLM/abbie-tucker


programmes-developers:

src/Controller/Profiles/ShowController.php
```diff
@@ -53,8 +53,9 @@ public function __invoke(
         }
 
         $context = null;
-        if ('' !== $profile->getParentPid()) {
-            $context = $coreEntitiesService->findByPidFull(new Pid($profile->getParentPid()));
+        $parentPid = $profile->getParentPid();
+        if (!is_null($parentPid)) {
```
cisco87: now that I think about it I guess this could also be better if checked about the Pid instance?;o

PR #851, opened 2018-12-03
[PROGRAMMES-6752] fixed dates for podcasts

tomwinnington:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6752

Link:
http://sandbox.bbc.co.uk/programmes/b006qpgr/episodes/downloads

Link with episodes that have no release date or broadcast date:
http://sandbox.bbc.co.uk/programmes/p03wy14r/episodes/downloads


src/Ds2013/Presenters/Domain/CoreEntity/Programme/SubPresenters/programme_body.html.twig
```diff
@@ -11,11 +11,9 @@
         {% endif %}
     </div>
 {% endif %}
-{% if programme_body.hasReleaseDate() %}
-    <div class="programme__service micro ">
+<div class="programme__service micro ">
         <time >{{ programme_body.getReleaseDate() }}</time>
```
Bobkidbob: This is invalid HTML. `<time>` must have the `datetime` attribute or only contain a text node which is valid datetime. There is more detail on [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time).

src/Ds2013/Presenters/Domain/CoreEntity/Programme/SubPresenters/ProgrammeBodyPresenter.php
```diff
@@ -7,6 +7,7 @@
 use App\DsShared\Helpers\LocalisedDaysAndMonthsHelper;
 use App\DsShared\Helpers\PlayTranslationsHelper;
 use BBC\ProgrammesPagesService\Domain\Entity\Clip;
+use BBC\ProgrammesPagesService\Domain\Entity\Episode;
```
Bobkidbob: I don't think this import is required.

tomwinnington: the $programme variable can be an instance of Episode, so it kicks up a fuss without the import.

src/Ds2013/Presenters/Domain/CoreEntity/Programme/SubPresenters/ProgrammeBodyPresenter.php
```diff
@@ -108,20 +109,28 @@ public function containerHasAvailableEpisodes()
...
-    public function hasReleaseDate(): bool
+    public function getDisplayDate(): string
     {
-        return ($this->getOption('show_release_date') && ($this->programme instanceof ProgrammeItem) && $this->programme->getReleaseDate());
+        if ($this->getOption('show_release_date') && $this->programme instanceof ProgrammeItem) {
+            if ($this->programme->getReleaseDate()) {
+                return $this->localisedDaysAndMonthsHelper->getFormatedDay(new Chronos($this->programme->getReleaseDate()));
+            }
+            if ($this->programme->getFirstBroadcastDate()) {
+                return $this->localisedDaysAndMonthsHelper->getFormatedDay(new Chronos($this->programme->getFirstBroadcastDate()));
```
BjornTwachtmann: You don't need to create a Chronos here, the domain object will already be returning that (despite the typehint)

PR #855, opened 2018-12-04
PROGRAMMES-6775 - changed outerWidth() to getBoundingClientRect()

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6775

This does not fix the issue as a full fix will take to much time, but it does make it less pronounced.


programmes-developers:

PR #856, opened 2018-12-05
Added more ignore rules

Tasiobg:
I thought this could be a good idea, what do you think? or should we instead add it to our gobal .gitignore?


programmes-developers:

Bobkidbob:
I prefer it like this, but I think others prefer it in global.

Tasiobg:
I research a bit and some people agree that if this ignored files are not created by the project is better practice to add it to the global so I'll close the PR and add it to my globals.

PR #859, opened 2018-12-05
[PROGRAMMES-6782] Fix 500 when profile has no attached project space

benolot:
Link: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6782


programmes-developers:

src/ExternalApi/Isite/Service/IsiteService.php
```diff
@@ -80,6 +80,11 @@ public function getByProgramme(
     ): PromiseInterface {
         /** @var string $project */
         $project = $programme->getOption('project_space');
+
+        if (!$project) {
+            return new FulfilledPromise([]);
```
BjornTwachtmann: The null result for this class is already defined below in the call to getHttpApiMultiClient as 
`new IsiteResult($page, $limit, 0, []),`
I think it would be more sensible that this method always returns an IsiteResult, you then don't need to modify anything in the controller either.
Make sense? Or am I full of it?

benolot: This method is currently set to always return a PromiseInterface, just to check do you think it's worthwhile updating the method to only return an iSite response or am I misunderstanding?

PR #861, opened 2018-12-06
This moves responsiblity for injecting page-wide analytics classes in…

BjornTwachtmann:
…to the SMP presenter from the template to the PresenterFactory. That involves some fairly nasty hacks...


programmes-developers:

PR #854, opened 2018-12-04
PROGRAMMES-6761 Incorrect promotion size

Tasiobg:
- Updated the logics that builds the CSS for "content block promotions", this is now based on v2 code
- Updated the image presenter to use aspect ratio "auto" which mean the height will be propotional to the with we set, example /programmes/articles/5Mstn5KJJpgnH4Rd3SYHHSK/stories-and-features-about-the-archers

I could walk you thought the PR if you preffer an easy way to understand the changes

PR #865, opened 2018-12-06
PROGRAMMES-6761 Incorrect size for Promotion box

Tasiobg:
When I updated the image presenter I didn't realised it was extended by ImageEntityPresenter, this code implement "auto" for ImageEntityPresenter as well

Previous PR where I updated only the ImagePresenter https://github.com/bbc/programmes-frontend/pull/854


programmes-developers:

PR #866, opened 2018-12-06
Fix ContentBlockMapper to work on non-streamable clips within clip bl…

BjornTwachtmann:
…ocks


programmes-developers:

PR #867, opened 2018-12-06
[NoTicket] fixed route

tomwinnington:
http://sandbox.bbc.co.uk/programmes/styleguide/ds2013

Paginator route was invalid


BjornTwachtmann:
Mmm, why change the link from promotion to pagination? Am I missing something?

PR #869, opened 2018-12-06
PROGRAMMES-6761 fix missing "auto"

Tasiobg:
I miss this one because ratio is not wrapped with quotes, I would have thought that something like this should crash because ratio is not a variable in this scope


programmes-developers:

PR #870, opened 2018-12-06
Programmes 6770 isite guid wrong refactoring

cisco87:


programmes-developers:

src/Controller/Articles/ShowController.php
```diff
@@ -5,6 +5,7 @@
 
 use App\Controller\BaseController;
 use App\Controller\Helpers\IsiteKeyHelper;
+use App\Controller\IsiteBaseController;
```
Tasiobg: Please check all "use" I think some of then are not need

src/Controller/Articles/ShowController.php
```diff
@@ -13,95 +14,75 @@
...
+        $this->initBranding($article);
 
         $parents = $article->getParents();
         $siblingPromise = $isiteService->setChildrenOn($parents, $article->getProjectSpace()); //if more than 48, extras are removed
         $childPromise = $isiteService->setChildrenOn([$article], $article->getProjectSpace(), $this->getPage());
         $response = $this->resolvePromises(['children' => $childPromise, 'siblings' => $siblingPromise]);
-
-        $paginator = $this->getPaginator(reset($response['children']));
+        $paginator = $this->getPaginator($article);
+//        $paginator = $this->getPaginator(reset($response['children']));
```
Tasiobg: I think commiting commented code is consider a bad practice, please do you mind to direcly ask someone or add a comment in github, alternatively if a block is mean to be merged commented, please explain why so we can understand the reason of this code

src/Controller/IsiteBaseController.php
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        return new PaginatorPresenter(
+            $this->getPage(),
+            self::MAX_LIST_DISPLAYED_ITEMS,
+            $isiteObject->getChildCount()
+        );
+    }
+
+    protected function initContext(BaseIsiteObject $isiteObject, CoreEntitiesService $coreEntitiesService): void
```
Tasiobg: Can CoreEntitiesService be magically injected in siteBaseController? it's feel a bit strang to see it this way but I don't mean it is wrong

BjornTwachtmann:
As discussed, please wait till after profile page launch before merging.

src/Controller/IsiteBaseController.php
cisco87: I think this one was a work in progress thing, now CoreEntitiesService is referenced in a instance property, will change it

src/Controller/Articles/ShowController.php
cisco87: Yeah, but we're human so I guess it's ok to miss a line there and there :D, thanks :D

src/Controller/Articles/ShowController.php
cisco87: Will have a look!

src/Controller/Articles/ShowController.php
```diff
@@ -13,95 +14,75 @@
...
+        $this->initContext($article, $coreEntitiesService);
+        $this->initBranding($article);
 
         $parents = $article->getParents();
         $siblingPromise = $isiteService->setChildrenOn($parents, $article->getProjectSpace()); //if more than 48, extras are removed
         $childPromise = $isiteService->setChildrenOn([$article], $article->getProjectSpace(), $this->getPage());
         $response = $this->resolvePromises(['children' => $childPromise, 'siblings' => $siblingPromise]);
-
-        $paginator = $this->getPaginator(reset($response['children']));
+        $paginator = $this->getPaginator($article);
```
Tasiobg: The paginator was previosly using `$response` istead of `$article`, is $response not need anymore in this code?

cisco87: It was different in both, I just picked a version and moved it to the common base class. That's why the other one had a comment, to keep track of both versions while I was refactoring, nowthey are the same

Tasiobg: Sorry I don't understand what you mean but the line 73 is setting `$response` which doesn't seem to be used at all

cisco87: Oh nice catch, I will remove it then, it was already there and I didn't want to change anything I didn't know what was doing and just jumped through that piece of code

src/Controller/IsiteBaseController.php
cisco87: It is called method injection anyway it's not a strange thing, some people prefer this over constructor injection, I think it is wrong because method injection is a sign of low cohesion and is a code smell, but in this case was just an experiment to keep method s working while refactoring

PR #877, opened 2018-12-10
PROGRAMMES-6783: Actual stats on article/profile pages would probably be a good idea.

BjornTwachtmann:


src/ValueObject/AnalyticsCounterName.php
```diff
@@ -95,8 +95,9 @@ private function getProgrammeCounterNamePart(Programme $context, string $relativ
         $partialString .= '.' . $context->getPid();
 
         // add the rest of the URL to the counter name
-        $restOfUrl = str_replace(['/programmes/', $context->getPid()], '', $relativePath);
+        $restOfUrl = str_replace(['/programmes', '/' . $context->getPid()], '', $relativePath);
```
BjornTwachtmann: This fixes article/profile pages with parent programmes coutnernames that are missing a . between the programmes and the article ID (due to no context PID being present)

src/ValueObject/AnalyticsCounterName.php
```diff
@@ -95,8 +95,9 @@ private function getProgrammeCounterNamePart(Programme $context, string $relativ
...
         $restOfUrl = str_replace('/', '.', $restOfUrl);
+        $restOfUrl = $this->replaceDisallowedCharacters($restOfUrl);
```
BjornTwachtmann: @Tasiobg this looks like it should be doing this anyway, but any reason not to?

src/ExternalApi/Isite/Domain/BaseIsiteObject.php
```diff
@@ -40,6 +40,9 @@
     /** @var string|null */
     protected $shortSynopsis;
 
+    /** @var string|null */
```
cisco87: ?string?

BjornTwachtmann: This is the right syntax for phpDoc I think. It's not the same as PHP, because why would it be consistent.

cisco87: LOL

PR #878, opened 2018-12-11
PROGRAMMES-6798: Fix alt text on 3rd party embeds

BjornTwachtmann:


programmes-developers:

PR #879, opened 2018-12-11
Fix bug blowing up profile pages without content blocks

BjornTwachtmann:
Look at www.bbc.co.uk/programmes/profiles/2ctC7mSFfppkpkKcqCDq78V/thyra . Check this again on sandbox (latest develop). 

Note the exception complaining (wrongly) that `An exception has been thrown during the rendering of a template (\"Profile content blocks have not been queried for yet.\").`

This PR teaches the profile mapper about the difference between unfetched and nonexistent.

This may also be an issue with articles, but that's a possible separate PR.


programmes-developers:

PR #881, opened 2018-12-11
[PROGRAMMES-6762] Refactor context

tomwinnington:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6762

All the links are in the ticket. Appears to all work.

### DIFF
--- a/src/Controller/Articles/ShowController.php
+++ b/src/Controller/Articles/ShowController.php
@@ -58,6 +58,7 @@ class ShowController extends BaseController
         }
 
         $this->setContext($context);
+        $this->setAtiContentId('isite', $guid);
 
         if ('' !== $article->getBrandingId()) {
             $this->setBrandingId($article->getBrandingId());
--- a/src/Controller/Articles/ShowController.php
+++ b/src/Controller/Articles/ShowController.php
@@ -58,7 +58,7 @@ class ShowController extends BaseController
         }
 
         $this->setContext($context);
-        $this->setAtiContentId('isite', $guid);
+        $this->setAtiContentId($guid, 'isite');
 
         if ('' !== $article->getBrandingId()) {
             $this->setBrandingId($article->getBrandingId());
--- a/src/Controller/BaseController.php
+++ b/src/Controller/BaseController.php
@@ -52,6 +52,10 @@ abstract class BaseController extends AbstractController
 
     private $istatsExtraLabels = [];
 
+    private $atistatsExtraLabels = [];
+
+    private $atiContentId;
+
     private $istatsProgsPageType;
 
     private $isInternational = false;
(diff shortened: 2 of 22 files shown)


programmes-developers:

benolot:
This isn't about refactor context - my branch name was wrong because I'm an idiot - it's just implementing ATI stats stuff.

src/Controller/Articles/ShowController.php
```diff
@@ -58,6 +58,7 @@ public function __invoke(string $key, string $slug, Request $request, ArticleSer
         }
 
         $this->setContext($context);
+        $this->setAtiContentId($guid, 'isite');
```
Tasiobg: Should this be "isite" or "isite2"?

tomwinnington: says isite on the ticket

Tasiobg: Do you mind to have another look or check with Claudio? It seems to say "isite2" in some places

tomwinnington: Ignore me! I'll change it now

src/Controller/BaseController.php
```diff
@@ -304,6 +308,19 @@ protected function setIstatsExtraLabels(array $labels): void
         $this->istatsExtraLabels = array_replace($this->istatsExtraLabels, $labels);
     }
 
+    protected function setAtiStatsExtraLabels(array $labels): void
+    {
+        $this->atistatsExtraLabels = array_replace($this->istatsExtraLabels, $labels);
```
Tasiobg: I think you mean
`$this->atistatsExtraLabels = array_replace($this->atistatsExtraLabels, $labels);`

tomwinnington:
```suggestion
        $this->atistatsExtraLabels = array_replace($this->atistatsExtraLabels, $labels);


```

src/Controller/BaseController.php
```diff
@@ -304,6 +308,19 @@ protected function setIstatsExtraLabels(array $labels): void
         $this->istatsExtraLabels = array_replace($this->istatsExtraLabels, $labels);
     }
 
+    protected function setAtiStatsExtraLabels(array $labels): void
+    {
+        $this->atistatsExtraLabels = $labels;
```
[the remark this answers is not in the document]
cisco87: `set` is notoriously used as a word to arbitrarily set a value, in your suggestion it would suggest something but do something else `replace`, I am not sure it would be the best thing tbh, I guess if we need to append we should `addIstatsExtraLabel(s)`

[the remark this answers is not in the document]
Tasiobg: `addIstatsExtraLabel` is a better name on my point of view, I didn't thought about it when I created the iStats ones

src/ValueObject/AtiAnalyticsLabels.php
```diff
@@ -14,35 +14,35 @@ class AtiAnalyticsLabels
...
 
-    /** @var array  */
+    /** @var array */
     private $extraLabels;
 
-    public function __construct($context, string $progsPageType, array $extraLabels, string $environment)
+    /** @var string */
+    private $contentId;
+
+    public function __construct($context, string $progsPageType, string $environment, array $extraLabels, string $contentId = null
```
BjornTwachtmann: Rather than pass the environment in as a string, perhaps just pass in the whole CosmosInfo?

src/Controller/BaseController.php
```diff
@@ -327,6 +329,16 @@ protected function setIstatsExtraLabels(array $labels): void
 
     protected function addAtiStatsExtraLabels(array $labels): void
     {
+        $this->atistatsExtraLabels = array_replace($this->atistatsExtraLabels, $labels);
+    }
+
+    protected function setAtiContentId(?string $identifier, string $authority = 'pips'): void
```
BjornTwachtmann: If you set the default $atiContentId to an empty string you can probably avoid making this nullable

src/Controller/BaseController.php
```diff
@@ -55,6 +55,8 @@
 
     private $atistatsExtraLabels = [];
 
+    private $atiContentId;
```
BjornTwachtmann: default value of '' ?

BjornTwachtmann: default value of 'pips' rather then?

BjornTwachtmann: bbc:urn:pips

src/Controller/BaseController.php
BjornTwachtmann: I can't see where null could actually be passed in here. But if it's needed, leave it I suppose.

PR #882, opened 2018-12-11
Fix a bug where assigning a parent article to a profile causes 500s

BjornTwachtmann:
Example: http://sandbox.bbc.co.uk/programmes/profiles/1vJmwcFRP2Wq4PQHp7jSdrr/three-personal-stories


programmes-developers:

PR #883, opened 2018-12-11
Fix a bug where a promotions content block has no layout parameter. T…

BjornTwachtmann:
…his can be null, despite being mandatory in iSite.

E.G: http://sandbox.bbc.co.uk/programmes/profiles/3HjNsmW9tbq9bqRTYYWLCFW/profiles-2014

### DIFF
--- a/src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
+++ b/src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
@@ -257,7 +257,7 @@ class ContentBlockMapper extends Mapper
             case 'promotions':
                 $contentBlockData = $form->content;
                 $title = $this->getString($contentBlockData->title);
-                $layout = $this->getString($contentBlockData->layout);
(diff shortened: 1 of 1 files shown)


programmes-developers:

src/ExternalApi/Isite/Mapper/ContentBlockMapper.php
```diff
@@ -257,7 +257,7 @@ public function getDomainModel(SimpleXMLElement $isiteObject)
             case 'promotions':
                 $contentBlockData = $form->content;
                 $title = $this->getString($contentBlockData->title);
-                $layout = $this->getString($contentBlockData->layout);
+                $layout = $this->getString($contentBlockData->layout) ?: 'list'; // Default to list if unset
```
Bobkidbob: This looks kinda weird. Does this work in PHP, I know it would in JS:
```suggestion
                $layout = $this->getString($contentBlockData->layout) || 'list'; // Default to list if unset
```

PR #884, opened 2018-12-12
PROGRAMMES-6789 invalid clips

Tasiobg:
Single player clip mapper should loop through all the clips because the first one coming from iSite may not be a valid one.


programmes-developers:

PR #885, opened 2018-12-12
Fix problem when a group is set as parent_pid of a profile/article in iSite

BjornTwachtmann:
This deals with branding ID and context for a group parent of a profile/article


programmes-developers:

PR #886, opened 2018-12-12
Fix Random layout issues with telescope and third party blocks

BjornTwachtmann:


programmes-developers:

PR #888, opened 2018-12-12
Kite fix

cisco87:


programmes-developers:

src/Ds2013/Presenters/Domain/ContentBlock/InteractiveActivity/interactive_activity.html.twig
```diff
@@ -1,8 +1,21 @@
...
-        embed({{ loader_url|json_encode|raw }}, {{ content_block.getParams()|json_encode|raw }});
-    });
-</script>
+{% block extraClasses %}
+    {% if not content_block.isInPrimaryColumn() %}
+        component--box-flushbody
+    {% endif %}
+{% endblock %}
+
+{% block componentTitle %}{% endblock %}
```
BjornTwachtmann: What is this block?

cisco87: Do you think it would be better to make InteractiveGame class always return an empty string?

cisco87: It is because the base template always shows the block title if getName() of component class doesn't return an empty string, I just made sure that it was possible to have the title block not rendered in certain conditions but without being too intrusive of other stuff

BjornTwachtmann: I see. Leave it this way then I suppose :(

PR #890, opened 2018-12-12
Fix recommendations bit

cisco87:
https://sandbox.bbc.co.uk/programmes/p02ct985/recommendations.2013inc


programmes-developers:

PR #892, opened 2018-12-13
[No ticket] Fixed 500 on article pages

benolot:
Happens when there is no children it tries to run a pagination element on it.


programmes-developers:

PR #893, opened 2018-12-14
PROGRAMMES-6809: Fix a bug where embedding the same clip multiple tim…

BjornTwachtmann:
…es into an article/profile only works once


programmes-developers:

PR #895, opened 2018-12-18
PROGRAMMES-6765 - Good luck reviewing this hahahahahahahahahahahahahaha

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6765


cisco87:
So the first time we duplicated all over the world `spacer-hidden` because it would never change, then in less than a week we had to change it all over the world, but I guess creating constants to change repeating values everywhere easily is too mainstream? :D

cisco87:
I would advise to remove the integers around the code base and show mercy towards the one who will have to squeeze his wits to understand what each integer represents

src/Ds2013/Presenters/Utilities/Paginator/PaginatorPresenter.php
```diff
@@ -59,6 +59,13 @@ public function shouldApplyNoHideClass(int $item): bool
         return $this->currentPageWithinThreeOfEnd($item);
     }
 
+    public function shouldApplyMobileClass(int $item): bool
+    {
+        $pages = $this->getPageCount();
+        $currentPage = $this->currentPage;
+        return ($item == 2 || $item == $pages - 1) && ($currentPage == 5 || $currentPage == $pages - 4);
```
Tasiobg: I think is a good practice to use the Identical operator `===` when possible

src/Ds2013/Presenters/Utilities/Paginator/PaginatorPresenter.php
```diff
@@ -59,6 +59,13 @@ public function shouldApplyNoHideClass(int $item): bool
         return $this->currentPageWithinThreeOfEnd($item);
     }
 
+    public function shouldApplyMobileClass(int $item): bool
```
Tasiobg: Some comments for this function could help, Is not clear why it should apply mobile class

PR #896, opened 2018-12-18
[PROGRAMMES-6787] Fixing bug in re-firing up the branding execution on fallback

cisco87:
http://sandbox.bbc.co.uk/programmes/articles/4QVGHVJT5JrknTjWpFcrv9G/touchcast-embedded-in-an-ancillary-page

http://sandbox.bbc.co.uk/programmes/b006q2x0


programmes-developers:

src/Controller/BaseController.php
```diff
@@ -411,7 +411,10 @@ private function rejectedBrandingPromise($reason, $usePreview, $previewId)
 
             $this->setBrandingId($this->fallbackBrandingId);
             $brandingClient = $this->container->get(BrandingClient::class);
-            return $brandingClient->getContent($this->brandingId, null);
+            $branding = $brandingClient->getContent($this->brandingId, null);
+            if ($branding instanceof Branding) {
+                return $this->fulfilledBrandingPromise($branding);
+            }
```
Bobkidbob: Is this `if` statement required? As far as I can tell `$brandingClient->getContent()` can only return an instance `Branding`.

cisco87: it is not hinted by the method return value, so I'd rather keep it this way instead of going to modify the return value of getContent(); to ensure it always returns Branding

cisco87: Aaaanyhow, I think that going to the thrown exception down is optimistic, I guess if this code is going there, it's because it's not being able to fetch the data of the default/hardcoded/fallback branding, so I guess Isite is down and we've worse problem in that case, so I guess this is ok this way

Tasiobg: If you check line 412
```
$this->setBrandingId($this->fallbackBrandingId);
```
The branding is set to fallback ID so when calling 
```
$branding = $brandingClient->getContent($this->brandingId, null);
```
It should always request the fallback branding which should always exist, I don't think there is any need to check `if ($branding instanceof Branding)`

So you only need to return ` $this->fulfilledBrandingPromise($branding);` 

On the other hand if for any reason the branding client falis to get branding it throws and exception and stops, doesn't look like we can handle it gracefully.

cisco87: I will try to be shorter and clearer, `getContent` doesn't assure me that it can't be modified and another entry point can be added to return another kind of object. We are adding a bit of cognitive overload to a future guy needing to modify that method to go through the code base to check if everyone was making sure that they were checking that the object return was suitable for his needs in order to save a clock cycle. But since we're 3 resources talking about an if, I guess also for today I'll have to save the world and remove it

cisco87: It is better to fail with `unknown error` exception than `trying to call method X of Y`

PR #897, opened 2018-12-18
PROGRAMMES-6811 - Added auto ratio to correct image height

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6811

I'm not sure whether this will break any other article pages.


programmes-developers:

PR #898, opened 2018-12-18
PROGRAMMES-6564 - Fix tab order

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6564

### DIFF
--- a/src/DsShared/Utilities/EntityContext/entity_context.html.twig
+++ b/src/DsShared/Utilities/EntityContext/entity_context.html.twig
@@ -1,5 +1,10 @@
 {%- for item in entity_context.getAncestry() -%}
-    <a tabindex="-1" class="context__item" href="{{ path('find_by_pid', {'pid': item.getPid()}) }}">{{ item.getTitle() }}</a>
+    <a
+        {{ not entity_context.getOption('is_tabbable') ? 'tabindex="-1"' }}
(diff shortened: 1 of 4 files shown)


programmes-developers:

src/DsShared/Utilities/EntityContext/entity_context.html.twig
```diff
@@ -1,5 +1,8 @@
 {%- for item in entity_context.getAncestry() -%}
-    <a tabindex="-1" class="context__item" href="{{ path('find_by_pid', {'pid': item.getPid()}) }}">{{ item.getTitle() }}</a>
+    <a class="context__item" href="{{ path('find_by_pid', {'pid': item.getPid()}) }}"
```
BjornTwachtmann: The problem with this is that it's not going to differentiate between making visually hidden links tabbable and visible links tabbable, is it?

E.g. the hidden "doctor who" link here http://sandbox.bbc.co.uk/programmes/b09kkt1m will still get tabbed to? Is that what we want?

BjornTwachtmann: The first context__item always seems to be hidden by the CSS

Bobkidbob:
Test this please.

PR #900, opened 2018-12-19
[PROGRAMMES-6822] Making error appear

cisco87:
https://sandbox.bbc.co.uk/programmes/articles/282P63C8jhnCXF9hDddCdmH/a-test-article-with-loads-of-stuff-in-the-right-column


programmes-developers:

src/Ds2013/Presenters/Domain/ContentBlock/Quiz/quiz.html.twig
```diff
@@ -1,9 +1,19 @@
...
 {% endblock %}
 
 {% block blockContent %}
-    {{ content_block.getBlock().getHtmlContent()|raw }}
+    {{ componentContent }}
+{% endblock %}
+
+{% block componentTitle %}
+    {% if componentContent == '' %}
+        <div class="br-box-subtle island"><h2>This content can't be found</h2></div>
```
Bobkidbob: Shouldn't this be done through the translation system? Also, it may be better to use "cannot" instead of "can't".

cisco87: I copied the text and implementation from V2

Tasiobg: We don't have translations for this one and becase is rare to display this error, BA agree that there is no need to request translations for this one. This error was mainly created to help editorial idenfiying issues when adding games.

Tasiobg:
@cisco87 Have you tested it? It doesn't seems to be working fine for me.
Also not necesarily wrong but a bit weird to display the error in the title

cisco87:
@Tasiobg I have tested it, what were you expecting, what do you see?

src/Ds2013/Presenters/Domain/ContentBlock/Quiz/quiz.html.twig
```diff
@@ -1,9 +1,19 @@
 {% extends "@Ds2013/Presenters/Domain/ContentBlock/content_block.html.twig" %}
 
+{% set componentContent = content_block.getBlock().getHtmlContent()|raw %}
```
BjornTwachtmann: The |raw here will not work as you think.

src/Ds2013/Presenters/Domain/ContentBlock/Quiz/quiz.html.twig
```diff
@@ -1,9 +1,19 @@
...
+
 {% block extraClasses %}
     {{ build_css_classes({'component--box-flushbody': not content_block.isInPrimaryColumn()}) }}
 {% endblock %}
 
 {% block blockContent %}
-    {{ content_block.getBlock().getHtmlContent()|raw }}
+    {{ componentContent }}
```
BjornTwachtmann: This will output escaped HTML, completely breaking the feature. It needs a |raw on the end.

See /programmes/articles/5mPMg7QW4G0V2RjQhDbs6Dy/celebrity-laps 

You should test that the basic functionality of what you're changing still works when you do these tickets.

PR #901, opened 2018-12-19
[PROGRAMMES-6749] Level 2 producer ATI Stats

benolot:
Ticket: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6749


src/Controller/BaseController.php
```diff
@@ -453,4 +453,90 @@ private function calculateOrbitSearchScope(?Network $network): string
...
+            'cbeebies_radio' => 'CBEEBIES',
+        ];
+
+        $masterbrand = $this->context->getMasterbrand();
+
+        if (array_key_exists((string) $masterbrand->getMid(), $producersMap)) {
+            return $producersMap[(string) $masterbrand->getMid()];
+        }
+
+        if ($masterbrand->getNetwork()->getMedium() === 'radio') {
```
cisco87: The approach of a dictinary based mapping is nice, why you decided that instead 2 if condition couldn't use the same approach? It could've been also forward thinking, if we needed more mapping for getNetwork->getMedium, no?

src/Controller/BaseController.php
```diff
@@ -453,4 +453,90 @@ private function calculateOrbitSearchScope(?Network $network): string
         }
         return '';
     }
+
+    private function calculateProducerVariable(): string
+    {
+        $producersMap = [
```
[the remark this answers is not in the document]
benolot: I'm not at all sure that would give us any benefit, can you think of any?

src/Controller/BaseController.php
```diff
@@ -453,4 +459,90 @@ private function calculateOrbitSearchScope(?Network $network): string
...
+            'cbeebies_radio' => 'CBEEBIES',
+        ];
+
+        $masterbrand = $this->context->getMasterbrand();
+
+        if (array_key_exists((string) $masterbrand->getMid(), $producersMap)) {
+            return $producersMap[(string) $masterbrand->getMid()];
+        }
+
+        if ($masterbrand->getNetwork()->getMedium() === 'radio') {
```
Tasiobg:
```suggestion
        if ($masterbrand->getNetwork()->getMedium() === NetworkMediumEnum::RADIO) {
```

Tasiobg:
Not strictly necesary but I suggest to move the calculateProducerVariable() outside of baseController if you can think of a nice place, probably a new class inside ValueObject.
Note tested but PR looks good, probably good idea to wait for bjorn to see if he has more feedback

src/Controller/BaseController.php
BjornTwachtmann: Yeah, this needs to use NetworkMediumEnum

src/Controller/BaseController.php
```diff
@@ -453,4 +459,90 @@ private function calculateOrbitSearchScope(?Network $network): string
...
+            return 'SOUNDS';
+        }
+
+        if ($masterbrand->getNetwork()->getMedium() === 'tv') {
```
BjornTwachtmann: This also needs to use NetworkMediumEnum

src/Controller/BaseController.php
```diff
@@ -453,4 +459,90 @@ private function calculateOrbitSearchScope(?Network $network): string
         }
         return '';
     }
+
+    private function calculateProducerVariable(): string
```
BjornTwachtmann: I think this logic might be better in AtiAnalyticsLabels

benolot: Updated it

src/Controller/BaseController.php
benolot: Updated

src/Controller/BaseController.php
benolot: Done

PR #910, opened 2018-12-24
[PROGRAMMES-6831] Removed empty div when no caption exists

tomwinnington:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6831

Below the image there was a block to contain a caption, but when there is no caption, empty space appears.

Before:
image

After:
image

Link:
sandbox.bbc.co.uk/programmes/articles/1Tg6t6YxdxyYmKykSfWp9cs/the-women-erased-from-musical-history


programmes-developers:

PR #912, opened 2019-01-02
[PROGRAMMES-6836] Extra line of podcast page

tomwinnington:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6836

Link:
http://sandbox.bbc.co.uk/programmes/b007qlvb/episodes/downloads?page=1

### DIFF
--- a/src/Ds2013/Presenters/Utilities/_highlight-box.scss
+++ b/src/Ds2013/Presenters/Utilities/_highlight-box.scss
@@ -61,6 +61,7 @@
         border-width: 1px;
         border-bottom-style: solid;
 
+
         // OK WTF is this doing, well gather around children...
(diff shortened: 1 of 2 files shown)


programmes-developers:

src/Ds2013/Presenters/Utilities/_highlight-box.scss
```diff
@@ -112,6 +112,10 @@
         padding: $spacing;
         border-style: none;
     }
+
+    .highlight-box--list:last-child  {
```
BjornTwachtmann: A quick search shows that this class is used on a lot of other pages. Changing the behaviour in the general case to solve the specific case seems dangerous. Have you looked at the effect on any of the other use cases?

PR #913, opened 2019-01-02
[No ticket] Added a css class for Automation QA

benolot:


programmes-developers:

PR #915, opened 2019-01-02
PROGRAMMES-6826: Article guid attached to /profiles/ causing 500 but should show 404

BjornTwachtmann:
__CLASS__ does not account for subclassing, causing key namespace clashes in our caching

To test:
Visit http://sandbox.bbc.co.uk/programmes/b006q2x0/articles, then http://sandbox.bbc.co.uk/programmes/b006q2x0/profiles . If you're on develop, and the first one is in cache, the second one will 500 due to retieving the first result from the cache.

This fixes by assigning the correct subclass to cache keys, preventing the cache clashes.


programmes-developers:

PR #916, opened 2019-01-02
[PROGRAMMES-6834] Adding some margin top?

cisco87:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6834
https://sandbox.bbc.co.uk/programmes/b006q2x0/articles?page=2


programmes-developers:

src/Ds2013/Presenters/Utilities/Paginator/_paginator.scss
```diff
@@ -9,6 +9,7 @@
 
 .pagination {
     margin-bottom: $basic-spacing;
+    margin-top: $basic-spacing;
```
BjornTwachtmann: This could potentially change the behaviour of the paginator everywhere it's used for this one case where it's broken. 

It probably won't change it much because margin collapse is a thing... but is that something you want to check every page type for?

BjornTwachtmann:
I would recommend to change article_list.html.twig and wrap `<ul class="grid-wrapper highlight-box-wrapper highlight-box-wrapper--grid">` in a `<div class="component">` , that seems to be how it works on profiles and elsewhere. 

That way you only need to test the one thing.

cisco87:
Tried that but wasn't working, but apparently if I wrap both the paginator and the list in separate div.component then it work

templates/profiles/partials/profile_list.html.twig
```diff
@@ -1,4 +1,4 @@
-{%  if profiles|length <= 0 %}
+v{%  if profiles|length <= 0 %}
```
Bobkidbob:
```suggestion
{%  if profiles|length <= 0 %}
```

templates/articles/partials/article_list.html.twig
```diff
@@ -3,14 +3,19 @@
...
+        <ul class="grid-wrapper highlight-box-wrapper highlight-box-wrapper--grid">
+            {% for article in articles %}
+                <li class="grid 1/2@bpb1 1/3@bpb2 1/3@bpw 1/4@bpw2 1/4@bpe">
+                    {{ ds2013('article', article) }}
+                </li>
+            {% endfor %}
+        </ul>
+    </div>
+
+    <div class="component">
```
BjornTwachtmann: This will add 18px padding below the div. It has no other effect. Seems strange that you would need this.

cisco87: I don't know why but without a div around it the space wasn't taken, and I couldn't really put much efforts in saving a div

cisco87: If you think I should find a way to remove the div and get the effect let me know

PR #918, opened 2019-01-02
[Programmes 6832] Fixed missing image issue

tomwinnington:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6832

Link:
https://www.bbc.co.uk/programmes/b006q2x0/articles


programmes-developers:

src/DsShared/Utilities/Image/ImagePresenter.php
```diff
@@ -56,6 +56,12 @@ public function __construct(
         $this->sizes = $this->buildSizes($sizes);
     }
 
+    public function hasImage(){
+        if ($this->imagePid !==''){
```
BjornTwachtmann: Is there ever a case where an empty image PID is passed in here? That seems like more of an error in the calling class to me than something we should be checking for here? Or is this a valid use case?

src/DsShared/Utilities/Image/image.html.twig
```diff
@@ -1,5 +1,11 @@
+{% if image.hasImage() %}
+    {% set Srcsets = image.getSrcsets() %}
```
Tasiobg: In PHP we use "camel case" for variables, not sure if this really matter for twig but could look more consisten with "camel case"
{% set srcsets = image.getSrcsets() %}
```suggestion
    {% set srcsets = image.getSrcsets() %}
```

Tasiobg:
Some information I'm not sure if could affect this ticket or PR, I think Images PIDs from iSite use the LIVE images not the ones displayed on test iSite

PR #919, opened 2019-01-02
Fix some highlight box classes that went wrong in porting and remove …

BjornTwachtmann:
…some superflous grid__inner

can test on /programmes/b071vl2l (related programmes)
/programmes/profiles/z5gWDtzYM5nmgZD7zfgDFy/test-profile (promo and gallery grids)


programmes-developers:

src/Ds2013/Presenters/Utilities/_highlight-box.scss
```diff
@@ -122,6 +122,10 @@
         margin-bottom: $wide-spacing;
     }
 
+    .highlight-box-wrapper--grid {
+        margin-bottom: -$wide-spacing !important;
```
Bobkidbob: Is there not a way to do this without `!important`?

BjornTwachtmann: Nope, sorry :(

BjornTwachtmann: It's already using !important for this (see definition above). So it's not a new thing

PR #924, opened 2019-01-03
PROGRAMMES-6840: Handle article/profile guids being passed to the wro…

BjornTwachtmann:
…ng controllers

I just saw a previous PR that fixed this by generically catching InvalidArgumentException, which is a shit idea, there needs to be a specific exception for this. Turns out there already was and it just wasn't being thrown, so this PR throws it.

Prevents links to /programmes/profiles/articleGUID and /programmes/articles/profileGUID from 500-ing. Which is not important. But the previous implementation made me nervous so I thought I'd fully fix it. 

See https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6840


programmes-developers:

PR #939, opened 2019-01-09
remove quiz title

Tasiobg:
I think we agree Quiz title shouldn't be displayed because it doesn't in v2, a previous PR as reintroduce the title probably by mistake https://github.com/bbc/programmes-frontend/pull/900/files so I'm removing it again


programmes-developers:

PR #940, opened 2019-01-09
fix titles

Tasiobg:
remove title from quiz so it behalves as v2
remove default title from clips stand alone and clip strem so it behalves as v2


programmes-developers:

PR #943, opened 2019-01-09
Fix article image size

Tasiobg:
Image slot is too small (490px) for a resultion were the image is mean to be around 945px.

I tested this fix it and is working fine but I'm not too familiar with the image sizing system in place so please let me know if any objections


Tasiobg:
It needs to be tested in the secondary column

PR #945, opened 2019-01-09
Decrease the severity error of an invalid content block

Tasiobg:
Decrease the severity error of an invalid content block because is not critical, is set by editorial and it doesn't stop the page from being rendered.


programmes-developers:

PR #953, opened 2019-01-10
Fix Ada related programmes. In some cases $uniqueProgrammes may conta…

Tasiobg:
…in programmes non existing in $programmeItems so we have to filter them


programmes-developers:

PR #957, opened 2019-01-14
[PROGRAMMES-6851] article profile controller refactoring

cisco87:
https://sandbox.bbc.co.uk/programmes/articles/5XY51BYR8zHVznBYpT8Zv2c/be-careful-what-you-wish-for-why-discovering-true-peace-and-quiet-can-be-noisier-and-more-terrifying-than-you-might-expect
https://sandbox.bbc.co.uk/programmes/profiles/z5gWDtzYM5nmgZD7zfgDFy/test-profile

If you could provide your point of view on this, lots of the redirecting reasons are common between article/profile, this one isn't, so I here I am making those common, and this outside.
```
        try {
            if ($redirect = $this->loadResourceIfUriIsCorrect('programme_article')) {
                return $redirect;
            }
        } catch (HasContactFormException $e) {
            if (!$this->slug) {
                $route = 'article_with_contact_form_noslug';
                $params = ['key' => $this->key];
            } else {
                $route = 'article_with_contact_form';
                $params = ['key' => $this->key, 'slug' => $this->slug];
            }

            return $this->cachedRedirectToRoute($route, $params, 302, 3600);
        }
```

Do you agree with it? They're a lot of boilerplate calls that do the same thing in both places, not reason I could think of to keep these duplicated function calls on both
```    protected function initPersonalisation(): void
    {
        $this->initContext();
        $this->initIstatsLabels();
        $this->overrideMetaTagsValues();
        $this->setAtiContentId($this->guid, 'isite2');
        $this->initBranding();
    }
```

While you come up with ideas for improvement I will keep QAing and getting my ideas too until EOD


programmes-developers:

src/Controller/BaseIsiteController.php
```diff
@@ -2,23 +2,169 @@
...
+                $this->isiteObject->getKey(),
+                $this->isiteObject->getSlug(),
+                $redirectRouteName
+            );
+        }
+
+        return false;
+    }
+
+    protected function initContext(): void
```
BjornTwachtmann: make this private if it's not called outside the base class

src/Controller/Articles/ShowController.php
```diff
@@ -25,100 +21,50 @@ public function __invoke(
...
+        $this->initPersonalisation();
 
+        $parents = $this->isiteObject->getParents();
+        $siblingPromise = $isiteService->setChildrenOn($parents, $this->isiteObject->getProjectSpace()); //if more than 48, extras are removed
+        $childPromise = $isiteService->setChildrenOn([$this->isiteObject], $this->isiteObject->getProjectSpace(), $this->getPage());
+        $response = $this->resolvePromises(['children' => $childPromise, 'siblings' => $siblingPromise]);
+        $children = reset($response['children']);
+        $paginatorPresenter = null;
+        if ($children) {
+            $paginatorPresenter = $this->getPaginator(reset($response['children'])->getTotal());
```
BjornTwachtmann:
```suggestion
            $paginatorPresenter = $this->getPaginator($children->getTotal());
```
calling reset(...) seems pointless

src/Controller/BaseIsiteController.php
```diff
@@ -2,29 +2,158 @@
...
+        if (!$isiteObjects) {
+            throw $this->createNotFoundException('No resource found for guid');
+        }
+
+        return reset($isiteObjects);
+    }
+
+    protected function getPreview(): bool
+    {
+        return $this->request()->query->has('preview') && (bool) $this->request()->query->get('preview');
```
BjornTwachtmann: bool is redundant here, it's being cast to a boolean anyway

src/Controller/BaseIsiteController.php
```diff
@@ -2,29 +2,158 @@
...
+    protected function getParentProgramme($context)
+    {
+        if ($context instanceof Group) {
+            return $context->getParent();
+        }
+
+        return $context;
+    }
+
+    protected function initPersonalisation(BaseIsiteObject $isiteObject, string $guid): void
```
BjornTwachtmann: Can you either rename this to something sensible or just copy the function calls into each controller.

PR #961, opened 2019-01-15
fix article image size

Tasiobg:


programmes-developers:

PR #963, opened 2019-01-16
[PROGRAMMES-6876] Subscribe text bug

tomwinnington:
Ticket:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6876

The gelicons were a bit too big, so when I made it smaller to match V2, and it all looks good.

Link:
sandbox.bbc.co.uk/programmes/p02nrvlw/episodes/downloads


programmes-developers:

tomwinnington:
> Do you think the text should be vertically aligned to the centre of the icon?

Yes they should. Tasio is looking at it in another ticket I believe.

PR #966, opened 2019-01-16
[PROGRAMMES-6868] Preview not working on != 200

cisco87:


programmes-developers:

src/Controller/ExceptionController.php
```diff
@@ -84,6 +82,11 @@ public function __invoke(Request $request, FlattenException $exception, DebugLog
             }
         }
 
+        if(!$request->query->has('preview') || !$request->query->get('preview')) {
```
benolot: What's the need for having both of these?

cisco87: Better this way?

PR #967, opened 2019-01-16
PROGRAMMES-6881 Change 204 to 404 cache lifetime for 404s is 60s anyway

BjornTwachtmann:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6881

PR #968, opened 2019-01-16
PROGRAMMES-6881: By fix I meant break.

BjornTwachtmann:
This fix fixes my fix from https://github.com/bbc/programmes-frontend/pull/967 , which was technically more of a break. This makes it so that partials return 204, while the full page returns a 404

Previous fix made it so that the lazy-loading icon-box remained on the page even when there was no content.


programmes-developers:

PR #970, opened 2019-01-18
[Programmes 6885] podcast structured data

tomwinnington:
Ticket:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6885

Link:
http://sandbox.bbc.co.uk/programmes/b006qpgr/episodes/downloads

Gives the following:
screen shot 2019-01-18 at 09 50 10


programmes-developers:

src/Controller/Podcast/PodcastController.php
```diff
@@ -87,4 +89,31 @@ public function __invoke(
             'genre' => $genre,
         ]);
     }
+
+    protected function getSchemaForProgrammeContainerAndParents(StructuredDataHelper $structuredDataHelper, Programme $programmeContainer): array
```
BjornTwachtmann: This looks identical to what we already have in BaseProgrammeEpisodesController. Is that right? If so, could we move this whole function into the StructuredDataHelper and call it from there, avoiding the duplication?

tomwinnington:
Test this please

Tasiobg:
I had a look again and the collection changes looks good to me

PR #973, opened 2019-01-21
PROGRAMMES-6794 fix icon position

Tasiobg:
Button icons in v3 are aling at the text baseline instead of looking centered to the button, this code fix the issue, I checked in some browsers including Edge and seems working fine but I'm not sure if this is worth implementing or if there is any better solution.

The difference is small, I have attached 2 screenshots to help understanding the issue

**Before**
screen shot 2019-01-21 at 16 50 45
**After**
screen shot 2019-01-21 at 16 51 05


programmes-developers:

PR #976, opened 2019-01-22
[PROGRAMMES-6896] non-podcastable episodes showing "Other podcasts you may like" 

tomwinnington:
Ticket:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6896

Link:
http://sandbox.bbc.co.uk/programmes/b006qj9z/episodes/downloads

Before:
screen shot 2019-01-22 at 11 31 20

After:
screen shot 2019-01-22 at 11 31 49


programmes-developers:

templates/podcast/podcast.html.twig
```diff
@@ -131,38 +131,40 @@
...
-                                    'body_options':{
-                                        'show_synopsis': false,
-                                    },
-                                    'image_options': {
-                                        'classes' : '1/3@bpb1 1/4@bpb2 1/3@bpw hidden@bpw2'
-                                    },
-                                }) }}
-                            </li>
-                        </ul>
+                {% if programme.isPodcastable() %}
```
BjornTwachtmann: shouldn't this be entity.isPodcastable() as above?

BjornTwachtmann: https://github.com/bbc/programmes-frontend/pull/976/files#diff-d6ceb9aa9878f15f0765d80bf3869eebR8

Tasiobg:
For this example V2 hides the whole secondary column
https://www.bbc.co.uk/programmes/b00f6hbp/episodes/downloads

This is because the next condition evaluates to false
https://github.com/bbc/programmes/blob/develop/webapp/php/sharedmodules/programmes/views/episodes/downloads.rv.spv#L139

And the condition evaluates to false becuase this response is null https://github.com/bbc/programmes/blob/develop/webapp/php/sharedmodules/programmes/controllers/EpisodesController.php#L206

I think we should hide the whole column if episode is non-podcastable

Tasiobg:
I think there are some issue with the merge, this PR would remove some fixed I did recently.

tomwinnington:
> I think there are some issue with the merge, this PR would remove some fixed I did recently.

Yeah I spoke to Bjorn about this. Probably easier for me to make a new branch as it's only a one line change.

PR #977, opened 2019-01-22
PROGRAMMES-6794 top episode size and remove some unused namespaces

Tasiobg:
Looks like in v2 only the top episode in the first page has an increased size, this is a fix to replicate v2 behaviour.
Also removed some unused namespaces
https://www.test.bbc.co.uk/programmes/p02p8mks/episodes/downloads?page=11
https://www.bbc.co.uk/programmes/p02p8mks/episodes/downloads?page=11


programmes-developers:

PR #979, opened 2019-01-23
[No ticket] Fix Telescope vote padding

benolot:


programmes-developers:

PR #980, opened 2019-01-23
PROGRAMMES-6794 fix download button classes

Tasiobg:
V3 was using same classes in Download and Subscribe buttons popup but they should be different, this fix updates the Download button to use the same v2 classes classes
**V2**
screen shot 2019-01-23 at 11 50 08
**V3 before fix**
screen shot 2019-01-23 at 11 49 53


programmes-developers:

BjornTwachtmann:
@Tasiobg Good spot, but on the episode page it is _meant_ to be using the other colour scheme, so it probably needs an option :(

PR #981, opened 2019-01-23
PROGRAMMES-6794 set iStats ProgsPageType to same value as v2

Tasiobg:


programmes-developers:

PR #983, opened 2019-01-24
[PROGRAMMES-6897] autowiring changes

cisco87:
I am not sure this is what you had in mind @BjornTwachtmann but autowiring more solves the problem


programmes-developers:

config/services.yaml
```diff
@@ -58,22 +58,38 @@ services:
                             # fetching services directly from the container via $container->get() won't work.
                             # The best practice is to be explicit about your dependencies anyway.
 
-    # makes classes in src/ available to be used as services
-    # this creates a service per class whose id is the fully-qualified class name
     App\:
-        resource: '../src/{ArgumentResolver,Ds2013,DsAmen,DsShared,ExternalApi,Branding,Translate,Twig}'
+        resource: '../src/*/{Mapper,Factory,Helpers,Service}'
```
BjornTwachtmann: I would avoid this wildcard and be specific about what you're mapping here.

config/services.yaml
```diff
@@ -58,22 +58,38 @@ services:
...
+
+    App\ExternalApi\:
+        resource: '../src/ExternalApi/*/{Mapper,Factory,Helpers,Service}'
 
-    # controllers are imported separately to make sure services can be injected
-    # as action arguments even if you don't extend any base controller class
     App\Controller\:
         resource: '../src/Controller'
         tags: ['controller.service_arguments']
```
BjornTwachtmann: Exclude the helpers from this tag and include them elsewhere if necessary

PR #986, opened 2019-01-24
[PROGRAMMES-6913] Remove UK only string from download button

benolot:
Ticket: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6913


programmes-developers:

PR #989, opened 2019-01-28
[PROGRAMMES-6913] UK Only download button

tomwinnington:
Ticket:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6913

Link:
http://sandbox.bbc.co.uk/programmes/p02nrspn/episodes/downloads

screen shot 2019-01-28 at 15 03 27

screen shot 2019-01-28 at 15 06 46


programmes-developers:

src/Ds2013/Presenters/Utilities/Download/DownloadPresenter.php
```diff
@@ -31,7 +31,7 @@ public function __construct(
         Version $version,
         ?Podcast $podcast,
         bool $secondary = true,
-        array $options = []
+        array $options = ['showUkOnly' => true]
```
cisco87: I think this is not sustainable, could you create a constant or a value for this? 
```
const DEFAULT_OPTIONS = ['showUkOnly' => true];

public function __construct(array $options = []) {
    if (empty($options) {
        $options = self::DEFAULT_OPTIONS;
    }
}
```

I know your solution is easy and simple, but I see it as a time bomb I'm not sure if you agree

src/Ds2013/Presenters/Utilities/Download/DownloadPresenter.php
```diff
@@ -87,4 +87,13 @@ public function getPositionType()
     {
         return ($this->secondary === true) ? 'secondary' : 'subtle';
     }
+
+    public function showUkOnly(): bool
+    {
+        if(array_key_exists('showUkOnly', $this->options) ){
+            return $this->getOption('showUkOnly');
```
cisco87: this looks like a lot of boilerplate code, can you change getOption to something like

```
public function getOption($key, $default = null) {
   if (array_key_exists($key, $this->options) {
      return $this->options[$key];
   }
   return $default;
}
```

so you can have showUkOnly only return 
```
return $this->getOption('showUkOnly', true);
```

src/Ds2013/Presenters/Utilities/Download/DownloadPresenter.php
```diff
@@ -25,6 +25,10 @@ class DownloadPresenter extends Presenter
     /** @var bool */
     private $secondary;
 
+    protected $options = [
```
cisco87: Nice!

PR #991, opened 2019-01-30
Programmes 6878 contact form redirect loop

cisco87:
If I understood the problem correctly, this should fix it 

https://sandbox.bbc.co.uk/programmes/articles/3mh1WBFYSqTSVN2znrmr8gM/testing-embeds-galore/contact?__flush_cache=1


programmes-developers:

PR #1004, opened 2019-02-07
PROGRAMMES-6800 Telescope vote: exclude bogus VPN's

Tasiobg:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6800

- Added "is_uk_only" to telescope mapper
- Check if vote can be displayed based on iSite option and GeoIP header
- Fix getBaseIsiteObject() returned value (it doesn't return a BaseIsiteObject)
- Added text as translation but not translated in any other language for now
- Unitest


programmes-developers:

src/Controller/Articles/ShowController.php
```diff
@@ -81,4 +83,43 @@ protected function getRouteName()
...
+     * @param array $rowGroups
+     * @param Request $request
+     * @return bool
+     */
+    private function canDisplayVote(array $rowGroups, Request $request): bool
+    {
+        foreach ($rowGroups as $rowGroup) {
+            foreach ($rowGroup->getPrimaryBlocks() as $primaryBlocks) {
+                if ($primaryBlocks instanceof Telescope) {
+                    if ($primaryBlocks->isUkOnly() === true) {
```
Bobkidbob: This if statement and the one above could be merged.

src/Controller/Articles/ShowController.php
```diff
@@ -81,4 +83,43 @@ protected function getRouteName()
...
+                        }
+                    }
+                }
+            }
+            foreach ($rowGroup->getSecondaryBlocks() as $secondaryBlocks) {
+                if ($secondaryBlocks instanceof Telescope) {
+                    if ($secondaryBlocks->isUkOnly() === true) {
+                        $this->response()->headers->set('vary', 'X-Ip_is_uk_combined');
+                        if ($request->headers->get('X-Ip_is_uk_combined') === 'no') {
+                            return false;
```
Bobkidbob: This code is repeated above and could be a separate function.

src/Controller/BaseIsiteController.php
```diff
@@ -59,7 +59,7 @@ protected function getRedirectToSlugIfNeeded(BaseIsiteObject $isiteObject, bool
         return null;
     }
 
-    protected function getBaseIsiteObject(string $guid, bool $preview): BaseIsiteObject
+    protected function getBaseIsiteObject(string $guid, bool $preview)
```
BjornTwachtmann: Why change the typehint?

Tasiobg: PhpStan complains
 
```
 Line   src/Controller/Articles/ShowController.php

  76     Call to an undefined method App\ExternalApi\Isite\Domain\BaseIsiteObject::getRowGroups().
```
`getRowGroups` exisit in Articles.php but it doesn't in the base class `BaseIsiteObject`

Any improvement suggestion?

templates/find_by_pid/tlec.html.twig
```diff
@@ -41,19 +41,23 @@
...
+                            </noscript>
+                        </div>
+                        <div id="account-wrapper-vote"></div>
                     </div>
-                    <div id="account-wrapper-vote"></div>
                 </div>
             </div>
-        </div>
+        {% else %}
+            <div class="br-box-subtle island">{{ tr('vote_not_available') }}</div>
```
BjornTwachtmann: This is going to need basically the same container classes as the actual vote. If you check this on sandbox, this will display full browser width (e.g. not respect the page width) and look pretty off.

templates/find_by_pid/tlec.html.twig
```diff
@@ -216,15 +220,18 @@
...
+                            <div class="grid">
+                                <div class="grid__item 1/2@gel3b">
+                                    {% include 'shared/telescope/target_element.html.twig' with {
+                                        container_html_id: 'account-wrapper-vote'
+                                    } %}
+                                </div>
                             </div>
-                        </div>
+                        {% else %}
+                            <div class="br-box-subtle island">{{ tr('vote_not_available') }}</div>
```
BjornTwachtmann: Same comment here really, I think this needs the same grid classes whether the vote is present or not or it will be in a different box.

Tasiobg: I move the condition inside the grid, I think it makes more sense, thanks

templates/find_by_pid/tlec.html.twig
```diff
@@ -44,13 +44,17 @@
...
                 <div class="grid__item">
-                    <div class="vote__noscript-container">
-                        <noscript class="gel-body-content">
-                            This vote does not support non-JavaScript. Please enable JavaScript in your browser and refresh the page in order to vote.
-                            <a href="https://www.bbc.co.uk/iplayer/help/enable_javascript">Find out how.</a>
-                        </noscript>
-                    </div>
-                    <div id="account-wrapper-vote"></div>
+                    {% if canDisplayVote %}
+                        <div class="vote__noscript-container">
```
BjornTwachtmann: One other question: Should this not do, 
``` {% include 'shared/telescope/target_element.html.twig' with {
                                        container_html_id: 'account-wrapper-vote'
                                    } %}```
like the other vote on this page. Realise that this code was written before this PR but we might as well tidy it up

File: src/Controller/BaseIsiteController.php (hunk 3 above)
Comment (cisco87) [reply src/Controller/BaseIsiteController.php]: You can put a local phpdoc above the variablr in the calling code afaik

/* @var Article $article */
$article = etc

Comment (Tasiobg) [reply src/Controller/BaseIsiteController.php]: Thanks for the feedback but still `getBaseIsiteObject` should not be typehind as returning `BaseIsiteObject` because the base class probably won't implement all the methods of the classes that extend this class. Let me know if you think this is a wrong approach and we can discuss

File: templates/find_by_pid/tlec.html.twig (hunk 6 above)
Comment (Tasiobg) [reply templates/find_by_pid/tlec.html.twig]: ~~That would make sense, not sure if fit in our code~~ Good idea :+1:

File: src/Controller/BaseIsiteController.php (hunk 3 above)
Comment (cisco87) [reply src/Controller/BaseIsiteController.php]: I think casting is quite a common practice, and my suggestion might just be seen like that in a PHP way, also you are removing the return type in order to keep the returned object consistent with it's nature but then the method name still says `getBaseIsiteObject` which then telling you that the class is returning a baseisiteobject no?

File: src/Controller/Articles/ShowController.php [hunk 7]
```diff
@@ -81,4 +83,39 @@ protected function getRouteName()
...
+     * "UK Only" is enabled the restriction will be applied to all of them.
+     *
+     * If the Telescope vote is set to UK Only we need to "vary" on "X-Ip_is_uk_combined" header to avoid returning
+     * same cached content to non UK users when there is a vote set to UK only
+     *
+     * @param array $rowGroups
+     * @param Request $request
+     * @return bool
+     */
+    private function canDisplayVote(array $rowGroups, Request $request): bool
```
cisco87: I am not sure but just to clarify, if one telescope vote `isUkOnly` you treat them all as `isUkOnly`?

src/Controller/Articles/ShowController.php
cisco87: What about something like
```
private function canDisplayVote(array $rowGroups, Request $request): bool
    {
        if ($request->headers->get('X-Ip_is_uk_combined') !== 'no') {
            return true;
        }
        

        foreach ($rowGroups as $rowGroup) {
            if (!$this->checkRowGroup($rowGroup->getSecondaryBlocks())
                || !$this->checkRowGroup($rowGroup->getPrimaryBlocks())) {
                
            }
        }

        return true;
    }

    private function checkRowGroup(array $blocksList)
    {
        foreach ($blocksList as $secondaryBlocks) {
            if ($secondaryBlocks instanceof Telescope && $secondaryBlocks->isUkOnly() === true) {
                $this->response()->headers->set('vary', 'X-Ip_is_uk_combined');
                return false;
            }
        }
    }
```

PR #1013, opened 2019-02-13
I have no idea how this could possibly have worked before

BjornTwachtmann:
This fixes a bug where a horizontal scrollbar is visible in firefox in the gallery page (that bug is currently on live, but I'm only fixing it here unless somebody objects). 

Look at the gallery JS where scrollbar height is calculated if you want to see how this works.


programmes-developers:

PR #1014, opened 2019-02-13
PROGRAMMES-6904 - created A to Z routes and basic layout

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6904

I need your eyeballs, my brain no longer functions correctly. Not that it ever did.


programmes-developers:

src/Ds2013/Presenters/Section/Atoz/SearchBar/atoz_search_bar.html.twig
```diff
@@ -0,0 +1,20 @@
+{% spaceless %}
+    <form class="component search-bar" action="/programmes/a-z/{{ atoz_search_bar.getSlice() }}" method="GET">
+        <label class="search-bar__label">
+            <span class="visually-hidden">{{ tr('atoz_search_label') }}</span>
+            <input
+                class="search-bar__input"
+                type="search"
+                name="q"
```
benolot: Should we maybe change this to something like "Query"? Just "q" when it's an A-Z listing by letter may be confusing.

config/routes.yaml
```diff
@@ -210,3 +210,18 @@ cloud_labs:
...
+
+atoz_list:
+    path: /programmes/a-z/{slice}
+    defaults: { _controller: App\Controller\Atoz\ListController }
+    requirements: { slice: 'all|player' }
+
+atoz_show:
+    path: /programmes/a-z/by/{search}/{slice}
+    defaults: { _controller: App\Controller\Atoz\ShowController }
+    requirements: { slice: 'all|player' }
```
BjornTwachtmann: Stuff without a slice is allowed in V2 and renders the player view. We should probably redirect a-z/by/a to a-z/by/a/player

src/Controller/Atoz/ListController.php
```diff
@@ -0,0 +1,26 @@
...
+    {
+        $searchQuery = $request->query->get('q');
+        if (isset($searchQuery)) {
+            return $this->cachedRedirectToRoute('atoz_show', [
+                'search' => $searchQuery,
+                'slice' => $slice,
+            ]);
+        }
+
+        return $this->renderWithChrome('atoz/list.html.twig', [
```
BjornTwachtmann: This should probably just redirect to the index, it's not doing anything much useful is it?

src/Controller/Atoz/ListController.php
```diff
@@ -0,0 +1,26 @@
...
+use Symfony\Component\HttpFoundation\Request;
+
+class ListController extends BaseController
+{
+    public function __invoke(Request $request, string $slice)
+    {
+        $searchQuery = $request->query->get('q');
+        if (isset($searchQuery)) {
+            return $this->cachedRedirectToRoute('atoz_show', [
+                'search' => $searchQuery,
```
BjornTwachtmann: Is it safe to pass the query string here unmodified? I can't think of any specific vulnerabilities, but I can see that V2 is restricting this to [A-Za-z0-9 '\-]

src/Ds2013/Presenters/Section/Atoz/LetterNav/atoz_letter_nav.html.twig
```diff
@@ -0,0 +1,21 @@
...
+            {{ self.item(letter, atoz_letter_nav.getSlice(), letter, atoz_letter_nav.getCurrent()) }}
+        {%- endfor -%}
+        <li class="br-page-link-onbg015 letter-nav__spacer"></li>
+    </ol>
+</div>
+
+{%- macro item(search, slice, text, current, classes) -%}
+    <li class="br-box-page br-page-bg-onborder letter-nav__page{% if classes is not empty %} {{ classes }}{% endif %}">
+        {% if text != current %}
+            <a class="br-page-link-onbg015 br-page-linkhover-onbg--hover br-page-bg-ontext--hover" href="/programmes/a-z/by/{{ search }}/{{ slice }}">{{ text }}</a>
```
BjornTwachtmann: This href should probably use the path function, no?

BjornTwachtmann: That will make sure to correctly URLencode things too

src/Controller/Atoz/ShowController.php
```diff
@@ -0,0 +1,28 @@
...
+    {
+        $current = '';
+        if (strlen($search) === 1) {
+            if ($search === '@') {
+                $current = '0-9';
+            } else {
+                $current = $search;
+            }
+        }
+
```
BjornTwachtmann: Just because this popped into my head and I will probably forget if you don't, can you stick a comment here saying that we should add meta robots noindex,nofollow for keyword searches, there's loads of them in google and it's pointless to have them crawled. Letters still need to be indexed though.

config/routes.yaml
```diff
@@ -210,3 +210,27 @@ cloud_labs:
...
+    path: /programmes/a-z/{slice}
+    defaults: { _controller: App\Controller\Atoz\ListController }
+    requirements: { slice: 'all|player' }
+
+atoz_naked_letter:
+    path: /programmes/a-z/by/{search}
+    defaults: { _controller: FrameworkBundle:Redirect:redirect, route: atoz_show, slice: "player" }
+    requirements: { search: '[a-zA-Z@]{1}' }
+
+atoz_naked_search:
```
BjornTwachtmann: this and the atoz_naked_letter route can be combined and redirected to the player slice I think, that seems to be the behaviour on live (well, not the actual redirect, but same effect)

You should also use `controller: App\Controller\RedirectAndCacheController::redirectAction`, which works the same way, but adds caching headers for BBC Varnish

BjornTwachtmann:
If that one minor change is made I'm happy for this to be merged.

PR #1016, opened 2019-02-18
PROGRAMMES-6940 - created redirects found in A to Z v2

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6940

You may want to check v2 (https://github.com/bbc/programmes/blob/develop/webapp/php/sharedmodules/programmes/controllers/AtozController.php) as well since I created the ticket requirements from that.


programmes-developers:

PR #1019, opened 2019-02-20
[PROGRAMMES-6815] Fix the fix that was added for the fix

benolot:
Ticket: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6815


programmes-developers:

benolot:
It's not actually fixed - it's been released upstream but not made it into the redhat mirrors yet.

PR #1020, opened 2019-02-20
PROGRAMMES-6907 - added missing A to Z meta information

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6907

@BjornTwachtmann I have also added the robots meta tag, let me know if there's a better way to do it.


programmes-developers:

PR #1024, opened 2019-02-26
Remove sensiolabs/security-checker

Tasiobg:
Remove `sensiolabs/security-checker` because is not working and we can't update to the last version, this is required or the APP won't build

All changes has beend done automatically after running
```
composer remove "sensiolabs/security-checker"
```


programmes-developers:

PR #1025, opened 2019-03-01
some fixes

Tasiobg:
- ProfileMapper.php was throwing an error when attemptin to instanciate `new Pid(null)`, I have to use `Throwable` as a generic way to catch any exception, in PHP7 the type errors are not reported as Exceptions anymore http://php.net/manual/en/language.errors.php7.php
Example: /programmes/b006m86d/profiles (In test DB)

- StructuredDataHelper.php was throwing this error `StructuredDataHelper::getSchemaForProgrammeContainer() must be an instance of instance ProgrammeContainer, instance of Episode given` This happen because the $programme for this article is a clip and the parent an Episode. I'm not sure if looping through the parents is a good idea, we probably should discuss this solution and also think if we should add any try / cath into the controllers when getting the schema
Example: /programmes/articles/grTP6yCQz3zy3n3fhBR8QT/about-the-apprentice (In test DB)


programmes-developers:

src/Controller/Helpers/StructuredDataHelper.php
```diff
@@ -143,8 +143,14 @@ public function getSchemaForArticle(Article $article, ?Programme $programme, boo
             $programme = $programme->getParent();
         }
         $schema = $this->schemaHelper->buildSchemaForArticle($article, $programme);
-        if ($showParent and $programme) {
-            $schema['isPartOf'] = $this->getSchemaForProgrammeContainer($programme);
+
+        if ($showParent) {
```
BjornTwachtmann: Just call $this->getSchemaForCoreEntity()

PR #1008, opened 2019-02-12
[PROGRAMMES-6870] Removed Artist Links For BBC Music

tomwinnington:
Ticket:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6870

Links:
http://sandbox.bbc.co.uk/programmes/p0215y1j

http://sandbox.bbc.co.uk/programmes/m00017k9

PR #1026, opened 2019-03-04
revert PROGRAMMES-6870 Removed Artist Links For BBC Music

Tasiobg:
I have reverted the changes made here https://github.com/bbc/programmes-frontend/pull/1008/files because we don't want to release it yet


programmes-developers:

PR #1027, opened 2019-03-05
PROGRAMMES-6905 - create A to Z All Programmes page

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6905


programmes-developers:

src/Controller/Atoz/ShowController.php
```diff
@@ -5,30 +5,105 @@
...
+    /** @var int */
+    private $currentPage;
+
+    public function __invoke(
+        string $search,
+        string $slice,
+        AtozTitlesService $atozTitlesService,
+        Request $request
+    ) {
+        $this->currentPage = (int) $request->query->get('page', 1);
```
BjornTwachtmann: We already have a $this-getPage in the baseController you can use instead of this

templates/atoz/show.html.twig
```diff
@@ -22,7 +22,35 @@
...
+                            'show_masterbrand': true,
+                            'show_child_availability': true
+                        }
+                    }) }}
+                </li>
+            {% endfor %}
+        </ol>
+    </div>
+    {% if count > resultsPerPage %}
+        {{ ds2013('paginator', currentPage, resultsPerPage, count) }}
```
BjornTwachtmann: I would probably instantiate the paginator in the controller, it saves you having to do this logic in the view and pass all these variables through. Check Controller/ProgrammeEpisodes/PlayerController.php for an example.

src/Controller/Atoz/ShowController.php
```diff
@@ -5,30 +5,105 @@
...
+        if ($this->currentPage < 1) {
+            $this->pageNotInRange();
         }
 
-        if ($slice === 'all') {
-            $descriptionSlice = 'all';
+        if (array_key_exists(strtolower($search), self::LETTERS)) {
```
BjornTwachtmann: `if (preg_match('/^[a-z@]$/', strtolower($search)` would acomplish the same and be less verbose. This is fine though if you prefer it.

PR #1029, opened 2019-03-06
PROGRAMMES-6906 - create A to Z Available Programmes page

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6906


programmes-developers:

src/Controller/Atoz/ShowController.php
```diff
@@ -45,7 +47,15 @@ public function __invoke(
...
-                default: // Case for 'player' will be added in available programmes ticket.
+                case 'player':
+                    $count = $this->verifyPageIsInRange($atozTitlesService->countAvailableTleosByFirstLetter($search));
+                    $results = $atozTitlesService->findAvailableTleosByFirstLetter(
+                        $search,
+                        self::RESULTS_PER_PAGE,
+                        $this->currentPage
+                    );
+                    break;
+                default:
```
BjornTwachtmann: Should we ever end up in the default block here? Should it be a 404 if so? Not quite sure in which circumstances we'd end up there.

BjornTwachtmann:
Only other comment would be that the listing page is still missing the count that live has. I'm not really that bothered by it, but it will be brought up in test, so be aware.

Bobkidbob:
@BjornTwachtmann If you are referring to the number of programmes in the slice subnav, the acceptance criteria explicitly says to remove it.

src/Controller/Atoz/ShowController.php
Bobkidbob: There is no reason for the default to exist, but the linter complains if it is possible for the variables inside it to not be defined. I thought it would be better to leave them in the default rather than setting them before the switch each time for no reason.

PR #1031, opened 2019-03-06
PROGRAMMES-6937 - create A to Z Results page

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6937

Requires https://github.com/bbc/programmes-pages-service/pull/280


programmes-developers:

src/Controller/Atoz/ShowController.php
```diff
@@ -52,8 +56,27 @@ public function __invoke(
         } else {
             $selectedLetter = '';
 
-            $count = 0; // This will be replaced in the results page ticket.
-            $results = [];  // This will be replaced in the results page ticket.
+            switch ($slice) {
+                case 'all':
+                    $count = $this->verifyPageIsInRange($programmesService->countByKeywords($search));
+                    $results = $programmesService->searchByKeywords(
```
BjornTwachtmann: Just thinking about it, is there an argument not to bother running the search queries when we know there are no results (count = 0)? It's a bit ugly, but these are rather expensive.

PR #1042, opened 2019-03-25
PROGRAMMES-6972 - descriptive descriptions

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6972


programmes-developers:

src/Controller/Profiles/ShowController.php
```diff
@@ -74,6 +74,8 @@ public function __invoke(
             ]);
         }
 
+        $this->overridenDescription = $programme->getTitle() . ' profiles: ' . $isiteObject->getTitle();
```
Tasiobg: In theory `$programme` could be null here so it could throw and exception.

Tasiobg:
Does it cover all cases mentioned in the ticket?
for example this page seems to not havea good descrition: https://sandbox.bbc.co.uk/programmes/profiles/3r2cnZ9Cvg2lZ7Ff0ZjrC4W/davros
I suggest to hold this merge until more feedback is provided

PR #1047, opened 2019-03-26
[PROGRAMMES-6900] Fix stuff being sent to live bucket accidentally

benolot:
Ticket: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6900

And yes this does work if context is null: IE homepage


programmes-developers:

PR #1048, opened 2019-03-26
Programmes 6981 only series with clips

Tasiobg:
Use the new PPS function to get series with clips only and remove the 300 limit because it doesn't exist in v2


programmes-developers:

PR #1051, opened 2019-04-02
[PROGRAMMES-6988] Added two things with no medium to use the TV default producer value

benolot:
Ticket: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6988


programmes-developers:

PR #1055, opened 2019-04-05
Composer update, and fix search parameter name on hompage

BjornTwachtmann:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6937?focusedCommentId=4270698&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4270698

The homepage uses "?q=" rather than "?query=" for some reason, which is different to the rest of the app. I've verified this change works in both V2 and V3, so won't break anything when released.


programmes-developers:

PR #1057, opened 2019-04-11
PROGRAMMES-6959 update composer

Tasiobg:
Update composer to get last PPS version with the updated findTleosByCategory


programmes-developers:

PR #1058, opened 2019-04-12
PROGRAMMES-6958 - create category metadata page

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6958

I have also fixed some routes that weren't changed after migration from v2.


programmes-developers:

config/routes/3rd_party.yaml
```diff
@@ -73,17 +73,6 @@ my_tracks:
 
 ### /programmes v2
 
-atoz_list_all:
-    path: /programmes/a-z/all
-    defaults: { _controller: FrameworkBundle:Redirect:redirect, route: '' }
-    schemes: [https]
-
-categories_show:
```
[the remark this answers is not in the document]
Bobkidbob: They are in the main routes file now.

Bobkidbob:
Just remembered I forgot the JS for the subcategory nav on mobile. I will fix it on Monday.

translations/programmes/en.po
```diff
@@ -79,6 +79,10 @@ msgstr "All podcasts"
 msgid "all_previous_episodes"
 msgstr "All previous episodes"
 
+#. English: "All programmes".
+msgid "all_programmes"
```
BjornTwachtmann: There is already an all programmes translation (atoz_all)

src/Controller/Category/MetadataController.php
```diff
@@ -0,0 +1,75 @@
...
+                if ($category === null) {
+                    throw $this->createNotFoundException('Format does not exist.');
+                }
+                $parent = null;
+                $title = $category->getTitle();
+                $this->overridenDescription = 'Find BBC programmes categorised as "' . $title . '".';
+                $children = [];
+                break;
+            default:
+                $category = null;
```
BjornTwachtmann: This case shouldn't really be possible given the routing. In any case it would be a 404, so might as well throw that exception.

src/Controller/Category/MetadataController.php
```diff
@@ -0,0 +1,75 @@
...
+            case 'genres':
+                $category = $categoriesService->findGenreByUrlKeyAncestry(
+                    array_reverse(explode('/', $categoryHierarchy)),
+                    CacheInterface::MEDIUM
+                );
+                if ($category === null) {
+                    throw $this->createNotFoundException('Genre does not exist.');
+                }
+                $title = $category->getTitle();
+                $this->overridenDescription = 'Find BBC programmes categorised as "' . $title . '"';
```
BjornTwachtmann: This could probably be pulled into a separate method, given it's repeated below.

src/Controller/Category/MetadataController.php
```diff
@@ -0,0 +1,75 @@
...
+            default:
+                $category = null;
+                $parent = null;
+                $title = '';
+                $children = [];
+        }
+
+        return $this->renderWithChrome('category/metadata.html.twig', [
+            'categoryType' => $categoryType,
+            'categoryHierarchy' => $categoryHierarchy,
```
BjornTwachtmann: I would use getUrlKeyHeirarchy from the domain object rather than passing through the user set values here.

src/Controller/Category/MetadataController.php
```diff
@@ -0,0 +1,75 @@
...
+            'category' => $category,
+            'parent' => $parent,
+            'title' => $title,
```
BjornTwachtmann: do we need to pass through parent and title, aren't these available from the domain object?

src/Ds2013/Css/PageSpecfic/_category.scss
```diff
@@ -0,0 +1,278 @@
...
+     * At large sizes navigation is a vertical list.
+     */
+
+    .categories-navigation {
+        width: 162.66016px;
+    }
+
+    .categories-navigation__heading {
+        margin-bottom: $basic-spacing;
+        padding: 0.25em $basic-spacing;
```
BjornTwachtmann: you're missing a  `margin-left: -$basic-spacing;` from the old CSS here that's making the nav indentation allok weird.

src/Controller/Category/MetadataController.php
BjornTwachtmann: or even add some kind of getHeirarchicalTitle to the domain object?

src/Controller/Category/MetadataController.php
BjornTwachtmann: There is also an existing translation for "Programmes categoriesed as %1" (category_title)

templates/category/metadata.html.twig
```diff
@@ -0,0 +1,81 @@
...
+                    {{ tr('filter_by') }}
+                </h2>
+                <div class="categories-navigation-filter-by__button island hidden@bpw br-box-page__link br-page-link-onbg015 br-page-linkhover-onbg--hover br-page-bg-ontext--hover js-subcategory-toggle component">
+                    {{ gelicon('core', 'list', 'gelicon--centi') }}
+                    {{ tr('filter_by') }}
+                </div>
+                <ol class="categories-navigation__options categories-navigation-filter-by__options hidden visible@bpw br-page-link-onbg015 br-page-link-onborder--beforeafter">
+                    {% for child in children %}
+                        <li class="category">
+                            <a href="{{ path('category_metadata', { categoryType: categoryType, categoryHierarchy: category.getUrlKey() ~ '/' ~ child.getUrlKey() }) }}" class="categories-navigation__link categories-navigation-filter-by__link centi br-page-linkhover-onbg--hover br-page-bg-ontext--hover deemphasize">
```
BjornTwachtmann: There is a getUrlKeyHierarchy function in Genre that you need to use here to generate the correct URLs for third level categories (see /programmes/genres/music/danceandelectronica for example ). It may be worth moving that function to the base category class and overriding it so that you don't need to do a lot of instanceof checks.

templates/category/metadata.html.twig
```diff
@@ -0,0 +1,81 @@
...
+        {{ ' ' ~ tr('formats') }}
+    {%- endif -%}
+    {{ ' - ' ~ title }}
+{%- endblock -%}
+{% block page_classes %}programmes-page text-base b-g-p br-box-page{% endblock %}
+
+{% block body %}
+    <h1 class="alpha categories-breadcrumb">
+        <span class="categories-breadcrumb__item deemphasize">{{ tr('categories') }}:</span>
+        {% if parent is not null %}
```
BjornTwachtmann: This is going to fail for categories with more than one parent, e.g. /programmes/genres/music/danceandelectronica/drumandbass

translations/programmes/en.po
Bobkidbob: `atoz_all` is for "All Programmes" not "All programmes". I don't think we can assume capitalisation rules are the same for different languages.

src/Controller/Category/MetadataController.php
Bobkidbob: Getting the translations into the controller is probably more effort than it's worth bearing in mind the page is only shown in English anyway.

BjornTwachtmann: Fair enough, we probably don't have translations for all of it anyway I suppose

translations/programmes/en.po
BjornTwachtmann: VERY WELL

src/Controller/Category/MetadataController.php
```diff
@@ -0,0 +1,60 @@
...
+                if ($category === null) {
+                    throw $this->createNotFoundException('Format does not exist.');
+                }
+                $children = [];
+                break;
+            default:
+                throw $this->createNotFoundException('Category does not exist.');
+        }
+
+        $this->overridenDescription = 'Find BBC programmes categorised as "' . $category->getHierarchicalTitle() . '".';
```
benolot: Should this come from a translation string?

src/Ds2013/Css/PageSpecfic/_category.scss
```diff
@@ -0,0 +1,280 @@
...
+        li {
+            @include inline-block();
+        }
+    }
+
+    .categories-navigation-top__link {
+        // Undo Faux columns overflow and background coloring
+        margin-bottom: 0;
+        border-bottom-style: none;
+        //background: inherit;
```
benolot: Any reason this is commented out?

Bobkidbob: Absolutely no idea.

src/Controller/Category/MetadataController.php
Bobkidbob: It is more effort getting the translations into the controller than it is worth, so I done it the same way it's done in A to Z pages.

src/Controller/Category/MetadataController.php
```diff
@@ -0,0 +1,60 @@
...
+
+        return $this->renderWithChrome('category/metadata.html.twig', [
+            'categoryType' => $categoryType,
+            'categoryHierarchy' => $category->getUrlKeyHierarchy(),
```
BjornTwachtmann: Does this need to be passed separately rather than just called from the view layer?

PR #1062, opened 2019-04-16
[PROGRAMMES-7002] Removed A-Z Search

benolot:
Ticket: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-7002

I just flat out removed it, iirc there was no plans to redirect to main BBC search.


programmes-developers:

Tasiobg:
test this please

config/routes.yaml
```diff
@@ -216,30 +216,21 @@ atoz_index:
     path: /programmes/a-z
     defaults: { _controller: App\Controller\Atoz\IndexController }
 
-atoz_list:
-    path: /programmes/a-z/{slice}
-    defaults: { _controller: App\Controller\Atoz\ListController }
-    requirements: { slice: 'all|player' }
-
 atoz_redirect_current_list:
```
Bobkidbob: This can also go, it is an old slice that would've been handled by `atoz_list`.

config/routes.yaml
```diff
@@ -216,30 +216,21 @@ atoz_index:
...
     defaults: { route: atoz_show, slice: "player", permanent: true }
+    requirements: { search: '^[a-zA-Z@]$' }
 
 atoz_show:
     path: /programmes/a-z/by/{search}/{slice}
     defaults: { _controller: App\Controller\Atoz\ShowController }
-    requirements: { slice: 'all|player' }
+    requirements: { slice: 'all|player',search: '^[a-zA-Z@]$' }
 
 atoz_redirect_current_show:
```
Bobkidbob: If we're getting rid of old redirects this can probably go as well.

src/Controller/Atoz/ShowController.php
```diff
@@ -36,76 +37,47 @@ public function __invoke(
...
-                        );
-                    } else {
-                        $results = [];
-                    }
-                    break;
-                default:
-                    $count = 0;
+        if (!preg_match('/^[a-zA-Z@]$/', $search)) {
+            // Throw a 404 if not a single letter or @ as that's a search and that's invalid now
+            throw new NotFoundHttpException();
```
Bobkidbob: Is this needed now that the routes have it set as a requirement?

benolot: Yeah it throws a 500 as this code makes sure it's single char or nothing (Rest of code assumes single char)

PR #1065, opened 2019-04-24
[PROGRAMMES-6815] Remove cURL version hack

benolot:


programmes-developers:

PR #1066, opened 2019-04-25
add security checker script

Tasiobg:
This seems helpful for testing locally and different branckes. It can be added into other projects, what do you think?


programmes-developers:

PR #1068, opened 2019-04-26
composer update

Tasiobg:
I have updated symfony and twig to fix the security dependencies, twig requieres some code changes due to deprecations


programmes-developers:

PR #1070, opened 2019-05-02
PROGRAMMES-6968: Stop lazy-loaded recommendation partial being redire…

BjornTwachtmann:
…cted and breaking some articles in edge case.

There is an edge case where a TLEO redirected by editorial which has an article attached will render a blank page to the user due to the lazy-loaded partial being redirected to a full HTML page, that breaks the page layout when loaded.


programmes-developers:

PR #1079, opened 2019-05-13
PROGRAMMES-7014 - add programmes list to topic page

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-7014

Feedback on the new pagination pls ty.


BjornTwachtmann:
@Bobkidbob did you meant to change the lock file: package-lock.json? Seems pointless unless I'm missing something

src/Controller/Topic/ShowController.php
```diff
@@ -39,16 +52,37 @@ public function __invoke(
...
+                '',
+                $page + 1,
+                $programmeContainer
+            );
         }
 
         $resolvedPromises = $this->resolvePromises([
             'relatedTopics' => $relatedAdaClasses,
+            'programmes' => $adaProgrammeItems,
+            'isAnotherPage' => $nextPage,
```
BjornTwachtmann: isAnotherPage sounds weird to me. nextPage, or hasNextPage maybe?

src/ExternalApi/Ada/Service/AdaClassService.php
```diff
@@ -91,6 +107,52 @@ function (array $responses): AdaClass {
         )->makeCachedPromise();
     }
 
+    public function findProgrammeItemsByClass(
```
BjornTwachtmann: I think this method belongs in AdaProgrammesService

src/ExternalApi/Ada/Service/AdaClassService.php
```diff
@@ -91,6 +107,52 @@ function (array $responses): AdaClass {
...
+                ($programmeContainer !== null ? '&programme=' . $programmeContainer->getPid() : '') .
+                ($mediaType !== '' ? '&media_type=' . $mediaType : '') .
+                '&page=' .
+                $page .
+                '&page_size=' .
+                self::PROGRAMME_LIMIT .
+                '&order=title&direction=ascending',
+            ],
+            function (array $responses): array {
+                $items = json_decode($responses[0]->getBody()->getContents(), true)['items'];
```
BjornTwachtmann: I think this is probably complex enough to merit its own function rather than being put inline. I'd also check that the bits of JSON you expect to be there are there, and throw an exception if not.

templates/topic/show.html.twig
```diff
@@ -66,7 +66,17 @@
                 </nav>
             {% endif %}
         </div>
+        {% if programmeContainer is not null %}
+            <div class="br-box-page island">
+                {{ self.results(programmes, page, isAnotherPage) }}
```
BjornTwachtmann: is having a macro here clearer than just setting the classes in a variable based on the value of programmeContainer? I don't really mind if you prefer this way.

Bobkidbob: It needs to be a macro as the results are in a different place in the DOM depending on `programmeContainer`.

src/ExternalApi/Ada/Service/AdaClassService.php
```diff
@@ -4,13 +4,15 @@
...
 use App\ExternalApi\Ada\Domain\AdaClass;
+use App\ExternalApi\Ada\Domain\AdaProgrammeItem;
 use App\ExternalApi\Ada\Mapper\AdaClassMapper;
 use App\ExternalApi\Client\Factory\HttpApiClientFactory;
 use App\ExternalApi\Client\HttpApiMultiClient;
 use App\ExternalApi\Exception\MultiParseException;
 use BBC\ProgrammesCachingLibrary\CacheInterface;
 use BBC\ProgrammesPagesService\Domain\Entity\Programme;
 use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeContainer;
+use BBC\ProgrammesPagesService\Domain\ValueObject\Pid;
```
BjornTwachtmann: don't think you need these imports

templates/topic/show.html.twig
```diff
@@ -84,3 +94,49 @@
...
+    {% if programmes is not empty %}
+        {% if page > 1 %}
+            {{ self.paginationLink('up', tr('previous'), page - 1, 'grid-unit') }}
+        {% endif %}
+        <ul class="grid">
+            {% for programme in programmes %}
+                <li class="1/1 grid__item">
+                    {{ ds_amen('programme', programme.getProgramme(), {
+                        'media_variant': 'media--row',
+                        'branding_name': 'page',
```
BjornTwachtmann: On the per-programme listing page you need to pass the programme as `context_programme` here, otherwise the name of the brand (e.g. "in our time") will be displayed against every episode.

src/ExternalApi/Ada/Service/AdaProgrammeService.php
```diff
@@ -40,6 +44,39 @@ public function __construct(
...
+    public function findProgrammeItemsByClass(
+        AdaClass $adaClass,
+        string $mediaType,
+        int $page,
+        ?ProgrammeContainer $programmeContainer
+    ): PromiseInterface {
+        $id = $adaClass->getId();
+
+        return $this->clientFactory->getHttpApiMultiClient(
+            $this->clientFactory->keyHelper(__CLASS__, __FUNCTION__, $id, $mediaType, $page),
```
BjornTwachtmann: You need to specify the programme PID here, otherwise you'll get cache conflicts

src/Controller/Topic/ShowController.php
```diff
@@ -5,32 +5,60 @@
...
                                         . $adaClass->getTitle()
                                         . '".';
             $relatedAdaClasses = $adaClassService->findRelatedClassesByClass($adaClass);
+            $adaProgrammeItems = $this->adaProgrammeService->findProgrammeItemsByClass(
+                $adaClass,
+                $this->slice,
+                $page,
+                null
+            );
+            $videoPage = self::getSlicePage('video', $adaClass, null);
```
BjornTwachtmann:
```suggestion
            $videoPage = $this->getSlicePage('video', $adaClass, null);
```

src/Controller/Topic/ShowController.php
```diff
@@ -5,32 +5,60 @@
...
+            $audioPage = self::getSlicePage('audio', $adaClass, null);
```
BjornTwachtmann:
```suggestion
            $audioPage = $this->getSlicePage('audio', $adaClass, null);
```

PR #1080, opened 2019-05-14
Stop podcast page throwing 500 error when a collection does not have a parent

BjornTwachtmann:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-7033


programmes-developers:

PR #1087, opened 2019-05-28
PROGRAMMES-7008 - topics list page

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-7008


programmes-developers:

src/Controller/Topic/ListController.php
```diff
@@ -0,0 +1,52 @@
...
+        if (count($resolvedPromises['topics']) < 1) {
+            throw $this->createNotFoundException('No topics matched your query.');
+        }
+
+        $resolvedPromises['hasNextPage'] = count($resolvedPromises['hasNextPage']) > 0;
+
+        return $this->renderWithChrome('topic/list.html.twig', array_merge($resolvedPromises, [
+            'programmeContainer' => $programmeContainer,
+            'page' => $page,
+            'routeCollection' => $this->container->get('router')->getRouteCollection(),
```
BjornTwachtmann: Rather than passing the routeCollection through the view, can we not pass it into the PresenterFactory via the constructor?

BjornTwachtmann: Actually, hang on, why do you need the routeCollection in the presenter at all when you're using encoded_path ?

Bobkidbob: Good catch. I don't, it's a hangover from before I wrote `encoded_path`.

src/Controller/Topic/ShowController.php
```diff
@@ -113,6 +113,7 @@ public function __invoke(
             'slice' => $this->slice,
             'programmeContainer' => $programmeContainer,
             'page' => $page,
+            'routeCollection' => $this->container->get('router')->getRouteCollection(),
```
BjornTwachtmann: This isn't needed is it?

src/ExternalApi/Ada/Service/AdaClassService.php
```diff
@@ -67,17 +69,43 @@ public function __construct(
...
+
+        return $this->clientFactory->getHttpApiMultiClient(
+            $this->clientFactory->keyHelper(__CLASS__, __FUNCTION__, $page, $pid),
+            [
+                $this->baseUrl .
+                '/classes?order=title&direction=ascending&page=' .
+                $page .
+                '&page_size=' .
+                self::ALL_CLASS_LIMIT .
+                ($pid !== null ? '&programme=' . $pid . '&count_context=' . $pid : ''),
```
BjornTwachtmann: it's better to urlencode these as a matter of course. I know it's not technically necessary here, but it helps to prevent potential future fuckups.

PR #1098, opened 2019-06-14
We should use this CSS everywhere in this situation

BjornTwachtmann:


programmes-developers:

Tasiobg:
Should QA have a look?

PR #1100, opened 2019-06-18
PROGRAMMES-7041 - series controller

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-7041


programmes-developers:

src/Controller/FindByPid/BaseProgrammeContainerController.php
```diff
@@ -173,17 +173,27 @@ public function __invoke(
             'clips' => $clips,
             'galleries' => $galleries,
             'mapPresenter' => $mapPresenter,
+            'shouldDisplayVote' => $this->shouldDisplayVote(),
             'isVotePriority' => $isVotePriority,
             'canDisplayVote' => $this->canDisplayVote($programme->getOption('telescope_block'), $this->request()),
```
BjornTwachtmann: canDisplayVote and shouldDisplayVote is a bit confusing as to what these variables represent. Perhaps rename canDisplayVote to canDisplayVoteInRegion or something?

BjornTwachtmann: Alternatively, you could pass an isSeriesPage value through to the template and use that for shouldDisplayVote and  shouldDisplayPriorityText I suppose.

PR #1105, opened 2019-06-26
Fix duplication of priority promotion

BjornTwachtmann:
see /programmes/b006m8wd on TEST


programmes-developers:

PR #1109, opened 2019-07-01
some fixes

Tasiobg:
- Improve getPendingEpisode. It is suppose to return only Episodes but we have seen cases in the logs where it was set to a clip (probably by mistake) and throws a 500

- In some rare cases hasPlayableDestination doesn't seems to be enough to check if there are streamable versions


programmes-developers:

src/DsAmen/Presenters/Section/Map/SubPresenter/OnDemandPresenter.php
```diff
@@ -79,7 +80,7 @@ public function getClass(): string
         return $this->class;
     }
 
-    public function getPendingEpisode(): ?Episode
+    public function getPendingEpisode(): ?ProgrammeItem
```
benolot: Could this now ever return anything other than an Episode?

BjornTwachtmann: It appears that some clips have been scheduled for broadcast and that was causing a few issues. That shouldn't really happen, but this is perfectly safe to do.

PR #1112, opened 2019-07-04
stop build after code analyser if build is trigger through a webhook

Tasiobg:


programmes-developers:

PR #1117, opened 2019-07-15
run npm audit fix

Tasiobg:
I have run `npm audit fix` to solve the vulnerability issues.
Looks like this packages got a patch update: mixin-deep, set-value and union-value


programmes-developers:

PR #1119, opened 2019-07-15
PROGRAMMES-7051 - collection page pagination

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-7051


programmes-developers:

src/DsShared/Utilities/Paginator/_paginator.scss
```diff
@@ -7,8 +7,46 @@
  * more pages as the viewport size increases.
  */
 
+$basic-spacing: 8px;
```
BjornTwachtmann: Is re-defining common SASS variables here a good idea? I think it might be safer to rename these to something paginator specific.

src/DsShared/Utilities/Paginator/paginator.html.twig
```diff
@@ -1,11 +1,11 @@
-<ol class="nav nav--banner pagination delta">
+<ol class="pagination">
     {%- if paginator.getCurrentPage() <= 1 -%}
         <li class="pagination__previous pagination--disabled">
             {{ gelicon('core', 'previous', 'pagination__next-previous gelicon--leading') }}
             <span>{{ tr('previous') }}</span>
         </li>
     {%- else -%}
-        <li class="pagination__previous">
+        <li class="br-box-page pagination__previous">
```
BjornTwachtmann: I would remove this, place the relevant branding class on the container outside of this presenter

templates/find_by_pid/collection.html.twig
```diff
@@ -33,6 +33,11 @@
                 </li>
             {% endfor %}
         </ul>
+        <div class="br-box-page island--vertical">
```
BjornTwachtmann: Does the if statement need to wrap the container class here too? We don't want an empty box on collections with <24 items.

Bobkidbob: 🤦‍♀

PR #1125, opened 2019-07-19
Make related links box less ugly and more in line with the rest of the page

BjornTwachtmann:


programmes-developers:

Tasiobg:
Code looks OK but not tested

PR #1129, opened 2019-07-25
accessibility nvda

Tasiobg:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-7066
This is an issue that only seems to happen in some windows 10 machines with NVDA, we don't think the issue is in our side but this change stop the issue from happening.


programmes-developers:

PR #1134, opened 2019-07-30
Fix broken promos on Dr. Who etc

BjornTwachtmann:
A recent commit introduced some CSS for this class, which was unused previously. This just renames this div to stop that breaking it.


programmes-developers:

PR #1133, opened 2019-07-30
7102 use details/summary pairs to render FAQ

gmarinov:
7102

 - switches FAQ module markup to `<details><summary>` pairs (changes HTML markup)
 - example pids w/ FAQ: `b006qpgr`, `b006t1q9`

QA: 
worth checking that the following components also look and function correctly
- https://www.bbc.co.uk/programmes/b062jzdq mobile view, list of tracks
- https://www.bbc.co.uk/programmes/m0001vjc synopsis
- https://www.bbc.co.uk/programmes/b006q2x0/clips filter by

dev notes:

- I suggest that we do not polyfill this. IE11 renders it open by default. Other browsers actually understand the markup.
- adds a `target-ie` SASS mixin to use where we need to target IE>=9
- there's likely a bunch of CSS to remove from `_moreless.scss` - guidance welcome

#### modern browser output (`sandbox` branding)

Screenshot 2019-07-30 at 16 25 03

#### ie11 output (`test.` branding)

ie11-safe-ss

#### side by side (mobile-ish view: old ⟷ new )

Screenshot 2019-07-31 at 10 25 26

PR #1135, opened 2019-07-31
7102+ tidy FAQ CSS

gmarinov:
note base #1133 
removes CSS no longer in use (FAQ refs in ds2013/moreless)


programmes-developers:

PR #1141, opened 2019-08-02
PROGRAMMES-6987 - fix article/profile page heading order

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6987


src/Ds2013/Presenters/Domain/ContentBlock/ContentBlockPresenter.php
```diff
@@ -50,4 +52,16 @@ public function getBrandingContext(): string
...
+
+    public function getCurrentHeading(): string
+    {
+        return 'h' . $this->currentHeading;
+    }
+
+    public function newHeadingLevel(): string
+    {
+        $this->currentHeading++;
+        // Error if heading higher than 6
```
Bobkidbob: Would probably help if I actually handled this haha.

PR #1142, opened 2019-08-02
PROGRAMMES-6299 - fixed weekly schedule page title order

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6299

The times and programme titles were already headings, but I have corrected the ordering.


src/Ds2013/Presenters/Domain/Broadcast/broadcast.html.twig
```diff
@@ -58,6 +58,10 @@
                             800: '148px',
                         }
                     },
+                    'title_options': {
+                        'title_tag': broadcast.getOption('programme_title_tag'),
```
gmarinov: `broadcast.getOption('programme_title_tag')` unDRY this?

PR #1145, opened 2019-08-03
PROGRAMMES-7113 - don't output null linked data

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-7113


templates/base_ds2013.html.twig
```diff
@@ -116,7 +116,7 @@
         <script>
             require(['rv-bootstrap']);
         </script>
-        {% if schema is defined %}
+        {% if schema is defined and schema is not null %}
```
BjornTwachtmann: Would `and schema is not empty` be better here? Would check for null, but also empty arrays

Bobkidbob:
```suggestion
        {% if schema is defined and schema is not empty %}
```

templates/base_ds_amen.html.twig
```diff
@@ -92,7 +92,7 @@
     <script>
         require(['dsamen-bootstrap']);
     </script>
-    {% if schema is defined %}
+    {% if schema is defined and schema is not null %}
```
Bobkidbob:
```suggestion
    {% if schema is defined and schema is not empty %}
```

PR #1147, opened 2019-08-04
PROGRAMMES-6879 - structured data for player page

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6879


templates/base_unbranded.html.twig
```diff
@@ -8,6 +8,11 @@
     </head>
     <body>
         {% block body %}{% endblock %}
+        {% if schema is defined and schema is not null %}
+            <script type="application/ld+json">
+                {{ schema|json_encode()|raw }}
```
benolot: Ahhh twig... how we don't love you

templates/base_unbranded.html.twig
```diff
@@ -8,6 +8,11 @@
     </head>
     <body>
         {% block body %}{% endblock %}
+        {% if schema is defined and schema is not null %}
```
Bobkidbob:
```suggestion
        {% if schema is defined and schema is not empty %}
```

PR #1148, opened 2019-08-05
6995 clipslist accessibility

gmarinov:
- clips list filter can be navigated to using `Tab`
- compromise dotted outline otherwise the focus isn't really visible
- changes markup on clips list (details/summary).
- extracts sass details/summary markup to reusable helper (previously #1133)

example pid: `b006q2x0`

Screenshot 2019-08-05 at 11 07 51
Screenshot 2019-08-05 at 11 07 56
faq-nav


src/Ds2013/Css/Components/_details.scss
```diff
@@ -0,0 +1,66 @@
+
+@mixin mixin-details($inner-indent: 26px) {
```
BjornTwachtmann: Is there a reason to make this a mixin rather than just a reusable class name? Seems like this way the CSS is likely to be repeated in the output.

gmarinov: hmmm.... looking

PR #1150, opened 2019-08-06
PROGRAMMES-7126 - add breadcrumbs

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-7126


templates/base_ds2013.html.twig
```diff
@@ -116,6 +116,11 @@
         <script>
             require(['rv-bootstrap']);
         </script>
+        {% if breadcrumbs is defined %}
```
Bobkidbob:
```suggestion
        {% if breadcrumbs is defined and breadcrumbs is not empty %}
```

templates/base_ds_amen.html.twig
```diff
@@ -92,6 +92,11 @@
     <script>
         require(['dsamen-bootstrap']);
     </script>
+    {% if breadcrumbs is defined %}
```
Bobkidbob:
```suggestion
    {% if breadcrumbs is defined and breadcrumbs is not empty %}
```

src/Controller/Articles/IndexController.php
```diff
@@ -41,6 +45,12 @@ public function __invoke(CoreEntity $coreEntity, ArticleService $isiteService)
 
         $this->overridenDescription = 'Articles about ' . $coreEntity->getTitle();
 
+        $this->breadcrumbs = $breadcrumbs
+            ->forNetwork($coreEntity->getNetwork())
```
BjornTwachtmann: $coreEntity is nullable

src/Controller/Articles/IndexController.php
```diff
@@ -41,6 +45,12 @@ public function __invoke(CoreEntity $coreEntity, ArticleService $isiteService)
...
+            ->forEntityAncestry($coreEntity)
+            ->forRoute('Features', 'programme_article_listings', ['pid' => $coreEntity->getPid()])
```
BjornTwachtmann: This route won't exist without a coreentity either

src/Controller/Articles/ShowController.php
```diff
@@ -75,6 +77,14 @@ public function __invoke(
         if ($children) {
             $paginatorPresenter = $this->getPaginator($children->getTotal());
         }
+
+        $this->breadcrumbs = $breadcrumbs
+            ->forNetwork($programme->getNetwork())
```
BjornTwachtmann: $programme is nullable

src/Controller/Articles/ShowController.php
```diff
@@ -75,6 +77,14 @@ public function __invoke(
...
+            ->forEntityAncestry($programme)
+            ->forRoute('Features', 'programme_article_listings', ['pid' => $programme->getPid()])
```
BjornTwachtmann: This won't exist for articles without a programme

src/Controller/Profiles/ShowController.php
```diff
@@ -60,6 +62,14 @@ public function __invoke(
                 self::MAX_LIST_DISPLAYED_ITEMS
             );
         }
+
+        $this->breadcrumbs = $breadcrumbs
+            ->forNetwork($programme->getNetwork())
```
BjornTwachtmann: Same as for article, programme is nullable

src/Controller/BaseController.php
```diff
@@ -264,13 +270,18 @@ protected function renderWithChrome(string $view, array $parameters = [])
...
+        $structuredDataHelper = $this->container->get(StructuredDataHelper::class);
+
         $parameters = array_merge([
             'appVersion' => $this->container->get(CosmosInfo::class)->getAppVersion(),
             'orb' => $orb,
             'meta_context' => $this->createMetaContextFromContext(),
             'branding' => $this->branding,
             'with_chrome' => true,
             'is_international' => $this->isInternational,
+            'breadcrumbs' => $structuredDataHelper->prepare(
```
BjornTwachtmann: This should probably not be invoked if breadcrumbs hasn't been set (which it won't always be, right?). We don't want empty-ish LD+JSON blocks in the output

src/Controller/Category/ListingController.php
```diff
@@ -47,6 +51,21 @@ public function __invoke(
             $paginator = new PaginatorPresenter($page, self::RESULTS_PER_PAGE, $programmeCount);
         }
 
+        $opts = ['categoryType' => $categoryType];
```
BjornTwachtmann: Why not just include this directly in the array below?

Bobkidbob: @gmarinov wrote it. It's because it's used in a parameter as well as merged into that array.

src/Controller/BaseController.php
Bobkidbob: The twig has if defined and if not empty.

src/Controller/Schedules/ByWeekController.php
```diff
@@ -113,6 +116,23 @@ public function __invoke(
...
+            ->forRoute($service->getName(), 'schedules_by_day', $opts)
+            ->forRoute(
+                $broadcastWeekStart->format('Y'),
+                'schedules_by_year',
+                ['year' => $broadcastWeekStart->format('Y')] + $opts
+            )
+            ->forRoute(
+                'Week ' . $broadcastWeekStart->format('W'),
+                'schedules_by_week',
+                ['date' => $broadcastWeekStart->format('Y/\wW')] + $opts
```
BjornTwachtmann: ->format('o/\wW') or you'll get it wrong around the year boundry.

src/Controller/BaseController.php
BjornTwachtmann: Yeah, but the structured data helper prepare method will return an array with @context in all cases, so it will never be empty in practice.

Bobkidbob: ah yh good point

PR #1154, opened 2019-08-08
6925 fix embed instagram

gmarinov:
jira 6925

- removes background from instagram embeds on the side (they're all white per IG html)
- in content_block twig, exposes `contentBodyClasses` and `contentHeaderClasses` blocks which the individual block templates can override
- cleans up presenter factory
- reverts constructor signature on FAQ and Prose presenters
- adds FixIsiteMarkupInterface and associated FixIsiteMarkupTrait (see changed files for usage)

after:
Screenshot 2019-08-08 at 14 05 35

before:
Screenshot 2019-08-08 at 14 16 21


src/Ds2013/Factory/PresenterFactory.php
```diff
@@ -268,55 +269,42 @@ public function broadcastEventPresenter(
...
-        if ($contentBlock instanceof Telescope) {
-            return new TelescopePresenter($contentBlock, $inPrimaryColumn, $isPrimaryColumnFullWith, $options);
-        }
-        if ($contentBlock instanceof ThirdParty) {
-            return new ThirdPartyPresenter($contentBlock, $inPrimaryColumn, $isPrimaryColumnFullWith, $options);
-        }
-        if ($contentBlock instanceof Touchcast) {
-            return new TouchcastPresenter($contentBlock, $inPrimaryColumn, $isPrimaryColumnFullWith, $options);
+        // iterate as to allow the content classes to be extended
+        foreach (self::PRESENTER_MAP as $contentClass => $presenterClass) {
```
BjornTwachtmann: I think it might be better just to do get_class and index into the mapper array you've set up, these content blocks aren't subclassed anywhere that I know.

BjornTwachtmann: Nice to get rid of all the verbose stuff above anyway

src/Ds2013/Presenters/Domain/ContentBlock/ThirdParty/ThirdPartyPresenter.php
```diff
@@ -7,8 +7,19 @@
 
 class ThirdPartyPresenter extends ContentBlockPresenter
 {
+    /** @var ThirdParty */
+    protected $block;
```
BjornTwachtmann: isn't this defined in the base class already?

gmarinov: it is, just annotating it here..

src/Ds2013/Presenters/Domain/ContentBlock/ThirdParty/third_party.html.twig
```diff
@@ -4,6 +4,11 @@
     {{ build_css_classes({'component--box-flushbody': not content_block.isInPrimaryColumn()}) }}
 {% endblock %}
 
+{% if (not content_block.isInPrimaryColumn()) and (content_block.getThirdParty() == 'instagram.com')  %}
```
BjornTwachtmann: the third-party.js is using the regex `"instagr\.?am(\.com)?/.+"` to identify an instagram URL. so it's possible that there are instagr.am links in the data set

src/Ds2013/Factory/PresenterFactory.php
gmarinov: took it out in favour of `get_class`, if it's ever needed, the subclasses can be added to the map explicitly

src/Ds2013/Presenters/Domain/ContentBlock/ThirdParty/third_party.html.twig
gmarinov: ah, of course. thanks. added

PR #1158, opened 2019-08-15
7137 - /recommendations route

gmarinov:
### /recommendations
Screenshot 2019-08-16 at 11 32 23

### footer on dsamen pages:

Screenshot 2019-08-16 at 15 06 51

### footer on ds2013 pages:

Screenshot 2019-08-16 at 15 06 42


src/Controller/RecommendationsController.php
```diff
@@ -15,6 +15,9 @@
 
 class RecommendationsController extends BaseController
 {
+    const ITEMS_PAGE = 10;  // items on main page
```
Tasiobg: since php7.1 visibility modifier are available for PHP and recommended to use
```suggestion
    private const ITEMS_PAGE = 10;  // items on main page
```
Same can be done for ITEMS_BOX

gmarinov:
test

src/Ds2013/Presenters/Section/Footer/FooterPresenter.php
```diff
@@ -99,4 +100,10 @@ public function isWorldNews(): bool
     {
         return ($this->getNid() === 'bbc_world_news');
     }
+
+    public function shouldShowRecommendations(Request $req): bool
```
Tasiobg: Would it be worth to add unitest for this? or maybe the code inside is too simple to be worth unitesting it?

templates/recommendations/list.html.twig
```diff
@@ -0,0 +1,30 @@
...
+{% block page_classes %}text-base programmes-page--smallpush prog-layout{% endblock %}
+{% block title %}{{ meta_context.titlePrefix() }} - {{ tr("You may also like") }}{% endblock %}
+
+{% block body %}
+    <div class="programmes-page prog-layout__primary br-box-page">
+        <h1>
+            <span class="context">
+                {{ ds_shared('entityContext', programme, {}) }}
+            </span>
+            {{ tr('You may also like') }}
```
Tasiobg: check it

Tasiobg:
```suggestion
            {{ tr('you_might_also_like') }}
```

src/Ds2013/Presenters/Section/Footer/FooterPresenter.php
gmarinov: yes → 27855b6

templates/recommendations/list.html.twig
```diff
@@ -0,0 +1,30 @@
+{% extends 'base_ds2013.html.twig' %}
+
+{% block page_classes %}text-base programmes-page--smallpush prog-layout{% endblock %}
+{% block title %}{{ meta_context.titlePrefix() }} - {{ tr("You may also like") }}{% endblock %}
```
Tasiobg:
```suggestion
{% block title %}{{ meta_context.titlePrefix() }} - {{ tr("you_might_also_like") }}{% endblock %}
```

gmarinov: sorry.. 86f4fbc

src/DsAmen/Presenters/Section/Footer/FooterPresenter.php
```diff
@@ -111,4 +113,9 @@ public function isWorldNews(): bool
     {
         return ($this->getNid() === 'bbc_world_news');
     }
+
+    public function shouldShowRecommendations(Request $req): bool
```
BjornTwachtmann: Is this needed as part of the Amen presenter? It seems like the recommendations full page view is only using ds2013

gmarinov: this is also used on the other pages, like the brand page.. @BjornTwachtmann

src/DsShared/Helpers/FooterLogicHelper.php
```diff
@@ -0,0 +1,14 @@
...
+
+namespace App\DsShared\Helpers;
+
+use Symfony\Component\HttpFoundation\Request;
+
+class FooterLogicHelper
+{
+    public static function shouldShowRecommendations(Request $req, $isWorldNews): bool
+    {
+        $isRecommendationsRoute = ($req->get('_route') === 'programme_recommendations');
```
BjornTwachtmann: I'm probably being overly fussy and this is purely stylistic preference, but it might be nicer to have parameter called "show_recommendations" or something passed to the presenter than to have it make decisions based on the route. That does involve overriding the call to the presenter with a block or something then, but I'm just not sure I like the idea of presenters having to make decisions based on the route.

gmarinov: I'll take a look following our chat.

gmarinov: @todo 
- remove the "isWorldNews" logic, 
- then the rest of footer "show recommendations" logic
- make a recommendations block and override it on the recommendations page

gmarinov: ↑ were done in 37a1cf8 and subsequent commits

PR #1162, opened 2019-08-27
ATI Publisher events groundwork

gmarinov:
- adds a twig helper to output an array to html attributes markup
- adds a helper to infer the current ATI Publisher container from the top-level controller class, and a `get_ati_attributes` twig helper to use it. This is called via Twig, because that's where the App environment is always available. It could sit in a base Presenter class if all templates had presenters, but the top ones don't.
- adds ~ATI Publisher Events attributes mapping in~ `ATIPublisherEventsHelper.php` and related methods to get the `data-bbc-container` and `data-bbc-title` attributes by providing a `$container` (inferred from top-level controller) and `$elementId` (arbitrary as per template). The returned `data-bbc-title` is now just the element Id.
- adds the "`ATI_prefix`" presenter option to various presenters and templates. The prefix is used to prefix all `$elementId`-s used on a same page, and by the same template building blocks but in various contexts (which can be other templates or different same-template sections)
- defines `title`, `cta`, `topic`, and other element IDs, used with or without prefix such as `related:title` or `related:topic` or just the element ID (`cta`), for programme templates and for other minor elements such as social bar buttons, rss links, etc; in the relevant templates that use them.
- **NEW**: if an `ATI_prefix` option is not supplied, then ProgrammePresenters and GroupPresenters in both design systems infer an ATI_prefix by looking at the entity type they are presenting - i.e. none for Programme, `clip` for Clip, `gallery` etc. These then usually trickle down to any subpresenters constructed. Likely there are spots missing in this PR. I've left this for revisiting in a subsequent PR.


src/Ds2013/Presenters/Domain/CoreEntity/Group/GroupPresenter.php
```diff
@@ -62,7 +66,7 @@ public function __construct(
         }
     }
 
-    public function getCoreEntityTitlePresenter(): CoreEntityTitlePresenter
+    public function getCoreEntityTitlePresenter($options = []): CoreEntityTitlePresenter
```
BjornTwachtmann: Why add this options parmeter here? Isn't the code below already handling passing the ATI prefix?

src/Ds2013/Presenters/Section/Episode/Map/Panels/Main/DetailsPresenter.php
```diff
@@ -31,6 +31,10 @@ class DetailsPresenter extends Presenter
     /** @var Version[] */
     private $alternateVersions;
 
+    public $options = [
```
BjornTwachtmann: Is this ever set? I can't see where if so.

src/Ds2013/Presenters/Section/Footer/footer.html.twig
```diff
@@ -6,7 +6,7 @@
                 <!-- (Recommendations) You may also like -->
                 {%- if not footer.isWorldNews() -%}
                     <div class="icon-box br-box-page lazy-module"
-                        data-lazyload-inc="{{ path('programme_recommendations', {pid: footer.getPid(), extension: '.2013inc'}) }}"
+                        data-lazyload-inc="{{ path('programme_recommendations', {pid: footer.getPid(), extension: '.2013inc'}) ~ '?' ~random() }}"
```
BjornTwachtmann: Cachebusting? Was this something you used for developing this, or why is it needed?

src/DsAmen/Presenters/Domain/CoreEntity/Base/SubPresenter/base_cta.html.twig
```diff
@@ -1,5 +1,9 @@
 <div class="icon-link {{ cta.getOption('cta_class') }} {% block extra_container_classes %}{% endblock %}">
-    <a class="link-complex" href="{{ cta.getUrl() }}" data-linktrack="{{ cta.getLinkLocation() }}"
+    <a
+        {% block ati_attributes %}
```
BjornTwachtmann: Maybe use the whitespace surpressing {%- -%} syntax here?

src/Ds2013/Presenters/Section/Footer/footer.html.twig
gmarinov: uuugh, yes. nice catch. thanks

src/DsAmen/Presenters/Domain/CoreEntity/Base/SubPresenter/base_cta.html.twig
gmarinov: aha, that did cause the tests to fail, I think because Twig gets a bit eager, cutting everything from "<a" to "class"

BjornTwachtmann: Ah, fair enough

PR #1163, opened 2019-08-30
tooling: Twig template labels

gmarinov:
adds `config/packages/dev/twig.yaml` in which a custom DevTemplate can be enabled. This is normally commented out as we'd like to use the `prod` i.e. stock template in development.

uncommenting the `base_template_class` line produces commented-out template markers in html  output like the below, based on @benolot's code.

Screenshot 2019-08-30 at 12 17 57

also uncommenting `debug_print_template_labels: true` adds behaviour as on the gif below:

debug-labels

This change does not affect production, is disabled by default, and must be explicitly enabled on dev.


.gitignore
```diff
@@ -18,3 +18,6 @@ config/parameters.yaml
 .phpunit
 /phpunit.xml
 ###< symfony/phpunit-bridge ###
+
+# prevent accidental changes to dev template config going out to github
+config/packages/dev/twig.yaml
```
Bobkidbob: I feel like this might come back and bite you on the arse.

gmarinov: ⚖️  😄

benolot: We usually catch things like this in PR's, we have the same issue technically with services.yaml to point at things like test iSite

gmarinov: aha, @benolot, that feels nice for weight

PR #1164, opened 2019-09-04
7115 remove iStats

gmarinov:
* WIP * WIP * WIP *


src/Controller/BaseController.php
```diff
@@ -243,7 +238,7 @@ protected function renderWithChrome(string $view, array $parameters = [])
         $this->preRender();
         $cosmosInfo = $this->container->get(CosmosInfo::class);
 
-        // Didn't iStats die? Why is this still here?
+        // @todo check this can be removed
```
BjornTwachtmann: It can't be removed entirely, unfortunately. Just had a chat with head of the SMP team and they still need that value for monitoring and analytics.

src/Ds2013/Factory/PresenterFactory.php
```diff
@@ -392,7 +389,7 @@ public function smpPresenter(
             $streamableVersion,
             $segmentEvents,
             $this->analyticsCounterName ? (string) $this->analyticsCounterName : null,
-            $this->istatsAnalyticsLabels ? $this->istatsAnalyticsLabels->getLabels() : null,
+            null,
```
BjornTwachtmann: It looks like SmpPresenter is still relying on istatsAnalyticsLabels['app_name'] being set. I would modify that class not to take this parameter and hardcode that app_name to 'programmes' in there, I think there might be some problems with playout stats otherwise

src/ValueObject/IstatsAnalyticsLabels.php
```diff
@@ -11,6 +11,9 @@
 use BBC\ProgrammesPagesService\Domain\Entity\ProgrammeItem;
 use BBC\ProgrammesPagesService\Domain\Entity\Service;
 
+/**
+ * @deprecated
+ */
 class IstatsAnalyticsLabels
```
BjornTwachtmann: Can this not be removed? Only thing that creates it seems to be the unit test for it.

templates/find_by_pid/programme_container.html.twig
```diff
@@ -251,7 +251,7 @@
 {% macro trailing_link(route, translation, programme, link_location_prefix) %}
```
BjornTwachtmann: Can probably remove the last parameter from this function, unless I'm reading it wrong

src/Ds2013/Presenters/Domain/CoreEntity/SharedSubPresenters/core_entity_title.html.twig
```diff
@@ -4,7 +4,7 @@
     <{{ title_tag }} class="programme__titles">
         <a href="{{ core_entity_title.getTitleLinkUrl() }}"
            class="br-blocklink__link block-link__target"
-           data-linktrack="{{ core_entity_title.getOption('link_location_track') }}" {% block aria_title %}{% endblock %}
+           {% block aria_title %}{% endblock %}
```
BjornTwachtmann: Is it worth going through the presenters and removing the link_location_track (and similar) options, I'm assuming they'll no longer be in use

src/ValueObject/ComscoreAnalyticsLabels.php
```diff
@@ -17,14 +17,14 @@ class ComscoreAnalyticsLabels
     /** @var Comscore|null */
     private $comscore;
 
-    public function __construct($context, CosmosInfo $cosmosInfo, IstatsAnalyticsLabels $istatsAnalyticsLabels, string $canonicalUrl)
+    public function __construct($context, CosmosInfo $cosmosInfo, string $canonicalUrl)
```
BjornTwachtmann: I think this class is unused and can be removed.

templates/base_ds_amen.html.twig
```diff
@@ -47,7 +47,7 @@
         </script>
 
         <!--[if (gt IE 9)|!(IE)]><!-->
-        <link rel="stylesheet" href="{{ preload(asset('css/dsAmen.css'), { as: 'style', nopush: true }) }}" />
+        <link rel="stylesheet" href="{{ preload(asset('css/dsAmen.css'), { as: 'style', nopush: true }) }}?2" />
```
BjornTwachtmann: typo?

src/ValueObject/IstatsAnalyticsLabels.php
gmarinov: should be able to. I marked it so that the IDE can cross it out if it's used, and then I was supposed to remove it. nice catch.

PR #1165, opened 2019-09-05
7137 - in recommendations take out view-all link

gmarinov:
FAO @benolot


benolot:
@gmarinov  - can you merge this into master instead of develop?

PR #1170, opened 2019-09-11
[PROGRAMMES-7133] Logos for R2 Beatles

benolot:
Ticket: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-7133

I _think_ this is all there is to it...


BjornTwachtmann:
Does R2 beatles need live playout? Does it need ondemand playout? If so is that going to sounds.

benolot:
No R2 beatles doesn't need live playback, but we will need to update programmes-faucet as a second part of this ticket to point it at sounds for on demands.

PR #1184, opened 2019-09-27
contact: tests and refinements to contact markup and controller

gmarinov:
Screenshot 2019-09-27 at 14 28 56

 - extracts contact markup to its own tempalte (also so it can be tested)
 - adds test for contact html rendering and UGC contact option priority
 - configures PHPStan to ignore the controller mock (!) [see here](https://github.com/bbc/programmes-frontend/compare/6886-contact...6886-contact-test?expand=1#diff-e094b47b5c82f5ed26f4e4013f602ab1R21)


templates/contact/contact_markup.html.twig
```diff
@@ -0,0 +1,63 @@
...
+{% endmacro %}
+
+{% if contactDetails %}
+    <ul>
+        {% for contactDetail in contactDetails %}
+            {{ _self.renderContact(contactDetail, currentUrl) }}
+        {% endfor %}
+    </ul>
+{% else %}
+    <p>{{ tr('coming_soon') }}</p>
```
BjornTwachtmann: doesn't the page 404 without contact details anyway?

gmarinov: it does! it didn't use to.. I'll take it out.

PR #1186, opened 2019-10-01
7056 - throw 500 instead of 404 on disabled recipe dial

gmarinov:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-7056

- in AbstractRecipesController, adds a guard to check if the recipe dial is disabled before calling the API. if disabled, throws 500, instead of getting 0 results back from API and saying "found no recipes".

NB to test code in sandbox, in _dials.json_ set `"api-recipe": "disabled"

---

#### before (test site):
Screenshot 2019-10-01 at 15 03 53

#### after (sandbox, this won't be visible on test site):
Screenshot 2019-10-01 at 15 03 44


src/Controller/Recipes/AbstractRecipesController.php
```diff
@@ -5,7 +5,7 @@
 
 use App\Controller\BaseController;
 use App\Controller\Helpers\Breadcrumbs;
-use App\ExternalApi\Recipes\Domain\RecipesApiResult;
```
benolot: Don't need this anymore?

benolot: Doesn't look used but just checking

gmarinov: yea, looked like a leftover

PR #1190, opened 2019-10-08
7181 contact page translations

gmarinov:
WIP


BjornTwachtmann:
I think this is still missing the Irish translation for contact us - "Déan teagmháil linn"

BartekKloza:
> I think this is still missing the Irish translation for contact us - "Déan teagmháil linn"

I double checked and it's there - messages.ga.po file, line 522. The reason it's not showing up in the diff is because it's already in the master branch.

PR #1192, opened 2019-10-09
6903 AV tracking for segments

gmarinov:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6903

the gist of this PR is that it rewires the Snippets code to take a `statsObject` from outside (in the segments_list view), and it makes the Snippets code retain and pass that statsObject to SMP every time a new playlist is loaded (= when a snippet starts playing), slightly modifying the playlist to include the title of the asset playing (for stats purposes). Outside, the statsObject is "helped" by a pair of Twig helpers in the (renamed) ATIAnalyticsExtension. For that to work, the SegmentListPresenter allows access to its `.context` (typically a CoreEntity). Other changes are mostly cosmetic.

The variables sent are as follows:

- destination (site)
- producer (L2) - this might not show in AV events (see screenshot) but _is_ set and resolves to "Media Level 2"
- episodePID - contains the Tupac Clip ID, this is supposed to not work but results in a Media Name at the top, and shows in the ATI Analytics UI. Tupac IDs can be searched for in PIPS inspector.
- media theme 1 title - this is restricted to 60 characters, and any special characters and space are replaced with '_'
- media theme 2 - app name is set to "programmes-snippets"
- counter name - set to clip/episode "%pid%-segments"
- playlist name - set to clip/episode "%pid%-segments" (as counter name)
- app version (apvr) - set to 1
- action - this is set to play or pause automatically by SMP
- type - this is fixed to "audio" or "video" only, sent automatically by SMP.

Screenshot 2019-10-09 at 15 06 56


src/Twig/ATIAnalyticsExtension.php
```diff
@@ -67,10 +76,41 @@ public function getATIPublisherAttributes(?string $elementId, ?string $prefix =
         ];
     }
 
+    /**
+     * get the Site parameter for ATI event stats
+     * @see https://confluence.dev.bbc.co.uk/pages/viewpage.action?pageId=175885338#tab--2022260337
+     * @param CoreEntity $coreEntity
+     * @return string
+     */
+    public function getATIDestination($coreEntity)
```
Tasiobg:
```suggestion
    public function getATIDestination(ProgrammeItem $coreEntity): string
```

segments_list.getContext() returns instance of ProgrammeItem, maybe worth adding this

src/Twig/ATIAnalyticsExtension.php
```diff
@@ -67,10 +76,41 @@ public function getATIPublisherAttributes(?string $elementId, ?string $prefix =
...
+            ->getDestinationFromContext($coreEntity, $this->cosmosInfo->getAppEnvironment());
+    }
+
+    /**
+     * get the L2 parameter for ATI event stats
+     * @see https://confluence.dev.bbc.co.uk/pages/viewpage.action?pageId=175885338#tab-79895020
+     * @param CoreEntity $coreEntity
+     * @return string
+     */
+    public function getATIProducer($coreEntity)
```
Tasiobg:
```suggestion
    public function getATIProducer(ProgrammeItem $coreEntity): string
```
segments_list.getContext() returns instance of ProgrammeItem, maybe worth adding this

assets/js/snippet-player/snippets/driver/smp.js
```diff
@@ -30,125 +54,98 @@ define(['../playables'], function(playables) {
...
+                        href: 'http://emp.bbci.co.uk/emp/media/blank.mp3'
+                    }]
+                },
+
+                autoplay: false,
+
+                // Additional data for DAx reporting
+                appName: 'snippets-smp-driver',
+                appType: 'responsive',
+                counterName: '?',
```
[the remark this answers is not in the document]
gmarinov: it should. this is the default value for the SMP driver of Snippets. since we'd like it always set, the default looks stupid, so as to force us to set it.

assets/js/snippet-player/snippets/component/transform.js
```diff
@@ -4,64 +4,47 @@ define([
...
 
-    var getRecordIdFromSnippetWebComponent = function(elem) {
-        var recordId = dom.getData(elem, 'record-id');
-        if (!recordId) {
-            var recordIdElement = elem.querySelector('record-id');
-            recordId = recordIdElement && recordIdElement.innerHTML;
-        }
-        return recordId;
+    var getRecordIdFromSnippet = function(el) {
+        if ('recordId' in el.dataset) return el.dataset['recordId'];
```
Tasiobg: I think you mention you would like to do some changes to the selectors.
On the other hand I haven't done JS for a while but changes makes sense to me.

src/Twig/ATIAnalyticsExtension.php
gmarinov: → 36598b8c

src/Twig/ATIAnalyticsExtension.php
gmarinov: → 36598b8c

assets/js/snippet-player/snippets/component/transform.js
gmarinov: → 84bd7bfe

src/Ds2013/Presenters/Section/Segments/segments_list.html.twig
```diff
@@ -2,12 +2,29 @@
...
+    {% set ctx = segments_list.getContext() %}
+    {% set destination = get_ati_destination(ctx) %}
+    {% set producer = get_ati_producer(ctx) %}
+    {% set pid = ctx.getPid() %}
     <script>
         require(["snippets", "bump-3"], function(snippets, bump) {
             snippets.init({
+                theme: "music-track",
                 base_url: '/programmes/snippet',
-                context: "programmes",
```
BjornTwachtmann: Is this parameter not needed, the default in the JS seems to be empty, but no idea what it does

PR #1195, opened 2019-10-14
[PROGRAMMES-7164] Update PHP to 7.3

benolot:
Ticket: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-7164


script/codebuild/build
```diff
@@ -10,6 +10,8 @@ set -e
 # INSTALL #
 ###########
 
+php --version
```
BjornTwachtmann: did you mean to leave this in?

benolot: Don't see much harm in leaving it in for future debugging

script/codebuild/build
```diff
@@ -10,6 +10,8 @@ set -e
...
+
 # Fail if we're not using php71+
 if ! (php --version | grep -q "PHP 7.") ; then
     echo "You're not using PHP 7.1+, please install it."
```
Tasiobg:
```suggestion
    echo "You're not using PHP 7.3+, please install it."
```

script/codebuild/build
```diff
@@ -10,6 +10,8 @@ set -e
 # INSTALL #
 ###########
 
+php --version
+
 # Fail if we're not using php71+
```
Tasiobg:
```suggestion
# Fail if we're not using php73+
```

PR #1197, opened 2019-10-15
Fix rendering of social bar icons. 

BjornTwachtmann:
* Fix rendering of UGC in social bar
* Make contact form URLs come from router
* Remove <title> from SVG icons messing up tooltips


script/update-gelicons.php
```diff
@@ -67,6 +67,8 @@ private function fixSvg(string $inputPath): void
         $contents = preg_replace('#^\s*(<svg[^>]*?) width="[0-9\.]+"([^>]*>)#i', '${1}${2}', $contents);
         // get rid of height
         $contents = preg_replace('#^\s*(<svg[^>]*?) height="[0-9\.]+"([^>]*>)#i', '${1}${2}', $contents);
+        // get rid of <title>
+        $contents = preg_replace('#<title>[^<]+</title>#i', '', $contents);
```
Tasiobg: Not familiar with this script but, could this now or at any point in the future remove the title for elemens with hidden tab `tabindex="-1"` ? if so it could affect accesibility.

BjornTwachtmann: Currently all SVGs have tabindex=-1 . But yeah, it could have that impact. I tried using CSS to remove the tooltip, but it also disabled tooltips on the parent elements, so not sure how else to do it.

PR #1200, opened 2019-10-17
[PROGRAMMES-6149] Add logging for all failed HTTP calls

benolot:
Ticket: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6149


src/Metrics/MetricsMiddleware.php
```diff
@@ -29,20 +28,27 @@ public function __construct(MetricsManager $metricsManager, LoggerInterface $log
...
-                    return;
-                }
                 $responseTime = (int) round($stats->getTransferTime() * 1000);
                 // No response/timeout is logged as a 504 (gateway timeout). Which isn't correct. But whatever.
                 $responseCode = 504;
                 if ($stats->hasResponse()) {
                     $responseCode = $stats->getResponse()->getStatusCode();
                 }
+
+                // Log all HTTP requests that fail - except 404's because sometimes crawlers make mistakes
```
Tasiobg: What do you mean by crawlers make mistakes? Isn't this code only loging curl request from frontend to APIs?

benolot: Yes but things like iSite or topic pages make API calls based on the route hit, often a 404 by a crawler. It was just added as a joke, so I'll remove it if it's causing confusion.

BjornTwachtmann:
Thinking about this, if we log an error, is that going to dump out the debug logging for the whole request? Do we need to try and prevent that if so?

benolot:
> Thinking about this, if we log an error, is that going to dump out the debug logging for the whole request? Do we need to try and prevent that if so?

What do you mean?

BjornTwachtmann:
> 
> What do you mean?

The app uses the "fingers crossed" error handler, which dumps out all logging for a request once an error is encountered. See monolog.yml. If we log an error when an HTTP request fails, does that result in a full output of all logs for that request? 

Make any more sense?

benolot:
Yeah, thought thats what you meant but wanted to be sure 😄  - I think in this case that will be useful, might help us track down why certain queries are failing, what do you think?

BjornTwachtmann:
{

> Yeah, thought thats what you meant but wanted to be sure 😄 - I think in this case that will be useful, might help us track down why certain queries are failing, what do you think?

I don't think the extra lines would be particularly useful in debugging the cause of a request failing, probably more likely to just clutter the logs, but we can merge this and fix that issue if it becomes problematic I suppose

PR #1203, opened 2019-10-24
Clifton: {pid}/episodes.json

gmarinov:
base: `clifton`
WIP
merely cleans up the episodes.json route controller


src/CliftonBundle/Controller/BroadcastYearsAndMonthsController.php
```diff
@@ -2,32 +2,27 @@
...
 
 class BroadcastYearsAndMonthsController extends BaseApsController
 {
-    public function yearsAndMonthsAction(
+    public function __invoke(
         ProgrammesService $ps,
-        CollapsedBroadcastsService $bs,
-        Request $request,
+        CollapsedBroadcastsService $cbs,
         string $pid
```
BjornTwachtmann: You can actually simplify this to a typehint of
`Pid $pid` or even omit programmeservice and $pid althogether and just `ProgrammeContainer $programme` and it will still work. Not sure how nicely that will play with the tests, but just to flag up that various forms of Symfony magic exist to deal with that stuff already.

gmarinov: thanks. used `Pid`

PR #1205, opened 2019-10-25
Clifton: A to Z json routes

gmarinov:
base: `clifton`

enables:

- `/programmes/a-z.json`
- `/programmes/a-z/{slice}.json`
- `/programmes/a-z/by/{search}.json`
- `/programmes/a-z/by/{search}/{slice}.json`

Also refactors the A-Z controller:
- no longer allowed unlimited query
- default limit of 100, default page of 1
- take limit between 10 and 100 incrementing in steps of 10
- include limit value in the returned results
- zero array of letters in json response when searching by keyword

Screenshot 2019-10-25 at 13 24 10

Screenshot 2019-10-25 at 13 28 32

Screenshot 2019-10-25 at 13 27 49

Screenshot 2019-10-25 at 12 18 19


src/CliftonBundle/Controller/AtozController.php
```diff
@@ -4,98 +4,106 @@
...
-        }
+        $isByKeyword = strlen($search) >= 2;
 
-        if (strlen($search) < 2) {
-            list ($items, $itemCount) = $this->letterSearch($search, $slice, $page, $limit);
-        } else {
-            list ($items, $itemCount) = $this->keywordSearch($search, $slice, $page, $limit);
-        }
+        list ($items, $itemCount) = ($isByKeyword)
+            ? $this->keywordSearch($progService, $search, $slice, $page, $limit)
```
BjornTwachtmann: We actually disabled keyword search on the main site due to performance issues with crawlers ( basically searching for certain things that should probably be stopwords caused a lot of load and we couldn't really be bothered to fix it for the traffic it got). Might as well remove it here too.

src/CliftonBundle/Controller/AtozController.php
```diff
@@ -151,18 +160,14 @@ private function filterParams(Request $request)
...
-        if ($inputPage && $inputLimit) {
-            // Limit is only valid with a page (APS behaviour)
-            $page = (int) $inputPage;
-            $limit = (int) $inputLimit;
-        }
+
+        $page = $inputPage ?: 1;
+        $limit = $inputLimit ?: self::MAX_LIMIT;
+        // e.g. 10, 20, 30, etc
+        $limit = round($limit / 10) * 10;
```
BjornTwachtmann: I can see the intention here, if we're doing this should we generalise it though, most of these accept a limit and a page.

src/CliftonBundle/Controller/AtozController.php
gmarinov: have created a separate ticket for this - 7211

src/CliftonBundle/Controller/AtozController.php
gmarinov: have created a separate ticket for this - 7212

gmarinov: #1221

PR #1213, opened 2019-10-28
Clifton: {pid}.json

gmarinov:
WIP!
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-7208
base `clifton`
(fails Codebuild)

NB this fails tests for a reason I can't work out - the same code works fine on `programmes-clifton` but not here. 

`./bin/phpunit --no-coverage --stop-on-failure tests/CliftonBundle/Controller/FindByPidControllerTest.php `

```
1) Tests\App\CliftonBundle\Controller\FindByPidControllerTest::testFindByPidActionWithSeries
Failed asserting that an array does not have the key 'ownership'.

/mnt/hgfs/workspace/programmes-frontend/tests/CliftonBundle/Controller/FindByPidControllerTest.php:62
```

apparently ownership is attached automatically here in `programmes-frontend`

Screenshot 2019-10-28 at 17 35 19


src/CliftonBundle/Controller/FindByPidController.php
```diff
@@ -12,66 +12,82 @@
...
         if ($programme->getRelatedLinksCount()) {
-            $rls = $this->get('pps.related_links_service');
             $relatedLinks = $rls->findByRelatedToProgramme($programme);
         }
 
         // Peers
         $nextSibling = null;
         $previousSibling = null;
+
+        /** @noinspection PhpUnhandledExceptionInspection */
```
BjornTwachtmann: Is this a PHPStan thing? Never seen that be necessary before.

gmarinov: phpstorm/webstorm

BjornTwachtmann: fair enough

PR #1220, opened 2019-11-01
7211 remove by-keyword a-z search

gmarinov:
- removes byKeyword functionality from A-Z JSON
- silently ignores more than 1 letter

not sure if `substr()` is the right way to go here. open to `routes.yaml` or other suggestions.


src/Controller/JSON/AtozController.php
```diff
@@ -33,25 +33,15 @@ public function byAction(
         string $slice = 'player'
     ): JsonResponse {
         list ($page, $limit) = $this->filterParams($request);
-        $isByKeyword = strlen($search) >= 2;
-
-        list ($items, $itemCount) = ($isByKeyword)
-            ? $this->keywordSearch($progService, $search, $slice, $page, $limit)
-            : $this->letterSearch($azService, $search, $slice, $page, $limit);
+        $letter = substr($search, 0, 1);
```
BjornTwachtmann: Would it be better here to throw a 404 if $search is anything other than `^[a-z\@]$`? Otherwise you could get something like `/programmes/a-z/by/myOldSearchTerm.json returning results for the letter M, which seems weird, and also splits the FE cache

gmarinov: how about `4018c2b`?

PR #1233, opened 2019-11-20
redirect v2 route to episodes list

Tasiobg:


gmarinov:
related https://github.com/bbc/programmes-findbypid-proxy/pull/84

PR #1237, opened 2019-12-16
[No ticket] Update /schedules footer links

benolot:


BjornTwachtmann:
Not sure that destination is massively useful. The old link used to redirect to /faqs, which now redirects to https://www.bbc.co.uk/contact/questions. Which I'm not convinced is super useful either, but is at least more general.

Either go with the /contact/questions link or delete it if you also think that's useless would be my steer.

src/Ds2013/Presenters/Pages/Schedules/ByDayPage/schedules_by_day_page.html.twig
```diff
@@ -129,7 +129,7 @@
                 <a class="deemphasize" href="/showsandtours/">Shows &amp; Tours</a>
             </li>
             <li>
-                <a class="deemphasize" href="http://faq.external.bbc.co.uk/questions/television/">Help &amp; FAQ</a>
+                <a class="deemphasize" href="contact/questions">Help &amp; FAQ</a>
```
Tasiobg:
```suggestion
                <a class="deemphasize" href="/contact/questions">Help &amp; FAQ</a>
```
does yours works fine?

benolot: Yeah copy paste fail. 😄 My b

PR #1251, opened 2020-04-02
Hacky canonical fix until we can find the real problem allowing pages…

BjornTwachtmann:
… to be served on bbc.com


gmarinov:
jira: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-7258

PR #1252, opened 2020-04-15
PROGRAMMES-7287 fix mapping of ada responses

Tasiobg:
Similar to a fix I did before https://github.com/bbc/programmes-frontend/pull/953
There is no guaranty that an item comming from ada would map to a progamme in our side, a common example is a programme that flag as embargoed in our DB.


src/ExternalApi/Ada/Service/AdaProgrammeService.php
```diff
@@ -32,16 +33,21 @@ class AdaProgrammeService
...
 
+    /** @var CoreEntitiesService */
+    private $coreEntitiesService;
+
     public function __construct(
         HttpApiClientFactory $clientFactory,
         string $baseUrl,
         AdaProgrammeMapper $mapper,
-        ProgrammesService $programmesService
+        ProgrammesService $programmesService,
```
BjornTwachtmann, line 43: Do you still need this? I think the tests are blowing up due to the constructor signature change.

Tasiobg: PR is not finish, I want to know if this looks good to you and then I can also change the other place where we are using ProgrammesService to use CoreEntitiesService and get ride of the nested loops

BjornTwachtmann: Ah, fair enough, yes this looks good, I'd switch them both over to this method myself

tests/ExternalApi/Ada/Service/AdaProgrammeServiceTest.php
```diff
@@ -44,63 +44,6 @@ public function testNoRelatedItemsAreReturnedWhenNoRelatedItems()
...
-            ->expects($this->exactly(2))
-            ->method('mapItem')
-            ->willReturn(
-                $this->createMock(AdaProgrammeItem::class)
-            );
-        $programmesService = $this->createMock(ProgrammesService::class);
-        $programmesService->expects($this->once())
-            ->method('findByPids')
-            ->with([new Pid('b007rlb6'), new Pid('b007rlb4')])
-            ->willReturn([$this->mockProgramme('b007rlb6'), $this->mockProgramme('b007rlb6')]);
```
Tasiobg, line 91: coreEntitiesService->findByPids now returns and array of pid indexed by pid so there is no way the same pid index can exist twice, I don't think this test is need anymore.

src/ExternalApi/Ada/Service/AdaProgrammeService.php
```diff
@@ -166,17 +166,15 @@ function ($carry, $item) {
...
+
+        $programmeItems = $this->coreEntitiesService->findByPids($pids);
         $relatedProgrammes = [];
-        // in some cases $uniqueProgrammes may contain programmes non existing in $programmeItems so we have to filter them
         foreach ($uniqueProgrammes as $item) {
-            foreach ($programmeItems as $programmeItem) {
-                if ((string) $programmeItem->getPid() === $item['pid']) {
-                    $relatedProgrammes[] = $this->mapper->mapItem($programmeItem, $item);
-                }
+            if (isset($programmeItems[$item['pid']])) {
```
gmarinov, line 173: I still think this is unnecessary

src/ExternalApi/Ada/Service/AdaProgrammeService.php
```diff
@@ -166,17 +166,15 @@ function ($carry, $item) {
...
+                $relatedProgrammes[] = $this->mapper->mapItem($programmeItems[$item['pid']], $item);
```
gmarinov, line 174: ... and it can simply map the whole of `$programmeItems` as it found them as the pids it searched by came from `$uniqueProgrammes` anyway

src/ExternalApi/Ada/Service/AdaProgrammeService.php
```diff
@@ -189,15 +187,17 @@ private function parseResponse(array $responses): array
...
+        $programmes = $this->coreEntitiesService->findByPids(array_map(static function ($item) {
             return new Pid($item['pid']);
-        }, $items));
+        }, $items), 'Programme');
 
         $adaProgrammeItems = [];
-        for ($i = 0, $l = count($items); $i < $l; $i++) {
-            $adaProgrammeItems[] = $this->mapper->mapItem($programmes[$i], $items[$i]);
+        foreach ($items as $item) {
+            if (isset($programmes[$item['pid']])) {
```
gmarinov, line 196: same here as previously. 
in what scenario does this need an extra check?

src/ExternalApi/Ada/Service/AdaProgrammeService.php
BjornTwachtmann: Just because ADA returns a PID doesn't mean it exists in our DB. That condition is the whole reason for this PR isn't it?

Tasiobg: that's right, some ada pids are not wont be returned because reasons like embargoed

tests/ExternalApi/Ada/Service/AdaProgrammeServiceTest.php
BjornTwachtmann: I'm not convinced this isn't still a potential problem. Reading https://github.com/bbc/programmes-frontend/blob/c90d86b6b9dace9715bd0b627982876d46bb1391/src/ExternalApi/Ada/Service/AdaProgrammeService.php#L146-L176 , which is basically what this is testing, I think it is quite possible for ADA duplicates to be reflected as duplicates in the return value despite the PID indexed array. Do you disagree?

Tasiobg: $this->coreEntitiesService->findByPids($pids);  returns an array of programmes indexed by pid. Array indexes are unique so the last ID would probably override the previous one which mean we can't never return the same programme twice in a recomendations.
Does it make sense?

BjornTwachtmann: > Does it make sense?

So, on line 172 it's looping through $uniqueProgrammes, which is a numerically indexed array made by combining various ADA queries in code. Each item in $uniqueProgrammes is associated to a PIPs programme and then mapped and pushed onto another array. Therefore if $uniqueProgrammes (which doesn't come from pages-service), has the same item in it more than once, it will be returned more than once. 

I think this test makes sure that $uniqueProgrammes actually contains only unique programmes and is still necessary. I could be wrong though.

src/ExternalApi/Ada/Service/AdaProgrammeService.php
gmarinov: yes, I mean mapping from the `$programmeItems` it found, not from `$uniqueProgrammes` would make the extra check unnecessary? the `$pids` to look for already came from `$uniqueProgrammes` above.

BjornTwachtmann: Oh, I see what you mean. You could flip it I suppose yeah, it seems easier to me to use the existing code for creating a PID indexed array in pages-service than to mess around with the slightly weird code creating $uniqueProgrammes to make that an associative array.

tests/ExternalApi/Ada/Service/AdaProgrammeServiceTest.php
Tasiobg: I see what you mean, I have fixed it now, I got confused because in this line https://github.com/bbc/programmes-frontend/blob/develop/tests/ExternalApi/Ada/Service/AdaProgrammeServiceTest.php#L91 is moking the API response with the same pid twice which makes no sense to me, I think the mooking was wrong, instead it should have mook 2 different pids.
What do you think?

BjornTwachtmann: Yeah, you're right, that old mock looks a bit screwey. These tests could probably be simplified a fair bit, but let's not go there...

PR #1273, opened 2020-09-18
Update caching headers

Tasiobg:
Move "stale-while-revalidate=30" from "X-Cache-Control" to "Cache-Control"
Remove X-Webapp which is no longer in use


gmarinov (approved):
lgtm

PR #1283, opened 2020-11-10
DATCAP-88: removing /music/images src from contributor images

ajroberts10:
* Removing musicBrainz image urls for contributor images
* image `src` will now be the direct response from TUPAC
* image `data-src` will now be a placeholder image

Before:
Screenshot 2020-11-10 at 14 59 31

After:
Screenshot 2020-11-10 at 14 59 58

https://jira.dev.bbc.co.uk/browse/DATCAP-87


src/Ds2013/Presenters/Section/Segments/SegmentItem/AbstractMusicSegmentItemPresenter.php
```diff
@@ -74,10 +74,6 @@ public function getPageType(): string
 
     public function getImageUrl(): string
```
llh1, line 75: How about renaming this function to something like `getImagePlaceholderUrl` as this will only return a placeholder image from now on.

llh1 (approved):
Looks good to me 👍

ColinBarnwell (approved):
🧙

PR #1284, opened 2020-11-13
DATCAP-117: updating UrlKeyExtenstion with new masterbrands

ajroberts10:
Adding the following masterbrands to UrlKeyExtension file:

- bbc_voice
- bbc_radio_mixes
- bbc_world_service_africa
- bbc_world_service_asia
- bbc_world_service_north_america
- bbc_world_service_south_america

https://jira.dev.bbc.co.uk/browse/DATCAP-117


gmarinov (commented):
is it `true` @BjornTwachtmann ?

BjornTwachtmann:
> is it `true` @BjornTwachtmann ?

I think it will sort of depend if these are being read as radio networks or not. Which some of them might not be as they don't have services. There's certainly no page at /urlkey for any of them.

PR #1297, opened 2021-02-01
De-index episode pages for specified brands

ColinBarnwell:
Uses existing controller hook to set a no-indexing flag when an episode belonging to certain brands is rendered via `/programmes/{pid}` route.

Screenshot 2021-02-01 at 15 38 45

## Notes

I have added two unit tests:

- Check that episode pages for non-indexed brands return 200 and have meta noindex element.
- Check that episode pages for other brands return 200 and do not have meta noindex element.

These tests are in a new file which is isolated, however, to support them I've had to add to the existing episode and brand fixtures, which increases the footprint of this experiment slightly.

## Jira

https://jira.dev.bbc.co.uk/browse/DATCAP-181


murkas (approved):

DouglasJE-bbc (approved):

ajroberts10 (approved):
nice 👍🏻

PR #1301, opened 2021-03-03
DATCAP-223: swapping out podcast urls to fix broken links

ajroberts10:
Updating podcast links based on the following logic:

- if genre key is childrens => https://www.bbc.co.uk/sounds/category/childrens
- if genre key is anything other than childrens => https://www.bbc.co.uk/sounds/category/podcasts

Tests to check the various urls will be added as part of the `programmes-aws-live` automation testing

https://jira.dev.bbc.co.uk/browse/DATCAP-223

### DIFF
--- a/config/routes/3rd_party.yaml
+++ b/config/routes/3rd_party.yaml
@@ -79,11 +79,6 @@ podcast_download_low:
     defaults: { _controller: Symfony\Bundle\FrameworkBundle\Controller\RedirectController::redirectAction, route: '' }
     schemes: [https]
 
-podcast_other_podcasts:
-    path: /radio/categories/redirect
-    defaults: { _controller: Symfony\Bundle\FrameworkBundle\Controller\RedirectController::redirectAction, route: '' }
-    schemes: [https]
-
 podcast_childrens_podcasts:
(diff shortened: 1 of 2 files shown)


murkas (approved):

DouglasJE-bbc (approved):

config/routes/3rd_party.yaml
```diff
@@ -79,12 +79,21 @@ podcast_download_low:
     defaults: { _controller: Symfony\Bundle\FrameworkBundle\Controller\RedirectController::redirectAction, route: '' }
     schemes: [https]
 
-
```
ColinBarnwell, line 82: `podcast_other_podcasts` seems no longer to be in use?

ColinBarnwell (approved):

llh1 (approved):

config/routes/3rd_party.yaml
ajroberts10: Yeah, I was a bit nervous about removing it in case it was re-used somewhere. Searching in the code doesn't bring back any results so I guess it's safe to get rid

PR #1302, opened 2021-03-04
DATCAP-225: Fix brand and collection broken links

murkas:


spacekitcat (approved):

ajroberts10 (approved):
Looks good, nice fix. Has this been broken for a while? Looks like it would never have worked unless something has changed behind the scenes?

src/Controller/Podcast/PodcastController.php
```diff
@@ -88,7 +88,7 @@ public function __invoke(
         switch ($coreEntity->getType()) {
             case 'collection':
                 $soundsSubscribeUrl = $router->generate('sounds_collection', [
-                    'collectionPid' => $coreEntity->getPid(),
+                    'collectionBrandPid' => $coreEntity->getParent()->getPid(),
```
ColinBarnwell, line 91: Should this be `getTleo`?

murkas: Now modified as discussed! Thanks! 😄

murkas:
> Looks good, nice fix. Has this been broken for a while? Looks like it would never have worked unless something has changed behind the scenes?

It looks to me as though this may be due to a change to Sounds' routing/redirects that will have been in place previously but have since been removed. Albeit, I wouldn't be able to guess at how long this may have been the case for.

ColinBarnwell (approved):
👍

PR #1308, opened 2021-03-10
Spike: Noindexing for series pages

ColinBarnwell:
We have previously implemented no-indexing for targeted episode pages (#1297). This spike looks at the same thing for series pages.

## Findings

- We can follow roughly the same approach, however the base template used by Series pages didn't have the noindex meta tag option, so I have added that.
- There is more than one type of series page. This PR implements the change for both 'series' and 'series episodes'
  - https://www.bbc.co.uk/programmes/p040tlqx
  - https://www.bbc.co.uk/programmes/p040tlqx/episodes/player
- Potentially series could be routed to brand pages (if the series is also the TLEO). I'm assuming such cases would not be part of the experiment.
- Although the logic to add the noindexing is trivial, we would probably want to share the noindex list among controllers to avoid duplication of data.

## Images

Screenshot 2021-03-10 at 16 05 31
Screenshot 2021-03-10 at 15 34 10

## Jira

https://jira.dev.bbc.co.uk/browse/DATCAP-219

### DIFF
--- a/src/Controller/FindByPid/SeriesController.php
+++ b/src/Controller/FindByPid/SeriesController.php
@@ -38,6 +38,19 @@ class SeriesController extends BaseProgrammeContainerController
         Breadcrumbs $breadcrumbs
     ) {
         $this->setAtiContentLabels('series', 'series');
+        $noIndexBrands = [
+            'b006pfjx', // North West Tonight
+            'b007t9y1', // Match of the Day
+            'p00yzlr0', // Line of Duty
+            'p070npjv', // Fleabag
+            'b006v5y2', // Saturday Kitchen
+            'b006mgyl', // BBC News
+            'm000lxp1', // Powering Britain
+            'b08s3bgz',  // Impossible
+        ];
+        if (in_array($programme->getTleo()->getPid(), $noIndexBrands)) {
+            $this->metaNoIndex = true;
+        }
         return parent::__invoke(
             $presenterFactory,
             $request,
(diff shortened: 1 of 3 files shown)


nicgarner:
- There is more than one type of series page. This PR implements the change for both 'series' and 'series episodes'

Dan M has another experiment in mind for 

- Potentially series could be routed to brand pages (if the series is also the TLEO). I'm assuming such cases would not be part of the experiment.

This is a point of confusion for me at the moment too and I'm not sure what we have in mind - there's a meeting with Dan M in the diary on Monday to discuss and hopefully get more clarity. But yes, we'll be able to achieve what we want by being selective about the PIDs we use.

- Although the logic to add the noindexing is trivial, we would probably want to share the noindex list among controllers to avoid duplication of data.

Would sharing the noindex list mean the same list of PIDs would be entered into each experiment? If so, I think that's undesirable.

ColinBarnwell:
> Would sharing the noindex list mean the same list of PIDs would be entered into each experiment? If so, I think that's undesirable.

That's fine then. We can keep them self contained. This will mean the experiments are fragmented throughout the codebase, so before we proceed, we should probably document these changes somewhere, even just collecting the Jira tickets and PRs should we decide to change or revert anything.

nicgarner:
Yes, good idea. I'll see whether Dan M already has a document or anything started to document the results of these experiments - if not I'll create one and add the relevant tickets.

PR #1329, opened 2022-01-20
fix: correcting brand manifest cache behaviour for failed network requests

spacekitcat:
### Context

The caching code in `LogoVersionStrategy` works in such a way that it calls the specified callback function if the specified key cannot be found. I think the theory is that the cache will always return a value because it will use the callback to resolve values that either aren't in the cache or that have exceeded the eviction TTL.

It's a nice theory, but what happens if the callback returns null? What if it doesn't return anything? Well, null gets cached for the specified key in both scenarios. The next time you try to get the same key out of the cache, it will treat it as a cache hit and return null. `LogoVersionStrategy` operates on the assumption that the cache will always return a valid JSON payload for the manifest and will attempt to parse whatever it gets from the cache like a valid JSON string. Since the cache TTL is 10 minutes in this context, one failed network request for the brand manifest can cascade into thousands of 500 errors. Most of those errors will be hidden from the end user by Belfrage, fortunately. 

The fix is to ensure that only valid JSON payloads get cached. If we throw an exception from the callback, the cache won't store anything for the specified key. Additionally, if we can't parse the JSON payload as JSON for some reason, we explicitly remove it from the cache and re-throw the exception. 

### Testing

Please see the investigation paper from the ticket (DATCAP-456) or Slack. The reproduction steps were used to verify that this fixes the behaviour. In addition to those manual test steps, the unit test verifies the logic changes.


gmarinov:
I wrote this. The intent for caching a failed response is two-fold:

1/ you probably don't want to repeatedly query the network if the network is down.
2/ I set the cache time at 600 seconds at the time. It seemed a reasonable choice, should something fail, for the issue to go away by itself after a while

So, a choice between a very loud complaint – burst of 500 errors to leave to the (caching) pipes – or potentially a lot of timeouts, some/many of which may simply get swallowed, and the site performance is degraded but not to the extent that it triggers an alarm.

spacekitcat:
I appreciate you providing some context.

This fix is the outcome of an investigation we carried out after two different OPS tickets about a burst of 500 errors (around 5000 I believe). I think at this point in time, my fix the correct approach going forward.

gmarinov:
Since you own this code now, I have no opinion about what goes forward.
I am aware of the tickets. Catching this in one of the traffic management layers is part of the design.
You may like to investigate the root cause. 
Over to the reviewers.

spacekitcat:
Once we fix the design flow in the cache logic, we can focus in on why the initial network request timed out in the first place. I have some theories, but I think reducing the overall severity of the issue should be our first priority.

llh1 (approved):
Looks good to me 👍 This overall should reduce the number of errors significantly, until we know more about the root cause.

ajroberts10 (approved):

gerardbbc (approved):

murkas (approved):
Very nice investigation, explanation and solution!

PR #1333, opened 2022-02-04
fix: should catch Throwable rather than Exception

spacekitcat:
In PHP 7, you should use Throwable if you want to catch all errors.


murkas (approved):
Thank you for the description added, I was not aware of this feature in PHP7. This seems like a good catch-all.

src/Branding/LogoVersionStrategy.php
```diff
@@ -116,7 +116,7 @@ protected function loadLogoConfig($forceRefresh = false): LogoConfig
                 return $json;
             });
             $this->logoConfig = LogoConfig::fromJSON($json);
-        } catch (\Exception $e) {
+        } catch (\Throwable $e) {
```
spacekitcat, line 119: It is arguably bad practice to catch Throwable here, but the problem is that `fromJSON` appears to throw a TypeError when it fails to parse the contents of `$json`. We need to remove the unparsable payload from the cache, so I think this context is an exception.

ajroberts10: We could always just add an extra catch block for `TypeError` if we didn't want to be so absolute although we probably just want to delete this key in the event of any error right?

ajroberts10: Maybe having an additional catch block here means we could add some logging specific to this issue, possibly making it easier to identify in the logs?

PR #1334, opened 2022-05-03
DATCAP-546" Adding orbit overrides to reset max width changes

ajroberts10:
Orbit is being updated such that the default maximum width will be 1280px instead of the current 1008px. (1280px has been available as an opt-in max width for a long time, but is now becoming the default max width.)

This impacts the Programmes pages as they're not currently opted in to handle the new 1280px width.

This PR adds a number of Orbit overrides to the `_main.scss` to revert these new width changes, whilst maintaining functionality for any pages that are setup to use 1280px.

### Jira 
https://jira.dev.bbc.co.uk/browse/DATCAP-546

Screenshot 2022-05-03 at 11 37 14
Screenshot 2022-05-03 at 11 36 27


bbc-kellys (approved):
📡 🪐

murkas (approved):

gerardbbc (approved):

PR #1335, opened 2022-05-09
DATCAP-546: removing padding for orbit updates

ajroberts10:
Taking out padding for orbit overrides. Fixes issue on the programmes homepage (and linked pages) where text isn't appropriately padded.

Screenshot 2022-05-09 at 13 47 25


gerardbbc (approved):
Looks good 👍

PR #1352, opened 2022-09-05
DATCAP-602: Remove Sounds Nav from codebase

grahamkeenan:
This PR removes all references to the `Sounds Nav` service from the Frontend codebase as it is being decomissioned.

This will remain in draft until these changes can be tested with someone who has a working Programmes Sandbox

Changes:

* All references to `Sounds Nav` have been removed


murkas (commented):
Provisionally saying that these changes look good and thorough, let's see how it performs in a working environment next!

murkas (approved):

PR #1356, opened 2022-10-04
Fix Critical Security Issues found in Programmes Frontend - Lodash

darceyh03:


murkas (approved):
Looks all good. We have a number of changes here so I would advise at least one more review before merging!

bbc-kellys (approved):

PR #1362, opened 2023-01-23
DATCAP-747 swap out gulp-rev for gulp-rev-all2 as this correctly repl…

ducampbe:
…aces paths in the CSS file (#1360)


murkas (approved):
Nice! Clean fix.

jatinkaswa (approved):

